PR #9, opened 2022-08-09
Support authenticate to private helm registry with username and password

xinnywinne:
b/239483790

Change-Id: Ia30d7af792a0634188238d54ebac43259b384913


nan-yu (commented):
/hold

PR #11, opened 2022-08-10
upsert privateCertSecret from RepoSync

sdowell:
This follows the pattern of the user-provided secrets for git
credentials, where the user is expected to create the RepoSync secret in
the same namespace as the RepoSync. The secrets are then upserted to the
config-management-system namespace by the Reconciler. This is to support
use cases where the RepoSync user does not have access to the c-m-s
namespace.

### DIFF
--- a/e2e/nomostest/ssh.go
+++ b/e2e/nomostest/ssh.go
@@ -237,7 +237,7 @@ func downloadSSHKey(nt *NT) string {
 	return privateKeyPath(nt)
 }
 
-// CreateNamespaceSecret creates secret in a given namespace using privateKeyPath.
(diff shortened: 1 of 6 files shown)


nan-yu (approved):
/lgtm

janetkuo:
I saw `ErrImagePull` errors from the test.

sdowell:
> I saw `ErrImagePull` errors from the test.

Yeah it looks like another kind-related issue. It seems to be less frequent after the side loading but is still flaky. I'm thinking it might be best to proceed with upgrading the kind version.

nan-yu (approved):
/lgtm

PR #22, opened 2022-08-12
Support authenticate to private helm artifact registry with GCPServic…

xinnywinne:
…eAccount

### DIFF
--- a/cmd/helm-sync/main.go
+++ b/cmd/helm-sync/main.go
@@ -36,6 +36,12 @@ var (
 		"the name of the helm chart being synced")
 	flVersion = flag.String("version", os.Getenv(reconcilermanager.HelmChartVersion),
 		"the version of the helm chart being synced")
+	flValuesFiles = flag.String("values-files", os.Getenv(reconcilermanager.HelmValuesFiles),
+		"helm values files url")
+	flValues = flag.String("values", os.Getenv(reconcilermanager.HelmValues),
+		"set the helm chart values, will be used in helm template --set key=value")
+	flIncludeCRDs = flag.String("include-crds", os.Getenv(reconcilermanager.HelmIncludeCRDs),
+		"include CRDs in the helm rendering output")
 	flAuth = flag.String("auth", util.EnvString(reconcilermanager.HelmAuthType, string(configsync.AuthNone)),
 		fmt.Sprintf("the authentication type for access to the Helm repository. Must be one of %s, %s, %s or %s. Defaults to %s",
 			configsync.AuthGCPServiceAccount, configsync.AuthToken, configsync.AuthGCENode, configsync.AuthNone, configsync.AuthNone))
@@ -67,7 +73,8 @@ func main() {
 	utillog.Setup()
 	log := utillog.NewLogger(klogr.New(), *flRoot, *flErrorFile)
 	log.Info("rendering Helm chart with arguments", "--repo", *flRepo,
-		"--chart", *flChart, "--version", *flVersion, "--root", *flRoot, "--dest", *flDest, "--wait", *flWait,
+		"--chart", *flChart, "--version", *flVersion, "--root", *flRoot, "--values-files", *flValuesFiles,
+		"--values", *flValues, "--include-crds", *flIncludeCRDs, "--dest", *flDest, "--wait", *flWait,
 		"--error-file", *flErrorFile, "--timeout", *flSyncTimeout,
 		"--one-time", *flOneTime, "--max-sync-failures", *flMaxSyncFailures)
 
@@ -103,6 +110,9 @@ func main() {
 			Version:     *flVersion,
 			ReleaseName: *flReleaseName,
 			Namespace:   *flNamespace,
+			Values:      *flValues,
+			ValuesFiles: *flValuesFiles,
+			IncludeCRDs: *flIncludeCRDs,
 			Auth:        configsync.AuthType(*flAuth),
 			HydrateRoot: *flRoot,
 			Dest:        *flDest,
--- a/e2e/nomostest/sync.go
+++ b/e2e/nomostest/sync.go
@@ -257,35 +257,46 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType v1beta1.Sourc
 	if status.Source.ErrorSummary != nil && status.Source.ErrorSummary.TotalCount > 0 {
 		return fmt.Errorf("status.source contains %d errors:\n%s", status.Source.ErrorSummary.TotalCount, objJSON)
 	}
-	if sourceType == v1beta1.OciSource {
-		if ociDir := status.Source.Oci.Dir; ociDir != dir {
-			return fmt.Errorf("status.source.ociStatus.dir %q does not match the provided directory %q:\n%s", ociDir, dir, objJSON)
-		}
-	} else if gitDir := status.Source.Git.Dir; gitDir != dir {
-		return fmt.Errorf("status.source.gitStatus.dir %q does not match the provided directory %q:\n%s", gitDir, dir, objJSON)
-	}
 	if status.Sync.ErrorSummary != nil && status.Sync.ErrorSummary.TotalCount > 0 {
 		return fmt.Errorf("status.sync contains %d errors:\n%s", status.Sync.ErrorSummary.TotalCount, objJSON)
-	}
-	if sourceType == v1beta1.OciSource {
-		if ociDir := status.Sync.Oci.Dir; ociDir != dir {
-			return fmt.Errorf("status.sync.ociStatus.dir %q does not match the provided directory %q:\n%s", ociDir, dir, objJSON)
-		}
-	} else if gitDir := status.Sync.Git.Dir; gitDir != dir {
-		return fmt.Errorf("status.sync.gitStatus.dir %q does not match the provided directory %q:\n%s", gitDir, dir, objJSON)
 	}
 	if status.Rendering.ErrorSummary != nil && status.Rendering.ErrorSummary.TotalCount > 0 {
 		return fmt.Errorf("status.rendering contains %d errors:\n%s", status.Rendering.ErrorSummary.TotalCount, objJSON)
 	}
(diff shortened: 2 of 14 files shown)


cmd/helm-sync/main.go
```diff
@@ -36,6 +36,14 @@ var (
 		"the name of the helm chart being synced")
 	flVersion = flag.String("version", os.Getenv(reconcilermanager.HelmChartVersion),
 		"the version of the helm chart being synced")
+	flValuesFiles = flag.String("values-files", os.Getenv(reconcilermanager.HelmValuesFiles),
+		"helm values files url")
```
janetkuo, line 40: Is this local path or remote url? This isn't consistent with the R*Sync API field description.

xinnywinne: Now I think it should be better to be a url, this way user will have more flexibility and they can also use public values files, instead of must use a values file in the same repo. The downside is that, if user use the values file in the same repo, they need to write the full url again. If you agree, I can update the API description to be consistent.

janetkuo: Users should be able to specify either local file path or remote url, and the tool should be able to detect it, instead of requiring local file path to be written in the url format. For example, Kustomize Helm plugin takes either a local file path or a remote URL to a values file.

If the usage of a field is changed here, the API doc needs to be updated in the same commit.

pkg/reconcilermanager/controllers/util.go
```diff
@@ -206,8 +208,25 @@ const (
 )
 
 // helmSyncEnvs returns the environment variables for the oci-sync container.
```
janetkuo, line 210:
```suggestion
// helmSyncEnvs returns the environment variables for the helm-sync container.
```

cmd/helm-sync/main.go
```diff
@@ -36,6 +36,14 @@ var (
...
+	flValuesValue = flag.String("values-v", os.Getenv(reconcilermanager.HelmValuesValue),
+		"the value of the helm chart values key value pair, will be used in helm template --set key=value")
+	flValuesKey = flag.String("values-k", os.Getenv(reconcilermanager.HelmValuesKey),
+		"the key of the helm chart values key value pair, will be used in helm template --set key=value")
```
janetkuo, line 44: Seems weird to keep keys and values separated?

cmd/helm-sync/main.go
xinnywinne: https://docs.google.com/document/d/1lDMU0lulNYhIQ_8nk330r6eI72jp6_95qufTXy-LwR4/edit?resourcekey=0-YP0D5bF2S0bCJhnXaBeV-A&disco=AAAAdL1P-gY. Had a discussion with Morten before, since user cannot put the values file into our container directly, we will only support remote URL. I will update the API description to make it more clear.

e2e/nomostest/sync.go
```diff
@@ -257,35 +257,46 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType v1beta1.Sourc
...
 		return fmt.Errorf("status.sync contains %d errors:\n%s", status.Sync.ErrorSummary.TotalCount, objJSON)
 	}
-	if sourceType == v1beta1.OciSource {
-		if ociDir := status.Sync.Oci.Dir; ociDir != dir {
-			return fmt.Errorf("status.sync.ociStatus.dir %q does not match the provided directory %q:\n%s", ociDir, dir, objJSON)
-		}
-	} else if gitDir := status.Sync.Git.Dir; gitDir != dir {
-		return fmt.Errorf("status.sync.gitStatus.dir %q does not match the provided directory %q:\n%s", gitDir, dir, objJSON)
+	if message := status.Rendering.Message; message != parse.RenderingSucceeded && message != parse.RenderingSkipped {
+		return fmt.Errorf("status.rendering.message %q does not indicate a successful state:\n%s", message, objJSON)
```
janetkuo, line 264: This if statement was originally after the next line that checks rendering error - is it still correct to move it before? It seems that rendering could be either in progress or failed, moving this changes the original behavior.

cmd/helm-sync/main.go
xinnywinne: Current solution is to process the values ( map[string]interface{} )in the util.go, combine all the keys together, all the values together and then pass to env. I am wondering if it would be better to pass the map to env directly.  Use Unstructured.MarshalJSON in util.go (unstructure -> byte[]) and then UnmarshalJSON (byte[] -> unstructure)in helm-sync container. Do you have any preference or suggestion?

pkg/reconcilermanager/controllers/util.go
```diff
@@ -206,8 +208,25 @@ const (
...
-func helmSyncEnvs(repo, chart, version, releaseName, namespace string, auth configsync.AuthType, period float64) []corev1.EnvVar {
+func helmSyncEnvs(repo, chart, version, releaseName, namespace string, values *unstructured.Unstructured, valuesFiles []string, includeCRDs bool, auth configsync.AuthType, period float64) []corev1.EnvVar {
```
janetkuo, line 211: There are too many parameters in this function. Consider creating a struct for it (e.g. `helmOptions`). Can be done in a separate commit or in a follow-up PR.

pkg/helm/helm.go
```diff
@@ -58,26 +65,61 @@ func (h *Hydrator) templateArgs(destDir string) []string {
...
+		args = append(args, "--include-crds")
+	}
 	args = append(args, "--output-dir", destDir)
 	return args
 }
 
-func (h *Hydrator) registryLoginArgs() []string {
+func (h *Hydrator) registryLoginArgs(ctx context.Context) ([]string, error) {
 	args := []string{"registry", "login"}
 	if h.Auth == configsync.AuthToken {
```
janetkuo, line 91: Seems more idiomatic to use `switch h.Auth` here

cmd/helm-sync/main.go
janetkuo: Separating the map k,v pair and then assembling them later is error prone. Could we combine the map as a flag with string type, such as `k1=v1,k2=v2`?

cmd/helm-sync/main.go
janetkuo: No, we shouldn't ask users to put values file in the container. For local file path - can users put multiple values files in the chart, and then reference different values files (as local file path in the remote chart) at runtime?

mortent: So users don't really have a way to provide values files to the container without including them in the Helm charts. But I'm not sure if we have good use-cases why users would want to do that.
Remote files are possible, but I'm a bit skeptical about that one too since that in a lot of cases will require auth, which is likely to need different credentials from fetching the chart itself. Asking users to provide multiple sets of credentials gets kinda complicated.

My take on this is that we should not support values files in this way for now. We can let users provide the values through the API directly and we can revisit this if we have a good use-case. It is easier to leave it out than trying something and getting it wrong.

pkg/reconcilermanager/controllers/util.go
xinnywinne: I think we can, I am following the style of other env functions, I was wondering why the other env functions do not just pass in the spec directly.

cmd/helm-sync/main.go
janetkuo: My thought of the local values file usage is that users would include multiple values file in a chart, and specify different files in different clusters/environment at runtime. I'm not 100% sure if that's how users would use it, though. I think it's fine if we don't support it in the first version.

xinnywinne: My idea is that we can support the valuesfiles under the same repo which need Auth, or under different public repo. I changed the API field description to: 
// valuesFiles is a list of URL to Helm value files.
// Values files must be in the same repository with the Helm chart or in a public repository.

xinnywinne: And in the future we can decide if we want to support local file path.

xinnywinne: Will remove it in separate PR.

cmd/helm-sync/main.go
xinnywinne: Done

e2e/nomostest/sync.go
xinnywinne: Done

pkg/reconcilermanager/controllers/util.go
xinnywinne: Done!

pkg/helm/helm.go
xinnywinne: Done

pkg/reconcilermanager/controllers/util.go
```diff
@@ -205,30 +206,52 @@ const (
...
+	helmValuesFiles := ""
+	if len(helm.ValuesFiles) > 0 {
+		helmValuesFiles = strings.Join(helm.ValuesFiles, ",")
+	}
+	if helm.Values != nil {
+		var vals []string
+		for key, val := range helm.Values.Object {
+			vals = append(vals, fmt.Sprintf("%s=%v", key, val))
+		}
+		helmValues = strings.Join(vals, ",")
```
janetkuo, line 222: nit: there will be a trailing "," (i.e. `k1=v1,k2=v2,`) - would that cause any issues?

janetkuo (commented):
One comment, lgtm otherwise

pkg/reconcilermanager/controllers/util.go
janetkuo: Per discussion, `Join` won't add trailing `,`

janetkuo (approved):
/lgtm

karlkfi:
Please don't manually merge PRs. Wait for the tests to pass and github will auto-merge it once the checks are passed.

xinnywinne:
After all the tests pass on #22, the tide task bundle #22 with #33 to retest, and when this process almost finish, it separate  #22 from #33 and start rerun them separately again. Since #33 only change one line and does not relate to my change, I decided to manually merge it.

PR #24, opened 2022-08-12
added a design doc template

mikebz:
Created a design document template similar to the one that is used in kpt.  I suggest that we start using the design document process in the public repo instead of internal Google docs.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,6 +1,6 @@
 # How to Contribute
 
-ConfigSync follows the fork and PR model that a lot of OSS projects do.  If you
+ConfigSync follows the [Kubernetes contribution process].  If you
 are looking to contribute to this project please fork it, create a change
 and then submit a PR against the `main` branch which is the HEAD of development.
 
@@ -33,3 +33,4 @@ Guidelines](https://opensource.google/conduct/).
 
 [developer guide]: development.md
 [design document]: design-docs/00-template.md
+[Kubernetes contribution process]: https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,15 +1,11 @@
 # How to Contribute
 
-Config Sync is currently being mirrored from internal Google systems to Github,
-but we are working on migrating all the infrastructure for the project onto
-github. Our goal is that all development of Config Sync will happen in this
-repository. We expect to complete this effort in a couple of months, so stay tuned.
+ConfigSync follows the fork and PR model that a lot of OSS projects do.  If you
+are looking to contribute to this project please fork it, create a change
+and then submit a PR against the `main` branch which is the HEAD of development.
 
-In the meantime, we will try to upstream any PRs made against this github repo
-on a best-effort basis. If you run into any issues using Config Sync or have
-any suggestions or questions to the Config Sync team, file an issue.
-
-When you are ready, take a look at our [developer guide] to get started.
+For larger changes and for anything that has a change to the CLI and API please
+write a [design document] first.
 
 ## Contributor License Agreement
 
@@ -36,3 +32,4 @@ This project follows [Google's Open Source Community
 Guidelines](https://opensource.google/conduct/).
 
 [developer guide]: development.md
+[design document]: design-docs/00-template.md
--- a/docs/design-docs/00-template.md
+++ b/docs/design-docs/00-template.md
@@ -1,0 +1,90 @@
+# Title
+
+* Author(s): \<your name\>, \<your github alias\>
+* Approver: \<kpt-maintainer\>
+
+>    Every feature will need design sign off an PR approval from a core
+>    maintainer.  If you have not got in touch with anyone yet, you can leave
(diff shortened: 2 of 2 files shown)


docs/contributing.md
```diff
@@ -35,4 +31,5 @@ information on using pull requests.
 This project follows [Google's Open Source Community
 Guidelines](https://opensource.google/conduct/).
 
-[developer guide]: development.md
\ No newline at end of file
+[developer guide]: development.md
+[design document]: design-docs/00-template.md
```
sdowell, line 35: If we want to use the same template as `kpt`, perhaps we could point directly to the template in the `kpt` repo so that its only defined in one place. WDYT?

mikebz: I think the design changes in kpt are not the same as ConfigSync.  I am open to decoupling the approvers but it seems like we are starting with combined group first.

I vote for design changes to ConfigSync being in the same repo as the code.

sdowell: I'm only suggesting that we point to the same template, I agree that config sync design docs themselves should be committed to the config sync repo

mikebz: I think cross repo template pointing is probably a little bit strange.  Once we get one design doc PR I am actually going to change the reference to the local repo for a PR example too.

docs/design-docs/00-template.md
```diff
@@ -0,0 +1,90 @@
+# Title
+
+* Author(s): \<your name\>, \<your github alias\>
+* Approver: \<kpt-maintainer\>
+
+>    Every feature will need design sign off an PR approval from a core
+>    maintainer.  If you have not got in touch with anyone yet, you can leave
+>    this blank and we will try to line someone up for you.
+
+## Why
```
janetkuo, line 10:
```suggestion
## Motivation
```

docs/design-docs/00-template.md
```diff
@@ -0,0 +1,90 @@
...
+* interface changes
+* design assumptions
+
+For a new config change, please mention:
+
+* Is it backwards compatible? If not, what is the migration and deprecation 
+  plan?
+
+
+## User Guide
```
janetkuo, line 34: Suggest: 

`## Design Overview` --> `## User Guide` (or `## User Stories`) --> `## Design Details`

docs/design-docs/00-template.md
```diff
@@ -0,0 +1,90 @@
...
+descriptions, and success and failure messages (Failure messages are an 
+important part of a good UX). This level of detail serves two functions:
+
+It forces the author and the readers to explicitly think about possible friction
+points, pitfalls, failure scenarios, and corner cases (“A measure of a good 
+engineer is not how clever they are, but whether they think about all the 
+corner cases”). Makes it easier to author the user-facing docs as part of the 
+development process (Ideally as part of the same PR) as opposed to it being an 
+afterthought.
+
```
janetkuo, line 55: Suggest adding `## Risks and Mitigations` and `## Test Plan`

docs/design-docs/00-template.md
```diff
@@ -0,0 +1,90 @@
+# Title
+
+* Author(s): \<your name\>, \<your github alias\>
+* Approver: \<kpt-maintainer\>
+
+>    Every feature will need design sign off an PR approval from a core
+>    maintainer.  If you have not got in touch with anyone yet, you can leave
+>    this blank and we will try to line someone up for you.
+
```
janetkuo, line 9: Suggest adding `## Summary`

docs/design-docs/00-template.md
```diff
@@ -0,0 +1,90 @@
+# Title
+
+* Author(s): \<your name\>, \<your github alias\>
+* Approver: \<kpt-maintainer\>
+
```
janetkuo, line 5: How about adding 
`* Status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced`

mikebz: adding thnx

docs/design-docs/00-template.md
mikebz: adding

docs/design-docs/00-template.md
mikebz: updated the title.  Not sure if we need to sections for design though.

mikebz: I think asking people to write too much might diminish the participation.

docs/contributing.md
```diff
@@ -1,15 +1,11 @@
 # How to Contribute
 
-Config Sync is currently being mirrored from internal Google systems to Github,
-but we are working on migrating all the infrastructure for the project onto
-github. Our goal is that all development of Config Sync will happen in this
-repository. We expect to complete this effort in a couple of months, so stay tuned.
+ConfigSync follows the fork and PR model that a lot of OSS projects do.  If you
+are looking to contribute to this project please fork it, create a change
+and then submit a PR against the `main` branch which is the HEAD of development.
```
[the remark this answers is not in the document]
mikebz: adding.  thnx

janetkuo (approved):
/lgtm

PR #25, opened 2022-08-12
Use kstatus for reconciler deployment status

karlkfi:
- This reduces duplicate code in CS and makes the messages more
  consistent.


pkg/reconcilermanager/controllers/deployment_conditions.go
```diff
@@ -1,118 +1,29 @@
-// Copyright 2022 Google LLC
```
sdowell, line 1: Keep license header

karlkfi: fixed

PR #28, opened 2022-08-15
Return blocking discover error to prevent resource pruning

tiffanny29631:
This change modifies the GetResource to only tolerate NotFound,
any other error is treated blocking and will be returned to
prevent resource pruning when backend is temporarily unavailable.

### DIFF
--- a/e2e/testcases/apiservice_test.go
+++ b/e2e/testcases/apiservice_test.go
@@ -17,36 +17,114 @@ package e2e
 import (
 	"testing"
 
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+)
+
+const (
+	adapterNamespace = "custom-metrics"
+	adapterName      = "custom-metrics-stackdriver-adapter"
 )
 
 func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
-	nt := nomostest.New(t)
+	nt := nomostest.New(t, ntopts.Unstructured, ntopts.RequireGKE(t))
+	t.Cleanup(func() {
+		if t.Failed() {
+			nt.PodLogs(adapterNamespace, adapterName, "pod-custom-metrics-stackdriver-adapter", true)
+		}
+	})
 	nt.T.Log("Creating commit with APIService and Deployment")
+	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/namespace.yaml", "acme/namespaces/custom-metrics/namespace.yaml")
 	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/namespace-custom-metrics.yaml", "acme/namespaces/custom-metrics/namespace-custom-metrics.yaml")
+	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/rbac.yaml", "acme/cluster/rbac.yaml")
 	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/apiservice.yaml", "acme/cluster/apiservice.yaml")
-	nt.RootRepos[configsync.RootSyncName].CommitAndPush("add testing resources")
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("adding apiservice resources")
 	nt.T.Log("Waiting for nomos to sync new APIService")
 	nt.WaitForRepoSyncs()
 
-	nt.T.Log("Removing APIService")
+	err := validateStackdriverAdapterStatusCurrent(nt)
+	if err != nil {
+		nt.T.Fatalf("validate failed %s", err)
+	}
+
 	nt.RootRepos[configsync.RootSyncName].Remove("acme/cluster/apiservice.yaml")
-	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Remove API service")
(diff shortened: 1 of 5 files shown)


e2e/testcases/apiservice_test.go
```diff
@@ -17,36 +17,117 @@ package e2e
...
-func TestImporterAndSyncerResilientToBadAPIService(t *testing.T) {
-	nt := nomostest.New(t)
-	nt.T.Log("Adding bad APIService")
+func TestImporterAndSyncerResilientToFlakyAPIService(t *testing.T) {
+	nt := nomostest.New(t, ntopts.RequireGKE(t), ntopts.Unstructured)
+	nt.T.Log("Adding API service with backend temporarily unavailable")
 	nt.MustKubectl("apply", "-f", "../testdata/apiservice/apiservice.yaml")
 	nt.T.Cleanup(func() {
 		nt.MustKubectl("delete", "-f", "../testdata/apiservice/apiservice.yaml", "--ignore-not-found")
+		t.Cleanup(func() {
```
karlkfi, line 74: Does this cleanup inside a cleanup even work? It seems like it might break the for loop iterating over the cleanups by modifying the array while looping.

If you're worried about the kubectl delete failing, then put the pod logs in a separate sequential cleanup function. otherwise, I think you can just put the kubectl and pod logs calls next to each other.

e2e/testcases/apiservice_test.go
```diff
@@ -17,36 +17,117 @@ package e2e
...
+	nt.MustKubectl("apply", "-f", "../testdata/apiservice/namespace-custom-metrics.yaml")
 	nt.T.Log("Waiting for nomos to stabilize")
 	nt.WaitForRepoSyncs()
+	err := validateObjects(nt)
+	if err != nil {
+		nt.T.Fatalf("validate failed %s", err)
+	}
+}
+
+func validateObjects(nt *nomostest.NT) error {
```
karlkfi, line 96: given this package is large, maybe make this name more specific.

e2e/testcases/apiservice_test.go
```diff
@@ -17,36 +17,117 @@ package e2e
...
+	// Add retry for backward compatibility for mono repo mode as the status update
+	// does not wait for reconciliation
+	t, e := nomostest.Retry(nt.DefaultWaitTimeout, func() error {
```
karlkfi, line 99: I don't think we need to wait this long (6m) for reconciliation. You can probably use the ReconciliationTimeout, which is 1m, I think.

e2e/testdata/apiservice/namespace.yaml
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: custom-metrics
```
karlkfi, line 18: add trailing linebreak to all files.

karlkfi (changes requested):

e2e/testcases/apiservice_test.go
tiffanny29631: Removed the nested Cleanup.

e2e/testcases/apiservice_test.go
tiffanny29631: Renamed.

e2e/testcases/apiservice_test.go

e2e/testdata/apiservice/namespace.yaml
tiffanny29631: Done

e2e/testcases/apiservice_test.go
```diff
@@ -17,36 +17,114 @@ package e2e
...
-	nt := nomostest.New(t)
-	nt.T.Log("Adding bad APIService")
+func TestImporterAndSyncerResilientToFlakyAPIService(t *testing.T) {
+	nt := nomostest.New(t, ntopts.RequireGKE(t), ntopts.Unstructured)
+	nt.T.Log("Adding API service with backend temporarily unavailable")
 	nt.MustKubectl("apply", "-f", "../testdata/apiservice/apiservice.yaml")
 	nt.T.Cleanup(func() {
 		nt.MustKubectl("delete", "-f", "../testdata/apiservice/apiservice.yaml", "--ignore-not-found")
+		if t.Failed() {
+			nt.PodLogs(adapterNamespace, adapterName, "pod-custom-metrics-stackdriver-adapter", true)
```
karlkfi, line 75: Is this pod one of the ones in `apiservice.yaml`? If so, print he pod logs before deleting.

pkg/util/discovery/server_resourcer.go
```diff
@@ -52,20 +54,22 @@ func GetResources(discoveryClient ServerResourcer) ([]*metav1.APIResourceList, s
...
-			klog.Warningf("failed to discover some APIGroups: %s", discoveryErr)
+		// b/238836947 ServerGroupsAndResources still returns the resources it discovered when there was an error.
+		// Most errors are fatal, but we want to ignore NotFound errors. This allows for CRDs and CRs to be applied
+		// with a two-pass apply & retry strategy, where objects with a NotFound resources are skipped until the
+		// resource is registered. Other errors must cause failure, otherwise they might cause existing objects to be pruned.
+		var discoErr *discovery.ErrGroupDiscoveryFailed
+		if errors.As(discoveryErr, &discoErr) {
+			for gv, subErr := range discoErr.Groups {
+				if apierrors.IsNotFound(subErr) {
+					klog.Warningf("Failed to discover APIGroups %s due to not found, moving on: %v", gv, subErr)
```
karlkfi, line 65: nit: s/APIGroups/APIGroup or "API group" (not plural)

karlkfi:
/lgtm

Just a couple nits/questions

karlkfi (approved):

PR #31, opened 2022-08-16
pass caCertPath to shared test envs

sdowell:
This fixes a failure we are seeing in shared test environments where the
once created caCertPath was not being properly passed to the shared test
NT.


xinnywinne (approved):
/lgtm

PR #33, opened 2022-08-16
add wait for ready to kind cluster creation

sdowell:
There is a flaky condition in the kind e2e tests where the nodes are not
yet available when the git server pod is created. This may be a race
condition where the nodes are not ready yet, so this change attempts to
address it by adding a wait for ready step to the cluster creation.

### DIFF
--- a/e2e/nomostest/ntopts/kind.go
+++ b/e2e/nomostest/ntopts/kind.go
@@ -211,6 +211,8 @@ apiServer:
 			}),
 			// Retain nodes for debugging logs.
 			cluster.CreateWithRetain(true),
+			// Wait for cluster to be ready before proceeding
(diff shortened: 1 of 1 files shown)


e2e/nomostest/ntopts/kind.go
```diff
@@ -211,6 +211,8 @@ apiServer:
 			}),
 			// Retain nodes for debugging logs.
 			cluster.CreateWithRetain(true),
+			// Wait for cluster to be ready before proceeding
+			cluster.CreateWithWaitForReady(10*time.Minute),
```
karlkfi, line 215: 5m is probably good enough, but 10 is fine.

karlkfi (approved):
/lgtm

janetkuo (approved):
/lgtm

PR #34, opened 2022-08-17
Support authenticate to private helm repo with token

xinnywinne:


pkg/helm/helm.go
```diff
@@ -147,3 +171,33 @@ func (h *Hydrator) HelmTemplate(ctx context.Context) error {
...
+func (h *Hydrator) authTokenArgs(args []string) []string {
+	args = append(args, "--username", h.UserName)
+	args = append(args, "--password", h.Password)
+	if !h.isOCI() {
+		certFilePath := filepath.Join(HelmCredentialMountPath, HelmSecretKeyCert)
+		if _, err := os.Stat(certFilePath); err == nil {
+			args = append(args, "--cert-file", certFilePath)
+		}
+		caFilePath := filepath.Join(HelmCredentialMountPath, HelmSecretKeyCa)
+		if _, err := os.Stat(certFilePath); err == nil {
```
sdowell, line 184: `caFilePath`

pkg/helm/helm.go
```diff
@@ -147,3 +171,33 @@ func (h *Hydrator) HelmTemplate(ctx context.Context) error {
...
+			args = append(args, "--ca-file", caFilePath)
+		}
+		keyFilePath := filepath.Join(HelmCredentialMountPath, HelmSecretKeyCa)
+		if _, err := os.Stat(certFilePath); err == nil {
```
sdowell, line 188: `keyFilePath`

pkg/helm/helm.go
```diff
@@ -147,3 +171,33 @@ func (h *Hydrator) HelmTemplate(ctx context.Context) error {
...
+	args = append(args, "--username", h.UserName)
+	args = append(args, "--password", h.Password)
+	if !h.isOCI() {
+		certFilePath := filepath.Join(HelmCredentialMountPath, HelmSecretKeyCert)
+		if _, err := os.Stat(certFilePath); err == nil {
+			args = append(args, "--cert-file", certFilePath)
+		}
+		caFilePath := filepath.Join(HelmCredentialMountPath, HelmSecretKeyCa)
+		if _, err := os.Stat(certFilePath); err == nil {
+			args = append(args, "--ca-file", caFilePath)
```
sdowell, line 185: I would expect CA cert to be decoupled from the auth type. A user should be able to supply their own CA regardless of auth method (e.g. for authtype `none`). If there is a discriminator I'd expect it might be the application layer protocol being used (e.g. `https`). This code path is only reached for auth type `token`

xinnywinne: Thanks Sam,  I think for the case you mentioned I need to add check for username and password. I can't assume they aways exist for token auth.  The idea is let the user decide what they want. They can do any combination. The token auth here means read from the secret.

sdowell: My suggestion would be to move the `--ca-file` argument outside of the `switch h.Auth` statement on line 110, and put it in a new if statement that checks application layer protocol (e.g. `h.isHTTPS()`)

xinnywinne: So the plan was handle the -ca-file in the same secret with other files and password, since they all options for helm template .. Currently, in the auth types helm support, only auth token can pass in data through secret. I agree that reuse token auth make it a little bit unclear. I used to create a new auth type , which name is "secret", means get data from secret.

sdowell: This is the main motivation for the `privateCertSecret` API for `git-sync` being managed with a separate secret from the `git-creds` secret. I think we should consider being consistent with the APIs to avoid user confusion. @janetkuo WDYT?

xinnywinne: I think helm-sync is different, the ca-file here is a option for helm template, same as other options(key-file, cert-file, username, password), put them in separate secrets for me it is not a good user experience, and I think we can handle them  all within one secret, the only question is that if we need an extra auth type.

sdowell: Seems like it's pretty much the same model as `privateCertSecret`, we are mapping the user provided secret to the container so that it can be passed to the underlying binary (`git`, `helm`, etc). Regardless of whether we choose to share a secret with the credentials or use a separate secret I think it is confusing to have an inconsistent API between git and helm for an analogous feature.

xinnywinne: privateCertSecert is part of git, it is not a field shared between oci, helm, git. Helm support is totally separate from git,   so I prefer not to separate template options into different secrets to follow the git design.

janetkuo: Separating repo credential and certificates into 2 secret refs in the API has the benefit that users can still choose to either put the information in the same Secret or separate Secrets. When we design this we just need to make sure the key of the Secrets don't overlap (i.e. login creds and certs use different keys.)

pkg/helm/helm.go
xinnywinne: Removed the cert file support

pkg/helm/helm.go
xinnywinne: Decided to push the ca, cert, key files support to after 1.13.

e2e/testcases/helm_sync_test.go
```diff
@@ -83,6 +84,39 @@ func TestHelmARFleetWISameProject(t *testing.T) {
 	})
 }
 
+// TestHelmARTokenAuth verifies Config Sync can pull Helm chart from private Artifact Registry with Token auth type.
+// This test will work only with following pre-requisites:
+// Google service account `e2e-test-ar-reader@stolos-dev.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for accessing images in Artifact Registry.
```
sdowell, line 89: I'd like to start translating these manually provisioned test resources to scripts so that we can easily automate creation of new test environments. None of the other tests have it though so it's not necessary for this PR

sdowell (approved):
/lgtm

e2e/testcases/helm_sync_test.go
xinnywinne: Sounds good!

PR #35, opened 2022-08-17
fix: Handle empty initContainers in pod resources

rquitales:
When describing a pod, we can have the initContainers field to be null.
This causes an error when we use unstructured.NestedSlice to access this
field as the unstructured library tries to assert nil to []interface{},
but fails, since we can't assert a nil value to be of []interface{}.
This PR ignores empty initContainers since it can still be a valid
resource yaml. It also ignores an empty port field gracefully.

This PR also uses the official conversion library function to convert an
unstructured object to its concrete type for validation purposes.

Tested by running reconciler-manager on a live cluster.

### DIFF
--- a/pkg/core/structured.go
+++ b/pkg/core/structured.go
@@ -15,7 +15,7 @@
 package core
 
 import (
-	"encoding/json"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


nan-yu (approved):
/lgtm

PR #37, opened 2022-08-17
fix lint-license-headers make target

sdowell:
the exit code was being lost in the previous make target due to the
pipe. This change moves the command to a bash script with pipefail
enabled so that the exit code is not lost.

### DIFF
--- a/scripts/lint-license-headers.sh
+++ b/scripts/lint-license-headers.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


janetkuo (approved):
/lgtm

scripts/lint-license-headers.sh
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+set -euo pipefail
+
+if [ "${GOBIN:-"unset"}" == "unset" ]; then
+  echo "GOBIN unset"
+  exit 1
+fi
+
+"${GOBIN}/addlicense" -check -ignore=vendor/** -ignore=third_party/** -ignore=.output/** . 2>&1 | sed '/ skipping: / d'
```
mikebz, line 23: do we want to add the license holder and type of license here?  

```
-l license type: apache, bsd, mit, mpl (defaults to "apache")
-c copyright holder (defaults to "Google LLC")
```

mikebz: oh I guess the defaults are what we want.

PR #39, opened 2022-08-18
[WIP] Use predicate to filter reconciles

karlkfi:
- Don't enqueue a reconcile if the ResourceVersion hasn't changed
- Remove lock and resourceVersion cache that was attempting to dedupe
  the hard way
- Reduce watch scope from cluster to namespace for Deployments and
  RootSync Secrets
- Add RootSync watch for managed ClusterRoleBindings (shared), to
  ensure that changes to the crb are reverted without needing to update
  another watched object.
- Add RootSync ownership for managed ServiceAccounts, to ensure that
  changes to the sa are reverted without needing to update another
  watched object.

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_base.go
+++ b/pkg/reconcilermanager/controllers/reconciler_base.go
@@ -82,15 +82,6 @@ type reconcilerBase struct {
 
 	// syncKind is the kind of the sync object: RootSync or RepoSync.
 	syncKind string
-
-	// lastReconciledResourceVersions is a cache of the last reconciled
-	// ResourceVersion for each R*Sync objects.
-	//
-	// This is used for an optimization to avoid re-reconciling.
-	// However, since ResourceVersion must be treated as opaque, we can't know
-	// if it's the latest or not. So this is just an optimization, not a guarantee.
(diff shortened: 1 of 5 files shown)


sdowell:
LGTM but someone else should take a look

karlkfi:
/hold

It occurs to me that this might not do what I expected.

The reason I initially added the dedupping was because the Reconcile was causing RSync status updates to happen using the same initial generation. They weren't always duplicate updates, because some other robject might have changed, but the GET at the beginning of the Reconcile was returning an RSync with a generation we'd already seen. My assumption was that this was due to the reconciler-manager using the controller-manager's caching client. Apparently the caching client doesn't invalidate the cache on writes, only on watch updates. So if another watched object (like the Deployment) triggered a reconcile, the client cache might still give you the same RSync it had cached, not the new one from the previous reconcile.

But the predicate in this PR doesn't check the RSync in the cache. It only checks that the object that triggered the watch event has changed (old RV != new RV). This may not actually do what the existing deduping code was doing, because the existing code was always deduping the RSync, not the watched object that triggered the reconcile.

So we probably do want this predicate added, but we probably don't want to remove the other dedupping code.

So I'm gonna put this on hold for now. I'll probably need to add tests to confirm it actually does what I expected.

karlkfi:
Obsolete

PR #42, opened 2022-08-18
add make targets for building release candidates

sdowell:
These make targets are intended to be used for tagging, building, and
publishing release candidates. Make targets for promoting release
candidates will be added in a subsequent change.

### DIFF
--- a/scripts/tag-release-candidate.sh
+++ b/scripts/tag-release-candidate.sh
@@ -26,21 +26,27 @@ set -eo pipefail
 # It's crucial that error messages in these functions be sent to stderr with 1>&2.
 # Otherwise, they will not bubble up to the make target that calls this script.
 
+err () {
+  local msg="$1"
+  echo "${msg}" 1>&2
+  exit 1
+}
+
 latest_rc_of_version () {
   local version="$1"
   if [ -z "$version" ]; then
-    echo "Error: function expects version argument" 1>&2
-    exit 1
+    err "Error: function expects version argument"
   fi
 
 
   # 1. Get all the versions with the right prefix,
   # 2. order them by their "rc.X" values
   # 3. Take the last/highest one
-  rc=$(git tag --list "$version*" | sort -V | tail -n 1)
+  rc=$(git tag --list "${version}-rc.*" | sort -V | tail -n 1)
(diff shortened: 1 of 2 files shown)


scripts/tag-release-candidate.sh
```diff
@@ -79,9 +80,10 @@ if [ -z "$CS_VERSION" ]; then
   exit 1
 fi
 
-# If a user omits the leading "v" (enters 1.2.3 instead of v1.2.3), handle it.
-if ! [[ "${CS_VERSION}" = v* ]]; then
-  CS_VERSION="v${CS_VERSION}"
```
janetkuo, line 84: Why not keeping this?

scripts/tag-release-candidate.sh
```diff
@@ -79,9 +80,10 @@ if [ -z "$CS_VERSION" ]; then
...
+# If a user omits the leading "v" (enters 1.2.3 instead of v1.2.3), throw an error.
+if ! [[ "${CS_VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+  echo "Error: CS_VERSION must be in the form v1.2.3 - got ${CS_VERSION}"
```
janetkuo, line 85: Combine this with L78-81 if we choose to throw an error here

scripts/tag-release-candidate.sh
sdowell: IMO having a stricter check is preferred, as it leaves less room for error. The `__build` make target also builds on the assumption of the well formed version for finding the name of the RC tag.

The plan is for this to all be run via automation, so it shouldn't be a large inconvenience users.

scripts/tag-release-candidate.sh
sdowell: I think having separate conditions and associated error messages is valid, it makes it clear whether the error was that
- The `CS_VERSION` was not provided
- The `CS_VERSION` was provided but is in an incorrect format

WDYT?

sdowell: I removed L78-81

janetkuo (approved):
/lgtm

PR #45, opened 2022-08-19
enable gh access tokens for tag-release-candidate

sdowell:


scripts/tag-release-candidate.sh
```diff
@@ -84,8 +84,11 @@ if ! [[ "${CS_VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
   err "Error: CS_VERSION must be in the form v1.2.3 - got ${CS_VERSION}"
 fi
 
+REMOTE="git@github.com:GoogleContainerTools/kpt-config-sync.git"
+# Use Github access token if provided
+[[ -n "${GH_TOKEN}" ]] && REMOTE="https://${GH_TOKEN}@github.com/GoogleContainerTools/kpt-config-sync.git"
 # Fetch all existing tags
-git fetch --tags > /dev/null
+git fetch "${REMOTE}" --tags > /dev/null
```
karlkfi, line 91: This puts the token in the bash history, which isn't ideal for security. But I think the only way to do it more securely would be to use the `gh` CLI, which will allow reading it from an environment variable. Another option is reading from disk with `gh auth login --with-token < mytoken.txt`. I guess you could also disable bash history temporarily and re-enable afterwards, which I've seen before to prevent leakage.

sdowell: I can disable history for the scope of the script to be safe

karlkfi (approved):
/lgtm

PR #47, opened 2022-08-22
add OSS manifest staging dir to make target

sdowell:
This change adds a step to the building of config sync manifests which
produces the subset of manifests for multirepo/oss from the already
generated ACM manifests.


janetkuo (approved):
/lgtm

PR #48, opened 2022-08-22
reconciler-manager cleanup

karlkfi:
- Make logs a little more consistent
- Add a few missing logs to catch all managed object upserts
- Log namespace/name instead of the raw NamespacedName object
- Use -Ref for ObjectKey and NamespacedName variable names
- Reduce number of pointers to types objects, where possible.


sdowell (approved):
/lgtm

PR #49, opened 2022-08-22
Fix apiservice bats test

tiffanny29631:
This change applies to the recent CI failures in mono repo test under GKE environment.

Also updates the Gcloud CLI used in test containers and installs gke-gcloud-auth-plugin.

### DIFF
--- a/e2e/testcases/apiservice.bats
+++ b/e2e/testcases/apiservice.bats
@@ -11,19 +11,37 @@ load "../lib/wait"
 YAML_DIR=${BATS_TEST_DIRNAME}/../testdata/apiservice
 
 test_setup() {
+  kubectl apply -f "${MANIFEST_DIR}/source-format_unstructured.yaml"
+  nomos::restart_pods
   setup::git::initialize
-  setup::git::commit_minimal_repo_contents
 }
 
 test_teardown() {
+  # Remove test APIService to prevent Discovery failure blocking test repo cleanup
   git::rm acme/cluster/apiservice.yaml
   git::commit
+  wait::for -t 120 -- nomos::repo_synced
+
+  # Delete resources applied by Kubectl
   kubectl delete -f "${YAML_DIR}/apiservice.yaml" --ignore-not-found
+  kubectl delete -f "${YAML_DIR}/rbac.yaml" --ignore-not-found
+  kubectl delete -f "${YAML_DIR}/namespace.yaml" --ignore-not-found
+  kubectl delete -f "${YAML_DIR}/namespace-custom-metrics.yaml" --ignore-not-found
+
   setup::git::remove_all acme
+  # Restore the resource format to hierarchical mode in shared test env
+  kubectl apply -f "${MANIFEST_DIR}/source-format_hierarchy.yaml"
+  nomos::restart_pods
 }
 
 @test "Create APIService and endpoint in same commit" {
   debug::log "Creating commit with APIService and Deployment"
+  git::add \
+      "${YAML_DIR}/namespace.yaml" \
+      acme/namespaces/custom-metrics/namespace.yaml
+  git::add \
+    "${YAML_DIR}/rbac.yaml" \
+    acme/cluster/rbac.yaml
   git::add \
     "${YAML_DIR}/namespace-custom-metrics.yaml" \
     acme/namespaces/custom-metrics/namespace-custom-metrics.yaml
@@ -38,18 +56,27 @@ test_teardown() {
   wait::for -t 240 -- nomos::repo_synced
 }
 
-@test "importer and syncer resilient to bad APIService" {
+@test "importer and syncer resilient to flaky APIService" {
+  # Create an APIService without backend so that Config Sync DiscoveryClient will
+  # get service unavailable errors when trying to parse resource
   debug::log "Adding bad APIService"
   kubectl apply -f "${YAML_DIR}/apiservice.yaml"
 
-  debug::log "Creating commit with Deployment"
+  debug::log "Creating a commit with some test resources"
   git::add \
     "${YAML_DIR}/namespace-resilient.yaml" \
     acme/namespaces/resilient/namespace-resilient.yaml
   git::commit
 
+  debug::log "Waiting for APIServerError to be present in sync status"
+  nomos::source_error_code "2002"
+
+  # Restore the backend for the bad APIservice and see if Config Sync can sync
+  # the previous commit
+  debug::log "Adding APIService backend"
+  kubectl apply -f "${YAML_DIR}/namespace.yaml"
+  kubectl apply -f "${YAML_DIR}/rbac.yaml"
+  kubectl apply -f "${YAML_DIR}/namespace-custom-metrics.yaml"
   debug::log "Waiting for nomos to stabilize"
   wait::for -t 240 -- nomos::repo_synced
-
-  kubectl delete -f "${YAML_DIR}/apiservice.yaml" --ignore-not-found
 }
--- a/e2e/testcases/apiservice_test.go
+++ b/e2e/testcases/apiservice_test.go
@@ -23,7 +23,8 @@ import (
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/pkg/api/configsync"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"kpt.dev/configsync/pkg/status"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
 )
 
 const (
@@ -52,14 +53,14 @@ func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
 		nt.T.Fatalf("validate failed %s", err)
 	}
 
+	// Test done, removing the test APIService first to prevent Discovery failure blocking
+	// the test repo from cleaning up
 	nt.RootRepos[configsync.RootSyncName].Remove("acme/cluster/apiservice.yaml")
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Remove custom metric stackdriver adapter API service")
 	nt.WaitForRepoSyncs()
 
+	// Remove the backend Deployment of test APIService
 	nt.RootRepos[configsync.RootSyncName].Remove("acme/namespaces/custom-metrics/namespace-custom-metrics.yaml")
-	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Remove custom metric stackdriver adapter deployment")
-	nt.WaitForRepoSyncs()
-
 	nt.RootRepos[configsync.RootSyncName].Remove("acme/namespaces/custom-metrics/namespace.yaml")
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Remove custom metric stackdriver adapter namespace")
 	nt.WaitForRepoSyncs()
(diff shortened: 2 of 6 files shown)


e2e/testcases/apiservice.bats
```diff
@@ -13,17 +13,32 @@ YAML_DIR=${BATS_TEST_DIRNAME}/../testdata/apiservice
 test_setup() {
   setup::git::initialize
   setup::git::commit_minimal_repo_contents
+  wait::for -t 120 -- nomos::repo_synced
```
nan-yu, line 16: In the end of `setup::git::commit_minimal_repo_contents`, it waits for the repo to be synced, so the additional wait is unnecessary.

e2e/testcases/apiservice.bats
```diff
@@ -13,17 +13,32 @@ YAML_DIR=${BATS_TEST_DIRNAME}/../testdata/apiservice
...
+kubectl delete -f "${YAML_DIR}/apiservice.yaml" --ignore-not-found
+  git::rm acme/namespaces/custom-metrics/namespace.yaml
+  git::rm acme/namespaces/custom-metrics/namespace-custom-metrics.yaml
+  git::commit
+  kubectl delete -f "${YAML_DIR}/namespace.yaml" --ignore-not-found
+  kubectl delete -f "${YAML_DIR}/namespace-custom-metrics.yaml" --ignore-not-found
+  kubectl delete -f "${YAML_DIR}/rbac.yaml" --ignore-not-found
+
   setup::git::remove_all acme
```
nan-yu, line 31: `setup::git::remove_all acme` deletes all cluster and namespace resources. I don't think you need to delete them explicitly.

e2e/testcases/apiservice.bats
```diff
@@ -13,17 +13,32 @@ YAML_DIR=${BATS_TEST_DIRNAME}/../testdata/apiservice
...
 }
 
 @test "Create APIService and endpoint in same commit" {
+  kubectl apply -f "${YAML_DIR}/namespace.yaml"
+  kubectl apply -f "${YAML_DIR}/rbac.yaml"
```
nan-yu, line 36: Why do you apply them manually using kubectl? This testcase is supposed to test the resources being managed by Config Sync.

e2e/testcases/apiservice.bats
```diff
@@ -13,17 +13,32 @@ YAML_DIR=${BATS_TEST_DIRNAME}/../testdata/apiservice
...
+
   debug::log "Creating commit with APIService and Deployment"
+  git::add \
+      "${YAML_DIR}/namespace.yaml" \
```
nan-yu, line 40: This file is already included in `namespace-custom-metrics.yaml`.

e2e/testcases/apiservice.bats
```diff
@@ -48,8 +63,11 @@ test_teardown() {
     acme/namespaces/resilient/namespace-resilient.yaml
   git::commit
 
+  debug::log "Adding APIService backend"
+  kubectl apply -f "${YAML_DIR}/namespace.yaml"
+  kubectl apply -f "${YAML_DIR}/rbac.yaml"
+  kubectl apply -f "${YAML_DIR}/namespace-custom-metrics.yaml"
```
nan-yu, line 69: What is the intention of applying these resources manually? The purpose of this test case is to verify that config-sync mono repo mode can survive with a bad apiservice.

e2e/testcases/apiservice_test.go
```diff
@@ -37,6 +37,12 @@ func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
 		if t.Failed() {
 			nt.PodLogs(adapterNamespace, adapterName, "pod-custom-metrics-stackdriver-adapter", true)
 		}
+		if !nt.MultiRepo {
+			nt.MustKubectl("delete", "-f", "../testdata/apiservice/apiservice.yaml", "--ignore-not-found")
+			nt.MustKubectl("delete", "-f", "../testdata/apiservice/rbac.yaml", "--ignore-not-found")
+			nt.MustKubectl("delete", "-f", "../testdata/apiservice/namespace-custom-metrics.yaml", "--ignore-not-found")
+			nt.MustKubectl("delete", "-f", "../testdata/apiservice/namespace.yaml", "--ignore-not-found")
+		}
```
nan-yu, line 45: Why not remove them from the repository and let Config Sync prune them?

e2e/testcases/apiservice_test.go
```diff
@@ -57,9 +63,6 @@ func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
 	nt.WaitForRepoSyncs()
 
 	nt.RootRepos[configsync.RootSyncName].Remove("acme/namespaces/custom-metrics/namespace-custom-metrics.yaml")
```
nan-yu, line 65: If you don't plan to commit the change, you should remove `nt.RootRepos.Remove` as well.

nan-yu (changes requested):

e2e/testcases/apiservice.bats
tiffanny29631: Done! Didn't notice that.

e2e/testcases/apiservice.bats
tiffanny29631: Done!

e2e/testcases/apiservice.bats
tiffanny29631: Moved into commit by Config Sync. To not having to breakdown the test data and create more artifacts I'm using unstructured source format for this test bats.

e2e/testcases/apiservice.bats
tiffanny29631: I split it out to bring more flexibility when applying resources.

e2e/testcases/apiservice.bats
tiffanny29631: The change in https://github.com/GoogleContainerTools/kpt-config-sync/pull/28 make errors like 503 blocking, where Config Sync will always retry syncing on bad APIs, thus previous test setup will timeout. Manually applying the deployment a while after APIService is deployed is to mock a temporary flaky API communication.

e2e/testcases/apiservice_test.go
tiffanny29631: Don't know why I added them, they are getting removed from git through Config Sync.

e2e/testcases/apiservice_test.go
tiffanny29631: It is getting committed in the second line from here.

e2e/testcases/apiservice.bats
```diff
@@ -11,19 +11,23 @@ load "../lib/wait"
 YAML_DIR=${BATS_TEST_DIRNAME}/../testdata/apiservice
 
 test_setup() {
+  kubectl apply -f "${MANIFEST_DIR}/source-format_unstructured.yaml"
```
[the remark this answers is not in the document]
tiffanny29631: Added revert to hierarchical mode in test_testdown()

e2e/testcases/apiservice.bats
```diff
@@ -11,19 +11,23 @@ load "../lib/wait"
...
+  nomos::restart_pods
   setup::git::initialize
-  setup::git::commit_minimal_repo_contents
 }
 
 test_teardown() {
-  git::rm acme/cluster/apiservice.yaml
-  git::commit
-  kubectl delete -f "${YAML_DIR}/apiservice.yaml" --ignore-not-found
```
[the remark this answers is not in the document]
tiffanny29631: Resumed

e2e/testcases/apiservice.bats
tiffanny29631: Added comments to explain the manual application and cleanups during test teardown.

e2e/testcases/apiservice.bats
```diff
@@ -41,15 +45,15 @@ test_teardown() {
...
   git::commit
-
+  debug::log "Adding APIService backend"
+  kubectl apply -f "${YAML_DIR}/namespace.yaml"
+  kubectl apply -f "${YAML_DIR}/rbac.yaml"
+  kubectl apply -f "${YAML_DIR}/namespace-custom-metrics.yaml"
   debug::log "Waiting for nomos to stabilize"
   wait::for -t 240 -- nomos::repo_synced
-
-  kubectl delete -f "${YAML_DIR}/apiservice.yaml" --ignore-not-found
```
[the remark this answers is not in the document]
tiffanny29631: Resumed

e2e/testcases/apiservice_test.go
tiffanny29631: Discussed: let Config Sync remove the test APISerivce first so it's not blocked by Discovery failure later when cleanup rest of repo.

e2e/testcases/apiservice.bats
```diff
@@ -38,18 +56,24 @@ test_teardown() {
   wait::for -t 240 -- nomos::repo_synced
 }
 
-@test "importer and syncer resilient to bad APIService" {
+@test "importer and syncer resilient to flaky APIService" {
```
sdowell, line 59: Is the bats test still needed given we have an equivalent go test?

tiffanny29631: So far it's still needed, the mono repo test on GKE clusters are still running bats tests and this is the quickest way to bring CI back to green. We are going to fully deprecate the bats tests in coming Fixit.

e2e/testcases/apiservice.bats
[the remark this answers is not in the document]
tiffanny29631: Also doing cleanup to the manually applied resources here, plus letting Config Sync remove the test APIService first so it's not blocked by discovery failures when trying to cleanup acme.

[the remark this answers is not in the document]
nan-yu: ACK

e2e/testcases/apiservice.bats
```diff
@@ -38,18 +56,24 @@ test_teardown() {
...
+  # Create an APIService without backend so that Config Sync DiscoveryClient will
+  # get service unavailable errors when trying to parse resource
```
nan-yu, line 61: Is the service unavailable error visible to the user? Can we add a validation for that if possible?

If we don't add any validation before restoring the backend, it is basically the same as the first test case.

e2e/testcases/apiservice_test.go
```diff
@@ -52,36 +52,40 @@ func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
...
+		nt.MustKubectl("delete", "-f", "../testdata/apiservice/namespace-custom-metrics.yaml", "--ignore-not-found")
 		if t.Failed() {
 			nt.PodLogs(adapterNamespace, adapterName, "pod-custom-metrics-stackdriver-adapter", true)
 		}
 	})
 
+	nt.T.Log("Adding APIService with backend temporarily unavailable to trigger discovery errors")
+	nt.MustKubectl("apply", "-f", "../testdata/apiservice/apiservice.yaml")
+
 	nt.T.Log("Creating commit with resources")
```
nan-yu, line 82: Similarly as the bats test, can we validate the backend unavailable error?

e2e/testcases/apiservice.bats
```diff
@@ -38,18 +56,27 @@ test_teardown() {
...
   kubectl apply -f "${YAML_DIR}/apiservice.yaml"
 
-  debug::log "Creating commit with Deployment"
+  debug::log "Creating a commit with some test resources"
   git::add \
     "${YAML_DIR}/namespace-resilient.yaml" \
     acme/namespaces/resilient/namespace-resilient.yaml
   git::commit
 
+  debug::log "Waiting for APIServerError to be present in Rootsync status"
```
nan-yu, line 71: nit: The `nomos::source_error_code` function works for both mono-repo and multi-repo. Suggest removing `Rootsync` in the comment as that is specific to the multi-repo mode.

e2e/testcases/apiservice_test.go
```diff
@@ -52,36 +53,48 @@ func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
...
+	nt.T.Log("Creating commit with test resources")
+	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/namespace-resilient.yaml", "acme/namespaces/resilient/namespace.yaml")
+	//nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/apiservice.yaml", "acme/cluster/apiservice.yaml")
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("add testing resources")
 
-	nt.T.Log("Adding backend")
+	nt.T.Log("Waiting for APIServerError to be present in Rootsync status")
+	if !nt.MultiRepo {
+		nt.WaitForRepoSourceError(status.APIServerErrorCode)
+	}
```
nan-yu, line 93: else?

The test runs on both mono-repo mode and multi-repo mode.

e2e/testcases/apiservice_test.go
```diff
@@ -52,36 +53,48 @@ func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
...
 	})
 
-	nt.T.Log("Creating commit with resources")
-	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/rbac.yaml", "acme/cluster/rbac.yaml")
+	nt.T.Log("Adding APIService with backend temporarily unavailable to trigger discovery errors")
+	nt.MustKubectl("apply", "-f", "../testdata/apiservice/apiservice.yaml")
+
+	nt.T.Log("Creating commit with test resources")
+	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/namespace-resilient.yaml", "acme/namespaces/resilient/namespace.yaml")
+	//nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/apiservice.yaml", "acme/cluster/apiservice.yaml")
```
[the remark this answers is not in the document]

e2e/testcases/apiservice.bats
tiffanny29631: Done

e2e/testcases/apiservice_test.go
tiffanny29631: Just confirmed in debugging env that under multirepo mode the test resource can be correctly reconciled even with bad APIService. Added RG status check for the test resource to make sure.

e2e/testcases/apiservice_test.go
```diff
@@ -52,36 +53,62 @@ func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
...
+			err = nt.Validate("resilient", "", &corev1.Namespace{},
+				nomostest.StatusEquals(nt, kstatus.CurrentStatus))
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+		if e != nil {
+			nt.T.Fatal("validate failed test resource to have status CURRENT")
+		}
```
nan-yu, line 103: Just to confirm, if the APIService is unavailable in the multi-repo mode, there won't be any effect to the sync and reconciliation status. Basically, user won't notice any difference. Is that the expected behavior?

tiffanny29631: In this test scenario where APIService is unavailable what you described is correct. I was also working on producing a case where multi-repo mode fails to reconcile due to timeout discovery failure until this bats test starts failing. Will switch back to that part once CI is fixed.

nan-yu: SG

nan-yu (approved):
/lgtm

nan-yu (approved):
/lgtm

PR #53, opened 2022-08-23
Remove spec.helm.valuesFiles field and update spec.helm.period field

xinnywinne:
description

### DIFF
--- a/pkg/api/configsync/v1alpha1/helmconfig.go
+++ b/pkg/api/configsync/v1alpha1/helmconfig.go
@@ -44,11 +44,6 @@ type Helm struct {
 	// +optional
 	Values *unstructured.Unstructured `json:"values,omitempty"`
 
-	// valuesFiles is a list of URL to Helm value files.
-	// Values files must be in the same repository with the Helm chart or in a public repository.
-	// +optional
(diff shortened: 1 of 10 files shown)


pkg/api/configsync/v1alpha1/helmconfig.go
```diff
@@ -44,11 +44,6 @@ type Helm struct {
 	// +optional
 	Values *unstructured.Unstructured `json:"values,omitempty"`
 
-	// valuesFiles is a list of URL to Helm value files.
-	// Values files must be in the same repository with the Helm chart or in a public repository.
-	// +optional
-	ValuesFiles []string `json:"valuesFiles,omitempty"`
```
mikebz, line 50: if this is removed, how do I provide values?

nan-yu (approved):
/lgtm

pkg/api/configsync/v1alpha1/helmconfig.go
xinnywinne: Here are some discussions relate to this field https://github.com/GoogleContainerTools/kpt-config-sync/pull/22#discussion_r946253655. User will provide values through "values" field

tiffanny29631 (approved):
/lgtm

PR #56, opened 2022-08-23
Change helm.values field type from unstructured to JSON

xinnywinne:
This change is to resolve the issue that kubebuilder does not work with
map[string]interface{} and consider unstructured type as Object.

apiextentionsv1.JSON represents any valid JSON value, it supports
map[string]interface{}. After change the field type, user will be able to set the values in this format:
values: {key : val}


xinnywinne:
Already verify this change will work with e2e test. Haven't add error handling logic for unmarshal json, push this PR to get feedback.

tiffanny29631 (approved):
/lgtm

PR #57, opened 2022-08-23
rc tag creation separate from rc build

sdowell:
This change separates the creation of RC tagging from the building of
the RC artifacts. This makes the targets a bit for flexible and enables
the use of tag-based triggers to run the RC build step.

A prompt is added to the RC tagging target with the intent of making it
an interactive step.

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -9,11 +9,9 @@ __tag-release-candidate:
 	CS_VERSION="$(CS_VERSION)" ./scripts/tag-release-candidate.sh
 
 .PHONY: __build-release-candidate
-__build-release-candidate: __tag-release-candidate
+__build-release-candidate:
 	@echo "+++ Building release candidate"
 	@$(eval GCR_PREFIX := "config-management-release")
-	@test -n "$(CS_VERSION)" || (echo "CS_VERSION unset" ; exit 1)
-	$(eval RC_TAG != git tag --points-at "HEAD" --list "$(CS_VERSION)-rc.*" | sort -V | tail -n 1)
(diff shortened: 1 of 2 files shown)


Makefile.release
```diff
@@ -9,11 +9,9 @@ __tag-release-candidate:
 	CS_VERSION="$(CS_VERSION)" ./scripts/tag-release-candidate.sh
 
 .PHONY: __build-release-candidate
-__build-release-candidate: __tag-release-candidate
+__build-release-candidate:
 	@echo "+++ Building release candidate"
 	@$(eval GCR_PREFIX := "config-management-release")
-	@test -n "$(CS_VERSION)" || (echo "CS_VERSION unset" ; exit 1)
-	$(eval RC_TAG != git tag --points-at "HEAD" --list "$(CS_VERSION)-rc.*" | sort -V | tail -n 1)
 	@test -n "$(RC_TAG)" || (echo "RC_TAG unset" ; exit 1)
```
janetkuo, line 15: Is `RC_TAG` is used?

sdowell: Yes it is, if you expand the diff below a little you can see that it is used to tag the images/manifests/binaries (as this target builds RC artifacts).

Before the assumption was that the `CS_VERSION` is provided and this target would invoke the tag target to create the `RC_TAG` using `CS_VERSION`. This PR changes the assumption so that the tag target is invoked separately and that `RC_TAG` is provided to the build target. This makes it such that the build target can be easily run separately from the tag target via a tag-based trigger

xinnywinne (approved):
/lgtm

PR #59, opened 2022-08-24
remove exclusion of acm-psp from oss manifest

sdowell:
This ensures CS works on clusters with PSP enabled. Although PSP is
deprecated, there might still be users for it until we stop supporting
GKE versions older than 1.25.


janetkuo (approved):
/lgtm

PR #60, opened 2022-08-25
ensure branch is correct and up to date for RC tag

sdowell:
Determine the correct branch based on the patch version and
tag the latest commit on that branch for release candidates. Minor
releases are cut from main and patch releases are cut from release
branches. This is intended to minimize user error and make it easier
to tag release candidates without worrying about the local repo state.


karlkfi (approved):
/lgtm

PR #61, opened 2022-08-25
register the status reader for rollout objects

chunglu-chou:
This change utilizes the new functions in cli-utils and kpt to register the status reader for rollout objects in the applier builder.
Note: since it depends on the new version of cli-utils and kpt, this PR can be tested only after the new dependencies are released and imported.


janetkuo:
Closing outdated PR. Please reopen if it's still active.

PR #63, opened 2022-08-25
Validate Config Sync components health in the e2e test

nan-yu:
The resource-group-controller crashed on GKE Autopilot clusters with the
current default memory request. The e2e-test should be able to capture
the issue. This commit adds health validation for CS components in the
e2e test.

### DIFF
--- a/e2e/nomostest/config-sync.go
+++ b/e2e/nomostest/config-sync.go
@@ -213,7 +213,7 @@ func installConfigSync(nt *NT, nomos ntopts.Nomos) {
 // waitForConfigSync validates if the config sync deployment is ready.
 func waitForConfigSync(nt *NT, nomos ntopts.Nomos) error {
 	if nomos.MultiRepo {
-		return validateMultiRepoDeployments(nt)
+		return ValidateMultiRepoDeployments(nt)
 	}
 	return validateMonoRepoDeployments(nt)
 }
@@ -457,7 +457,8 @@ func validateMonoRepoDeployments(nt *NT) error {
 	return nil
 }
 
-func validateMultiRepoDeployments(nt *NT) error {
+// ValidateMultiRepoDeployments validates if all Config Sync Components are available.
+func ValidateMultiRepoDeployments(nt *NT) error {
 	for name := range nt.RootRepos {
 		// Create a RootSync to initialize the root reconciler.
 		rs := RootSyncObjectV1Beta1FromRootRepo(nt, name)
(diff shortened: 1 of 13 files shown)


e2e/nomostest/git-server.go
```diff
@@ -103,6 +103,16 @@ func isAvailableDeployment(o client.Object) error {
 		return fmt.Errorf("%w for deployment/%s in namespace %s: got %d available replicas, want %d\n\n%s",
 			ErrFailedPredicate, d.GetName(), d.GetNamespace(), available, want, string(jsn))
 	}
+	unavailable := d.Status.UnavailableReplicas
```
karlkfi, line 106: Can we just replace this whole isAvailableDeployment with kstatus.Compute? That's what I did in the reconciler-manager recently with `computeDeploymentStatus`.

e2e/nomostest/config-sync.go
```diff
@@ -511,15 +513,60 @@ func validateMultiRepoDeployments(nt *NT) error {
...
+		nt.T.Logf("took %v to wait for %s %s", took, "ValidatingWebhookConfiguration", "admission-webhook.configsync.gke.io")
 	}
-	nt.T.Logf("took %v to wait for %s %s", took, "ValidatingWebhookConfiguration", "admission-webhook.configsync.gke.io")
 
+	return validateMultiRepoPods(nt)
+}
+
+// validateMultiRepoPods validates if all Config Sync Pods are healthy. It doesn't retry
+// because it is supposed to be invoked after validateMultiRepoDeployments succeeds.
+func validateMultiRepoPods(nt *NT) error {
```
karlkfi, line 532: This looks like it replicates what's happening in kstatus StatusReader, which has exceptions for checking Deployment > ReplicaSet > Pod and lifting out extra status details.

e2e/nomostest/git-server.go
nan-yu: DONE

e2e/nomostest/config-sync.go
nan-yu: Do you mean creating a custom StatusReader for the Config Sync Pods? I feel like the validation is specific to the test, e.g., we only check the Pods without DeletionTimestamp that have no container terminated because of OOMKilled.

nan-yu:
This has been incorporated in https://github.com/GoogleContainerTools/kpt-config-sync/pull/65. Closing now

PR #64, opened 2022-08-26
Allow other declared resources in the config-management-system namespace

victorpras:
To vertically scale ACM pods, we want to manage autoscaler resource in the root repository. Right now, only RootSync object can be declared in the config-management-namespace, therefore we want to loosen the restriction in Config Sync by removing ObjectInIllegalNamespace error. This will allow user to sync other resource type in config-management-namespace using unstructured repo config

### DIFF
--- a/e2e/testcases/acme_test.go
+++ b/e2e/testcases/acme_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 	"time"
 
+	"kpt.dev/configsync/pkg/api/configmanagement"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/testing/fake"
 
@@ -30,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/pkg/kinds"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -188,6 +190,27 @@ func TestAcmeCorpRepo(t *testing.T) {
 	nt.RootRepos[configsync.RootSyncName].Add("acme/cluster/test-clusterrole.yaml", fake.ClusterRoleObject())
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Reset the acme directory")
 	nt.WaitForRepoSyncs()
+}
+
+// TestObjectInCMSNamespace will test that user can sync object to CMS namespace
+func TestObjectInCMSNamespace(t *testing.T) {
+	nt := nomostest.New(t, ntopts.SkipMonoRepo, ntopts.Unstructured)
+
+	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/object-in-cms-namespace", "acme")
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("adding resource to config-management-system namespace")
+	nt.WaitForRepoSyncs()
+
+	// validate the resources synced successfully in CMS namespace
+	namespace := configmanagement.ControllerNamespace
+	err := nt.Validate("cms-configmap", namespace, &corev1.ConfigMap{})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	err = nt.Validate("cms-roles", namespace, &rbacv1.Role{})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
 }
 
 func checkResourceCount(nt *nomostest.NT, gvk schema.GroupVersionKind, namespace string, count int, labels, annotations map[string]string) {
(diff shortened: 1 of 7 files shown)


pkg/validate/raw/validate/namespace_validator.go
```diff
@@ -27,7 +27,8 @@ import (
 // to the following rules:
 // - if the object is a Namespace, it must not be `config-management-system`
 // - if the object has a metadata namespace, it must be a valid k8s namespace
-//   and it must not be in `config-management-system`
+//
```
nan-yu, line 30: nit: remove the empty line to keep the diff only relevant to the changes.

nan-yu (commented):
Can you update one of the existing test case to add a resource in the c-m-s namespace to the root repository and confirm the resource can be synced?

pkg/validate/raw/validate/namespace_validator.go
victorpras: this seems to happen automatically when i ran `make fmt-go`, will keep an eye out next time

victorpras:
> Can you update one of the existing test case to add a resource in the c-m-s namespace to the root repository and confirm the resource can be synced?

Added new e2e test instead. Test will copy unstructured repo configs that add Roles and ConfigMap to CMS namespace and validate their existence.

pkg/validate/raw/validate/namespace_validator.go
```diff
@@ -27,7 +27,7 @@ import (
 // to the following rules:
 // - if the object is a Namespace, it must not be `config-management-system`
 // - if the object has a metadata namespace, it must be a valid k8s namespace
-//   and it must not be in `config-management-system`
+// and it must not be in `config-management-system`
```
haiyanmeng, line 30: This should be updated to reflect that "we allow other declared resources in the `config-mangemente-system` namespace`.

e2e/testcases/acme_test.go
```diff
@@ -190,6 +191,17 @@ func TestAcmeCorpRepo(t *testing.T) {
 	nt.WaitForRepoSyncs()
 }
 
+func TestObjectInCMSNamespace(t *testing.T) {
+	nt := nomostest.New(t, ntopts.SkipMonoRepo, ntopts.Unstructured)
+
+	nt.RootRepos[configsync.RootSyncName].Copy("../../examples/object-in-cms-namespace", "acme")
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("adding resource to config-management-system namespace")
+	nt.WaitForRepoSyncs()
+	namespace := "config-management-system"
```
haiyanmeng, line 200: Use `configmanagement.ControllerNamespace` from the `pkg/api/configmangement` pkg instead.

victorpras: changed to use ` configmanagement.ControllerNamespace `

pkg/validate/raw/validate/namespace_validator.go
victorpras: the comment that indicates the restriction has been removed

e2e/testcases/acme_test.go
```diff
@@ -190,6 +192,18 @@ func TestAcmeCorpRepo(t *testing.T) {
...
+// TestObjectInCMSNamespace will test that user can sync object to CMS namespace
+func TestObjectInCMSNamespace(t *testing.T) {
+	nt := nomostest.New(t, ntopts.SkipMonoRepo, ntopts.Unstructured)
+
+	nt.RootRepos[configsync.RootSyncName].Copy("../../examples/object-in-cms-namespace", "acme")
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("adding resource to config-management-system namespace")
+	nt.WaitForRepoSyncs()
+	namespace := configmanagement.ControllerNamespace
+	checkResourceCount(nt, kinds.Role(), namespace, 1, nil, nil)
+	checkResourceCount(nt, kinds.ConfigMap(), namespace, 1, map[string]string{"app.kubernetes.io/managed-by": "configmanagement.gke.io"}, nil)
```
nan-yu, line 204: Instead of checking the resource count, why not checking the particular object? That will ensure the resource is the one you expected.

e2e/testcases/acme_test.go
```diff
@@ -190,6 +192,18 @@ func TestAcmeCorpRepo(t *testing.T) {
 	nt.WaitForRepoSyncs()
 }
 
+// TestObjectInCMSNamespace will test that user can sync object to CMS namespace
+func TestObjectInCMSNamespace(t *testing.T) {
+	nt := nomostest.New(t, ntopts.SkipMonoRepo, ntopts.Unstructured)
+
+	nt.RootRepos[configsync.RootSyncName].Copy("../../examples/object-in-cms-namespace", "acme")
```
nan-yu, line 199: This is actually test data, not example. I would suggest you move to the `e2e/testdata` directory.

sdowell:
I would like to review this change but there is no context in the commit messages or PR description to help me understand the motivation of the change

victorpras:
Added context description

e2e/testcases/acme_test.go
victorpras: validation has been changed to use `nt.validate` directly

e2e/testcases/acme_test.go
victorpras: moved to testdata folder

nan-yu:
@victorpras Can you squash the commits in this PR?

nan-yu (approved):
/lgtm

PR #66, opened 2022-08-26
Update the IsGKEAutopilotCluster to understand the new autopilot webhook

nan-yu:
We use the policycontrollerv2.config.common-webhooks.networking.gke.io webhook to detect if a cluster is an Autopilot cluster. However, the webhook was renamed to gkepolicy.config.common-webhooks.networking.gke.io`.
This commit updates the function to return true if the new webhook is found.

b/243938243

### DIFF
--- a/pkg/util/autopilot.go
+++ b/pkg/util/autopilot.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"strings"
 
+	"go.uber.org/multierr"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/util/autopilot.go
```diff
@@ -49,6 +49,11 @@ var AutopilotManagedKinds = []schema.GroupVersionKind{
 	admissionregistrationv1.SchemeGroupVersion.WithKind("MutatingWebhookConfigurationList"),
 }
 
+var autopilotWebhooks = []string{
+	"policycontrollerv2.config.common-webhooks.networking.gke.io",
+	"gkepolicy.config.common-webhooks.networking.gke.io",
```
haiyanmeng, line 54: Please add some comments here: which one is new and when it is introduced, which one is legacy.

pkg/util/autopilot.go
```diff
@@ -76,15 +81,17 @@ func IsGKEAutopilotCluster(c client.Client) (bool, error) {
...
-	} else if apierrors.IsNotFound(err) {
-		return false, nodesErr
+	for _, webhookName := range autopilotWebhooks {
+		webhook := &admissionregistrationv1.ValidatingWebhookConfiguration{}
+		objectKey := client.ObjectKey{Name: webhookName}
+		err := c.Get(context.Background(), objectKey, webhook)
+		if err == nil {
+			return true, nil
+		} else if !apierrors.IsNotFound(err) {
+			return false, err
```
janetkuo, line 91: The range order would affect the result, if only some of the webhooks are successfully got. We should make the behavior more deterministic.

janetkuo (approved):
/lgtm

pkg/util/autopilot.go
nan-yu: Added a comment about which one is new and which is legacy. I'm still confirming with the autopilot team to see which version was this introduced (waiting for the response). Will add the start version when I get an answer.

pkg/util/autopilot.go
nan-yu: Updated with a multierror to collect the apiserver errors.

PR #67, opened 2022-08-26
fix: Fix index out of bounds error in nomos status command

rquitales:
This PR fixes nomos status to prevent it from panicing by ensuring the slice it references is of the right length.

### DIFF
--- a/cmd/nomos/status/client.go
+++ b/cmd/nomos/status/client.go
@@ -255,10 +255,16 @@ func (c *ClusterClient) multiRepoClusterStatus(ctx context.Context, cs *ClusterS
 		}
 	}
 
-	if len(rootSyncs) != 0 {
+	if len(rootSyncs) != len(rootRGs) {
+		errs = append(errs, fmt.Sprintf("expected the number of RootSyncs and ResourceGroups to be equal, but found %d RootSyncs and %d ResourceGroups", len(rootSyncs), len(rootRGs)))
+	} else if len(rootSyncs) != 0 {
 		var repos []*RepoState
 		for i, rs := range rootSyncs {
 			rg := rootRGs[i]
+			if rg == nil {
+				// We always expect a ResourceGroup, even if we have no managed resources.
+				errs = append(errs, rgNotFoundErrMsg(rootSyncNsAndNames[i].Name, rootSyncNsAndNames[i].Namespace))
+			}
 			repos = append(repos, RootRepoStatus(rs, rg, syncingConditionSupported))
 		}
 		sort.Slice(repos, func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


cmd/nomos/status/client.go
```diff
@@ -279,7 +279,11 @@ func (c *ClusterClient) multiRepoClusterStatus(ctx context.Context, cs *ClusterS
 		}
 	}
 
-	if len(repoSyncs) != 0 {
+	// We need to do an empty slice check on namespaceRGs as well in case there was an error
+	// only getting the resourceGroups. If there were no errors getting both, they are guranteed
+	// to be identical in length due to how we create the namespaceRGs slice by initializing it to
+	// be the same length as repoSyncNsAndNames before populating it.
+	if len(repoSyncs) != 0 && len(namespaceRGs) != 0 {
 		var repos []*RepoState
```
janetkuo, line 287: This is implementation detail in `namespaceRGs` creation and not guaranteed to be always true (someone might change the function and break it in the future). Check that the length are identical and append an error if not, so that we won't hit panic errors.

rquitales: Good idea. Added a check to also ensure both slices are of the same length.

cmd/nomos/status/client.go
```diff
@@ -255,10 +255,16 @@ func (c *ClusterClient) multiRepoClusterStatus(ctx context.Context, cs *ClusterS
...
-	if len(rootSyncs) != 0 {
+	if len(rootSyncs) != len(rootRGs) {
+		errs = append(errs, fmt.Sprintf("expected the number of RootSyncs and ResourceGroups to be equal, but found %d RootSyncs and %d ResourceGroups", len(rootSyncs), len(rootRGs)))
+	} else if len(rootSyncs) != 0 {
 		var repos []*RepoState
 		for i, rs := range rootSyncs {
 			rg := rootRGs[i]
+			if rg == nil {
+				// We always expect a ResourceGroup, even if we have no managed resources.
+				errs = append(errs, fmt.Sprintf("resourcegroups.kpt.dev %q not found", rs.Name))
```
nan-yu, line 266: Can you add namespace to the error so that we know if that's for root-sync or repo-sync, and all other resourcegroups as well?

rquitales: Done.

nan-yu (approved):
Do you want to merge the change into the main branch? 

I added a `/hold` comment to hold the merge.  Feel free to `/unhold` to get it into the current release.

/lgtm
/hold

cmd/nomos/status/client.go
```diff
@@ -255,10 +255,16 @@ func (c *ClusterClient) multiRepoClusterStatus(ctx context.Context, cs *ClusterS
...
 
-	if len(rootSyncs) != 0 {
+	if len(rootSyncs) != len(rootRGs) {
+		errs = append(errs, fmt.Sprintf("expected the number of RootSyncs and ResourceGroups to be equal, but found %d RootSyncs and %d ResourceGroups", len(rootSyncs), len(rootRGs)))
+	} else if len(rootSyncs) != 0 {
 		var repos []*RepoState
 		for i, rs := range rootSyncs {
 			rg := rootRGs[i]
+			if rg == nil {
+				// We always expect a ResourceGroup, even if we have no managed resources.
```
karlkfi, line 265: I imagine it's possible that there's no RG if there was an error before it was created. Will this suppress or expose that kind of error?

rquitales: If there is no RG because of an error before it was created, we should still report that the RG is missing in `nomos status` right? We should always have a ResourceGroup per R*Sync I belive.

rquitales: Previously when nomos CLI used a GET method, we'd always surface errors when we can't find the ResourceGroup. A regression occurred when we updated the status command to LIST all ResourceGroups, thus we didn't surface that the ResourceGroup is missing after that change was merged.

nan-yu (approved):
/lgtm

PR #70, opened 2022-08-29
Change VPA e2e test to use dynamic generation validation

victorpras:
Changed generation validation to be more dynamic instead of hardcoding the expected number. While individual test runs always starts from 1, CI runs use shared environment and generation numbers may have been changed by other test runs


sdowell (changes requested):
Please make the commit message more descriptive. I can infer what was broken by looking at the diff but the commit message should provide context as well

sdowell:
fyi looks like the PR description was updated but not the commit message

victorpras:
Added more description and commit message has been updated

sdowell (approved):
/lgtm

PR #71, opened 2022-08-29
Use the mutating webhook to determine if the cluster is an autopilot cluster

nan-yu:
The validating webhook name is subject to change. A better solution is to use the mutating webhook `workload-defaulter`, which exists and only exists on Autopilot cluster.

b/243938243


sdowell (approved):
/lgtm

PR #73, opened 2022-08-31
Use kubectl to deploy necessary RBAC in mono repo apiservice_test

tiffanny29631:
The mentioned RBAC includes necessary ClusterRole and ClusterRoleBindings
for the backend deployment to work in GKE environment. In Config Sync CI
after each test it will try to reset the test git repo to default in the sharedTestEnv.
Removing this RBAC from git directly is blocked by importer as it detects the 
commit will reduce the count of cluster resource from more than 1 down
to 0.

Thus manually deploying RBAC when test is running in mono repo mode.

### DIFF
--- a/e2e/testcases/apiservice_test.go
+++ b/e2e/testcases/apiservice_test.go
@@ -35,14 +35,24 @@ const (
 func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
 	nt := nomostest.New(t, ntopts.Unstructured, ntopts.RequireGKE(t))
 	t.Cleanup(func() {
+		nt.MustKubectl("delete", "-f", "../testdata/apiservice/rbac.yaml", "--ignore-not-found")
+		nt.MustKubectl("delete", "-f", "../testdata/apiservice/namespace.yaml", "--ignore-not-found")
 		if t.Failed() {
 			nt.PodLogs(adapterNamespace, adapterName, "pod-custom-metrics-stackdriver-adapter", true)
 		}
 	})
 	nt.T.Log("Creating commit with APIService and Deployment")
+	if nt.MultiRepo {
+		nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/rbac.yaml", "acme/cluster/rbac.yaml")
+	} else {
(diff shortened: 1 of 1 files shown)


sdowell (approved):
/lgtm

e2e/testcases/apiservice_test.go
```diff
@@ -35,14 +35,24 @@ const (
...
 		if t.Failed() {
 			nt.PodLogs(adapterNamespace, adapterName, "pod-custom-metrics-stackdriver-adapter", true)
 		}
 	})
 	nt.T.Log("Creating commit with APIService and Deployment")
+	if nt.MultiRepo {
+		nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/rbac.yaml", "acme/cluster/rbac.yaml")
+	} else {
+		// In mono repo mode the necessary ClusterRole and ClusterRolebinding
+		// in rbac.yaml is blocked from being removed from git repo by reconciler in
```
janetkuo, line 49: Should be git importer instead of reconciler

e2e/testcases/apiservice_test.go
```diff
@@ -35,14 +35,24 @@ const (
...
+	if nt.MultiRepo {
+		nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/rbac.yaml", "acme/cluster/rbac.yaml")
+	} else {
+		// In mono repo mode the necessary ClusterRole and ClusterRolebinding
+		// in rbac.yaml is blocked from being removed from git repo by git-importer in
+		// sharedTestEnv, so manually deploying them ahead.
+		nt.MustKubectl("apply", "-f", "../testdata/apiservice/namespace.yaml")
+		nt.MustKubectl("apply", "-f", "../testdata/apiservice/rbac.yaml")
+	}
 	nt.RootRepos[configsync.RootSyncName].Copy("../testdata/apiservice/namespace.yaml", "acme/namespaces/custom-metrics/namespace.yaml")
```
sdowell, line 54: Does the committing of this namespace also need to be moved up to the `if` statement on line 46? It looks like this would get applied twice in monorepo mode (manually applied and committed to git repo)

PR #75, opened 2022-08-31
Update TestHelmNamespaceRepo to make it works in shared test environment

xinnywinne:
Change the RepoSync to sync from original git source at the end of test to make test works in the shared test environment.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -81,11 +81,9 @@ func TestHelmNamespaceRepo(t *testing.T) {
 		Auth:    configsync.AuthNone,
 		Version: publicHelmChartVersion,
 	}}
-	rs.Spec.Git = nil
 	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from a public Helm Chart with cluster-scoped type")
 	nt.WaitForRepoSyncSourceError(repoSyncNN.Namespace, repoSyncNN.Name, nonhierarchical.BadScopeErrCode, "must be Namespace-scoped type")
-
 	nt.T.Log("Fetch password from Secret Manager")
 	key, err := gitproviders.FetchCloudSecret("config-sync-ci-ar-key")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


nan-yu (approved):
/lgtm

e2e/testcases/helm_sync_test.go
```diff
@@ -105,14 +103,16 @@ func TestHelmNamespaceRepo(t *testing.T) {
 		ReleaseName: "test",
 		SecretRef:   v1beta1.SecretReference{Name: "foo"},
```
sdowell, line 104: Consider adding a `Cleanup` block to delete the `foo` Secret as part of the test cleanup

e2e/testcases/helm_sync_test.go
```diff
@@ -105,14 +103,16 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
 	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from a private Helm Chart without cluster scoped resources")
 	nt.WaitForRepoSyncs(nomostest.WithRepoSha1Func(helmChartVersion(privateNSHelmChart+":"+privateNSHelmChartVersion)), nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{repoSyncNN: privateNSHelmChart}))
 	if err := nt.Validate(rs.Spec.Helm.ReleaseName+"-"+privateNSHelmChart, testNs, &appsv1.Deployment{}); err != nil {
 		nt.T.Error(err)
 	}
+	// Change the RepoSync to sync from the original git source to make test works in the shared test environment.
+	rs.Spec.SourceType = string(v1beta1.GitSource)
+	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from the original git repository")
```
sdowell, line 115: Consider putting this snippet in a `Cleanup` block so that test teardown can succeed even if the test fails before this step. It also makes it a bit more self documenting that this snippet is associated with test cleanup

e2e/testcases/helm_sync_test.go
xinnywinne: Probably not needed for RepoSync test, since the repo sync namespace will be deleted and this secret is under that namespace.  But for RootSync test, since the secret is under config-management-system namespace,  It is better that clean up the secret after test finish. Will add this with other helm e2e test changes.

e2e/testcases/helm_sync_test.go
xinnywinne: Sounds good, will make the change in next helm e2e tests PR.

PR #76, opened 2022-09-01
Fix InvalidSourceType error message

xinnywinne:
Add "helm" to error message: must specify spec.sourceType to be one of
"git", "oci", "helm"


tiffanny29631 (approved):
/lgtm

PR #87, opened 2022-09-07
update make targets to support generic registry

sdowell:
This change maintains the default behavior of using GCR if no REGISTRY is provided to avoid breaking changes. If REGISTRY is provided then the full registry prefix will be used. This helps support use cases other than a GCR prefix (e.g. GAR, other registries, etc).

Additionally refactors some make targets to reuse common targets and enable building images/manifests without pushing.


haiyanmeng (approved):

Makefile.build
```diff
@@ -121,10 +121,9 @@ image-nomos: license
 		--build-arg VERSION=${VERSION} \
 		.
 
-
 # Pushes the specified component's docker image to gcr.io.
-push-to-gcr-nomos: image-nomos
```
nan-yu, line 126: I'm curious about the change. Why did we remove the dependency?
The original intent is what we push is what we build.

After the change, if we run `make image-nomos` first and followed by `make push-to-gcr-nomos`, the image URL will be different with different timestamp. We have to manually provide `GCR_PREFIX` with the timestamp of `make image-nomos`, which is not obvious to get.

nan-yu: @sdowell ^^

PR #88, opened 2022-09-07
Fix the test failure for TestResourceConditionAnnotations

nan-yu:
TestResourceConditionAnnotations failed on clusters with K8s 1.22+ because the v1beta1 API for the RBAC resources are deprecated and the k8s-api-server only returns one copy of the RBAC resource.

b/240334486

### DIFF
--- a/e2e/testcases/resource_conditions_test.go
+++ b/e2e/testcases/resource_conditions_test.go
@@ -92,16 +92,29 @@ func TestResourceConditionAnnotations(t *testing.T) {
 	nt.MustKubectl("annotate", "configmap", cmName, "-n", ns,
 		`configmanagement.gke.io/errors=["CrashLoopBackOff"]`)
 
+	support, err := nt.SupportV1Beta1CRDAndRBAC()
+	if err != nil {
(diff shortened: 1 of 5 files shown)


e2e/testcases/resource_conditions_test.go
```diff
@@ -92,16 +92,29 @@ func TestResourceConditionAnnotations(t *testing.T) {
 	nt.MustKubectl("annotate", "configmap", cmName, "-n", ns,
 		`configmanagement.gke.io/errors=["CrashLoopBackOff"]`)
 
+	support, err := nt.SupportV1Beta1CRDAndRBAC()
```
sdowell, line 95: I'm wondering if we should just support one version and skip the test otherwise. Supporting both adds complexity to the assertions which I expect will confuse future people. WDYT?

sdowell (approved):
/lgtm

e2e/testcases/resource_conditions_test.go
nan-yu: We can drop the support for v1beta1 version when we don't need to support k8s 1.22-. Currently, Config Sync supports k8s 1.20 to 1.24.

PR #89, opened 2022-09-08
Add unit test for parseAndUpdate

tiffanny29631:
This test is intended to validate the fix in https://github.com/GoogleContainerTools/kpt-config-sync/pull/28.

When discovery call fails due to any possible reason and some objects in the inventory are with scope Unknown, the parser previously would ignore the error and skip sending these objects to the applier, causing the objects to be pruned in later process. 

The above mentioned fix makes the discovery failure a blocking error so no object with scope Unknown will be skipped, as error will be early returned instead of moving on.

### DIFF
--- a/pkg/parse/root_test.go
+++ b/pkg/parse/root_test.go
@@ -25,8 +25,11 @@ import (
 	"github.com/pkg/errors"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	discovery "k8s.io/client-go/discovery"
+	"k8s.io/utils/pointer"
 	"kpt.dev/configsync/pkg/api/configmanagement"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
@@ -44,6 +47,9 @@ import (
 	syncertest "kpt.dev/configsync/pkg/syncer/syncertest/fake"
 	"kpt.dev/configsync/pkg/testing/fake"
 	"kpt.dev/configsync/pkg/testing/testmetrics"
+	discoveryutil "kpt.dev/configsync/pkg/util/discovery"
+	"sigs.k8s.io/cli-utils/pkg/testutil"
+
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	}
+	return fake.FileObject(crd, "cluster/crd.yaml")
+}
+
+func TestRoot_Parse_Discovery(t *testing.T) {
```
karlkfi, line 495: Can you also add a test that fakes a timeout error from the discovery client? And one that shows how it handles multiple different errors? Maybe a 500 error?

pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
...
+				fake.Unstructured(kinds.Anvil(), core.Name("deploy")),
+			},
+			want: []ast.FileObject{},
+		},
+		{
+			// unknown scoped object does not get skipped when sending go applier
+			name:            "unknown scoped object with discovery failure",
+			format:          filesystem.SourceFormatHierarchy,
+			discoveryClient: syncertest.NewDiscoveryClientWithErrors(true, kinds.Namespace(), kinds.Role()),
+			expectError:     true,
```
karlkfi, line 522: Can you swap these out with an error comparison? That would be better than just a boolean. You can use testutil.AssertEqual with an expected error wrapped with testutil.EqualError to verify both error type and string equivalency, without worrying about exact pointer equivalency.

https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/testutil/matcher.go#L115

haiyanmeng:
Please update the commit message and the PR title to sth more specific.

pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
...
+		},
+		{
+			// happy path condition
+			name:            "known scoped object without discovery failure",
+			format:          filesystem.SourceFormatUnstructured,
+			discoveryClient: syncertest.NewDiscoveryClientWithErrors(false, kinds.Namespace(), kinds.Role()),
+			expectError:     false,
+			parsed: []ast.FileObject{
+				fake.Role(core.Namespace("foo")),
+				fakeCRD(core.Name("anvils.acme.com")),
```
haiyanmeng, line 576: I am a little confused here. The test case name says `known scoped object ...`, but `fakeCRD` defined a CRD whose scope is unknown.

haiyanmeng (changes requested):

pkg/parse/root_test.go
tiffanny29631: Done, added two error types: context deadline exceeded and net/http timeout error. Also added a case with 500 errors. PTAL

pkg/parse/root_test.go
tiffanny29631: Done, using expectedError and doing testutil.AssertEqual with actual parse error.

pkg/syncer/syncertest/fake/discovery_client.go
```diff
@@ -64,6 +83,11 @@ func NewDiscoveryClient(gvks ...schema.GroupVersionKind) discovery.ServerResourc
 		}
 	}
 
+	if wantErrors {
+		return discoveryClient{
+			errors: discoveryError(gvks...),
+		}
+	}
```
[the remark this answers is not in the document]
tiffanny29631: Done, moved to the top for early return.

pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
...
+	testCases := []struct {
+		name            string
+		format          filesystem.SourceFormat
+		parsed          []ast.FileObject
+		want            []ast.FileObject
+		discoveryClient discovery.ServerResourcer
+		expectError     bool
+	}{
+		{
+			// unknown scoped object does not get skipped when sending go applier
```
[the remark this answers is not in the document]
tiffanny29631: Done

pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
...
+			expectError:     true,
+			parsed: []ast.FileObject{
+				fake.Role(core.Namespace("foo")),
+				// add a faked obect in parser.parsed without CRD so it's scope will be unknown when validating
+				fake.Unstructured(kinds.Anvil(), core.Name("deploy")),
+			},
+			want: []ast.FileObject{},
+		},
+		{
+			// unknown scoped object does not get skipped when sending go applier
```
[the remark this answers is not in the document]
tiffanny29631: Done

pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
...
+			name:            "unknown scoped object with discovery failure",
```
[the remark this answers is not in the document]
tiffanny29631: This case is removed as sourceFormat does not affect this test.

pkg/parse/root_test.go
tiffanny29631: The `fakeCRD` generates a CRD with group `acme.com` so that the `fake.Unstructured(kinds.Anvil())` which is also in tc.parsed can have a known scope, previous cases are not using fakeCRD to force the `fake.Unstructured(kinds.Anvil())` to be unknown, then get skipped when doing `splitObjects` in `cache.go`

pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
 	}
 }
 
+func fakeCRD(opts ...core.MetaMutator) ast.FileObject {
```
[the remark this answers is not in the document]
tiffanny29631: See the comment at line 576

pkg/parse/root_test.go
```diff
@@ -451,6 +454,221 @@ func TestRoot_Parse(t *testing.T) {
...
+	}
+
+	converter, err := declared.ValueConverterForTest()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// test against namespace.go!!!
```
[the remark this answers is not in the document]
tiffanny29631: Not for this test, thanks for pointing out.

pkg/syncer/syncertest/fake/discovery_client.go
```diff
@@ -19,28 +19,45 @@ import (
...
 
 // NewDiscoveryClient returns a discoveryClient that reports types available
-// to the API Server.
+// to the API Server with no errors, or ErrGroupDiscoveryFailed error with the
+// specified wantedErr
 //
 // Does not report the scope of each GVK as no tests requiring a discoveryClient
 // use scope information.
-func NewDiscoveryClient(gvks ...schema.GroupVersionKind) discovery.ServerResourcer {
+func NewDiscoveryClient(wantedErr error, gvks ...schema.GroupVersionKind) discoveryutil.ServerResourcer {
```
karlkfi, line 55: nit: Should we make a NewDiscoveryClientWithError constructor so we dont need to use a nil first arg and refactor the old use cases? I don't feel strongly.

pkg/parse/root_test.go
```diff
@@ -451,6 +457,245 @@ func TestRoot_Parse(t *testing.T) {
...
+		parsed          []ast.FileObject
+		want            []ast.FileObject
+		discoveryClient discoveryutil.ServerResourcer
+		expectedError   error
+	}{
+		{
+			// unknown scoped object should not be skipped when sending to applier when discovery call fails
+			name:            "unknown scoped object with discovery failure of deadline exceeded error",
+			discoveryClient: syncertest.NewDiscoveryClient(context.DeadlineExceeded, kinds.Namespace(), kinds.Role()),
+			expectedError:   fakeParseError(context.DeadlineExceeded, kinds.Namespace(), kinds.Role()),
```
karlkfi, line 534: Q: Is context.DeadlineExceeded actually the error that gets returned by the real discovery client? It seems like that would be a client-side error, not a server-side error, but I'm not sure. Maybe both use it. Do you know where in the code this is being returned? Maybe add a comment linking to it?

pkg/syncer/syncertest/fake/discovery_client.go
tiffanny29631: I was doing that previously, a wrapper over the main constructor, I can revert it so the other callers remains as they were.

pkg/parse/root_test.go
tiffanny29631: I've seem both errors from Spotify's log, and all of them are folded in the same discovery error, then folded in the same APIServerError

pkg/syncer/syncertest/fake/discovery_client.go
karlkfi: Seems cleaner to make the error argument optional, yeah.

tiffanny29631: Done

karlkfi (approved):
/lgtm
/unhold

PR #90, opened 2022-09-08
Increase the `parallel` flag of e2e presubmit jobs

haiyanmeng:
Increase the `parallel` flag from 18 to 24 to reduce the excution time of the `kpt-config-sync-presubmit-e2e-multi-repo` job, which runs in the `large-job-pool` and each node in the pool has 30 vCPUs.

Each vCPU corresponds to a hardware thread rather than a core, this PR is to figure out whether setting the `parallel` flag to 24 could reduce the execution time.

### DIFF
--- a/Makefile.e2e
+++ b/Makefile.e2e
@@ -170,10 +170,10 @@ test-e2e-go-ephemeral: __push-local-image
 	@docker run -v /var/run/docker.sock:/var/run/docker.sock --network="host" prow-image make __test-e2e-go-nobuild-prow
 test-e2e-go-ephemeral-mono-repo: __push-local-image
 	kind delete clusters --all
-	./scripts/e2e-go-ephemeral.sh --timeout 60m --parallel 18 --image-tag=$(LATEST_IMAGE_TAG)
+	./scripts/e2e-go-ephemeral.sh --timeout 60m --parallel 23 --image-tag=$(LATEST_IMAGE_TAG) --test.run TestIncreaseParallel*
 test-e2e-go-ephemeral-multi-repo: __push-local-image
 	kind delete clusters --all
-	./scripts/e2e-go-ephemeral.sh --multirepo --timeout 60m --parallel 18 --image-tag=$(LATEST_IMAGE_TAG)
+	./scripts/e2e-go-ephemeral.sh --multirepo --timeout 60m --parallel 23 --image-tag=$(LATEST_IMAGE_TAG) --test.run TestIncreaseParallel*
 
 # Runs the Go e2e tests against multi-repo Config Sync.
 # Rebuilds the nomos image, so use this when you're actively modifying the Nomos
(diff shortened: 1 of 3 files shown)


haiyanmeng:
> Perhaps we can create another PR with no changes as a baseline comparison?

We can compare the performance based on the presubmit job history: https://oss.gprow.dev/job-history/gs/oss-prow-build-kpt-config-sync/pr-logs/directory/kpt-config-sync-presubmit-e2e-multi-repo

haiyanmeng:
Setting the `parallel` flag to 24 causes lots of tests to fail due to the lack of resources. Here are the errors:
```
new.go:416: ERROR: waiting for ConfigSync Deployments to become available: 2 error(s)
        
        
        [1] KNV9999: deployments.apps "root-reconciler" not found
        
        For more information, see https://g.co/cloud/acm-errors#knv9999
        
        
        [2] KNV9999: failed predicate for deployment/admission-webhook in namespace config-management-system: got status Failed, want Current

        {
          "metadata": {
            "name": "admission-webhook",
            "namespace": "config-management-system",
          ...
          "status": {
            "observedGeneration": 1,
            "replicas": 2,
            "updatedReplicas": 2,
            "unavailableReplicas": 2,
            "conditions": [
              {
                "type": "Available",
                "status": "False",
                "lastUpdateTime": "2022-09-08T14:04:40Z",
                "lastTransitionTime": "2022-09-08T14:04:40Z",
                "reason": "MinimumReplicasUnavailable",
                "message": "Deployment does not have minimum availability."
              },
              {
                "type": "Progressing",
                "status": "False",
                "lastUpdateTime": "2022-09-08T14:14:41Z",
                "lastTransitionTime": "2022-09-08T14:14:41Z",
                "reason": "ProgressDeadlineExceeded",
                "message": "ReplicaSet \"admission-webhook-5c79b59f86\" has timed out progressing."
              }
            ]
          }
        }
        
        For more information, see https://g.co/cloud/acm-errors#knv9999
```

sdowell:
> > Perhaps we can create another PR with no changes as a baseline comparison?
> 
> We can compare the performance based on the presubmit job history: https://oss.gprow.dev/job-history/gs/oss-prow-build-kpt-config-sync/pr-logs/directory/kpt-config-sync-presubmit-e2e-multi-repo

@haiyanmeng Only issue I see with looking at the job history is that any given job could include changes to the job itself (more tests, functional changes, etc). I suggested created a PR from the same base as this one so that we can confident we have an accurate control when comparing the difference in performance

sdowell:
@haiyanmeng I suspect the second optional is the viable option if we really want to start pushing the scale of the test parallelism. My only concern is that we are already pretty aggressive on the hardware requirements to run the presubmit tests. I'm not sure how big of a concern that cost is to us, but this direction does increase test infra costs. This becomes a greater concern if we want to start scaling horizontally to support more consecutive presubmit jobs.

I think it may warrant taking a step back and asking whether we want to throw more hardware at the problem

mikebz:
should we just close this?

PR #92, opened 2022-09-08
Fix mono-repo "failed to create object" error

karlkfi:
- Rename Update to Apply to reflect existing client-side apply behavior
- Add an Update method to update directly (and send metrics) when the Get has already been performed
- Fix "failed to create object" error in the mono-repo finalizer by switching from Upsert to Apply
- Remove unused Upsert function
- Add additional equivelency checks to avoid updates when using Apply and ApplyStatus

### DIFF
--- a/pkg/syncer/client/client.go
+++ b/pkg/syncer/client/client.go
@@ -26,7 +26,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	m "kpt.dev/configsync/pkg/metrics"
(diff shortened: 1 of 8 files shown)


sdowell (approved):
The changes I see here make sense to me, but I don't have much context on the workings of monorepo. `LGTM` but adding `hold` label in case you want someone with more context to take a look.

/lgtm
/hold

pkg/syncer/client/client.go
```diff
@@ -197,33 +202,28 @@ func (c *Client) update(ctx context.Context, obj client.Object, updateFn update,
...
-		if apierrors.IsNotFound(err) {
-			return c.Create(ctx, obj)
-		}
-		return status.ResourceWrap(err, "failed to get resource for update", obj)
-	}
-
 	klog.V(1).Infof("Will update %s to %s", description, spew.Sdump(obj))
+	oldV := resourceVersion(obj)
 	start := time.Now()
 	err := c.Client.Update(ctx, obj)
```
nan-yu, line 212: `c.Client.Apply` after the rename? Otherwise, it calls itself recursively.

pkg/syncer/client/client.go
```diff
@@ -197,33 +202,28 @@ func (c *Client) update(ctx context.Context, obj client.Object, updateFn update,
...
 	c.recordLatency(start, "update", obj.GetObjectKind().GroupVersionKind().Kind, metrics.StatusLabel(err))
 	m.RecordAPICallDuration(ctx, "update", m.StatusTagKey(err), obj.GetObjectKind().GroupVersionKind(), start)
-
-	if err != nil {
+	switch {
+	case apierrors.IsNotFound(err):
+		return ConflictUpdateDoesNotExist(err, obj)
```
nan-yu, line 217: `c.Client.Apply` returns `ConflictUpdateDoesNotExist` error directly, no need to wrap it again.

pkg/syncer/sync/reconciler.go
```diff
@@ -188,29 +192,22 @@ func (r *metaReconciler) reconcileSyncs(ctx context.Context, request reconcile.R
...
 	}
 
-	sync = sync.DeepCopy()
-	sync.Finalizers = newFinalizers
 	klog.Infof("beginning Sync finalize for %s", sync.Name)
 	if err := r.gcResources(ctx, sync, gvks); err != nil {
 		return err
 	}
-	err := r.client.Upsert(ctx, sync)
+	_, err := r.client.Apply(ctx, sync, func(obj client.Object) (client.Object, error) {
```
nan-yu, line 205: Should this be `Apply` or `Update`?

haiyanmeng (approved):

pkg/syncer/client/client.go
karlkfi: Nah, this is the real k8s client, not syncer.Client. I only renamed the syncer package functions.

pkg/syncer/sync/reconciler.go
karlkfi: I changed it to an Apply because we don't actually care about the original state of the Sync. All we want to do is remove the finalizer, if it still exists. Plus, since the Sync was last Get before a bunch of Deletes in gcResources, we want to do an Apply instead of just an Update to avoid conflicting on any sort of drift that's happened since the Get.

pkg/syncer/client/client.go
karlkfi: There is no c.Client.Apply. You're confusing the upstream client library with the syncer package functions.

nan-yu: Ahh, I see. That's the k8s client.

pkg/syncer/sync/reconciler.go
nan-yu: That makes sense.

nan-yu (approved):
/lgtm

PR #93, opened 2022-09-08
Add e2e tests for Helm support

xinnywinne:
Add more e2e tests to verify `gcenode` and `gcpserviceaccount` auth types work with Artifact Registry.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -23,6 +23,7 @@ import (
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/importer/analyzer/validation/nonhierarchical"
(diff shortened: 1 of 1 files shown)


e2e/testcases/helm_sync_test.go
```diff
@@ -110,9 +111,11 @@ func TestHelmNamespaceRepo(t *testing.T) {
 		nt.T.Error(err)
 	}
 	// Change the RepoSync to sync from the original git source to make test works in the shared test environment.
-	rs.Spec.SourceType = string(v1beta1.GitSource)
-	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
-	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from the original git repository")
+	nt.T.Cleanup(func() {
```
sdowell, line 114: The cleanup should be declared earlier in the test before a failure might occur. In this case it should probably be done when the `SourceType` is changed to use helm

xinnywinne: Sounds good, done.

e2e/testcases/helm_sync_test.go
```diff
@@ -172,6 +259,7 @@ func TestHelmARTokenAuth(t *testing.T) {
 		// Change the rs back so that it works in the shared test environment.
 		nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": null, "git": {"dir": "acme", "branch": "main", "repo": "%s", "auth": "ssh","gcpServiceAccountEmail": "", "secretRef": {"name": "git-creds"}}}}`,
 			v1beta1.GitSource, origRepoURL))
+		nt.MustKubectl("delete", "secret", "foo", "-n", v1.NSConfigManagementSystem, "--ignore-not-found")
```
sdowell, line 262: This can be moved to its own cleanup block that is declared above when the secret is created

xinnywinne: Done!

PR #96, opened 2022-09-09
[WIP] Add resource attributes

karlkfi:
This is still WIP, but demonstrates how to add resource attributes at runtime using metric tags that are converted to attributes by the otel agent sidecar before sending to the otel collector. The otel collector can then convert them back into tags for the prometheus exporter, because prometheus doesn't support attributes.

This allows us to use convert high-cardinality tags into attributes for GCP monitoring and Google monarch, while still being able to test them using prometheus.

We will need to decide, however, if we want to keep the tag to attribute conversion on by default or whether we just enable it for tests and provide docs or a flag for customers to turn them on if desired.

I've only made the change to the reconciler, but the more important change would be to do the same to the reconciler-manager and the resource-group-controller, because they are 1-to-Many when it comes to reconcilers and sync objects.

I added a log to the e2e tests to print the metrics as yaml for debugging.

### DIFF
--- a/e2e/nomostest/metrics/parse.go
+++ b/e2e/nomostest/metrics/parse.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	"go.opencensus.io/tag"
 )
 
 // ParseMetrics connects to the local port where metrics are being forwarded to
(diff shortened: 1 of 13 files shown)


karlkfi:
/hold

Not ready for review but I want tests to run.

karlkfi:
obsolete

PR #98, opened 2022-09-09
Fail test during reset if sync still in-progress

karlkfi:
This should prevent mono-repo test suite timeout when a test wedges the syncer, causing all subsequent tests to fail.


nan-yu (approved):
/lgtm

PR #101, opened 2022-09-12
Split the sync ordering e2e test into several shorter ones

haiyanmeng:


nan-yu (commented):
Looks good. Just want to confirm how long does each test take?

haiyanmeng:
> Looks good. Just want to confirm how long does each test take?

--- PASS: TestSyncOrdering_CM_Error (359.35s)
--- PASS: TestSyncOrdering_CM_HappyPath (387.70s)
--- PASS: TestSyncOrdering_Pod (464.56s)

e2e/testcases/sync_ordering_test.go
```diff
@@ -35,7 +35,7 @@ import (
 // This file includes e2e tests for sync ordering.
 // The sync ordering feature is only supported in the multi-repo mode.
 
-func TestSyncOrdering(t *testing.T) {
+func TestSyncOrdering_CM_HappyPath(t *testing.T) {
```
karlkfi, line 38: Not a fan of HappyPath as a name. Can you make it more meaningful? Like TestMultipleDependencies?

Also not sure why you added underscores. Most of the other test names just use camel case, right?

e2e/testcases/sync_ordering_test.go
```diff
@@ -308,6 +308,17 @@ func TestSyncOrdering(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
+}
+
+func TestSyncOrdering_CM_Error(t *testing.T) {
```
karlkfi, line 313: TestExternalDependencyError

e2e/testcases/sync_ordering_test.go
```diff
@@ -391,6 +402,14 @@ func TestSyncOrdering(t *testing.T) {
 	if err := nt.ValidateNotFound("cm4", namespaceName, &corev1.ConfigMap{}); err != nil {
 		nt.T.Fatal(err)
 	}
+}
+
+func TestSyncOrdering_Pod(t *testing.T) {
```
karlkfi, line 407: TestDependencyWithReconciliation

karlkfi (changes requested):
naming suggestions

nan-yu:
I think 5-7 minutes is okay. FYI, from my last analysis, there're 11 tests taking longer than 5 minutes, and we are planning to reduce the running time of the top 5 first.

nan-yu (approved):
/lgtm

karlkfi:
The longest tests are the ones that fail and wait 6m for sync timeout, or worse, use multiple long timeouts with Error instead of Fatal. Personally, I'm more concerned about those than the ones that run long on success. But this is a good example of a test that is way too long even when it succeeds. So it was a good candidate to break up.

I don't think we need a definitive rule to follow. We can just continue fixing the longest ones and making them shorter. 

It might be nice to know what the average & median times are tho, so we can target that as a guideline when shortening the long ones.

e2e/testcases/sync_ordering_test.go
haiyanmeng: We do have a dozen of tests using underscores like `TestConflictingDefinitions_NamespaceToNamespace`. It is not consistent. 
I renamed it to TestMultipleDependencies.

karlkfi (approved):
/lgtm

PR #102, opened 2022-09-12
Remove the ResourceGroup option that is not used

nan-yu:
The ResourceGroup option is not used anywhere. The option is supposed to be used to determine whether to install the resource-group-controller. However, in the installation function, it uses nt.MultiRepo to determine whether to install the resource-group-controller:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.13.0-rc.7/e2e/nomostest/config-sync.go#L184.

### DIFF
--- a/e2e/nomostest/ntopts/multi_repo.go
+++ b/e2e/nomostest/ntopts/multi_repo.go
@@ -54,9 +54,6 @@ type MultiRepo struct {
 	// SkipMonoRepo will skip the test if run in mono repo mode.
 	SkipMonoRepo bool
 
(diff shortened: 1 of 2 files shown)


sdowell (approved):
/lgtm

nan-yu:
Got the `There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors` error from the presubmit jobs.

/retest

sdowell:
> Got the `There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors` error from the presubmit jobs.

@haiyanmeng FYI

PR #105, opened 2022-09-13
Use opencensus based exporter for monitoring

rquitales:
When we upgraded to otel v0.58, we introduced a regression in the metrics being exported with a `generic_node` type instead of `k8s_pod`. This happened as our codebase uses opencensus semantic conventions for resource attributes instead of opentelemetry semantic conventions.

This workaround runs our collector with a flag to disable the newer exporter logic until we can upgrade our codebase to use opentelemetry conventions.

See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/breaking-changes.md for more details about the rewritten exporter.

### DIFF
--- a/e2e/testdata/reconciler-manager-configmap-updated.yaml
+++ b/e2e/testdata/reconciler-manager-configmap-updated.yaml
@@ -192,6 +192,9 @@ data:
           - /otelcol-contrib
           args:
           - "--config=/conf/otel-agent-config.yaml"
+          # TODO: Remove this feature gate when opentelemetry semantic conventions are used
+          # in the collector code.
+          - "--feature-gates=-exporter.googlecloud.OTLPDirect"
           resources:
             limits:
               cpu: 500m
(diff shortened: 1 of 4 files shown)


rquitales:
This is a workaround to revert to the older logic for our GCP metrics exporter. Tested that metrics were correctly exported with this change using a GKE cluster.

e2e/testdata/reconciler-manager-configmap-updated.yaml
```diff
@@ -192,6 +192,7 @@ data:
           - /otelcol-contrib
           args:
           - "--config=/conf/otel-agent-config.yaml"
+          - "--feature-gates=-exporter.googlecloud.OTLPDirect"
```
karlkfi, line 195: Please add an inline comment as to why this flag does and why we need it. Or at least a TODO to remove it when the codebase is upgraded.

tiffanny29631 (commented):
Were you able to test this change and verify that the metrics are back to `k8s_pod`?

rquitales:
> Were you able to test this change and verify that the metrics are back to `k8s_pod`?

Yes, I've verified that metrics are back to `k8s_pod` with this change.

tiffanny29631 (approved):
/lgtm

PR #108, opened 2022-09-15
Split e2e tests into multiple test groups by features

nan-yu:
This allows the tests to be divided into separate groups and all groups to run in parallel.

### DIFF
--- a/Makefile.e2e.ci
+++ b/Makefile.e2e.ci
@@ -134,3 +134,58 @@ manual-rotate-prober-sa-key: __download_prober_creds
 # Manually rotate the prober's service account key in OSS prow.
 manual-rotate-prober-sa-key-oss: __download_prober_creds
 	GCP_PROBER_CREDS=$(SA_KEY_FILE) ./scripts/oss-prow/prober-runner-sa-key-rotate.sh
+
+# The CI target that runs the first group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group1:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=acm-controller,nomos-cli,override-api" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the second group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group2:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo" --test-features=cluster-selector,hydration" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the third group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group3:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=drift-control,lifecycle" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the fourth group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group4:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=multi-repos" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the fifth group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group5:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=reconciliation1" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the sixth group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group6:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=reconciliation2" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the seventh group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group7:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=sync-source" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the eighth group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group8:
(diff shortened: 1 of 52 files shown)


e2e/nomostest/testing/wrapper.go
```diff
@@ -14,18 +14,70 @@
...
+		features := strings.Split(testFeaturesFlag, ",")
+		for _, f := range features {
+			sanitizedFeature := strings.TrimSpace(f)
+			if !KnownFeature(Feature(sanitizedFeature)) {
+				t.Fatalf("Test failed because the test-features flag has an unknown feature %q", sanitizedFeature)
+			} else if Feature(f) == testFeature {
+				skip = false
+			}
+		}
+	}
```
haiyanmeng, line 76: Considering both `New` and `NewShared` checks whether the `TestFeatures` flag includes any unknown features,let us move the logic of checking  the `TestFeatures` flag into `TestMain`. 

We can move the logic of checking whether `testFeature` is supported into the `New` function in`e2e/nomostest/new.go`.

After these two changes, we no longer need both `New` and `NewShared` in this file, can just keep the `func New(t NTB) *Wrapper {` func.

haiyanmeng:
@nan-yu , I left one comment for the change. Otherwise, it looks good to me.

Makefile.e2e.ci
```diff
@@ -134,3 +134,58 @@ manual-rotate-prober-sa-key: __download_prober_creds
...
+		test-e2e-go-gke-ci
+
+# The CI target that runs the fourth group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group4:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=multi-repos" \
+		test-e2e-go-gke-ci
+
+# The CI target that runs the fifth group of tests with the multi-repo mode on GKE cluster.
+test-e2e-gke-multi-repo-test-group5:
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --share-test-env --multirepo --test-features=reconciliation1" \
```
haiyanmeng, line 160: Can we use an environment variable somehow to track the features to test? If we do that, we won't need these `test-e2e-gke-multi-repo-test-groupXX` make rules.

nan-yu: Yes, we can. But then we will configure the env variable in the Prow job. So the grouping logic is outside the project. 

The purpose of creating separate targets explicitly here is to keep the grouping logic inside the `kpt-config-sync` project and the prow jobs only need to invoke corresponding targets.

It will be easier to rebalance the testing groups within the same project in the future.

e2e/nomostest/testing/wrapper.go
haiyanmeng: > We have to split *e2e.TestFeatures into an array and iterate through the array to search for testFeature. If we move it to TestMain, the logic has to repeat twice.

You can split *e2e.TestFeatures into an array once and track the array somehow. Then the `New` function in `e2e/nomostest/new.go` can reuse the array instead of repeating the logic.

nan-yu: That requires us to have a global variable. Even with the global variable, we still need to iterate through the array twice, once in `TestMain` and once in `New`. Why don't we have two separate functions, but have the logic executed only once?

Makefile.e2e.ci
nan-yu: With the multiple similar but slightly different targets, we can have the following benefits:
- Keep the grouping logic self-contained within the kpt-config-sync project.
- It is easier to run a particular test group locally by invoking the target directly, instead of specifying the features.
- It is in line with the nomos-operator settings. The operator has multiple similar targets for different platforms.

For presubmit jobs, we can have separate targets.

e2e/nomostest/testing/wrapper.go
haiyanmeng: To me, the logic in these two functions duplicate a lot, which is why I made the recommendation: https://github.com/GoogleContainerTools/kpt-config-sync/pull/108#discussion_r973513316.

nan-yu: I'm not a big fan of using global variable to track an internal value across packages, but like you said, it's a personal preference. If you insist, I'll update this PR.

nan-yu: Just notice `New` and `NewShared` cannot be merged into a single function, because `NewShared` doesn't take `testFeature` as an argument, but `New` does.

nan-yu: If we go with [comment](https://github.com/GoogleContainerTools/kpt-config-sync/pull/108#discussion_r973513316) , the changes include:
1. Declare a global variable in `e2e/nomostest/testing/feature.go`: `var TestFeatureArray []Feature`. The variable has to be declared here to avoid cyclic import, but it is intialized in `TestMain`, a different package.
2. Check `e2e.TestFeatures` in `TestMain`:
```
	testFeaturesFlag := *e2e.TestFeatures
	if testFeaturesFlag != "" {
		features := strings.Split(testFeaturesFlag, ",")
		for _, f := range features {
			sanitizedFeature := strings.TrimSpace(f)
			nomostesting.TestFeatureArray = append(nomostesting.TestFeatureArray, nomostesting.Feature(sanitizedFeature))
			if !nomostesting.KnownFeature(nomostesting.Feature(sanitizedFeature)) {
				fmt.Printf("Error: Test failed because the test-features flag has an unknown feature %q\n", sanitizedFeature)
				os.Exit(1)
			}
		}
	}
```
3. Keep both `NewShared` and `New` functions:
```
// NewShared creates a new instance of the Testing Wrapper that provides additional
// functionality beyond the standard testing package.
// It doesn't require the testFeature param.
func NewShared(t NTB) *Wrapper {
	t.Helper()

	tw := &Wrapper{
		t: t,
	}
	return tw
}

// New creates a new instance of the Testing Wrapper that provides additional
// functionality beyond the standard testing package.
func New(t NTB, testFeature Feature) *Wrapper {
	t.Helper()

	if !KnownFeature(testFeature) {
		t.Fatalf("Test failed because the feature %q is unknown", testFeature)
	}

	skip := true
	for _, f := range TestFeatureArray {
		if f == testFeature {
			skip = false
		}
	}
	if skip {
		t.Skipf("Skip the test because the feature %q is not included in the test-features flag %q", testFeature, *e2e.TestFeatures)
	}

	tw := &Wrapper{
		t: t,
	}

	return tw
}
```

From simplicity's perspective, I couldn't see too much gain, but the check for allowed features spreads in multiple places.

PR #109, opened 2022-09-16
Fix helm.values field to support mutiple values

xinnywinne:
helm.values field type used to be map[string]interface{}. Looping through the map to generate a env variable leads to a inconsistent env variable issue. Since we change the helm.values field type to JSON already, we can pass it through env variable directly, and process it later in helm-sync.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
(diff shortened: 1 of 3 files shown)


pkg/helm/helm.go
```diff
@@ -80,6 +84,18 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
 	return args, nil
 }
 
+func (h *Hydrator) appendValuesArgs(args []string) ([]string, error) {
+	var values map[string]interface{}
```
sdowell, line 88: For the `values` key/value pairs, what type(s) can the value be? I expected it would only be `string`

e2e/testcases/helm_sync_test.go
```diff
@@ -266,7 +269,7 @@ func TestHelmARTokenAuth(t *testing.T) {
 	nt.WaitForRepoSyncs(nomostest.WithRootSha1Func(helmChartVersion(privateHelmChart+":"+privateHelmChartVersion)),
 		nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: privateHelmChart}))
 	if err := nt.Validate("my-coredns-coredns", "coredns", &appsv1.Deployment{},
-		containerImagePullPolicy("IfNotPresent")); err != nil {
+		containerImagePullPolicy("Always"), nomostest.HasCorrectResourceRequestsLimits("coredns", resource.MustParse("150m"), resource.MustParse("200m"), resource.MustParse("250Mi"), resource.MustParse("300Mi"))); err != nil {
```
sdowell, line 272: Is this change related?

e2e/testcases/helm_sync_test.go
```diff
@@ -256,7 +257,9 @@ func TestHelmARTokenAuth(t *testing.T) {
 		nt.MustKubectl("delete", "secret", "foo", "-n", v1.NSConfigManagementSystem, "--ignore-not-found")
 	})
 	nt.T.Log("Update RootSync to sync from a private Artifact Registry")
-	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": {"repo": "%s", "chart": "%s", "auth": "token", "version": "%s", "releaseName": "my-coredns", "namespace": "coredns", "secretRef": {"name" : "foo"}}, "git": null}}`,
+	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "git": null, "helm": {"repo": "%s", "chart": "%s", "auth": "token", "version": "%s", "releaseName": "my-coredns", "namespace": "coredns",
+                                                                                                "values": {"image.pullPolicy": "Always", "resources.requests.memory": "250Mi", "resources.requests.cpu": "150m", "resources.limits.memory": "300Mi", "resources.limits.cpu": "200m"},
```
nan-yu, line 261: Suggest replacing `"200m"` with `1` (string -> integer) to test different value types.

pkg/helm/helm.go
xinnywinne: value can be different types, string, integer, bool ...

e2e/testcases/helm_sync_test.go
xinnywinne: Sounds good, Done!

e2e/testcases/helm_sync_test.go
```diff
@@ -237,7 +238,7 @@ func TestHelmARTokenAuth(t *testing.T) {
 	nt := nomostest.New(t,
 		ntopts.SkipMonoRepo,
 		ntopts.Unstructured,
-		ntopts.RequireGKE(t),
+		//ntopts.RequireGKE(t),
```
nan-yu, line 241: Is this test required to run on GKE cluster? If not, remove the comment.

nan-yu (commented):
lgtm with one minor comment.

e2e/testcases/helm_sync_test.go
xinnywinne: done.

nan-yu (approved):
/lgtm

PR #111, opened 2022-09-16
Make `setupTestCase` faster

haiyanmeng:


e2e/nomostest/config-sync.go
```diff
@@ -524,7 +524,7 @@ func ValidateMultiRepoDeployments(nt *NT) error {
 		nt.T.Logf("took %v to wait for %s %s", took, "ValidatingWebhookConfiguration", "admission-webhook.configsync.gke.io")
 	}
 
-	return validateMultiRepoPods(nt)
```
nan-yu, line 527: We need to validate the Pod status because it is possible that the Deployment is validated, but the Pod is stuck in a crashloopbackoff state.

e2e/nomostest/nt.go
```diff
@@ -579,9 +579,6 @@ func (nt *NT) WaitForRepoSyncs(options ...WaitForRepoSyncsOption) {
 	syncTimeout := waitForRepoSyncsOptions.timeout
 
 	if nt.MultiRepo {
-		if err := ValidateMultiRepoDeployments(nt); err != nil {
-			nt.T.Fatal(err)
-		}
```
nan-yu, line 584: The validation here guarantees no Pod is in a CrashLoopBackoff state because of OOMKilled.

haiyanmeng: ACKed. Skipping the valiation here may make the tests more flaky.
I am trying to see how much time skipping this validation here can save, and how much worse it makes the flakiness. 
Mostly likely we won't need this change after https://github.com/GoogleContainerTools/kpt-config-sync/pull/116 is merged.

e2e/nomostest/config-sync.go
haiyanmeng: Same comment as https://github.com/GoogleContainerTools/kpt-config-sync/pull/111#discussion_r974657873.

haiyanmeng:
No longer needed.

PR #113, opened 2022-09-16
Fix helm.values field to support mutiple values (#109)

xinnywinne:
helm.values field type used to be map[string]interface{}. Looping through the map to generate a env variable leads to a inconsistent env variable issue. Since we change the helm.values field type to JSON already, we can pass it through env variable directly, and process it later in helm-sync.


nan-yu (approved):
/lgtm

PR #100, opened 2022-09-10
Fix metrics validation

karlkfi:
Fix metrics validation

- Enable resource_to_telemetry_conversion on the otel-collector.
- Use the k8s.deployment.name resource attribute to filter and
  validate metrics from specific reconcilers.
- Add a safety ClusterRole to e2e tests.
  This cluster-scoped resource is required to allow deletion of
  all other cluster-scoped resources in mono-repo mode.
  This allows for generic cleanup code, rather than needing each
  test to work around the deletion validation code.
- Rewrite e2e validation of the declared_resources metric to
  dynamically account for default resources, whether the test is
  using centralized or delegated setup.
- Fix a race condition in the TestAddUpdateRemoveClusterScopedCRDV1
  tests when deleting both the CRD and CR in the same commit.
- Shorten the metrics wait timeout to 1m from 6m.

Depends on:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/114
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/115
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/117
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/118
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/119
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/120

PR #114, opened 2022-09-19
Add GCP+GKE metric attributes

karlkfi:
- Enable the GCP resource processor on the opentelemetry agent. This adds the following resource attributes:
  * cloud.provider
  * cloud.platform
  * cloud.account.id
  * cloud.region OR cloud.availability_zone
  * host.id
  * host.name (when not using workload identity)
- When then GCE metadata server is unavailable, the resource processor exits without modifying the resource attributes. So it's safe to enable globally.
- Enabling the resource processor on the agent sidecars allows for more accurate attributes, like the host.id of the node the reconciler is on, instead of the host.id of the otel-collector.

For more details, see the docs:
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor

### DIFF
--- a/docs/metric-resource-attributes.md
+++ b/docs/metric-resource-attributes.md
@@ -1,0 +1,53 @@
+# Config Sync Metric Resource Attributes
+
+Config Sync uses the Open Telemetry Collector (otel-collector) to collect and
+export metrics from its various components to one or more destinations. 
+
+Open Telemetry resource attributes uniquely identify a metrics source. 
+
+## Pipeline
+
+The metric pipeline:
+
+component binary -> otel agent sidecar -> otel collector
+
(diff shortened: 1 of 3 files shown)


karlkfi:
/retest

TestOverrideGitSyncDepthV1Beta1 flaked

nan-yu (approved):
/lgtm

karlkfi:
/retest

TestConflictingDefinitions_NamespaceToNamespace flaked with `ERROR: validating metrics: expected metric "reconciler_errors" (tags: [{{component} sync}]) to equal 1 but got 0`, which should (i think) be fixed by https://github.com/GoogleContainerTools/kpt-config-sync/pull/100 later.

PR #115, opened 2022-09-19
Don't send sync metrics unless status has changed

karlkfi:
This reduces redundant metrics and allows us to wait until the sync metric is received to confirm that sync has occurred.


nan-yu (approved):
/lgtm

PR #117, opened 2022-09-19
Log metrics recorded and received for debugging

karlkfi:
- Enable recorded metric logging with -v=5 (component binary)
- Enable received metric logging with --debug (nomostest)
- Fix existing Tag printout in test validation failure by using a custom Tag struct which is serializable with fmt %v (no private fields)

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -16,6 +16,7 @@ package nomostest
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
@@ -29,6 +30,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
+	"kpt.dev/configsync/e2e/nomostest/metrics"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/reposync"
 	"kpt.dev/configsync/pkg/rootsync"
@@ -38,6 +40,7 @@ import (
 	"github.com/pkg/errors"
 	"go.opencensus.io/tag"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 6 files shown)


karlkfi:
/hold

TestChangeCustomResourceDefinitionSchema flaked

pkg/kmetrics/record.go
```diff
@@ -46,76 +47,85 @@ func RecordKustomizeFieldCountData(ctx context.Context, fieldCountData *Kustomiz
 	recordKustomizeTopTierMetrics(ctx, fieldCountData.TopTierCount)
 }
 
+func record(ctx context.Context, ms ...stats.Measurement) {
+	stats.Record(ctx, ms...)
+	if klog.V(5).Enabled() {
+		for _, m := range ms {
+			klog.V(5).Infof("Metric recorded: { \"Name\": %q, \"Value\": %#v, \"Tags\": %s }", m.Measure().Name(), m.Value(), tag.FromContext(ctx))
```
nan-yu, line 54: `.V(5)` is not needed as `klog.V(5)` is already enabled in this case.

pkg/metrics/record.go
```diff
@@ -22,18 +22,28 @@ import (
...
+	"k8s.io/klog/v2"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/reconcilermanager"
 )
 
+func record(ctx context.Context, ms ...stats.Measurement) {
+	stats.Record(ctx, ms...)
+	if klog.V(5).Enabled() {
+		for _, m := range ms {
+			klog.V(5).Infof("Metric recorded: { \"Name\": %q, \"Value\": %#v, \"Tags\": %s }", m.Measure().Name(), m.Value(), tag.FromContext(ctx))
```
nan-yu, line 34: `klog.Infof`

pkg/kmetrics/record.go
karlkfi: The reason I use the levels log inside the if enabled statement is because it makes it easier to programatically refactor later without requiring manual changes. I remember an email thread a few years back about how K8s was hard to refactor in these cases when they decided what the levels should mean. But you're right that it's redundant. It's only redundant when the log level is high tho, so it's not a big tax. Do you still want me to simplify it?

karlkfi: Fixed, for consistency with other -v=5 logs.

pkg/metrics/record.go
karlkfi: Fixed

e2e/nomostest/nt.go
```diff
@@ -430,6 +433,22 @@ func (nt *NT) ValidateNotFound(name, namespace string, o client.Object) error {
 	return err
 }
 
+// ValidateNotFoundOrNoMatch returns an error if the indicated object is
+// either NotFound or NoMatchFound (GVK not found).
```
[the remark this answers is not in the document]
karlkfi: Good catch. Fixed.

nan-yu (approved):
/lgtm

PR #118, opened 2022-09-19
Send correct number of declared_resources

karlkfi:
- This fixes an edge case where declared_resources was zero when more than zero resources were declared.
- The reason this wasn't caught before is because TestDontDeleteAllNamespaces expects zero resources when it tests deletion of all namespaces.
- This unblocks adding a new cluster-scoped safety resource in a future change.

### DIFF
--- a/pkg/declared/resources.go
+++ b/pkg/declared/resources.go
@@ -63,6 +63,9 @@ func (r *Resources) Update(ctx context.Context, objects []client.Object) ([]clie
 		newObjects = append(newObjects, obj)
 	}
 
+	// Record the declared_resources metric, after parsing but before validation.
+	metrics.RecordDeclaredResources(ctx, len(newObjects))
+
 	previousSet := r.getObjectSet()
 	if err := deletesAllNamespaces(previousSet, newSet); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


nan-yu (commented):
Can we update the `TestDontDeleteAllNamespaces` test to include a cluster-scoped resource, and update the expected declared resources from 0 to 1 to validate the change?

karlkfi:
> Can we update the TestDontDeleteAllNamespaces test to include a cluster-scoped resource, and update the expected declared resources from 0 to 1 to validate the change?

I could, but https://github.com/GoogleContainerTools/kpt-config-sync/pull/100 does it anyway with the new safety ClusterRole, so it would be a bit redundant. Do you want both?

nan-yu:
> I could, but https://github.com/GoogleContainerTools/kpt-config-sync/pull/100 does it anyway with the new safety ClusterRole, so it would be a bit redundant. Do you want both?

I haven't reviewed https://github.com/GoogleContainerTools/kpt-config-sync/pull/100 yet. If https://github.com/GoogleContainerTools/kpt-config-sync/pull/100  covers it, this PR looks good.

nan-yu (approved):
/lgtm

karlkfi:
/retest

TestConflictingDefinitions_RootToRoot flaked with `ERROR: roles.rbac.authorization.k8s.io "pods" not found`

PR #119, opened 2022-09-19
Validate metrics after Sync, not just Apply

karlkfi:
- Add `"status"` tag to `last_sync_timestamp` so it can be used to check whether a sync had either source or sync errors. Because, `last_apply_timestamp` only represents sync errors, not source errors.
- Change e2e metrics validation to wait for `last_sync_timestamp`, instead of `last_apply_timestamp`. This more correctly waits until sync is complete, not just apply, and has an error status that includes source errors.
- Add constants for status metric tag values.
- These changes should reduce flakey errors.

Depends onhttps://github.com/GoogleContainerTools/kpt-config-sync/pull/115

### DIFF
--- a/pkg/metrics/record.go
+++ b/pkg/metrics/record.go
@@ -86,9 +86,12 @@ func RecordParserDuration(ctx context.Context, trigger, source, status string, s
 }
 
 // RecordLastSync produces a measurement for the LastSync view.
-func RecordLastSync(ctx context.Context, commit string, timestamp time.Time) {
+func RecordLastSync(ctx context.Context, status, commit string, timestamp time.Time) {
+	tagCtx, _ := tag.New(ctx,
+		tag.Upsert(KeyStatus, status),
+		tag.Upsert(KeyCommit, commit))
 	measurement := LastSync.M(timestamp.Unix())
-	record(ctx, measurement)
+	record(tagCtx, measurement)
 }
 
 // RecordDeclaredResources produces a measurement for the DeclaredResources view.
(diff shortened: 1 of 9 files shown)


pkg/metrics/record.go
```diff
@@ -76,9 +76,12 @@ func RecordParserDuration(ctx context.Context, trigger, source, status string, s
 }
 
 // RecordLastSync produces a measurement for the LastSync view.
-func RecordLastSync(ctx context.Context, commit string, timestamp time.Time) {
+func RecordLastSync(ctx context.Context, status, commit string, timestamp time.Time) {
+	tagCtx, _ := tag.New(ctx,
+		tag.Upsert(KeyStatus, status),
+		tag.Upsert(KeyCommit, commit))
```
nan-yu, line 82: Just want to confirm the commit tag will not increase the cardinality?

karlkfi: It will, but no more than already present in the last_apply_timestamp, which has the same tags.

My hope is that the resource attributes added by https://github.com/GoogleContainerTools/kpt-config-sync/pull/100 and https://github.com/GoogleContainerTools/kpt-config-sync/pull/114 will mitigate the issue so that it's only a problem if you have 5k+ commits in the same repo PER RECONCILER POD.

And if that becomes a problem, you can simply delete the reconciler pod and re-create it to start over on a new time series with new resource attributes.

More likely tho, a node replacement or upgrade will cause replacement first.

nan-yu: Good to know.

karlkfi: Also, the 5k+ limit is a recommendation is for Monarch. Cloud Monitoring actually documents 30k as the limit: https://cloud.google.com/logging/docs/logs-based-metrics/troubleshooting#too-many-time-series

karlkfi:
/retest

mono-repo tests appear to be flakey for non-metrics reasons

nan-yu (approved):
/lgtm

PR #125, opened 2022-09-21
Added additional annotation validation to TestPreventDeletionRole

victorpras:
The test `TestPreventDeletionRole` has been flakey on MonoRepo. The test sets up a Role with `client.lifecycle.config.k8s.io/deletion: detach` annotation so it can't be deleted, but when the test is flaky it gets deleted anyway. The other tests in this suite have more validation to the annotation, so I added the annotation validation to this test too. If the test is still failing/flaky, this additional validation should help us isolate the root cause


nan-yu (approved):
/lgtm

Issue #127, opened 2022-09-22
Namespaces not included in OSS-release yaml


wmuizelaar:
Hi,

I downloaded the latest release yaml (https://github.com/GoogleContainerTools/kpt-config-sync/releases/download/v1.13.0/config-sync-manifest.yaml) and it did not include the Namespace definitions where it needs to be deployed.

I did find the release-oss script though where it states that it needs to include those, so probably somewhere something went wrong while running this script on the last release. Could this be validated?

Kind regards,
Wietse

sdowell:
Thank you for reporting this issue, I'll look into this

sdowell:
A fix has been applied which will include the missing resources for future releases. The `v1.13.0` release has also been updated to include the missing resources. Thank you for your bug report!

wmuizelaar:
That is a real quick and nice response, I can confirm it works. Thanks!

PR #128, opened 2022-09-22
add test-resources to OSS manifest

sdowell:
This includes both the Config Management namespaces and the ResourceGroup manifests, which are needed for OSS installation.

https://github.com/GoogleContainerTools/kpt-config-sync/issues/127


nan-yu (approved):
/lgtm

PR #129, opened 2022-09-22
Re-enable a disabled metrics validation test

karlkfi:
The new DefaultRootSyncObjectCount should solve the flakiness that caused this check to be disabled.


nan-yu (approved):
/lgtm

PR #130, opened 2022-09-22
Add missing test groups for the multi-repo mode

haiyanmeng:


Makefile.e2e
```diff
@@ -201,7 +201,7 @@ test-e2e-kind-multi-repo-test-group2:
 
 # This target runs the third group of e2e tests with the multi-repo mode.
 test-e2e-kind-multi-repo-test-group3:
-	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=reconciliation-2" test-e2e-kind-multi-repo
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=reconciliation-2,multi-repos,override-api,workload-identity,hydration" test-e2e-kind-multi-repo
```
nan-yu, line 204: This is not needed as the workload identity tests only run on GKE clusters.

haiyanmeng: I see. Good catch. I didn't noticed it has `ntopts.RequireGKE(t)`.
Removed it.

nan-yu (approved):
/lgtm

sdowell:
I wonder if we can add a presubmit check to make sure all the test groups are covered by an e2e target. Might be nice to programmatically make sure we don't accidentally miss anything

PR #131, opened 2022-09-22
Use timers instead of tickers

karlkfi:
Tickers can cause memory leaks and continuous execution, when execution takes longer than the tick duration. This is especially problematic in e2e where the timers are set very short and at scale when the sync can take longer than the retry/polling/resync period.

Timers with resets function more as a delay between executions, which is preferable, rather than something tied to a wall-clock.


nan-yu (approved):
/lgtm

PR #133, opened 2022-09-23
Avoid unnecessary status updates

karlkfi:
- Update source and reconciling status updates to match existing sync statups update dedupe behavior.
- Log source, reconciling, and status errors when they change.
- Only send status metrics when there's a change, and then record them even if there's no errors. This should help reset reconciler_errors back to zero when the errors have been fixed.

### DIFF
--- a/pkg/parse/namespace.go
+++ b/pkg/parse/namespace.go
@@ -158,23 +158,38 @@ func (p *namespace) setSourceStatusWithRetries(ctx context.Context, newStatus so
 		return status.APIServerError(err, "failed to get RepoSync for parser")
 	}
 
+	currentRS := rs.DeepCopy()
+
 	setSourceStatus(&rs.Status.Source, p, newStatus, denominator)
 
-	continueSyncing := true
-	if rs.Status.Source.ErrorSummary.TotalCount > 0 {
-		continueSyncing = false
-	}
-	metrics.RecordPipelineError(ctx, configsync.RepoSyncName, "source", rs.Status.Source.ErrorSummary.TotalCount)
+	continueSyncing := (rs.Status.Source.ErrorSummary.TotalCount == 0)
 	var errorSource []v1beta1.ErrorSource
-	if rs.Status.Source.Errors != nil {
+	if len(rs.Status.Source.Errors) > 0 {
 		errorSource = []v1beta1.ErrorSource{v1beta1.SourceError}
 	}
 	reposync.SetSyncing(&rs, continueSyncing, "Source", "Source", newStatus.commit, errorSource, rs.Status.Source.ErrorSummary, newStatus.lastUpdate)
 
-	metrics.RecordReconcilerErrors(ctx, "source", status.ToCSE(newStatus.errs))
+	// Avoid unnecessary status updates.
+	if cmp.Equal(currentRS.Status, rs.Status, compare.IgnoreTimestampUpdates) {
+		klog.V(5).Infof("Skipping source status update for RepoSync %s/%s", rs.Namespace, rs.Name)
+		return nil
+	}
+
+	csErrs := status.ToCSE(newStatus.errs)
+	metrics.RecordReconcilerErrors(ctx, "source", csErrs)
+	metrics.RecordPipelineError(ctx, configsync.RepoSyncName, "source", len(csErrs))
+	if len(csErrs) > 0 {
+		klog.Infof("New source errors for RepoSync %s/%s: %+v",
+			rs.Namespace, rs.Name, csErrs)
+	}
+
+	if klog.V(5).Enabled() {
+		klog.Infof("Updating source status for RepoSync %s/%s:\nDiff (- Expected, + Actual):\n%s",
+			rs.Namespace, rs.Name, cmp.Diff(currentRS.Status, rs.Status))
+	}
 
 	if err := p.client.Status().Update(ctx, &rs); err != nil {
-		// If the update failure was caused by the size of the RootSync object, we would truncate the errors and retry.
+		// If the update failure was caused by the size of the RepoSync object, we would truncate the errors and retry.
 		if isRequestTooLargeError(err) {
 			klog.Infof("Failed to update RepoSync source status (total error count: %d, denominator: %d): %s.", rs.Status.Source.ErrorSummary.TotalCount, denominator, err)
 			return p.setSourceStatusWithRetries(ctx, newStatus, denominator*2)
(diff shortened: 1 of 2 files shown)


pkg/parse/root.go
```diff
@@ -169,20 +169,35 @@ func (p *root) setSourceStatusWithRetries(ctx context.Context, newStatus sourceS
...
+
+	csErrs := status.ToCSE(newStatus.errs)
+	metrics.RecordReconcilerErrors(ctx, "source", csErrs)
+	metrics.RecordPipelineError(ctx, configsync.RootSyncName, "source", len(csErrs))
+	if len(csErrs) > 0 {
+		klog.Infof("New source errors for RootSync %s/%s: %+v",
+			rs.Namespace, rs.Name, csErrs)
+	}
+
+	if klog.V(5).Enabled() {
```
haiyanmeng, line 197: Why not use `klog.V(5).Infof` here?

pkg/parse/root.go
```diff
@@ -169,20 +169,35 @@ func (p *root) setSourceStatusWithRetries(ctx context.Context, newStatus sourceS
...
-	metrics.RecordReconcilerErrors(ctx, "source", status.ToCSE(newStatus.errs))
+	// Avoid unnecessary status updates.
+	if cmp.Equal(currentRS.Status, rs.Status, compare.IgnoreTimestampUpdates) {
+		klog.V(5).Infof("Skipping source status update for RootSync %s/%s", rs.Namespace, rs.Name)
+		return nil
+	}
+
+	csErrs := status.ToCSE(newStatus.errs)
+	metrics.RecordReconcilerErrors(ctx, "source", csErrs)
+	metrics.RecordPipelineError(ctx, configsync.RootSyncName, "source", len(csErrs))
```
haiyanmeng, line 191: We still should record these two metrics even if there is no need to update the status.

pkg/parse/root.go
karlkfi: the diff is expensive

pkg/parse/root.go
karlkfi: why? if they haven't changed the metric was recorded last time.

haiyanmeng: Recording these metrics for every syncs helps understand questions like the successful rate of syncs in the last 30 days.

karlkfi: These status update functions aren't being called after every sync tho. They're being called asynchronously every 5 seconds...

pkg/parse/namespace.go
```diff
@@ -280,8 +310,13 @@ func (p *namespace) setSyncStatusWithRetries(ctx context.Context, errs status.Mu
 		return nil
 	}
 
-	metrics.RecordReconcilerErrors(ctx, "sync", status.ToCSE(errs))
-	metrics.RecordPipelineError(ctx, configsync.RepoSyncName, "sync", rs.Status.Sync.ErrorSummary.TotalCount)
+	csErrs := status.ToCSE(errs)
+	metrics.RecordReconcilerErrors(ctx, "sync", csErrs)
+	metrics.RecordPipelineError(ctx, configsync.RepoSyncName, "sync", len(csErrs))
```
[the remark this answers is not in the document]
karlkfi: Pipeline errors? I don't think so. That would mean they would never zero out, and the latest value would always be non-zero. They need to zero out or we can't test there are no errors after correcting the problem. They all have a different component tag value, which makes them unique. This one is component=sync.

pkg/parse/namespace.go
```diff
@@ -158,23 +158,38 @@ func (p *namespace) setSourceStatusWithRetries(ctx context.Context, newStatus so
...
-	metrics.RecordReconcilerErrors(ctx, "source", status.ToCSE(newStatus.errs))
+	// Avoid unnecessary status updates.
+	if cmp.Equal(currentRS.Status, rs.Status, compare.IgnoreTimestampUpdates) {
+		klog.V(5).Infof("Skipping source status update for RepoSync %s/%s", rs.Namespace, rs.Name)
+		return nil
+	}
+
+	csErrs := status.ToCSE(newStatus.errs)
+	metrics.RecordReconcilerErrors(ctx, "source", csErrs)
+	metrics.RecordPipelineError(ctx, configsync.RepoSyncName, "source", len(csErrs))
```
sdowell, line 180: Seems we should define some constants for the component names (`source`, `rendering`, `sync`, etc)

pkg/parse/namespace.go
```diff
@@ -205,30 +220,45 @@ func (p *namespace) setRenderingStatusWithRetires(ctx context.Context, newStatus
 		return status.APIServerError(err, "failed to get RepoSync for parser")
 	}
 
-	if rs.Status.Rendering.Commit != newStatus.commit {
+	currentRS := rs.DeepCopy()
```
sdowell, line 223: Looks like we are repeating this same snippet with minor variation a number of times. Might be worth considering an interface for setting the status of each component type so that we could generalize this to a common `setStatus` type function. Not necessarily in scope for this change but just offering the suggestion.

WDYT?

pkg/parse/root.go
karlkfi: You're right about rendering. `setRenderingStatus` is called in 3 different places after rendering (done, not done, and error). So it does get called after each render. But rendering happens every time the run loop triggers, which is on every poll, re-sync, or retry timer. So even if there's nothing to sync, the rendering still happens. Why would we want to send metrics for all of those if it just re-renders the same way it rendered last time?

`setSourceStatus` tho is only called when the source status changes, because it's being wrapped by `needToSetSourceStatus`. This extra check inside is to confirm that nothing has changed after converting the input status to the actual RSync API.

`SetSyncStatus` is also normally wrapped by `needToSetSyncStatus`, but that didn't get added to the async status updates. So the equality check in `SetSyncStatus` is covering that gap.

So even before I added the cmp.Equal checks it wasn't possible to use these metrics to compute "successful rate of syncs in the last 30 days". 

I'm also not sure that metric would actually be useful unless it was a ratio of commits that succeeded, not just a ration of attempts that succeeded. Otherwise the 1s retry timer would just bork the metric any time a broken commit was left broken.

Plus, if you use Prometheus without timestamps enabled, the metric would just appear unchanged every time you scrape it, with the latest value, unless it dropped off the 5m cache timeout. So you wouldn't be able to get that metric with Prometheus, only Cloud Monitoring, assuming it keeps track of the historical values every time they're reported. 

Plus metrics are designed to be lossy. If there's a network issue they don't usually retry. Otherwise you get cascading metric loss where a pile up of old metrics can stop new metrics from being reported. So you can't really produce a useful ratio metric over time unless the reporter is doing the math.

pkg/parse/namespace.go
karlkfi: I considered that but it's the diff between a RootSync and RepoSync object, and we can't add methods to the API structs. So we would have to wrap them with indirection for everything we need to read/write, which sounds tedious...

If we did that tho, it would be useful for a lot of other RootSync vs RepoSync code, like the reconciler impls.

pkg/parse/namespace.go
karlkfi: Aye. Can do in a follow up.

sdowell:
LGTM but seems like @haiyanmeng still has some open questions so I'll let her approve
/lgtm

pkg/parse/namespace.go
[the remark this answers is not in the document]
karlkfi: Sorry, I misunderstood. I agree that the reconcile errors and pipeline errors should only be sent after the sync completes, not every 5s while syncing. So I can wrap these in the !syncing flag. However, while that will stop sending every 5s, I'm not sure if it means it will only send AFTER a sync. It might also send BEFORE the sync, especially on a re-sync, when the cache is wiped out before syncing. In fact, that case might be why we have some lost errors...

pkg/parse/root.go
haiyanmeng: I agree that today our code is not consistent on when to update status metrics for each component.
We should audit our metrics and make it consistent.

pkg/parse/namespace.go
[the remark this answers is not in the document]
haiyanmeng: Synced offline on this. We can clean this up in a separate PR.

PR #134, opened 2022-09-26
Port bats test to go test

tiffanny29631:
cli.bats
- test case converted and added check for multi-repo mode

foo_corp.bats
- test case already covered by acme_test.go, removed

repo_less.bats
- test case converted into mono repo only go test

The related BATS files are removed.


e2e/testcases/cli_test.go
```diff
@@ -37,6 +44,14 @@ import (
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/testing/fake"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
+)
+
+const (
+	KubeSystem                 = "kube-system"
+	ConfigManagementSystem     = "config-management-system"
```
haiyanmeng, line 52: Use `ControllerNamespace` from pkg/api/configsync

e2e/testcases/cli_test.go
```diff
@@ -37,6 +44,14 @@ import (
...
+	ConfigManagementMonitoring = "config-management-monitoring"
```
haiyanmeng, line 53: Use `MonitoringNamespace` from `pkg/metrics`

e2e/testcases/cli_test.go
```diff
@@ -37,6 +44,14 @@ import (
...
+	ResourceGroupSystem        = "resource-group-system"
```
haiyanmeng, line 54: Use `RGControllerNamespace` from `pkg/api/configmanagement`

e2e/testcases/cli_test.go
```diff
@@ -833,3 +848,235 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
 		})
 	}
 }
+
+func TestCLIBugreportNomosRunningCorrectly(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI, ntopts.RequireGKE(t))
```
haiyanmeng, line 853: `ntopts.RequireGKE(t)`: is this required?

e2e/testcases/cli_test.go
```diff
@@ -833,3 +848,235 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+	// Confirm all the namespaces and pods are up
+	time, e := nomostest.Retry(nt.DefaultReconcileTimeout, func() error {
+		err := nt.Validate(KubeSystem, "", &corev1.Namespace{})
+		if err != nil {
+			tw.Fatal(err)
+		}
+		err = nt.Validate(ConfigManagementSystem, "", &corev1.Namespace{})
+		if err != nil {
+			tw.Fatal(err)
+		}
```
haiyanmeng, line 873: Do sth to avoid the duplication of the following statements:
```
       err = nt.Validate(..., "", &corev1.Namespace{})
		if err != nil {
			tw.Fatal(err)
		}
```

e2e/testcases/cli_test.go
```diff
@@ -833,3 +848,235 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+			err = nt.Validate("otel-collector", ConfigManagementMonitoring, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus))
+			if err != nil {
+				tw.Fatal(err)
+			}
+		} else {
+			err := nt.Validate("git-importer", ConfigManagementSystem, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus))
+			if err != nil {
+				tw.Fatal(err)
+			}
+			err = nt.Validate("monitor", ConfigManagementSystem, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus))
```
haiyanmeng, line 904: Do sth to avoid duplication.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +848,235 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/namespaces/config-management-system/git-importer.*/importer.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/namespaces/config-management-system/monitor.*/monitor.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/namespaces/config-management-system/pods.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/namespaces/kube-system/pods.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/clusterconfigs.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/namespaceconfigs.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/repos.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/syncs.txt", bugReportDirName, context), files)
```
haiyanmeng, line 990: Do sth to avoid duplication.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +848,235 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+		tw.Fatal("did not find all expected files in bug report zip file")
+	}
+}
+
+// getBugReportZipName find and returns the zip name of bugreport under test dir
+// or error if no bugreport zip found
+func getBugReportZipName(dir string, nt *nomostest.NT) (string, error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		klog.Fatal(err)
```
haiyanmeng, line 1002: Why not use `nt.T.Fatal` here?

e2e/testcases/legacy_test.go
```diff
@@ -211,10 +208,11 @@ func TestBats(t *testing.T) {
...
 			skipAutopilotCluster: testNums(10), // ignore the test that mutates the kube-system namespace
 		},
-		{
-			fileName:      "cli.bats",
-			skipMultiRepo: allTests, // TODO: implement nomos status in CLI, this may be a go rewrite
-		},
+		// Fully converted into cli_test.go
+		//{
+		//	fileName:      "cli.bats",
+		//	skipMultiRepo: allTests, // TODO: implement nomos status in CLI, this may be a go rewrite
```
haiyanmeng, line 214: `// TODO: implement nomos status in CLI, this may be a go rewrite`: this should be removed.

e2e/testcases/repoless_test.go
```diff
@@ -0,0 +1,169 @@
...
+	testing2 "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/kinds"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
+)
+
+func TestSyncCorrectlyWithExplicitNamespaceDeclarations(t *testing.T) {
+	nt := nomostest.New(t, testing2.SyncSource, ntopts.Unstructured, ntopts.SkipMultiRepo)
+	nt.T.Cleanup(func() {
+		nt.MustKubectl("delete", "-f", "../raw-nomos/manifests/importer_repoless.yaml")
```
haiyanmeng, line 33: Put `../raw-nomos/manifests/importer_repoless.yaml` into a var since it is used at multiple places.

e2e/testcases/repoless_test.go
```diff
@@ -0,0 +1,169 @@
...
+		if err != nil {
+			nt.T.Fatal(err)
+		}
+		err = nt.Validate("pod-reader-default", "backend-default", &rbacv1.Role{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus))
+		if err != nil {
+			nt.T.Fatal(err)
+		}
+		checkResourceCount(nt, kinds.Namespace(), "", 0, map[string]string{"backend.tree.hnc.x-k8s.io": "0"}, nil)
+		checkResourceCount(nt, kinds.Namespace(), "", 0, map[string]string{"backend-default.tree.hnc.x-k8s.io": "0"}, nil)
+		checkResourceCount(nt, kinds.Namespace(), "", 0, nil, map[string]string{"hnc.x-k8s.io/managed-by": "configmanagement.gke.io"})
```
haiyanmeng, line 68: The logic here seems to be used in all the three tests in this file. Do sth to avoid duplication.

e2e/testcases/repoless_test.go
```diff
@@ -0,0 +1,169 @@
...
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	testing2 "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/kinds"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
+)
+
```
haiyanmeng, line 29: The three tests in this file share lots of common logic. Do sth to avoid duplication.

haiyanmeng (changes requested):

e2e/testcases/cli_test.go
tiffanny29631: Done

e2e/testcases/cli_test.go
tiffanny29631: Using status.MultiError instead to reduce the error catch.

e2e/testcases/repoless_test.go
tiffanny29631: Added const for three manifest paths.

e2e/testcases/repoless_test.go
tiffanny29631: Split resource validate out.

e2e/testcases/repoless_test.go
tiffanny29631: Split commit sample repo out.

e2e/testcases/cli_test.go
tiffanny29631: Done by using status.MultiError

e2e/testcases/cli_test.go
tiffanny29631: We want to validate all these files exist under the specific path, and the folder structure is different when running under mono or multi repo mode.

e2e/testcases/cli_test.go
tiffanny29631: The test errors out when running with kind cluster, that has trouble exporting the kubeconfig or dump the log. I got it running in GKE clusters.

haiyanmeng: Can the old bat test run on kind clusters?

e2e/testcases/cli_test.go
haiyanmeng: I see.

e2e/testcases/cli_test.go
tiffanny29631: It actually could.. 

```
-  # Fixes a problem when running the tests locally
-  if [[ -z "${KUBECONFIG+x}" ]]; then
-    export KUBECONFIG="${HOME}/.kube/config"
-  fi
```
This seemed to be doing the trick but looks confusing

e2e/testcases/cli_test.go
```diff
@@ -833,3 +850,206 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+
+	// get bugreport
+	out, err := exec.Command("nomos", "bugreport").CombinedOutput()
+	if err != nil {
+		tw.Log(string(out))
+		tw.Fatal(err)
+	}
+
+	// locate zip file
+	bugReportZipName, err = getBugReportZipName("./", nt)
```
nan-yu, line 897: It generates the bugreport in the current directory. If you move it to `nt.TmpDir`, you wouldn't need to delete it in the end. The test framework will blow away the entire `nt.TmpDir` after the test.

e2e/testcases/cli_test.go
nan-yu: Can we define a list of expected files and then iterate through the list using a for loop?

```
    expectedFiles := []string{
		fmt.Sprintf("%s/processed/%s/version.txt", bugReportDirName, context),
		fmt.Sprintf("%s/processed/%s/status.txt", bugReportDirName, context),
	}
	if nt.MultiRepo {
		expectedFiles = append(expectedFiles, ...)
		...
	} else {
		expectedFiles = append(expectedFiles, ...)
		...
	}

	for _, file := range expectedFiles {
		if checkFileExists(file, files) {
			nt.T.Errorf("did not find all expected files in bug report zip file")
		}
	}

```

Using the above code should get rid of the `found` flag.

nan-yu (changes requested):

e2e/testcases/repoless_test.go
```diff
@@ -0,0 +1,114 @@
...
+const (
+	importerRepoless            = "../raw-nomos/manifests/importer_repoless.yaml"
+	importerRepolessNoNS        = "../raw-nomos/manifests/importer_repoless-no-ns.yaml"
+	importerRepolessNoPolicyDir = "../raw-nomos/manifests/importer_no-policy-dir.yaml"
+)
+
+func TestSyncCorrectlyWithExplicitNamespaceDeclarations(t *testing.T) {
+	nt := nomostest.New(t, testing2.SyncSource, ntopts.Unstructured, ntopts.SkipMultiRepo)
+	nt.T.Cleanup(func() {
+		nt.MustKubectl("delete", "-f", importerRepoless)
```
nan-yu, line 40: This is not needed as the [RresetMonoRepoSpec](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/config-sync.go#L1211) will reset the policy_dir back to the initial one.

nan-yu (commented):
It is okay for this PR, but for future reference, I feel like the repoless conversion should go in a separate PR.

e2e/testcases/cli_test.go
tiffanny29631: Moving zip file and unzipping into nt.TmpDir

e2e/testcases/cli_test.go
tiffanny29631: Fixed, it's because of the resource validation predicate. Now the test works with kind.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +850,206 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
 		})
 	}
 }
+
+func TestCLIBugreportNomosRunningCorrectly(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI, ntopts.RequireGKE(t))
```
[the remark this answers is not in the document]
tiffanny29631: No it does not, removed as the test works on kind as well.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +850,206 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+			errs = status.Append(errs, nt.Validate(core.RootReconcilerPrefix, configsync.ControllerNamespace, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus)))
+			errs = status.Append(errs, nt.Validate(configmanagement.RGControllerName, configmanagement.RGControllerNamespace, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus)))
+			errs = status.Append(errs, nt.Validate(reconcilermanager.ManagerName, configsync.ControllerNamespace, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus)))
+			errs = status.Append(errs, nt.Validate(metrics.OtelCollectorName, metrics.MonitoringNamespace, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus)))
+		} else {
+			errs = status.Append(errs, nt.Validate(filesystem.GitImporterName, configsync.ControllerNamespace, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus)))
+			errs = status.Append(errs, nt.Validate(state.MonitorName, configsync.ControllerNamespace, &appsv1.Deployment{}, nomostest.StatusEquals(nt, kstatus.CurrentStatus)))
+		}
+		return errs
+	})
```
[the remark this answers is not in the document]
tiffanny29631: Agreed, removed this block.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +850,206 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
 		})
 	}
 }
+
+func TestCLIBugreportNomosRunningCorrectly(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI, ntopts.RequireGKE(t))
+	tmpDir := nomostest.TestDir(t)
```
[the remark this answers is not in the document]
tiffanny29631: Removed as it's duplicate.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +850,206 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+	tw := nomostesting.New(t, nomostesting.NomosCLI)
```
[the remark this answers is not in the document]
tiffanny29631: Removed the test wrapper, just using nt.

e2e/testcases/repoless_test.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+const (
+	importerRepoless            = "../raw-nomos/manifests/importer_repoless.yaml"
+	importerRepolessNoNS        = "../raw-nomos/manifests/importer_repoless-no-ns.yaml"
+	importerRepolessNoPolicyDir = "../raw-nomos/manifests/importer_no-policy-dir.yaml"
+)
+
+func TestSyncCorrectlyWithExplicitNamespaceDeclarations(t *testing.T) {
+	nt := nomostest.New(t, testing2.SyncSource, ntopts.Unstructured, ntopts.SkipMultiRepo)
```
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/testcases/repoless_test.go
tiffanny29631: Removed Cleanup for all three test cases.

e2e/testcases/cli_test.go
```diff
@@ -39,6 +46,11 @@ import (
 	"kpt.dev/configsync/pkg/testing/fake"
 )
 
+const (
+	KubeSystem          = "kube-system"
+	ResourceGroupSystem = "resource-group-system"
```
haiyanmeng, line 51: Use RGControllerNamespace from pkg/api/configmanagement, as commented in https://github.com/GoogleContainerTools/kpt-config-sync/pull/134#discussion_r981813531.

e2e/testcases/legacy_test.go
```diff
@@ -174,9 +174,6 @@ func TestBats(t *testing.T) {
 		t.Skip("Skip running bats test on non-local git providers")
```
haiyanmeng, line 174: Since this PR converts all the remaining bats tests to golang tests, we may remove this file and all the bats test files.

e2e/testcases/cli_test.go
nan-yu: KubeSystem is also declared as `metav1.NamespaceSystem`.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +845,192 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
 		})
 	}
 }
+
+func TestCLIBugreportNomosRunningCorrectly(t *testing.T) {
+	var bugReportZipName, bugReportDirName string
+	tmpDir := nomostest.TestDir(t)
+	tw := nomostesting.New(t, nomostesting.NomosCLI)
+	opts := nomostest.NewOptStruct(nomostest.TestClusterName(tw), tmpDir, tw)
+	nomostest.FreshTestEnv(t, opts)
```
nan-yu, line 854: Why not using `nomostest.New` directly? I think it should also apply to CI jobs with `--share-test-env`.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +845,192 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+
+	// get bugreport
+	out, err := exec.Command("nomos", "bugreport").CombinedOutput()
```
nan-yu, line 857: You need to switch to the test context, instead of the current context.

You can do so by two steps:
1. use the kubectl command to get the current test context: `nt.Kubectl*config", "current-context"), which takes `nt.kubeconfigPath` as `--kubeconfig`.
2. use the kubectl command to switch to the context: `nt.MustKubectl("config", "use-context", testContext)`

Then `nomos bugreport` is guaranteed to run with the test context, instead of the default one.

e2e/testcases/cli_test.go
```diff
@@ -833,3 +845,192 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+	if err != nil {
+		tw.Fatal(err)
+	}
+
+	// move zip file to nt.tmpDir and unzip
+	tw.Log(fmt.Sprintf("move %s to %s", bugReportZipName, tmpDir))
+	_, err = exec.Command("mv", bugReportZipName, path.Join(tmpDir, bugReportZipName)).CombinedOutput()
+	if err != nil {
+		tw.Fatal(err)
+	}
```
nan-yu, line 874: You can get rid of the move command by setting the working directory when running the `nomos bugreport` command:
```
cmd := exec.Command("nomos", "bugreport")
cmd.Dir = nt.TmpDir
out, err := cmd.Output()
```

e2e/testcases/cli_test.go
```diff
@@ -833,3 +845,192 @@ func TestNomosVetNamespaceRepo(t *testing.T) {
...
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/namespaces/kube-system/pods.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/clusterconfigs.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/namespaceconfigs.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/repos.txt", bugReportDirName, context), files)
+		found = found && checkFileExists(fmt.Sprintf("%s/raw/%s/cluster/configmanagement/syncs.txt", bugReportDirName, context), files)
+	}
+	if !found {
+		tw.Fatal("did not find all expected files in bug report zip file")
+	}
+	tw.Log("Found all expected files in bugreport zip")
```
nan-yu, line 946: I left a command a few days ago to simplify this block: https://github.com/GoogleContainerTools/kpt-config-sync/pull/134#discussion_r988440749.

examples/repoless/default-ns.yaml
```diff
@@ -15,4 +15,4 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: default
+  name: backend-default
```
nan-yu, line 18: Just curious why do we want to change the namespace? If so, we need to update the file name as it is not the default namespace anymore,

nan-yu (changes requested):

e2e/testcases/cli_test.go
tiffanny29631: Done, const replaced by specific declarations.

e2e/testcases/cli_test.go
tiffanny29631: I noticed consistent presubmit failures with `nomos` command fails at creating RestMap, and the error indicates attempts to load kubeconfig from other random test directories. The root cause points to the parallelism after several tests so using this logic but skip enableParallel.

e2e/testcases/cli_test.go
tiffanny29631: The test seems to have no issue with recognizing context, if it is for the error mentioned in [this comment](https://github.com/GoogleContainerTools/kpt-config-sync/pull/134#discussion_r998807686). I've tested it in both GKE and Kind.

e2e/testcases/cli_test.go
tiffanny29631: Done, setting work directory when running the `nomos bugreport`

e2e/testcases/cli_test.go
tiffanny29631: Done as recommended, checkFileExists now takes a list of file to verify, and returns nil or a list of files that are not found.

examples/repoless/default-ns.yaml
nan-yu: ACK

e2e/testcases/legacy_test.go
haiyanmeng: Sounds good. Thasnks!

haiyanmeng:
/lgtm
/hold 
Please unhold this PR after Nan approves it.

e2e/testcases/cli_test.go
nan-yu: I see. The `nomos` command loads the kubeconfig from either the `KUBECONFIG` env, or the `.kube/config` file under the current user HOME dir: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/client/restconfig/config.go#L43.

One last thing to try is to set the env variable in the test and then unsets it in `nt.T.cleanup`.

If we can use `nomostest.New` directly, then we don't need to create a Kind cluster, and install Config Sync on the Kind cluster in the periodic jobs, which will eventually avoid extending the CI running time.

tiffanny29631: Done by using nt and setting KUBECONFIG before running `nomos bugreport`, thanks for help debugging!

e2e/testcases/cli_test.go
tiffanny29631: As discussed in the mentioned thread, setting kubectl context turns out to be not relevant in this scenario.

nan-yu (approved):
/unhold
/lgtm

e2e/testcases/cli_test.go
tiffanny29631: Done!

haiyanmeng (approved):

PR #135, opened 2022-09-26
fix: Use opencensus based exporter for monitoring

rquitales:
When we upgraded to otel v0.58, we introduced a regression in the metrics being exported with a `generic_node` type instead of `k8s_pod`. This happened as our codebase uses opencensus semantic conventions for resource attributes instead of opentelemetry semantic conventions.

This workaround runs our collector with a flag to disable the newer exporter logic until we can upgrade our codebase to use opentelemetry conventions.

See https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/breaking-changes.md for more details about the rewritten exporter.

### DIFF
--- a/e2e/testdata/reconciler-manager-configmap-updated.yaml
+++ b/e2e/testdata/reconciler-manager-configmap-updated.yaml
@@ -192,6 +192,9 @@ data:
           - /otelcol-contrib
           args:
           - "--config=/conf/otel-agent-config.yaml"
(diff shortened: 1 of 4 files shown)


rquitales:
Cherry-picking commit for v1.13.1 release.

nan-yu (approved):
/lgtm

PR #137, opened 2022-09-26
Change the RootSync helm.namespace field default value

xinnywinne:
Change the helm.namespace default value from "config-management-system"
to "default" to be consistent with other tools: Helm, Kustomize etc.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -51,7 +52,7 @@ func TestPublicHelm(t *testing.T) {
 	origRepoURL := nt.GitProvider.SyncURL(nt.RootRepos[configsync.RootSyncName].RemoteRepoName)
 
 	rs := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
-	nt.T.Log("Update RootSync to sync from a public Helm Chart")
+	nt.T.Log("Update RootSync to sync from a public Helm Chart with specified release namespace")
 	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": {"repo": "%s", "chart": "ingress-nginx", "auth": "none", "version": "4.0.5", "releaseName": "my-ingress-nginx", "namespace": "ingress-nginx"}, "git": null}}`,
 		v1beta1.HelmSource, publicHelmRepo))
 	nt.T.Cleanup(func() {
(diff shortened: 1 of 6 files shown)


e2e/testcases/helm_sync_test.go
```diff
@@ -64,6 +64,18 @@ func TestPublicHelm(t *testing.T) {
 	if err := nt.Validate("my-ingress-nginx-controller", "ingress-nginx", &appsv1.Deployment{}); err != nil {
 		nt.T.Error(err)
 	}
+	// reset rs to sync from original git repo
+	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": null, "git": {"dir": "acme", "branch": "main", "repo": "%s", "auth": "ssh","gcpServiceAccountEmail": "", "secretRef": {"name": "git-creds"}}}}`,
```
nan-yu, line 68: Why do we need to reset the root sync? Can we patch it directly by setting `namespace` to empty?

Unnecessary source spec change will result in reconciler restart, which increases the running time.

e2e/testcases/helm_sync_test.go
```diff
@@ -64,6 +64,18 @@ func TestPublicHelm(t *testing.T) {
...
+		v1beta1.GitSource, origRepoURL))
+	nt.WaitForRepoSyncs()
+	nt.T.Log("Update RootSync to sync from a public Helm Chart without specified release namespace")
+	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": {"repo": "%s", "chart": "ingress-nginx", "auth": "none", "version": "4.0.5", "releaseName": "my-ingress-nginx"}, "git": null}}`,
+		v1beta1.HelmSource, publicHelmRepo))
+	nt.WaitForRepoSyncs(nomostest.WithRootSha1Func(helmChartVersion("ingress-nginx:4.0.5")),
+		nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: "ingress-nginx"}))
+	if err := nt.Validate("my-ingress-nginx-controller", configsync.DefaultHelmReleaseNamespace, &appsv1.Deployment{}); err != nil {
```
nan-yu, line 76: Suggest validating `my-ingress-nginx-controller` doesn't exist in the ingress-nginx namespace anymore.

e2e/testcases/helm_sync_test.go
xinnywinne: I tried to only change the namespace, but the WaitForRepoSyncs function does not work well with this condition. Since the RootSha1 and SyncDirectory are both same for these two syncs, it will not wait until the second sync finish.

nan-yu: Can you confirm if that's an issue of the `WaitForRepoSyncs` function or the `helm-sync` container? I think it might because the image version is the same, so even if we update the `namespace` field, it won't trigger a re-sync.

And in this case, we should probably support resync.

e2e/testcases/helm_sync_test.go
xinnywinne: Done.

e2e/testcases/helm_sync_test.go
xinnywinne: Both WaitForRepoSyncs and helm-sync container work fine. It turns out to be that I did not use the MustMergePatch() function correctly. Now I can patch root sync directly by setting namespace to empty.

nan-yu (approved):
/lgtm

PR #139, opened 2022-09-27
Fix pipeline_error_observed description to match

karlkfi:
Fix pipeline_error_observed description to match the description in the resource-group controller.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -29,10 +29,13 @@ var (
 		"The number of errors in the reconciler",
 		stats.UnitDimensionless)
 
-	// PipelineError metric measures the error by components when syncing a commit
+	// PipelineError metric measures the error by components when syncing a commit.
+	// Definition here must exactly match the definition in the resource-group
+	// controller, or the Prometheus exporter will error. b/247516388
+	// https://github.com/GoogleContainerTools/kpt-resource-group/blob/main/controllers/metrics/metrics.go#L88
 	PipelineError = stats.Int64(
 		"pipeline_error_observed",
-		"A boolean indicates if any error happens from different stages when syncing a commit",
+		"A boolean value indicates if error happened at readiness stage when syncing a commit",
 		stats.UnitDimensionless)
 
 	// ReconcileDuration metric measures the latency of reconcile events.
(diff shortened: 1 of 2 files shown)


karlkfi:
FWIW, I don't actually like the description test here, but it's easier to make them consistent now than try to fix resource-group and then need to make a new release branch & tag, because there's other stuff on head of master for resource-group.

pkg/metrics/metrics.go
```diff
@@ -29,10 +29,13 @@ var (
 		"The number of errors in the reconciler",
 		stats.UnitDimensionless)
 
-	// PipelineError metric measures the error by components when syncing a commit
+	// PipelineError metric measures the error by components when syncing a commit.
+	// Definition here must exactly match the definition in the resource-group
+	// controller, or the Prometheus exporter will error. b/247516388
+	// https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/metrics/metrics.go#L35
```
rquitales, line 35: Shouldn't we point to the resource-group controller metric text (https://github.com/GoogleContainerTools/kpt-resource-group/blob/ede29f6f10dca881ffe63b86acf5ca0e23ba6c5f/controllers/metrics/metrics.go#L88) instead?

karlkfi: lol oops. good catch

rquitales (approved):

tiffanny29631 (changes requested):

pkg/metrics/metrics.go
```diff
@@ -29,10 +29,13 @@ var (
...
 
-	// PipelineError metric measures the error by components when syncing a commit
+	// PipelineError metric measures the error by components when syncing a commit.
+	// Definition here must exactly match the definition in the resource-group
+	// controller, or the Prometheus exporter will error. b/247516388
+	// https://github.com/GoogleContainerTools/kpt-resource-group/blob/main/controllers/metrics/metrics.go#L88
 	PipelineError = stats.Int64(
 		"pipeline_error_observed",
-		"A boolean indicates if any error happens from different stages when syncing a commit",
+		"A boolean value indicates if error happened at readiness stage when syncing a commit",
```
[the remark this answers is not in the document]
karlkfi: :+1:

tiffanny29631 (approved):
/lgtm

karlkfi:
TestInvalidRepoSyncBranchStatus is broken. I've been trying to fix it in another PR but it's going to require filtering metrics more, which is not a quick fix...

PR #141, opened 2022-09-28
Make client-side timeout when talking to API Server configurable

tomasaschan:
We have a few hundreds of CRDs in our clusters, and are seeing Config Sync reconciliation fail with API Server timeouts due to the client-side timeout parameter being too aggressive.

This makes the client-side timeout for API server requests configurable under `.spec.override` (next to things like `gitSyncDepth` and `statusMode`; if you'd like it to be somewhere else, feel free to recommend something and I'll adjust the PR). I tried to mimic patterns already in use for other parameters, both in implementation and tests (but oh, god, does this test suite need some work...!) so I hope it looks OK.

I [added](https://github.com/GoogleContainerTools/kpt-config-sync/pull/141/files#diff-96d9758339603eab1bf5deb9d64abf4acb751b8474cd85e2cd4d790a76514291R2384-R2480) two [new tests](https://github.com/GoogleContainerTools/kpt-config-sync/pull/141/files#diff-4e452cd48e11abd5a08fa09505577300d00a536c7c57d7ebc492a9b166be3f55R2783-R2878) for the actual expansion of config into environment variables, because I realized that all the tests of the public API that already existed [were using](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/reposync_controller_test.go#L329) the [production implementation](https://github.com/GoogleContainerTools/kpt-config-sync/blob/5c250076186a138434505dbe54fa425a08a78189/pkg/reconcilermanager/controllers/reposync_controller.go#L670) also to create the expected results, which in practice means that a bug there would never be exposed by the tests. There's room to add a lot more test cases in these two, to ensure that the set of expected env vars in the all the other tests match the intention. I'll leave that work for someone else 😅 

A couple of questions:

* Given that 1.13 was released quite recently, I expect that 1.14 is still a few weeks out. What's the easiest way I can re-generate all the manifets and locally build a Docker image to use in a deployment of Config Sync to test this out in one of our clusters, and see if it works?
* I've run `make generate`; are there any other code-generation commands I should run?
* Should I - and if so, how do I - write a changelog entry somewhere?

### DIFF
--- a/pkg/api/configsync/v1alpha1/resource_override.go
+++ b/pkg/api/configsync/v1alpha1/resource_override.go
@@ -54,6 +54,14 @@ type OverrideSpec struct {
 	// +optional
 	ReconcileTimeout *metav1.Duration `json:"reconcileTimeout,omitempty"`
 
+	// apiServerTimeout allows one to override the client-side timeout for requests to the API server.
+	// Default: 5s.
+	// Use string to specify this field value, like "30s", "1m".
+	// More details about valid inputs: https://pkg.go.dev/time#ParseDuration.
+	// Recommended apiServerTimeout range is from "3s" to "1m".
+	// +optional
+	APIServerTimeout *metav1.Duration `json:"apiServerTimeout,omitempty"`
+
 	// enableShellInRendering specifies whether to enable or disable the shell access in rendering process. Default: false.
 	// Kustomize remote bases requires shell access. Setting this field to true will enable shell in the rendering process and
 	// support pulling remote bases from public repositories.
--- a/pkg/api/configsync/v1beta1/resource_override.go
+++ b/pkg/api/configsync/v1beta1/resource_override.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/client/restconfig"
 )
 
 // OverrideSpec allows to override the settings for a reconciler pod
@@ -95,6 +96,14 @@ type ContainerResourcesSpec struct {
 func GetReconcileTimeout(d *metav1.Duration) string {
 	if d == nil || d.Duration == 0 {
 		return configsync.DefaultReconcileTimeout.String()
+	}
+	return d.Duration.String()
+}
+
+// GetAPIServerTimeout returns the API server timeout in string, defaulting to 5s if empty
+func GetAPIServerTimeout(d *metav1.Duration) string {
+	if d == nil || d.Duration == 0 {
+		return restconfig.DefaultTimeout.String()
 	}
 	return d.Duration.String()
 }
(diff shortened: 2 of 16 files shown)


pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -54,6 +55,14 @@ type OverrideSpec struct {
...
 	ReconcileTimeout *metav1.Duration `json:"reconcileTimeout,omitempty"`
 
+	// apiServerTimeout allows one to override the client-side timeout for requests to the API server.
+	// Default: 5s.
+	// Use string to specify this field value, like "30s", "1m".
+	// More details about valid inputs: https://pkg.go.dev/time#ParseDuration.
+	// Recommended apiServerTimeout range is from "3s" to "1m".
+	// +optional
+	APIServerTimeout *metav1.Duration `json:"apiServerTimeout,omitempty"`
+
```
haiyanmeng, line 65: Please also update the `pkg/api/configsync/v1alpha1` package. We try to keep the v1alpha1 and v1beta1 APIs the same.

After updating the `pkg/api/configsync/v1alpha1` package, please run `make configsync-crds` to update the RootSync|RepoSync CRDs.

tomasaschan: Thanks; done!

tomasaschan:
The e2e test that failed before seems to have passed on retest without any changes on my part. Is it just flaky?

pkg/api/configsync/v1alpha1/resource_override.go
```diff
@@ -54,6 +54,14 @@ type OverrideSpec struct {
 	// +optional
 	ReconcileTimeout *metav1.Duration `json:"reconcileTimeout,omitempty"`
 
+	// apiServerTimeout allows one to override the client-side timeout for requests to the API server.
```
haiyanmeng, line 57: You need to run `make generate` here to update `pkg/api/configsync/v1alpha1/zz_generated.deepcopy.go`

tomasaschan: Thanks!

haiyanmeng:
> * Given that 1.13 was released quite recently, I expect that 1.14 is still a few weeks out. What's the easiest way I can re-generate all the manifets and locally build a Docker image to use in a deployment of Config Sync to test this out in one of our clusters, and see if it works?

You can build and install Config Sync from source following https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/docs/installation.md#building-and-installing-from-source.

> * Should I - and if so, how do I - write a changelog entry somewhere?

We don't have a changelog.

nan-yu (commented):
/lgtm

The changes in this PR look good. The default client-side timeout is also used in the `nomos` CLI. Do you plan to update the `nomos` as well?

tomasaschan:
> The changes in this PR look good. The default client-side timeout is also used in the nomos CLI. Do you plan to update the nomos as well?

I didn't have any plans to, at this time, no. We have too many syncs for e.g. `nomos status` to be particularly useful, so I don't use it much - but when I do, I use it from my local environment, which typically will have an already warmed-up API discovery cache from previous `kubectl` usage, which means it's probably a rare problem in practice even for users who use `nomos` more regularly.

Of course, if you'd like `nomos` to support this for consistency, I can add that in here anyway.

nan-yu:
> Of course, if you'd like nomos to support this for consistency, I can add that in here anyway.

Yes, we would like the CLI to have the same behavior as the reconciler. It's okay to make the change in a separate commit.

nan-yu (approved):

PR #142, opened 2022-09-28
Use kubectl to deploy necessary RBAC in mono repo apiservice_test (#73)

nan-yu:
The mentioned RBAC includes necessary ClusterRole and ClusterRoleBindings for the backend deployment to work in GKE environment. In Config Sync CI after each test it will try to reset the test git repo to default. Removing this RBAC from git directly is blocked by importer as it detects the commit of removal will reduce the count of cluster resource from more than 1 down to 0.

Thus manually deploying RBAC when test is running in mono repo mode.

### DIFF
--- a/e2e/testcases/apiservice_test.go
+++ b/e2e/testcases/apiservice_test.go
@@ -35,14 +35,24 @@ const (
 func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
 	nt := nomostest.New(t, ntopts.Unstructured, ntopts.RequireGKE(t))
 	t.Cleanup(func() {
+		nt.MustKubectl("delete", "-f", "../testdata/apiservice/rbac.yaml", "--ignore-not-found")
+		nt.MustKubectl("delete", "-f", "../testdata/apiservice/namespace.yaml", "--ignore-not-found")
 		if t.Failed() {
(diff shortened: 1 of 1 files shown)


karlkfi:
Note: This is a hack to fix a problem that's already been fixed on the main branch in a more robust way, with a 'safety' ClusterRole. Let's make sure this stays only on the v1.13 branch.

nan-yu:
@tiffanny29631 Can you update the test on the main branch given that we have already had a `safety` ClusterRole? The ClusterRole prevents from deleting more than 1 existing cluster-scoped resources from the cluster in one shot.

karlkfi:
@nan-yu I dont think main need updating. The new safety ClusterRole I added ALLOWS tests to delete all their cluster-scoped resources. I added it so that individual tests don't need to do this dance.

nan-yu:
re: https://github.com/GoogleContainerTools/kpt-config-sync/pull/142#issuecomment-1261344293

>  Note: This is a hack to fix a problem that's already been fixed on the main branch in a more robust way, with a 'safety' ClusterRole. Let's make sure this stays only on the v1.13 branch.

The same change is already in the main branch before the ClusterRole fix was merged. Using `kubectl apply` bypasses the safety check issue. We can switch back to use Config Sync to manage the RBAC, but it is not worth it.

PR #143, opened 2022-09-28
Update TestHelmNamespaceRepo to make it works in shared test environment (#75)

nan-yu:
Change the RepoSync to sync from original git source at the end of test to make test works in the shared test environment.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -82,11 +82,9 @@ func TestHelmNamespaceRepo(t *testing.T) {
 		Auth:    configsync.AuthNone,
 		Version: publicHelmChartVersion,
 	}}
-	rs.Spec.Git = nil
 	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from a public Helm Chart with cluster-scoped type")
 	nt.WaitForRepoSyncSourceError(repoSyncNN.Namespace, repoSyncNN.Name, nonhierarchical.BadScopeErrCode, "must be Namespace-scoped type")
-
 	nt.T.Log("Fetch password from Secret Manager")
 	key, err := gitproviders.FetchCloudSecret("config-sync-ci-ar-key")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


xinnywinne (approved):
/lgtm

e2e/testcases/helm_sync_test.go
```diff
@@ -106,14 +104,16 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
 	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from a private Helm Chart without cluster scoped resources")
 	nt.WaitForRepoSyncs(nomostest.WithRepoSha1Func(helmChartVersion(privateNSHelmChart+":"+privateNSHelmChartVersion)), nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{repoSyncNN: privateNSHelmChart}))
 	if err := nt.Validate(rs.Spec.Helm.ReleaseName+"-"+privateNSHelmChart, testNs, &appsv1.Deployment{}); err != nil {
 		nt.T.Error(err)
 	}
+	// Change the RepoSync to sync from the original git source to make test works in the shared test environment.
+	rs.Spec.SourceType = string(v1beta1.GitSource)
+	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from the original git repository")
```
tiffanny29631, line 116: WaitForRepoSyncs for this step?

tiffanny29631: Would resetting .Spec.Helm to nil be safer

tiffanny29631 (commented):
/lgtm

e2e/testcases/helm_sync_test.go
sdowell: This is a cherry pick - we shouldn't alter the diff if possible

nan-yu: This is the cherry-pick PR and we would like it to be consistent with the commit on the main branch.

> WaitForRepoSyncs for this step?

Xin has moved this snippet to a cleanup block in #93. It is not necessary to wait for repo syncs to be synced as the test infra will validate that after it reset everything.

> Would resetting .Spec.Helm to nil be safer

Not necessary. The `.spec.helm` field will be ignored when `.spec.sourceType` is not `helm`.

tiffanny29631: Ack thanks for the explanation

tiffanny29631 (approved):
/lgtm

PR #146, opened 2022-09-28
Update cert-controller from v0.2.0 to v0.5.0

haiyanmeng:
cert-controller v0.2.0 generates an error when a ValidatingWebhoookConfiguration|MutatingWebhookConfiguration object does not have the `webhooks` field or has an empty `webhooks` field, which prevents the Config Sync admission webhhok from being ready and causes log spamming for the Config Sync admission webhook until the `webhooks` field has one or more webhooks.

cert-controller v0.5.0 fixes this issue by avoid generating an error when a ValidatingWebhoookConfiguration|MutatingWebhookConfiguration object does not have the `webhooks` field or has an empty `webhooks` field.


tiffanny29631 (commented):
/lgtm

PR #147, opened 2022-09-28
use BUILD_ID as unique identifier for prow jobs

sdowell:
We currently pass in a custom USER variable to uniquely identify the image tags for each prow job. This switches the make target functionality to check for BUILD_ID from the environment and use that instead if present.


nan-yu (approved):
/lgtm

PR #148, opened 2022-09-29
redesign build related make targets

sdowell:
The intention of this change is to streamline the functionality of the make targets and make the naming more intuitive. A few key features:

- Make naming of targets more intuitive/consistent
- Use current gcloud context to determine default GCP project. This helps support the model of a configurable dev project.
- Use git describe to tag images instead of using timestamp in the image prefix with a latest tag. This makes it so that images can be separately built and published in an intuitive way. Using git describe in the tag version also provides more contextual information than a timestamp.
- Replace build-oss with config-sync-manifest. The build-oss script was an initial take at building OSS Config Sync, but duplicates logic from config-sync-manifest. The config-sync-manifest target is also what we use now for producing release artifacts, so this makes the development target consistent with the release target.
- Update development docs to provide more instruction on interacting with the make build targets.


Makefile.oss
```diff
@@ -12,21 +12,9 @@
...
-
-$(OSS_MANIFEST_DIR): $(OUTPUT_DIR)
-	@echo "+++ Creating the local output dir for manifests: $(OSS_MANIFEST_DIR)"
-	@mkdir -p \
-		$(OSS_MANIFEST_DIR)
-
-.PHONY: build-oss
-build-oss: $(OSS_MANIFEST_DIR) pull-buildenv
-	MANIFEST_DIR=$(OSS_MANIFEST_DIR) ./scripts/build-oss.sh
+OSS_MANIFEST_DIR := $(OUTPUT_DIR)/staging/oss
```
nan-yu, line 16: Do we need to create the folder if it doesn't exist?

sdowell: That's already handled [here](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/Makefile#L180)

Makefile.build
```diff
@@ -262,6 +273,10 @@ __config-sync-manifest-oss: __config-sync-manifest
 	rsync \
 		.output/manifests/acm-psp.yaml ${OSS_MANIFEST_STAGING_DIR}/acm-psp.yaml
 
+# Deprecated alias of build-manifests. Remove this once unused.
```
karlkfi, line 276: What's this still used by?

scripts/get-docker-tag.sh
```diff
@@ -13,22 +13,10 @@
...
-# Get the upstream tracking branch.  Note that this needs to work in multiple
-# checkout/fetch environments.  See b/159654901 for the various changes that we
-# tried before settling on this.
-BRANCH=$(git symbolic-ref --short --quiet HEAD);
-if [[ "${BRANCH}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-  echo "${BRANCH}-latest"
+describe=$(git describe --tags --always --dirty --long)
+if [[ "${describe}" == *-dirty ]]; then
+  # Append hash of diff to help enable unique dirty repo states
+  echo "${describe}-$(git diff | sha1sum | head -c 7)"
```
karlkfi, line 19: FWIW, docker tags have a limit of 128 chars, so this means we have a branch name limit of 128-22 chars. Do we want to have some sort of truncation? Error?

Makefile.build
sdowell: There are some CI jobs that use it. They'll be moved to the new target and this target can be removed once it's unused

scripts/get-docker-tag.sh
sdowell: I'm pretty sure this will only choose tags (not branches unless the [--all](https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---all) option is provided)

Makefile.build
```diff
@@ -48,68 +48,72 @@ build-cli: pull-buildenv buildenv-dirs
 		--version $(VERSION) \
 		$(PLATFORMS)
 
-# Creates a docker image for the specified nomos component.
-image-nomos: license
+# Build Config Sync docker images
+build-images: license
```
tomasaschan, line 52: Would it make sense to split out separate targets for these, and make this one just depend on those?

Makefile.oss
```diff
@@ -12,21 +12,6 @@
...
-OSS_MANIFEST_DIR := $(OUTPUT_DIR)/oss/manifests
-
-$(OSS_MANIFEST_DIR): $(OUTPUT_DIR)
-	@echo "+++ Creating the local output dir for manifests: $(OSS_MANIFEST_DIR)"
-	@mkdir -p \
-		$(OSS_MANIFEST_DIR)
-
-.PHONY: build-oss
-build-oss: $(OSS_MANIFEST_DIR) pull-buildenv
-	MANIFEST_DIR=$(OSS_MANIFEST_DIR) ./scripts/build-oss.sh
```
tomasaschan, line 28: The expectation is now that OSS builds also just use `make build-images`, right?

docs/development.md
```diff
@@ -31,23 +31,57 @@ make test-e2e-go-ephemeral-multi-repo
...
 ```
-make build-oss
+make config-sync-manifest
 ```
 
 This will build all the docker images needed for Config Sync and generate
 the manifests needed to run it. The images will by default be uploaded to 
 Google Container Registry under your current gcloud project and the manifests
-will be created in .output/oss/manifests under the Config Sync directory.
+will be created in .output/staging/oss under the Config Sync directory.
```
tomasaschan, line 43: In #189 I explored using Kustomize to bundle all of the OSS manifests into a single one, to make it easier to e.g. copy out into some other location and make sure you got everything. Is it worth including that approach here?

Also, when writing that PR I noticed that [one of the required manifests](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/ns-reconciler-cluster-role.yaml) was not included. I don't see anything addressing that in this PR!

Makefile.oss
sdowell: Yeah OSS builds would use:
- `build-images` to just build the docker images
- `build-manifests` to just build the manifests
- `(pull|push|retag)-images` to optionally manipulate docker images

`config-sync-manifest` is a convenience target that behaves similar to `build-oss` which will build manifests and build/push images.

The previous behavior by default used a timestamp in the registry prefix, so you couldn't separately run the targets without overriding the `REGISTRY` variable. This change makes it so that the default behavior allows the user to separately run each target at a given commit. This is because the image tag will default to `git describe`, so it will use the same tag for the same SHA.

docs/development.md
sdowell: This target is already using the [append_manifests](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/Makefile.build#L257) to consolidate into a single manifest. I think we could consider replacing append_manifests with kustomize as you proposed but that should be a separate change.

This target also already includes `ns-reconciler-cluster-role`

scripts/get-docker-tag.sh
sdowell: After thinking about this a bit more I think there might be value in removing the extra hash and treating `dirty` as a floating tag. This would enable users to make experimental changes and just rebuild the part they want to rebuild.

I'm imagining a use case like:
- User has built images and manifests on a given dirty commit
- User wants to make a change to just the manifests, so updates the manifests and runs `build-manifests`

This way the user can choose to use the previously built docker images without needing to rebuild them for the new manifests. If the user wants to rebuild both then they can still choose to run both `build-images` and `build-manifests`. I suppose they also could have run `retag-images` and `build-manifests` to avoid rebuilding, but that seems a bit less intuitive to me.

@karlkfi WDYT?

karlkfi: You can probably get the best of both worlds by using a tombstone that contains the names & tags of the last build images in .env format. That way you can rebuild the manifests without regenerating the image tags.

sdowell:
@tomasaschan 

There aren't any blockers AFAIK, just waiting for code review.

scripts/get-docker-tag.sh
sdowell: I switched this back to treating `-dirty` as a floating tag, since I expect that to be useful in development workflows. I think we can iterate on this some more after getting feedback.

Makefile.build
sdowell: Ah do you mean that you'd like to have the ability to build the images without running license? I could try that, although I think there's an assumption in the docker build that the license file exists.

nan-yu (approved):
/lgtm

Makefile.build
tomasaschan: I did mean separate targets for each image, so it's possible to e.g. `make build-reconciler-image` or something, and not have to build all images every time. But it's not super important :)

PR #149, opened 2022-09-29
run tidy as part of presubmit check

sdowell:
The license script runs tidy with -compat=1.17 flag, which results in a different set of modules from what is currently committed to the repo. This adds license as a presubmit check to ensure the command does not result in a dirty repo.


sdowell:
I don't have all the context as to why we run tidy with `-compat=1.17` flag in the `license` target, but it doesn't seem right that it currently results in a dirty repo. e.g. running `image-nomos` from a clean repo currently results in a dirty repo.

karlkfi:
I added -compat=1.17 because something didnt work if I didn't, but I don't remember what it was. We need to consistently either use -compat=1.17 or not use it anywhere.

karlkfi:
I suggest adding a make tidy:
```
tidy:
	go mod tidy
```

Then remove `.DEFAULT_GOAL := test` and add an all target:
```
all: test tidy
```

Then change `test-presubmit`to run `all` instead of just `test`.

That way devs just run `make` and the all the presubmit does in addition is validate all the automated changes were committed.

karlkfi:
Then remove the -compat=1.17 from the license script and see if ti all works. If it doesn't, add it back and also to make tidy.

karlkfi:
Correction: you need the compat in both places because all the dockerfiles are still FROM golang:1.17

karlkfi:
Once all those changes are in tho, you can make a new PR to upgrade Go in the dockerfiles.

sdowell:
@karlkfi Updated with a separate `tidy` target which is a dependency of both `license` and `test-presubmit`. PTAL

scripts/prepare-licenses.sh
```diff
@@ -27,8 +27,6 @@ licenseRegex="\(LICEN\(S\|C\)E\|COPYING\|README\|NOTICE\)"
 licenses="${tmp}"/LICENSES.txt
 
 # Ensure the vendor directory only includes the project's build dependencies.
-go mod tidy -compat=1.17
-go mod vendor
```
karlkfi, line 31: historically, the vendor was required here to make sure the licenses were in the vendor dir to aggregate together. However, now that the vendor dir is committed in the repo, we do need to make sure we do a go mod vendor before aggregating licenses and before checking for no diffs. So if we remove it here, it needs a top level make target to be a dependency.

sdowell: It is run as part of the `tidy` target

Makefile
```diff
@@ -221,7 +221,7 @@ test: test-unit lint
 # The presubmits have made side-effects in the past, which makes their
 # validation suspect (as the repository they are validating is different
 # than what is checked in).  Run `test` but verify that repository is clean.
-test-presubmit: test
+test-presubmit: test tidy
```
karlkfi, line 224: Can you add an `all` target with the tidy so we get it automatically without having to run `make tidy` or `make test-presubmit`? Then remove the default so it reverts to make all being default instead of make test.

scripts/prepare-licenses.sh
karlkfi: Ahh. I missed that... I don't know if that makes sense to me. I would expect a different `make vendor` to do that and then use tidy and vendor targets as needed...

But if you want one to do both, maybe name it `make deps`?

Makefile
sdowell: I'm not saying that actually should be the behavior, just that it's what I might expect if I was running a `make all` target without context. I don't feel that strongly about it so I can make the change you suggest.

Take a look at https://github.com/GoogleContainerTools/kpt-config-sync/pull/148 for some more changes aimed at improving the developer experience with the make targets

sdowell: Done

Makefile
```diff
@@ -167,7 +167,7 @@ endif
 
 ##### SETUP #####
 
-.DEFAULT_GOAL := test
+.DEFAULT_GOAL := all
```
karlkfi, line 170: You can delete this, I think? All is the default default goal.

karlkfi (approved):
/lgtm

PR #153, opened 2022-09-29
Support passing number only values as string directly through helm.va…

xinnywinne:
…lues

Currently, when users need to pass in number only value as string through helm.values, they
have to set the value as "\\"1234\\"". After this change, user will be able to pass in the number only value directly as "1234".


nan-yu (approved):
/lgtm

PR #154, opened 2022-09-29
Fix InvalidSourceType error message (#76)

xinnywinne:
Add "helm" to error message: must specify spec.sourceType to be one of "git", "oci", "helm"


nan-yu (approved):
/lgtm

PR #156, opened 2022-09-30
Filter some metrics by pod name for testing

karlkfi:
- Fix some failing tests that were erroring because their expectations
  were assuming all the metrics for the same deployment would be
  reported with the same tags. Now they filter by pod names instead of
  delpoyment name, using the latest reconciler pod for that deployment.
- Add debug logging for SyncMetricOptions
- Fix status updates that were being skipped too agressively.
  They now record metrics and status if the status has never been
  updated before.
- Remove unnecessary metrics diff on update
- Add NONE value for commit tag in last_sync_timestamp and
  last_apply_timestamp metrics, to work around a bug in the
  otel-collector which causes invalid output by the Prometheus exporter.
  Without this fix, the commit value gets parsed to the status value, and the status
  tag is ignored as invalid.
- Fix metric expectation in TestCRDDeleteBeforeRemoveCustomResourceV1Beta1
  and TestCRDDeleteBeforeRemoveCustomResourceV1 to expect source but not
  sync error. Only non-blocking source errors cause sync errors. Blocking source
  errors cause early exit, which skips reporting of sync status.
- Fix metric expectation in TestInvalidRepoSyncBranchStatus to expect
  source error in the RepoSync, not the RootSync.

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -26,6 +26,7 @@ import (
 	"sync"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -37,9 +38,7 @@ import (
 	"kpt.dev/configsync/pkg/rootsync"
 	"kpt.dev/configsync/pkg/util/repo"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
-	"go.opencensus.io/tag"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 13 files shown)


e2e/nomostest/nt.go
```diff
@@ -581,8 +583,8 @@ func (nt *NT) ValidateMetricNotFound(metricName string) error {
 }
 
 // ValidateReconcilerErrors validates that the `reconciler_error` metric exists
-// for the correct reconciler and the tagged component has the correct value.
-func (nt *NT) ValidateReconcilerErrors(reconcilerName, component string) error {
+// for the correct reconciler pod and the tagged component has the correct value.
+func (nt *NT) ValidateReconcilerErrors(podName, component string) error {
```
sdowell, line 587: Since the consumer effectively has to always call `nt.GetDeploymentPod` before calling `nt.ValidateReconcilerErrors`, could we accept `reconcilerName` in this function and handle the mapping (i.e. call `nt.GetDeploymentPod`) in this function?

e2e/nomostest/nt.go
```diff
@@ -1555,16 +1593,18 @@ func SyncMetricsToLatestCommitSyncedWithSuccess(nt *NT) MetricsSyncOption {
 }
 
 // SyncMetricsToReconcilerSourceError syncs metrics to a reconciler source error
-func SyncMetricsToReconcilerSourceError(reconcilerName string) MetricsSyncOption {
+func SyncMetricsToReconcilerSourceError(nt *NT, reconcilerName string) MetricsSyncOption {
 	return func(metrics *testmetrics.ConfigSyncMetrics) error {
-		return metrics.ValidateReconcilerErrors(reconcilerName, 1, 0)
+		pod := nt.GetDeploymentPod(reconcilerName, configmanagement.ControllerNamespace)
+		return metrics.ValidateReconcilerErrors(pod.Name, 1, 0)
```
sdowell, line 1599: Perhaps `nt.ValidateReconcilerErrors(pod.Name, "source")`

e2e/nomostest/nt.go
```diff
@@ -1555,16 +1593,18 @@ func SyncMetricsToLatestCommitSyncedWithSuccess(nt *NT) MetricsSyncOption {
...
 	}
 }
 
 // SyncMetricsToReconcilerSyncError syncs metrics to a reconciler sync error
-func SyncMetricsToReconcilerSyncError(reconcilerName string) MetricsSyncOption {
+func SyncMetricsToReconcilerSyncError(nt *NT, reconcilerName string) MetricsSyncOption {
 	return func(metrics *testmetrics.ConfigSyncMetrics) error {
-		return metrics.ValidateReconcilerErrors(reconcilerName, 0, 1)
+		pod := nt.GetDeploymentPod(reconcilerName, configmanagement.ControllerNamespace)
+		return metrics.ValidateReconcilerErrors(pod.Name, 0, 1)
```
sdowell, line 1607: Perhaps `nt.ValidateReconcilerErrors(pod.Name, "sync")`

e2e/testcases/custom_resources_test.go
```diff
@@ -119,7 +120,8 @@ func TestCRDDeleteBeforeRemoveCustomResourceV1Beta1(t *testing.T) {
 
 	err = nt.ValidateMetrics(nomostest.SyncMetricsToLatestCommit(nt), func() error {
 		// Validate reconciler error metric is emitted.
-		return nt.ReconcilerMetrics.ValidateReconcilerErrors(nomostest.DefaultRootReconcilerName, 1, 1)
+		pod := nt.GetDeploymentPod(nomostest.DefaultRootReconcilerName, configmanagement.ControllerNamespace)
+		return nt.ReconcilerMetrics.ValidateReconcilerErrors(pod.Name, 1, 1)
```
sdowell, line 124: Is there a generic case we can add to `nt.ValidateReconcilerErrors` to support `(1, 1)`?

e2e/nomostest/nt.go
karlkfi: never really liked this wrapper with a switch statement and hard coded strings... I just replaced it with a `ValidateReconcilerErrors(reconcilerName string, sourceCount, syncCount int)` that does the pod lookup for you. Now GetDeploymentPod is only used in 3 places.

karlkfi:
/hold

This was supposed to be the "quick fix", instead of my metrics validation rewrite. But it looks like this isn't enough. We also need to wait for the `last_sync_timestamp` with the right commit, otherwise the `reconciler_errors` validation will still sometimes fail. So I'll port that over too.

karlkfi:
Looks like the metric is getting sent but the validation isn't seeing it. Trying to figure out why.

karlkfi:
/unhold

I think I fixed it.

karlkfi:
Once more with feeling...

karlkfi:
/test kpt-config-sync-presubmit-e2e-mono-repo-test-group2

Ahh, the classic `ERROR: status.sync.inProgress contains changes that haven't been synced`. :cheff-kiss:

sdowell:
🤌

karlkfi:
/test kpt-config-sync-presubmit-e2e-multi-repo-test-group2

Looks like an infra issue with not enough resources for the pods to come up before timeout.

karlkfi:
/test kpt-config-sync-presubmit-e2e-multi-repo-test-group3

New looking and probably unrelated issue in TestConflictingDefinitions_RootToRoot:

```
ERROR: expected Role test-ns/pods to NOT have resourceVersion "1174", but got "1174"
```

haiyanmeng (approved):

PR #157, opened 2022-09-30
make prober creds optional in go-gke-ci target

sdowell:
This change makes it so that prober creds are used by default, but can be overridden with an empty string at runtime to not use them. This is intended to enable a transition to workload identity for the prow jobs.


nan-yu (approved):
/lgtm

PR #160, opened 2022-09-30
Fix the test failure for TestResourceConditionAnnotations (#88)

nan-yu:
TestResourceConditionAnnotations failed on clusters with K8s 1.22+ because the v1beta1 API for the RBAC resources are deprecated and the k8s-api-server only returns one copy of the RBAC resource.

b/240334486

PR #161, opened 2022-09-30
Fix cleanup for Gatekeeper tests (#84)

nan-yu:
b/240334486

Change-Id: I11a5e58e9c4b5653be80d5dad59f93c22de273bd


nan-yu:
This PR needs to be cherrypicked to v1.13 branch only because it is a dependency of PR #160. To avoid cherrypick conflict, this needs to be cherrypicked as well.

PR #170, opened 2022-10-03
Increase the wait timeout and add retries to a mono-repo test

nan-yu:
This is to stabilize one flaky mono-repo test.

### DIFF
--- a/e2e/testcases/resource_conditions_test.go
+++ b/e2e/testcases/resource_conditions_test.go
@@ -39,6 +39,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var waitTimeout = time.Minute
+
 func TestResourceConditionAnnotations(t *testing.T) {
 	// TODO: Re-enable this test if/when multi-repo supports resource condition annotations.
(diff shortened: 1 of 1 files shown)


haiyanmeng (approved):

e2e/testcases/resource_conditions_test.go
```diff
@@ -140,14 +145,18 @@ func TestResourceConditionAnnotations(t *testing.T) {
 		`configmanagement.gke.io/errors-`)
 
 	// Ensure error conditions are removed.
-	_, err1 = nomostest.Retry(20*time.Second, func() error {
-		return nt.Validate(repo.DefaultName, "", &v1.Repo{},
+	_, err1 = nomostest.Retry(waitTimeout, func() error {
+		return nt.Validate("repo", "", &v1.Repo{},
```
karlkfi, line 149: Do we expect delay between these status updates? Or would it be sufficient to wrap them all in the same retry?

nan-yu: Oops, it was merged before I saw the comment. I think wrap them all in the same retry is a good suggestion. We can make a separate PR.

PR #169, opened 2022-10-03
Force to enable client-side throttling for mono-repo mode

nan-yu:
https://github.com/GoogleContainerTools/kpt-config-sync/commit/732f94ea702a9a2f97335f0000306f4e17b363ee disabled client-side throttling, which then caused the monorepo CI jobs to start becoming flaky and then constantly failing.

This commit forces to enable the client-side throttling for mono-repo only. Compared with 1.12.x releases, the client-side throttling QPS was
increased from 20 (burst: 40) to 30 (burst: 60).

PR #171, opened 2022-10-03
Force to enable client-side throttling for mono-repo mode (#169)

nan-yu:
https://github.com/GoogleContainerTools/kpt-config-sync/commit/732f94ea702a9a2f97335f0000306f4e17b363ee disabled client-side throttling, which then caused the monorepo CI jobs to start becoming flaky and then constantly failing.

This commit forces to enable the client-side throttling for mono-repo only. Compared with 1.12.x releases, the client-side throttling QPS was
increased from 20 (burst: 40) to 30 (burst: 60).

### DIFF
--- a/cmd/monitor/main.go
+++ b/cmd/monitor/main.go
@@ -31,7 +31,7 @@ func main() {
 	log.Setup()
 	ctrl.SetLogger(klogr.New())
 
-	cfg, err := restconfig.NewRestConfig(restconfig.DefaultTimeout)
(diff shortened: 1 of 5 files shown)


haiyanmeng (approved):

nan-yu:
Failed because of `There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (skip schedule deleting pod: test-pods/9fe15e30-4356-11ed-9500-7af23e4dfda4)`.

/retest

PR #173, opened 2022-10-03
only create e2e psp resources on psp clusters

sdowell:
PSP is removed as of k8s 1.25, causing tests to fail when they try to apply these PSP resources on 1.25 clusters. This change intends to make all PSP-related resources conditionally created based on whether the cluster has PSP enabled. We currently perform this check by checking whether psp is in the cluster name.

### DIFF
--- a/e2e/nomostest/git-server.go
+++ b/e2e/nomostest/git-server.go
@@ -119,13 +119,18 @@ func isAvailableDeployment(o client.Object) error {
 func gitServer() []client.Object {
 	// Remember that we've already created the git-server's Namespace since the
 	// SSH key must exist before we apply the Deployment.
-	return []client.Object{
-		gitPodSecurityPolicy(),
-		gitRole(),
-		gitRoleBinding(),
+	objs := []client.Object{
 		gitService(),
 		gitDeployment(),
 	}
+	if isPSPCluster() {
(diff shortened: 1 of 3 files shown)


e2e/nomostest/git-server.go
```diff
@@ -119,13 +119,18 @@ func isAvailableDeployment(o client.Object) error {
 func gitServer() []client.Object {
 	// Remember that we've already created the git-server's Namespace since the
 	// SSH key must exist before we apply the Deployment.
-	return []client.Object{
-		gitPodSecurityPolicy(),
```
nan-yu, line 123: Do we need the PodSecurityPolicy on non-PSP clusters? If so, we need to consider using other alternatives.

sdowell: I don't have all the context here, but I don't think we need these resources for non-PSP clusters. Looking back at the commit where this was introduced it seems this was just trying to make it so the git-server is deployable on PSP-enabled clusters.

The e2e/kind presubmits deploy the git server and shouldn't have PSP enabled. Given that, the presubmit tests are exercising this without PSP.

nan-yu: SG

nan-yu (approved):
/lgtm

PR #177, opened 2022-10-04
pass ARTIFACTS env var into kind container

sdowell:
the artifacts env var was not being passed into the container, causing some places which look for the env var to not produce artifacts at the correct location. For example, the kind cluster logs are emitted to the artifacts dir on failure.


sdowell:
Example run where the kind cluster logs did not make it to the correct artifacts dir:
https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/148/kpt-config-sync-presubmit-e2e-multi-repo-test-group2/1577094976393711616

haiyanmeng (approved):

PR #178, opened 2022-10-04
Support passing different types of values directly through helm.values

xinnywinne:
Instead of using helm template option --set, using option --values to set values.

After this change, users do not need to flatten the nested fields to set values, and will be able to pass special string values like "1234", "true" directly through helm.values

Add test cases to test helm.values field with different types of values.

### DIFF
--- a/e2e/nomostest/predicates.go
+++ b/e2e/nomostest/predicates.go
@@ -169,8 +169,8 @@ func HasExactlyLabelKeys(wantKeys ...string) Predicate {
 	}
 }
 
-// HasExactlyImageName ensures a container has the expected image name
-func HasExactlyImageName(containerName string, expectImageName string) Predicate {
+// HasExactlyImage ensures a container has the expected image.
+func HasExactlyImage(containerName, expectImageName, expectImageTag, expectImageDigest string) Predicate {
 	return func(o client.Object) error {
 		dep, ok := o.(*appsv1.Deployment)
 		if !ok {
@@ -178,15 +178,24 @@ func HasExactlyImageName(containerName string, expectImageName string) Predicate
 		}
 		for _, container := range dep.Spec.Template.Spec.Containers {
 			if containerName == container.Name {
-				if !strings.Contains(container.Image, "/"+expectImageName+":") {
-					return errors.Errorf(" Expected %q container image name is %q, however the actual image is %q", container.Name, expectImageName, container.Image)
+				expectImage := ""
+				if expectImageName != "" {
+					expectImage = "/" + expectImageName
+				}
+				if expectImageTag != "" {
+					expectImage += ":" + expectImageTag
+				}
+				if expectImageDigest != "" {
+					expectImage += "@" + expectImageDigest
+				}
+				if !strings.Contains(container.Image, expectImage) {
+					return errors.Errorf("Expected %q container image contains %q, however the actual image is %q", container.Name, expectImage, container.Image)
 				}
 				return nil
 			}
 		}
 		return errors.Errorf("Container %q not found", containerName)
 	}
-
 }
 
 // HasCorrectResourceRequestsLimits verify a root/namespace reconciler container has the correct resource requests and limits.
--- a/e2e/testcases/csr_auth_test.go
+++ b/e2e/testcases/csr_auth_test.go
@@ -290,7 +290,7 @@ func testWorkloadIdentity(t *testing.T, testSpec workloadIdentityTestSpec) {
 		nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "oci": {"dir": "%s", "image": "%s", "auth": "gcpserviceaccount", "gcpServiceAccountEmail": "%s"}, "git": null}}`,
 			v1beta1.OciSource, tenant, testSpec.sourceRepo, testSpec.gsaEmail))
 	case v1beta1.HelmSource:
-		nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": {"chart": "%s", "repo": "%s", "version": "%s", "auth": "gcpserviceaccount", "gcpServiceAccountEmail": "%s", "releaseName": "my-coredns", "namespace": "coredns", "values": {"image.pullPolicy": "Always"}}, "git": null}}`,
+		nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": {"chart": "%s", "repo": "%s", "version": "%s", "auth": "gcpserviceaccount", "gcpServiceAccountEmail": "%s", "releaseName": "my-coredns", "namespace": "coredns"}, "git": null}}`,
 			v1beta1.HelmSource, testSpec.sourceChart, testSpec.sourceRepo, testSpec.sourceVersion, testSpec.gsaEmail))
 	}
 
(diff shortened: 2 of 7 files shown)


pkg/helm/helm.go
```diff
@@ -87,14 +87,11 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
...
-	if err != nil {
-		return []string{}, fmt.Errorf("failed to unmarshal helm.values, error: %w", err)
-	}
-	for key, val := range values {
-		args = append(args, "--set", fmt.Sprintf("%s=%v", key, val))
+	valuesPath := "tmp/chart-values.yaml"
+	if err := ioutil.WriteFile(valuesPath, []byte(h.Values), 0644); err != nil {
+		return []string{}, fmt.Errorf("failed to create values file: %w", err)
 	}
+	args = append(args, "--values", valuesPath)
```
haiyanmeng, line 94: https://github.com/GoogleContainerTools/kpt-config-sync/blob/16fc6bafc040b5de3bb11110c495dc91bf0cddbf/cmd/helm-sync/main.go#L40 needs to be updated to reflect this change.

pkg/helm/helm.go
```diff
@@ -87,14 +87,11 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
...
 
 func (h *Hydrator) appendValuesArgs(args []string) ([]string, error) {
-	var values map[string]interface{}
-	err := json.Unmarshal([]byte(h.Values), &values)
-	if err != nil {
-		return []string{}, fmt.Errorf("failed to unmarshal helm.values, error: %w", err)
-	}
-	for key, val := range values {
-		args = append(args, "--set", fmt.Sprintf("%s=%v", key, val))
+	valuesPath := "tmp/chart-values.yaml"
```
rquitales, line 90: Instead of hard-coding a temp file, could we use `os.CreateTemp` to generate a temp file for us instead? This would allow Golang to determine the best directory to place the temp file as well as create globally unique filenames. I'm not sure if we need to worry about other processes writing to this file as well?

pkg/helm/helm.go
rquitales, line 94: Do we need to clean-up/remove the temp file after we consume it later on?

e2e/testcases/csr_auth_test.go
```diff
@@ -290,7 +290,7 @@ func testWorkloadIdentity(t *testing.T, testSpec workloadIdentityTestSpec) {
 		nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "oci": {"dir": "%s", "image": "%s", "auth": "gcpserviceaccount", "gcpServiceAccountEmail": "%s"}, "git": null}}`,
 			v1beta1.OciSource, tenant, testSpec.sourceRepo, testSpec.gsaEmail))
 	case v1beta1.HelmSource:
-		nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": {"chart": "%s", "repo": "%s", "version": "%s", "auth": "gcpserviceaccount", "gcpServiceAccountEmail": "%s", "releaseName": "my-coredns", "namespace": "coredns", "values": {"image.pullPolicy": "Always"}}, "git": null}}`,
```
haiyanmeng, line 293: Why `"values": {"image.pullPolicy": "Always"}` is removed here?

rquitales (changes requested):

pkg/helm/helm.go
xinnywinne: The reason to have a hard-coding temp file is that each helm sync container only sync one chart, and only need one values file. If retry/re-sync needed, the same values file should be overwrite instead of creating new temp file.

nan-yu: Instead of using a temp file, can we generate the values file under the current chart folder? With this approach, the entire folder will be removed when retrying or resyncing from a different chart.

If we go with the temp file, at least let golang decide which is the temp directory, `os.TempDir()`.

e2e/testcases/csr_auth_test.go
nan-yu: This RootSync object syncs from a private helm repo. I talked with Xin offline. To verify the `spec.helm.values` field, it's better to move it to the public repo so that the presubmit can capture any failures.

pkg/helm/helm.go
xinnywinne: Done

pkg/helm/helm.go
xinnywinne: Change to use os.TempDir() instead of using /tmp directly. Since we are using helm template instead of download the chart folder first and do the rendering later, this values file can not be put under chart folder. So will go with temp file

e2e/nomostest/predicates.go
```diff
@@ -169,24 +169,33 @@ func HasExactlyLabelKeys(wantKeys ...string) Predicate {
 	}
 }
 
-// HasExactlyImageName ensures a container has the expected image name
-func HasExactlyImageName(containerName string, expectImageName string) Predicate {
+// HasExactlyImage ensures a container has the expected image.
+func HasExactlyImage(containerName string, expectImageName string, expectImageTag string, expectImageDigest string) Predicate {
```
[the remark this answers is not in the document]
xinnywinne: Sounds good

e2e/nomostest/predicates.go
```diff
@@ -169,24 +169,33 @@ func HasExactlyLabelKeys(wantKeys ...string) Predicate {
 	}
 }
 
-// HasExactlyImageName ensures a container has the expected image name
-func HasExactlyImageName(containerName string, expectImageName string) Predicate {
+// HasExactlyImage ensures a container has the expected image.
+func HasExactlyImage(containerName string, expectImageName, expectImageTag, expectImageDigest string) Predicate {
```
nan-yu, line 173: nit: the first `string` type can be omitted as well.

pkg/helm/helm.go
```diff
@@ -87,14 +86,11 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
...
 
 func (h *Hydrator) appendValuesArgs(args []string) ([]string, error) {
-	var values map[string]interface{}
-	err := json.Unmarshal([]byte(h.Values), &values)
-	if err != nil {
-		return []string{}, fmt.Errorf("failed to unmarshal helm.values, error: %w", err)
-	}
-	for key, val := range values {
-		args = append(args, "--set", fmt.Sprintf("%s=%v", key, val))
+	valuesPath := filepath.Join(os.TempDir(), "chart-values.yaml")
```
nan-yu, line 89: nit: extract it to be a constant

nan-yu (commented):
LGTM with 2 minor comments.

e2e/nomostest/predicates.go
```diff
@@ -169,24 +169,33 @@ func HasExactlyLabelKeys(wantKeys ...string) Predicate {
...
+					expectImage = "/" + expectImageName
+				}
+				if expectImageTag != "" {
+					expectImage += ":" + expectImageTag
+				}
+				if expectImageDigest != "" {
+					expectImage += "@" + expectImageDigest
+				}
+				if !strings.Contains(container.Image, expectImage) {
+					return errors.Errorf(" Expected %q container image contains %q, however the actual image is %q", container.Name, expectImage, container.Image)
```
rquitales, line 192: nit: remove leading space in `" Expected %q ...`

pkg/helm/helm.go
rquitales: Per other comments, no clean up necessary it seems.

pkg/helm/helm.go
```diff
@@ -87,14 +86,11 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
...
+	if err := os.WriteFile(valuesPath, []byte(h.Values), 0644); err != nil {
+		return []string{}, fmt.Errorf("failed to create values file: %w", err)
```
rquitales, line 91: nit: return `nil` instead of initializing`[]string{}`.

pkg/helm/helm.go
xinnywinne: Current solution is creating and overwriting a specific value file. We will not create more than one temp file under the temp folder. So I think it is ok to not keep removing the temp file. And since it is a temp file, it will be removed when it is not needed.

e2e/nomostest/predicates.go
xinnywinne: done

e2e/nomostest/predicates.go
xinnywinne: good catch :)

pkg/helm/helm.go
xinnywinne: Sounds good.

pkg/helm/helm.go
xinnywinne: Done.

nan-yu (approved):
/lgtm

rquitales (approved):
/lgtm
/approve

PR #179, opened 2022-10-04
Wrap multiple retry blocks in the same one

nan-yu:


e2e/testcases/resource_conditions_test.go
```diff
@@ -179,40 +175,38 @@ func TestResourceConditionAnnotations(t *testing.T) {
...
-	if err1 != nil || err2 != nil || err3 != nil {
-		if err1 != nil {
-			nt.T.Error(err1)
-		}
-		if err2 != nil {
-			nt.T.Error(err2)
-		}
-		if err3 != nil {
-			nt.T.Error(err3)
+		if err1 != nil || err2 != nil || err3 != nil {
```
karlkfi, line 195: I think this if statement is redundant. But I suspect a MultiError would be better anyway, if we're going to wait for all of them to Validate before returning.

nan-yu: DONE. Switched to use MultiError.

karlkfi (approved):
/lgtm

Much cleaner, thank you!

PR #187, opened 2022-10-05
Fix e2e test cleanup to reset git branch

karlkfi:


karlkfi:
Wow. Looks like neither monorepo nor multirepo moves past the broken commit here. Gonna need to figure out some different way to reset. Or fix it so it can recover itself.

karlkfi:
/retest

nt.go:706: ERROR: status.conditions[1] is True: {"type":"Reconciling","status":"True","lastUpdateTime":"2022-10-05T20:18:56Z","lastTransitionTime":"2022-10-05T20:18:56Z","reason":"Deployment","message":"Available: 0/1","errorSummary":{}}

Looks like the opencensus port was still in use again. Flakey bug unrelated to this fix.

nan-yu (approved):
/lgtm

e2e/nomostest/config-sync.go
```diff
@@ -1387,3 +1400,25 @@ func toMetav1Duration(t time.Duration) *metav1.Duration {
 		Duration: t,
 	}
 }
+
+// SetGitBranch updates the git branch
+func SetGitBranch(nt *NT, syncName, branch string) {
+	nt.T.Logf("Change git branch to %q", branch)
+	if nt.MultiRepo {
+		rs := fake.RootSyncObjectV1Beta1(syncName)
+		nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"git": {"branch": "%s"}}}`, branch))
```
sdowell, line 1409: `resetGitBranch` deleted the `reconciler-manager` pod. Do we need to do that here as well?

karlkfi: Nah, the old func was only ever being called for mono-repo. So that code branch was never executed. For multi-repo updating the RSync is sufficient to have the reconciler-manager update the reconciler deployment, which replaces the pod automatically.

sdowell: Ack, thanks for explaining

karlkfi:
@googlebot: rescan

PR #188, opened 2022-10-05
Support passing different types of values directly through helm.value…

xinnywinne:
…s (#178)

Instead of using helm template option --set, using option --values to set values.

After this change, users do not need to flatten the nested fields to set values, and will be able to pass special string values like "1234", "true" directly through helm.values

Add test cases to test helm.values field with different types of values.


nan-yu (approved):
/lgtm

PR #189, opened 2022-10-05
Use Kustomize + Make to generate OSS manifests

tomasaschan:
Following [the instructions to build from source](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/docs/installation.md#building-and-installing-from-source), I found that the process was a bit cumbersome and error-prone, for a few reasons:

* Not all required manifests were included (the [ns reconciler cluster role](../ns-reconciler-cluster-role.yaml) was missing)
* The setup required me to build the images every time I exported manifests, even if I knew that nothing had changed that would render different images (which hurt iteration speed on my scripting downstream)
* The output was a jumble of files, with unclear/nonexistent naming conventions

Using Kustomize and moving most of the actual code into `Makefile.oss`, I was able to greatly simplify `scripts/build-oss.sh`, and also make it possible to run parts of that process directly through `make` (as long as `GCP_PROJECT` and `TAG` are provided). For example, you can now bundle the manifests required to install Config Sync with `make -C <repo root> oss-manifests GCP_PROJECT=<your project> TAG=latest`, without triggering Docker builds or pushes.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -51,56 +51,56 @@ build-cli: pull-buildenv buildenv-dirs
 # Creates a docker image for the specified nomos component.
 image-nomos: license
 	@gcloud $(GCLOUD_QUIET) auth configure-docker
-	@echo "+++ Building the nomos image"
+	@echo "+++ Building the nomos image: $(NOMOS_TAG)"
 	@docker build $(DOCKER_BUILD_QUIET) \
 		--target $(NOMOS_IMAGE) \
 		-t $(NOMOS_TAG) \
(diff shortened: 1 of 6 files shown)


sdowell:
Take a look at https://github.com/GoogleContainerTools/kpt-config-sync/pull/148 as I think it should solve most of the same problems, as well as de-duping the make target for OSS manifests. I do like your usage of kustomize here, I agree that it adds readability and simplicity

tomasaschan:
I'm gonna close this since #148 landed 🎉

PR #190, opened 2022-10-05
Explicitly clone into directory called kpt.dev

tomasaschan:
Because e.g. [this code](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/declared/value.go#L147) assumes that the repo is in kpt.dev, tests exercising this code path end up in an infinite loop (and eventually time out) unless the repo folder is called `kpt.dev`.

Of course, a better fix would be to remove this assumption, but I don't know where else that assumption might be present, so this is easier for now...


mortent:
This looks good. Thanks for the fix.
/approve

sdowell:
Good find, definitely a strange assumption.. I think this change should fix it https://github.com/GoogleContainerTools/kpt-config-sync/pull/192

PR #192, opened 2022-10-05
make openapi test file not depend on dir structure

sdowell:
This previously assumed that there was a kpt.dev parent directory to reach, which adds an unnecessary constraint to dev environment setups. This could probably be improved further, but this makes it so that the file can be found regardless of the parent directory structure.

### DIFF
--- a/pkg/declared/value.go
+++ b/pkg/declared/value.go
@@ -16,8 +16,6 @@ package declared
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 
 	openapiv2 "github.com/google/gnostic/openapiv2"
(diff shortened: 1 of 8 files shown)


nan-yu (commented):
The `ValueConverterForTest` seems to be used only by tests. Should we move it to the `testing` package?

sdowell:
> The `ValueConverterForTest` seems to be used only by tests. Should we move it to the `testing` package?

Agreed it doesn't seem like it's in the right place. Seems it was there because not all of the fields of `ValueConverter` were exported. I did some refactoring to enable moving it out to a separate package. PTAL

tomasaschan (approved):
Nice! /lgtm

xinnywinne (approved):
/lgtm

PR #175, opened 2022-10-03
Delete the membership after unregistration failure

nan-yu:

PR #193, opened 2022-10-05
Delete the membership after unregistration failure (#175)

nan-yu:


nan-yu:
Cherrypicking this CL to fix the lingering membership which caused the multi-repo CI failure.

PR #194, opened 2022-10-05
add terraform config for release clusters

sdowell:
these clusters are intended to be used for release branch periodic jobs.


nan-yu (commented):
I feel like we won't use the OSS project for v1.13 release, so we can switch to test groups directly for now.

sdowell:
> I feel like we won't use the OSS project for v1.13 release, so we can switch to test groups directly for now.

In that case I think we can hold off on this change until we have the v1.14 branch, since the clusters would be sitting idle in the meantime. WDYT?

PR #195, opened 2022-10-05
Fix a test failure on autopilot cluster

nan-yu:
The gcenode-askpass-sidecar container's resource requirements will be adjusted on the autopilot cluster, so the generation should bumpt by 2.

### DIFF
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -76,7 +76,7 @@ func TestManagingReconciler(t *testing.T) {
 		d.Spec.Template.Spec.Containers[0].Image = newImage
 	})
 	nt.T.Log("Verify the container image should be reverted by the reconciler-manager")
-	generation += 2
+	generation += 2 // generation bumped by 2 because the change will be first applied then reverted by the reconciler-manager
 	nomostest.Wait(nt.T, "the container image to be reverted", nt.DefaultWaitTimeout, func() error {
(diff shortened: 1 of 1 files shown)


e2e/testcases/reconciler_manager_test.go
```diff
@@ -150,7 +150,11 @@ func TestManagingReconciler(t *testing.T) {
 	nt.T.Log("Switch the auth type from none to gcpserviceaccount")
 	nt.MustMergePatch(rs, `{"spec":{"git":{"auth":"gcpserviceaccount","secretRef":{"name":""},"gcpServiceAccountEmail":"test-gcp-sa-email@test-project.iam.gserviceaccount.com"}}}`)
 	nt.T.Log("Verify the gcenode-askpass-sidecar container should be added")
-	generation++
+	if nt.IsGKEAutopilot {
+		generation += 2 // generation bumped by 2 because the sidecar container will first be added and then the resource requirements will be adjusted by autopilot
```
sdowell, line 154: Do we know that autopilot will always adjust the resource requirements exactly once?

nan-yu: Yes, the reconciler-manager will [keep the resource requirements adjusted by Autopilot](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/reconciler_base.go#L313). If it already follows the Autopilot rules, no more adjustment will be performed by Autopilot.

PR #197, opened 2022-10-05
[WIP] Use the server's preferred CRD version (monorepo)

karlkfi:
- This fixes a bug where the syncer would default to CRD v1, even on old clusters that prefer CRD v1beta1.
- This should help detection of compatibility issues if the CRD version is ever increased.

Depend on fake.Client changes from https://github.com/GoogleContainerTools/kpt-config-sync/pull/77. Extracted as a follow up.


karlkfi:
This is obsolete. The mono-repo code has been removed from this repo.

PR #198, opened 2022-10-06
Fix repoless.bats test cleanup

karlkfi:
Previous test teardown was resetting the POLICY_DIR to "acme" without having created the acme directory. So if the git-importer started up before the nomostest cleanup finished, it could cause the Repo to enter an unrecoverable error state.

The fix is to let nomostest handle the cleanup, because it already handles initializing the repo and directory before resetting the POLICY_DIR.


nan-yu (approved):
/lgtm

PR #203, opened 2022-10-06
fetch all branches in release tagging script

sdowell:
By default fetch will only fetch HEAD, i.e. main. For patch releases we need to also fetch release branches. This change updates the refspec to fetch all heads from remote.

### DIFF
--- a/scripts/tag-promotion.sh
+++ b/scripts/tag-promotion.sh
@@ -50,7 +50,7 @@ esac
 
 REMOTE="git@github.com:GoogleContainerTools/kpt-config-sync.git"
 # Fetch all existing tags
-git fetch "${REMOTE}" --tags > /dev/null
+git fetch "${REMOTE}" "${RC_TAG}" --tags > /dev/null
 echo "+++ Successfully fetched tags"
 
 echo "+++ Promoting ${RC_TAG} to ${CS_VERSION}"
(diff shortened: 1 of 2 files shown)


scripts/tag-promotion.sh
```diff
@@ -49,9 +49,9 @@ case "${choice}" in
 esac
 
 REMOTE="git@github.com:GoogleContainerTools/kpt-config-sync.git"
-# Fetch all existing tags
-git fetch "${REMOTE}" --tags > /dev/null
-echo "+++ Successfully fetched tags"
+# Fetch all existing tags/heads
+git fetch "${REMOTE}" +refs/heads/*:refs/remotes/upstream/* --tags > /dev/null
```
nan-yu, line 53: What if the remote name is not set to `upstream`?

Can we instead using `git fetch --all`?

sdowell: This fetch will create a local namespace named `upstream` even if the user has not already defined a remote for it yet.

`git fetch --all` assumes the user already defined a remote for the GoogleContainerTools repo. Which is common and they probably should, but I'm trying to minimize assumptions for this script

sdowell:
experimenting with an alternative approach
/hold

sdowell:
@nan-yu Updated to use `FETCH_HEAD`
/unhold

nan-yu (approved):
/lgtm

PR #205, opened 2022-10-07
Use deletion propagation for e2e test cleanup

karlkfi:
- Rewrite test cleanup to use the RSync finalizer to delete all
  managed resources by default (for e2e tests only)
- Add deletion propagation helper functions
- Fix TestDeleteRootSyncAndRootSyncV1Alpha1 to handle longer RSync
  deletion timeout
- Log elapsed time for Clean and Reset
- Add more logging for test setup and teardown
- Remove test cleanup blocks that are no longer necessary.
  Most of these involved resetting the sync config, which is now
  being deleted and re-created with the required settings by test
  Reset.

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/351
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/352
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/354
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/355 (rejected. non-blocking)
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/356
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/357
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/358
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/361
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/363

### DIFF
--- a/e2e/nomostest/deletion_propagation.go
+++ b/e2e/nomostest/deletion_propagation.go
@@ -1,0 +1,79 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nomostest
+
+import (
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/metadata"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// EnableDeletionPropagation enables foreground deletion propagation on a
+// RootSync or RepoSync. The object is annotated locally, but not applied.
+// Returns true if a change was made, false if already enabled.
+func EnableDeletionPropagation(rs client.Object) bool {
+	return SetDeletionPropagationPolicy(rs, metadata.DeletionPropagationPolicyForeground)
+}
+
+// DisableDeletionPropagation disables foreground deletion propagation on a
+// RootSync or RepoSync. The object annotated is removed locally, but not applied.
+// Returns true if a change was made, false if already enabled.
+func DisableDeletionPropagation(rs client.Object) bool {
+	return RemoveDeletionPropagationPolicy(rs)
+}
+
+// IsDeletionPropagationEnabled returns true if deletion propagation annotation
+// is set to Foreground.
+func IsDeletionPropagationEnabled(rs client.Object) bool {
+	return HasDeletionPropagationPolicy(rs, metadata.DeletionPropagationPolicyForeground)
+}
+
+// HasDeletionPropagationPolicy returns true if deletion propagation annotation
+// is set to the specified policy. Returns false if not set.
+func HasDeletionPropagationPolicy(obj client.Object, policy metadata.DeletionPropagationPolicy) bool {
+	annotations := obj.GetAnnotations()
+	// don't panic if nil
+	if len(annotations) == 0 {
+		return false
+	}
+	foundPolicy, found := annotations[metadata.DeletionPropagationPolicyAnnotationKey]
+	return found && foundPolicy == string(policy)
+}
+
+// SetDeletionPropagationPolicy sets the value of the deletion propagation
+// annotation locally (does not apply). Returns true if the object was modified.
+func SetDeletionPropagationPolicy(obj client.Object, policy metadata.DeletionPropagationPolicy) bool {
+	if HasDeletionPropagationPolicy(obj, policy) {
+		return false
+	}
+	core.SetAnnotation(obj, metadata.DeletionPropagationPolicyAnnotationKey, string(policy))
+	return true
+}
+
+// RemoveDeletionPropagationPolicy removes the deletion propagation annotation
+// locally (does not apply). Returns true if the object was modified.
+func RemoveDeletionPropagationPolicy(obj client.Object) bool {
+	annotations := obj.GetAnnotations()
+	// don't panic if nil
+	if len(annotations) == 0 {
+		return false
+	}
+	if _, found := annotations[metadata.DeletionPropagationPolicyAnnotationKey]; !found {
+		return false
+	}
+	delete(annotations, metadata.DeletionPropagationPolicyAnnotationKey)
+	obj.SetAnnotations(annotations)
+	return true
+}
--- a/e2e/nomostest/reset.go
+++ b/e2e/nomostest/reset.go
@@ -1,0 +1,564 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nomostest
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
+	"kpt.dev/configsync/pkg/api/configmanagement"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/importer/filesystem"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/metadata"
+	"kpt.dev/configsync/pkg/metrics"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+	"kpt.dev/configsync/pkg/syncer/differ"
+	"kpt.dev/configsync/pkg/util/log"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Reset performs multi-repo test reset:
+// - Delete unmanaged RootSyncs & RepoSyncs (with deletion propagation)
+// - Validate managed RepoSyncs & RootSyncs are deleted
+// - Delete all test namespaces not containing config-sync itself
+// - Delete NonRootRepos & RootRepos (except the default RootRepo)
+//
+// This should cleanly delete or reset all registered RSyncs.
+// Any managed RSyncs must have deletion propagation enabled by the test that
+// created them, otherwise their managed resources will not be deleted when the
+// RSync is deleted.
+// Any unregistered Repos must be reset by individual test Cleanup.
+func Reset(nt *NT) error {
+	start := time.Now()
+	defer func() {
+		elapsed := time.Since(start)
+		nt.T.Logf("[RESET] Test environment reset took %v", elapsed)
+	}()
+
+	// Delete all existing RootSyncs with the test label.
+	// Enable deletion propagation first, to clean up managed resources.
+	rootSyncList, err := listRootSyncs(nt)
+	if err != nil {
+		return err
+	}
+	if err := ResetRootSyncs(nt, rootSyncList.Items); err != nil {
+		return err
+	}
+
+	// Delete all existing RepoSyncs with the test label.
+	// Enable deletion propagation first, to clean up managed resources.
+	repoSyncList, err := listRepoSyncs(nt)
+	if err != nil {
+		return err
+	}
+	if err := ResetRepoSyncs(nt, repoSyncList.Items); err != nil {
+		return err
+	}
+
+	// Delete all Namespaces with the test label (except shared).
+	nsList, err := listNamespaces(nt)
+	if err != nil {
+		return err
+	}
+	// Error if any protected namespace was modified by a test (test label added)
+	// and not reverted by the test.
+	protectedNamespaces, nsListItems := filterNamespaces(nsList.Items,
+		protectedNamespaceList()...)
+	if len(protectedNamespaces) > 0 {
+		return errors.Errorf("protected namespace(s) modified by test: %+v",
+			protectedNamespaces)
+	}
+	// Skip deleting namespaces that contain test infra or Config Sync itself.
+	_, nsListItems = filterNamespaces(nsListItems,
+		configsync.ControllerNamespace,
+		configmanagement.RGControllerNamespace,
+		metrics.MonitoringNamespace,
+		testGitNamespace)
+	if err := ResetNamespaces(nt, nsListItems); err != nil {
+		return err
+	}
+
+	// NOTE: These git repos are not actually being deleted here, just forgotten.
+	// Repos are deleted by `Clean` for environment setup and teardown.
+
+	nt.T.Log("[RESET] Resetting NonRootRepos")
+	nt.NonRootRepos = make(map[types.NamespacedName]*Repository)
+
+	nt.T.Log("[RESET] Resetting RootRepos")
+	// Retain the root-sync repo (speeds up reset a bit when there are no changes)
+	rootRepo := nt.RootRepos[configsync.RootSyncName]
+	nt.RootRepos = make(map[string]*Repository)
+	nt.RootRepos[configsync.RootSyncName] = rootRepo
+
+	return nil
+}
+
+func protectedNamespaceList() []string {
+	list := make([]string, 0, len(differ.SpecialNamespaces))
+	for ns := range differ.SpecialNamespaces {
+		list = append(list, ns)
+	}
+	return list
+}
+
+type resetRecord struct {
+	Objects          map[types.NamespacedName]struct{}
+	ManagedObjects   map[types.NamespacedName]struct{}
+	ObjectNamespaces map[string]struct{}
+}
+
+func newResetRecord() *resetRecord {
+	return &resetRecord{
+		Objects:          make(map[types.NamespacedName]struct{}),
+		ManagedObjects:   make(map[types.NamespacedName]struct{}),
+		ObjectNamespaces: make(map[string]struct{}),
+	}
+}
+
+// ResetRootSyncs cleans up one or more RootSyncs and all their managed objects.
+// Use this for cleaning up RootSyncs in tests that use delegated control.
+func ResetRootSyncs(nt *NT, rsList []v1beta1.RootSync) error {
+	if len(rsList) == 0 {
+		return nil
+	}
+
+	nt.T.Logf("[RESET] Deleting RootSyncs (%d)", len(rsList))
+
+	record := newResetRecord()
+
+	for _, item := range rsList {
+		rs := &item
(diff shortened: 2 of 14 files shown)


e2e/nomostest/config-sync.go
```diff
@@ -1255,96 +1257,147 @@ func resetRepository(nt *NT, repoType RepoType, nn types.NamespacedName, upstrea
...
+	defaultRS := RootSyncObjectV1Beta1FromRootRepo(nt, rsName)
+	if cmp.Equal(rs.Spec, defaultRS.Spec) {
+		// If found and unchanged, just reset the Repository.
+		// This can happen if the RootSync was not modified by the test.
+		nt.RootRepos[rsName] = resetRepository(nt, RootRepo, RootSyncNN(rsName), upstream, sourceFormat)
+		return
 	}
+
+	// If found and needs changes, we need to pause reconciliation to avoid a
+	// race condition between the reconciler-manager and the reconciler.
```
sdowell, line 1286: What is the race condition?

e2e/nomostest/config-sync.go
```diff
@@ -1255,96 +1257,147 @@ func resetRepository(nt *NT, repoType RepoType, nn types.NamespacedName, upstrea
...
+	// This allows us to reset the Git repo without syncing.
+	scaleDeployment(nt, core.RootReconcilerName(rsName), configmanagement.ControllerNamespace, 0)
+	nt.RootRepos[rsName] = resetRepository(nt, RootRepo, RootSyncNN(rsName), upstream, sourceFormat)
+
+	// Reset the RootSync spec to default
+	defaultRS.Spec.DeepCopyInto(&rs.Spec)
+	if err = nt.Update(rs); err != nil {
+		nt.T.Fatal(err)
+	}
+	// nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceFormat": "%s", "git": {"dir": "%s"}}}`, sourceFormat, AcmeDir))
```
sdowell, line 1301: remove

e2e/nomostest/config-sync.go
```diff
@@ -1255,96 +1257,147 @@ func resetRepository(nt *NT, repoType RepoType, nn types.NamespacedName, upstrea
...
-// The spec.sourceFormat is reset to the specified format, to ensure the
-// remaining objects are deleted without erroring from format mismatch.
-//
-// The spec.git.dir is reset to "acme", which is created by
-// NewRepository and Repository.ReInit and must still exist.
-//
-// TODO: reset everything else, including the branch
-func resetRootRepos(nt *NT, upstream string, sourceFormat filesystem.SourceFormat) {
-	// Reset the default root repo first so that other managed RootSyncs can be re-created and initialized.
+func resetNamespaceRepo(nt *NT, rsNN types.NamespacedName, upstream string, sourceFormat filesystem.SourceFormat) {
```
sdowell, line 1307: Would be a bit cooler if this was generic instead of the a duplicate of resetRootRepo

e2e/nomostest/config-sync.go
```diff
@@ -1255,96 +1257,147 @@ func resetRepository(nt *NT, repoType RepoType, nn types.NamespacedName, upstrea
...
+	// This allows us to reset the Git repo without syncing.
+	scaleDeployment(nt, core.NsReconcilerName(rsNN.Namespace, rsNN.Name), configmanagement.ControllerNamespace, 0)
+	nt.NonRootRepos[rsNN] = resetRepository(nt, NamespaceRepo, rsNN, upstream, sourceFormat)
+
+	// Reset the RepoSync spec to default
+	defaultRS.Spec.DeepCopyInto(&rs.Spec)
+	if err = nt.Update(rs); err != nil {
+		nt.T.Fatal(err)
+	}
+	// nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceFormat": "%s", "git": {"dir": "%s"}}}`, sourceFormat, AcmeDir))
```
sdowell, line 1349: remove

e2e/nomostest/config-sync.go
karlkfi: Same problem everywhere, no abstraction around RootSync/RepoSync...

Also, in this case the RootRepos are keyed by name and the NonRootRepos are keyed by NamespacedName. That needs a huge refactor to change.

e2e/nomostest/config-sync.go
karlkfi: Yup. Left it in to see if you were paying attention. ;)

karlkfi:
/hold

Turns out this doesn't actually fix the GKE tests, even tho it works in Kind. The reset code isn't well tested in Kind because it just throws away the cluster after each test.

e2e/nomostest/config-sync.go
sdowell: 😢

e2e/nomostest/config-sync.go
karlkfi: In multi-repo, when you change spec.git.repo or spec.sourceFormat, instead of the reconciler just detecting it, the reconciler-manager detects it (watch on RSync resource) and modifies the reconciler config in its Deployment, causing the reconciler pod to be replaced.

And when you reset the Git repo, adding more commits or changing the branch, the reconciler detects that (git-sync polling git server) and starts a Sync.

So if you change SourceFormat, that requires changing both Git (add or remove the Repo object) and the RSync (spec.sourceFormat). If the reconciler picks up the Git change before the reconciler Pod is stopped by the Deployment change, the reconciler will error, saying either that the Repo is required or not allowed, depending on the direction of the SourceFormat change.

The error should just recover, when the reconciler Pod comes up with a the matching configuration. But for some reason, it's still causing test failure. So I put in these defensive measures to avoid it. Since I had to rewrite a bunch of other reset code to actually get it to work, tho, this may or may not be necessary any more... So I'm doing more testing now.

karlkfi: I think the race condition I was trying to solve is actually more of a problem in the mono-repo mode. Unlike multi-repo, there's no reconciler-manager in mono-repo mode, and the config for the SourceFormat is stored in a `source-format` ConfigMap that's loaded into the `importer` container of the `git-importer` Deployment. If the importer picks up the Git change before the ConfigMap has been updated, it can also error. But when there's an error with a specific commit in mono-repo, it gets stuck and stops accepting new commits, causing all subsequent tests to fail.

karlkfi: Removed the deployment scale down for now. I don't think multi-repo actually needs it. But mono-repo might.

karlkfi:
Looks like I've still got a bug in here. I need to make it skip modifying managed RSyncs and store them in a queue to reset. Otherwise the webhook rejects the change. And even if I disable the webhook it's still faster to not have to re-create the RSync to reset it and delete it again.

karlkfi:
Added new reconciler finalizer to see if it passes all the e2e tests before I break this down into smaller PRs and add more unit tests.

karlkfi:
Extracted dependency updates: https://github.com/GoogleContainerTools/kpt-config-sync/pull/232

karlkfi:
Extracted reconciler shutdown imporvements: https://github.com/GoogleContainerTools/kpt-config-sync/pull/234

karlkfi:
Extracted UpstreamRepoURL removal: https://github.com/GoogleContainerTools/kpt-config-sync/pull/240

karlkfi:
Extracted finalizer to https://github.com/GoogleContainerTools/kpt-config-sync/pull/242

karlkfi:
```
$ make test-e2e-kind-multi-repo-test-group3 E2E_ARGS="--share-test-env -v --test.v -p 1 --test.parallel=6 --timeout 2h"
...
PASS
ok  	kpt.dev/configsync/e2e/testcases	1507.431s
Tests took 1566 seconds
make[1]: Leaving directory '/home/karlisenberg/workspace/config-sync-prototype'
```

karlkfi:
```
$ make test-e2e-kind-multi-repo-test-group2 E2E_ARGS="--share-test-env -v --test.v -p 1 --test.parallel=6 --timeout 2h"
...
PASS
ok  	kpt.dev/configsync/e2e/testcases	984.499s
Tests took 1027 seconds
make[1]: Leaving directory '/home/karlisenberg/workspace/config-sync-prototype'
```

karlkfi:
/retest

I think this is a pre-existing flakey error:
` Error: cannot build pipelines: failt to create "opencensus" receiver, in pipeline "metrics": failed to bind to address "0.0.0.0:55678": listen tcp 0.0.0.0:55678: bind: address already in use`

karlkfi:
All 3 e2e tests groups passed locally on kind with shared-env using the latest changes & rebase.

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,531 @@
...
+	if err := batchDeleteAndWait(nt, rbs...); err != nil {
+		return err
+	}
+
+	// Delete any ClusterRoleBindings left behind.
+	// CRBs are usually only applied if PSP was enabled, but clean them up regardless.
+	nt.T.Log("[RESET] Deleting test ClusterRoleBindings")
+	var crbs []client.Object
+	for rsNN := range record.Objects {
+		crbs = append(crbs, repoSyncClusterRoleBinding(rsNN))
```
sdowell, line 309: hmm, I didn't realize the tests were already using a `ClusterRoleBinding` for RepoSyncs. We might want to limit this down to a `RoleBinding` to validate more realistic use case scenarios, but I don't view this change as a regression since we were already using `ClusterRoleBinding`.

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,531 @@
...
+	_, nsListItems = filterNamespaces(nsListItems,
+		configsync.ControllerNamespace,
+		configmanagement.RGControllerNamespace,
+		metrics.MonitoringNamespace,
+		testGitNamespace)
+	if err := ResetNamespaces(nt, nsListItems); err != nil {
+		return err
+	}
+
+	// NOTE: These git repos are not actually being deleted here, just forgotten.
```
sdowell, line 104: What is the reason that we need to forget the git repos here? It seems a bit strange to me that these pointers are being removed without cleaning up the underlying git repo.

Minor nit on the second line of the comment: it looks like the repos should normally get deleted by [this cleanup block](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/new.go#L319). The logic in `Clean` seems to be for stale repos (older than 24 hours)

e2e/nomostest/reset.go
karlkfi: Both setup options use a CRB if PSP is enabled, because the RB will violate default policy which disallows cross-namespace permissions (granting reconciler in namespace A permission to read/write objects in namespace B).

Why they dont just modify the policy, i don't know....

e2e/nomostest/reset.go
karlkfi: Repos were never actually being **deleted** before either:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/new.go#L234-L239
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/config_sync.go#L1105
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/config_sync.go#L1131

It WAS "resetting" the repo's contents (partially) and syncing again, but that never worked well (only some known files were being removed) and it's no longer necessary because the new finalizer does a destroy based on what's in the inventory, not what's in the repo, and then the repo contents are reset in test setup with ReInit:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/config_sync.go#L1062

This new way, we skip resetting repos we don't plan to use in the next test.

karlkfi:
> multi_sync_test.go:107: ERROR: admission webhook "rbac.authorization.k8s.io.v1.admission-webhook.configsync.gke.io" denied the request: kubernetes-admin is not authorized to delete managed resource "rbac.authorization.k8s.io_rolebinding_test-ns_nr2"

Looks like the extra cleanup in `TestMultiSyncs_Unstructured_MixedControl` calling `ResetRepoSyncs` before `ResetRootSyncs` was violating the assumption that all RSyncs would be deleted before trying to delete all the remaining RoleBindings. So I added a check to skip deleting managed RoleBindings. 

If this causes a side effect, it will probably manifest as a namespace garbage collector deadlocking with a similar error.

If that happens, we may have to disable the admission webhook in reset, which I don't want to do, because it will slow things down re-enabling it every time.

Another option would be to revert the `TestMultiSyncs_Unstructured_MixedControl` cleanup change and let it keep doing it the hard way. But I don't want to do that, because `ResetRepoSyncs` is more robust and accounts for more edge cases.

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,563 @@
...
+		defer cancel()
+		go TailReconcilerLogs(ctx, nt, RootReconcilerObjectKey(rsNN.Name))
+
+		// DeletePropagationBackground is required when deleting RSyncs with
+		// dependencies that have owners references. Otherwise the reconciler
+		// and dependenencies will be garbage collected before the finalizer
+		// can delete the managed resources.
+		// TODO: Remove explicit Background policy after the reconciler-manager finalizer is added.
+		nt.T.Logf("[RESET] Deleting RootSync %s", rsNN)
+		if err := nt.Delete(rs, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
```
sdowell, line 195: Could this deletion be done async as a `taskgroup`? I'm thinking the pattern of `delete synchronously and wait asynchronously` could be changed to `delete and wait asynchronously`, which I expect would complete faster. WDYT?

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,563 @@
...
+	for _, item := range rsList {
+		rs := &item
+		rsNN := client.ObjectKeyFromObject(rs)
+		record.Objects[rsNN] = struct{}{}
+		record.ObjectNamespaces[rsNN.Namespace] = struct{}{}
+
+		if manager, found := rs.GetAnnotations()[string(metadata.ResourceManagerKey)]; found {
+			record.ManagedObjects[rsNN] = struct{}{}
+			nt.T.Logf("[RESET] RootSync %s managed by %q", rsNN, manager)
+			if !IsDeletionPropagationEnabled(rs) {
```
sdowell, line 162: I'm not opposed to failing hard here, but would it be possible to use server side apply to set the annotation even though it's managed by another `RootSync`? I suppose it might get blocked by `admission-webhook`

karlkfi: The admission webhook would deny it and even if it’s disabled, the remediator would revert it, causing a race condition. 

So I just made it hard error so we can fix the test if we see it.

e2e/nomostest/reset.go
karlkfi: That is in fact what it’s doing. 
the deletes are all running, then the waits are triggered in goroutines. 

I could merge it to a single for loop, but that would intermingle the delete and wait event debug logs which I figured might be harder to read.

karlkfi: If speed is a concern tho, we could switch it to full async in a follow up.

e2e/nomostest/deletion_propagation.go
```diff
@@ -0,0 +1,76 @@
...
+// IsDeletionPropagationEnabled returns true if deletion propagation annotation
+// is set to Foreground.
+func IsDeletionPropagationEnabled(rs client.Object) bool {
+	return HasDeletionPropagationPolicy(rs, metadata.DeletionPropagationPolicyForeground)
+}
+
+// HasDeletionPropagationPolicy returns true if deletion propagation annotation
+// is set to the specified policy. Returns false if not set.
+func HasDeletionPropagationPolicy(obj client.Object, policy metadata.DeletionPropagationPolicy) bool {
+	foundPolicy, found := obj.GetAnnotations()[string(metadata.DeletionPropagationPolicyAnnotationKey)]
```
nan-yu, line 45: nit: the conversion to string seems unnecessary.

e2e/nomostest/deletion_propagation.go
```diff
@@ -0,0 +1,76 @@
...
+	return found && foundPolicy == string(policy)
+}
+
+// SetDeletionPropagationPolicy sets the value of the deletion propagation
+// annotation locally (does not apply). Returns true if the object was modified.
+func SetDeletionPropagationPolicy(obj client.Object, policy metadata.DeletionPropagationPolicy) bool {
+	annotations := obj.GetAnnotations()
+	if len(annotations) == 0 {
+		annotations = make(map[string]string, 1)
+	} else if val, found := annotations[metadata.DeletionPropagationPolicyAnnotationKey]; found && val == string(policy) {
```
nan-yu, line 55: nit: can be replaced with the `HasDeletionPropagationPolicy` function.

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,563 @@
...
+
+func newResetRecord() *resetRecord {
+	return &resetRecord{
+		Objects:          make(map[types.NamespacedName]struct{}),
+		ManagedObjects:   make(map[types.NamespacedName]struct{}),
+		ObjectNamespaces: make(map[string]struct{}),
+	}
+}
+
+// ResetRootSyncs cleans up one or more RepoSyncs and all their managed objects.
```
nan-yu, line 142: RootSyncs

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,563 @@
...
+// Use this for cleaning up RepoSyncs in tests that use delegated control.
```
nan-yu, line 143: RootSyncs

e2e/nomostest/reset.go
nan-yu, line 195: If the RootSync is the default `root-sync`, I guess it will be re-created in the next test?

e2e/testcases/helm_sync_test.go
```diff
@@ -143,11 +143,6 @@ func TestHelmNamespaceRepo(t *testing.T) {
 	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs)
 	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from a public Helm Chart with cluster-scoped type")
 	// Change the RepoSync to sync from the original git source to make test works in the shared test environment.
```
nan-yu, line 145: The comment can be deleted as well.

Since the default RootSync and its managed objects will be deleted in the `Reset` function, we don't need to reset the default RootSync to its original source in other places.
The cleanup snippet can be removed from the following places:
- helm_sync_test.go: line 61
- helm_sync_test.go: line 271
- helm_sync_test.go: line 314
- kcc_test.go: line 87
- csr_auth_test.go: line 69
- csr_auth_test.go: line 246
- oci_sync_testL line 86
- oci_sync_testL line 120
- oci_sync_testL line 484

e2e/nomostest/reset.go
karlkfi: Yes, the root-sync RootSync is always created or confirmed to already exist near the end of the test env setup, for every test, unless `opts.SkipConfigSyncInstall == true`. So it's safe to delete every time.

confusingly, it's `ValidateMultiRepoDeployments` that creates the root-sync RootSync, if it doesn't exist.

`ValidateMultiRepoDeployments` is called by `WaitForRepoSyncs`, which is called by `setupCentralizedControl` and then again by `setupTestCase`, which is called by both `FreshTestEnv` and `SharedTestEnv`.

It might be worth creating it in some other easier to find function beforehand tho, if you feel like a follow up PR.

e2e/nomostest/reset.go
karlkfi: done

e2e/nomostest/reset.go
karlkfi: done

e2e/testcases/helm_sync_test.go
karlkfi: Thanks for tracking these down. I removed them all and found a couple more to nuke too.

e2e/nomostest/deletion_propagation.go
karlkfi: fixed

e2e/nomostest/deletion_propagation.go
karlkfi: done

karlkfi:
/retest

I'm not sure why this is is flakey. It seems like this test creates `test-ns` and fails if it exists, indicating the previous cleanup was incomplete, but I'm not sure why...

nan-yu (approved):
/lgtm
/unhold

PR #209, opened 2022-10-12
fix terraform config for fresh projects

sdowell:
A fresh GCP project may not have all required APIs enabled, leading to certain resources not being available (such as the default compute account).

terraform apply will enable all of the required APIs, but they may not be available in time to create the actual resources. This adds a note to instruct the user that they may need to run apply twice for new projects.


nan-yu (approved):
/lgtm

PR #210, opened 2022-10-12
test: Fix cleanup for TestSyncFailureAfterSuccessfulSyncs

rquitales:
This commit resets the branch back to main to ensure proper cleanup checks.

Tested locally using a GKE cluster.

### DIFF
--- a/e2e/testcases/invalid_git_branch_test.go
+++ b/e2e/testcases/invalid_git_branch_test.go
@@ -22,6 +22,7 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/status"
 	"kpt.dev/configsync/pkg/testing/fake"
@@ -29,6 +30,7 @@ import (
 
 func TestInvalidRootSyncBranchStatus(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipMonoRepo)
+	nt.T.Cleanup(resetToMain(nt))
 
 	// Update RootSync to invalid branch name
 	nomostest.SetGitBranch(nt, configsync.RootSyncName, "invalid-branch")
@@ -58,6 +60,30 @@ func TestInvalidRootSyncBranchStatus(t *testing.T) {
 
 func TestInvalidRepoSyncBranchStatus(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipMonoRepo, ntopts.NamespaceRepo(namespaceRepo, configsync.RepoSyncName))
+	nt.T.Cleanup(func() {
+		nt.T.Log("Resetting all R*Sync branches to main")
+		nn := nomostest.RepoSyncNN(namespaceRepo, configsync.RepoSyncName)
+
+		// Ensure RepoSync is switched back to main branch locally.
+		nt.T.Log("Switching to main for locally checked out RepoSync")
+		nt.NonRootRepos[nn].CheckoutBranch(nomostest.MainBranch)
+
+		// Ensure RepoSync is reset to main branch in the RootSync.
+		nt.T.Log("Switching to main for local RepoSync object")
+		obj := nt.RootRepos[configsync.RootSyncName].Get(nomostest.StructuredNSPath(namespaceRepo, configsync.RepoSyncName))
+		rs, ok := obj.(*v1beta1.RepoSync)
+		if !ok {
+			nt.T.Fatal("unable to cast commited RepoSync object")
+		}
+
+		if rs.Spec.Branch == nomostest.MainBranch {
+			return
+		}
+		rs.Spec.Branch = nomostest.MainBranch
+		nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(namespaceRepo, configsync.RepoSyncName), rs)
+		nt.RootRepos[configsync.RootSyncName].CommitAndPush("Reset RepoSync to main branch")
+		nt.WaitForRepoSyncs()
+	})
 
 	nn := nomostest.RepoSyncNN(namespaceRepo, configsync.RepoSyncName)
 	rs := nomostest.RepoSyncObjectV1Beta1FromNonRootRepo(nt, nn)
@@ -107,6 +133,7 @@ func TestInvalidMonoRepoBranchStatus(t *testing.T) {
 
 func TestSyncFailureAfterSuccessfulSyncs(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource)
+	nt.T.Cleanup(resetToMain(nt))
 
 	// Add audit namespace.
 	auditNS := "audit"
@@ -140,4 +167,15 @@ func TestSyncFailureAfterSuccessfulSyncs(t *testing.T) {
 	// Change the remote branch name back to the original name.
 	nt.RootRepos[configsync.RootSyncName].RenameBranch("invalid-branch", devBranch)
 	nt.WaitForRepoSyncs()
+}
+
+// resetToMain is a cleanup closure function that will reset the RootSync because
+// the cleanup stage will check if RootSync is synced from the main branch. We need
+// to locally checkout the main branch, AND update the RootSync.
+func resetToMain(nt *nomostest.NT) func() {
+	return func() {
+		nt.RootRepos[configsync.RootSyncName].CheckoutBranch(nomostest.MainBranch)
+		nomostest.SetGitBranch(nt, configsync.RootSyncName, nomostest.MainBranch)
+		nt.WaitForRepoSyncs()
+	}
 }
--- a/e2e/testcases/invalid_git_branch_test.go
+++ b/e2e/testcases/invalid_git_branch_test.go
@@ -22,7 +22,6 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/api/configsync"
-	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/status"
 	"kpt.dev/configsync/pkg/testing/fake"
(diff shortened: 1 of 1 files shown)


rquitales:
/hold

Adding the cleanup to all other tests in the file.

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -140,4 +140,9 @@ func TestSyncFailureAfterSuccessfulSyncs(t *testing.T) {
 	// Change the remote branch name back to the original name.
 	nt.RootRepos[configsync.RootSyncName].RenameBranch("invalid-branch", devBranch)
 	nt.WaitForRepoSyncs()
+
+	// Reset RootSync because the cleanup stage will check if RootSync is synced from the main branch.
+	nt.RootRepos[configsync.RootSyncName].CheckoutBranch(nomostest.MainBranch)
+	nomostest.SetGitBranch(nt, configsync.RootSyncName, nomostest.MainBranch)
```
sdowell, line 146: I suggest moving this to a `Cleanup` block after the branch is changed around line 123.

We could also potentially update the test scaffolding functions (e.g. `nomostest.SetGitBranch`) to add the Cleanup block which resets the branch after the test

karlkfi: I tried updating the test scafolding to do this generically, but it creates a lot of side effects that break other tests. That’s what inspired the deletion propagation.

karlkfi: I tried updating the test scafolding to do this generically, but it creates a lot of side effects that break other tests. That’s what inspired the deletion propagation.

rquitales: Updated all tests in the file to run the cleanup code as a `t.Cleanup` block.

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -58,6 +59,7 @@ func TestInvalidRootSyncBranchStatus(t *testing.T) {
 
 func TestInvalidRepoSyncBranchStatus(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipMonoRepo, ntopts.NamespaceRepo(namespaceRepo, configsync.RepoSyncName))
+	nt.T.Cleanup(resetToMain(nt))
```
karlkfi, line 62: I think this one also needs to reset the RepoSync to main, not just the RootSync. And the reset might need to happen in Git, not just by patching the RepoSync directly, because the RepoSync is centrally managed in Git by the RootSync. So this might need custom cleanup.

sdowell: It looks like this one actually _only_ needs cleanup for the RepoSync, since the test changes the RepoSync branch but not the RootSync branch. I think we might need a separate cleanup function for changing the branch directly on the RSync and for committing the change to a git repo.

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -58,6 +60,28 @@ func TestInvalidRootSyncBranchStatus(t *testing.T) {
...
+			nt.T.Fatal("unable to cast checked out RepoSync object to unstructured")
+		}
+		currBranch, found, err := unstructured.NestedString(uObj.Object, "spec", "branch")
+		if err != nil || !found {
+			nt.T.Fatal("unable to find repoSync.Spec.Branch field")
+		}
+		if currBranch != nomostest.MainBranch {
+			unstructured.SetNestedField(uObj.Object, nomostest.MainBranch, "spec", "branch")
+			nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(namespaceRepo, configsync.RepoSyncName), uObj)
+			nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to main branch")
```
karlkfi, line 81: "Reset RepoSync to main branch"

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -58,6 +60,28 @@ func TestInvalidRootSyncBranchStatus(t *testing.T) {
...
 func TestInvalidRepoSyncBranchStatus(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipMonoRepo, ntopts.NamespaceRepo(namespaceRepo, configsync.RepoSyncName))
+	nt.T.Cleanup(func() {
+		nn := nomostest.RepoSyncNN(namespaceRepo, configsync.RepoSyncName)
+
+		// Ensure RepoSync is switched back to main branch locally.
+		nt.NonRootRepos[nn].CheckoutBranch(nomostest.MainBranch)
+		// Ensure RepoSync is reset to main branch in the RootSync.
+		obj := nt.RootRepos[configsync.RootSyncName].Get(nomostest.StructuredNSPath(namespaceRepo, configsync.RepoSyncName))
+		uObj, ok := obj.(*unstructured.Unstructured)
```
karlkfi, line 70: Pretty sure Repo.Get will try to return the typed object if it's in the scheme. So you should get a RepoSync, it just won't have the GVK set.

karlkfi (changes requested):

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -22,13 +22,15 @@ import (
...
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/status"
 	"kpt.dev/configsync/pkg/testing/fake"
 )
 
 func TestInvalidRootSyncBranchStatus(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipMonoRepo)
+	nt.T.Cleanup(resetToMain(nt))
```
nan-yu, line 33: Line 49 already updates RootSync to sync from the main branch, no need to reset again.
If we want to guarantee the reset, we can move line 49 to 58 to the cleanup block.

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -58,6 +60,30 @@ func TestInvalidRootSyncBranchStatus(t *testing.T) {
 
 func TestInvalidRepoSyncBranchStatus(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipMonoRepo, ntopts.NamespaceRepo(namespaceRepo, configsync.RepoSyncName))
+	nt.T.Cleanup(func() {
```
nan-yu, line 63: I don't see a need to reset the RepoSync branch.

If the test failed:
1. before setting the invalid branch, no need to cleanup as the RepoSync is not created yet.
2. after setting the invalid branch, but before setting back to main. No resources will be managed by RepoSync and the RootSync will be reset by the test infra, which prunes the invalid RepoSync at the end.
3. after setting back to main, no need to reset again.

nan-yu (commented):
Haven't reviewed the cleanup in the `TestSyncFailureAfterSuccessfulSyncs` test yet, but the first two cleanups look unnecessary.

e2e/testcases/invalid_git_branch_test.go
karlkfi: > No resources will be managed by RepoSync and the RootSync will be reset by the test infra

This is not correct. The general reset does not change the local checkout branch. So if you last checked out some other branch for this repo, it won't reset correctly unless you check out the right branch. The generic reset also does not reset the branch in the the RootSync or RepoSync.

nan-yu: The general reset will reset all repos before and after each test case. Every test case will start with a new initialized git repository. In the [resetRepository](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/config-sync.go#L1248) function, both `ReInit` and `NewRepository` do a `git init` followed by `git checkout -b main`, which guarantees that each repo starts with the main branch.

rquitales:
/unhold
Will only address cleanup for `TestSyncFailureAfterSuccessfulSyncs` in this PR.

e2e/testcases/invalid_git_branch_test.go
rquitales: Removed the cleanup code for this PR.

e2e/testcases/invalid_git_branch_test.go
rquitales: Moved the reset and metrics validation into a cleanup block.

nan-yu (approved):
/lgtm

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -29,6 +29,18 @@ import (
 
 func TestInvalidRootSyncBranchStatus(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipMonoRepo)
+	nt.T.Cleanup(func() {
+		// Cleanup to Reset RootSync back to main.
+		nomostest.SetGitBranch(nt, configsync.RootSyncName, nomostest.MainBranch)
+		nt.WaitForRepoSyncs()
```
karlkfi, line 35: This will probably work, but there might be a gotcha...

WaitForRepoSyncs calls DefaultRootSha1Fn to get the commit sha, and DefaultRootSha1Fn uses the Repository.Hash() from nt.RootRepos, and Hash() only returns the local commit, without verifying the branch. This is something I've been trying to fix but caused a bunch of side effects.

So basically, you need to be sure that the local repo is on the right branch before using WaitForRepoSyncs, otherwise it will try to look for the wrong commit (which is what was causing the test to fail in CI).

Are we sure the local branch is "main" at all times that this test can fail? If not, we need to `Repo.CheckoutBranch(nomostest.MainBranch)` BEFORE calling WaitForRepoSyncs.

karlkfi: /cc @nan-yu

e2e/testcases/invalid_git_branch_test.go
```diff
@@ -29,6 +29,18 @@ import (
...
+
+		err := nt.ValidateMetrics(nomostest.SyncMetricsToLatestCommit(nt), func() error {
```
karlkfi, line 37: Do we need to validate the metrics in the cleanup? Ideally the metrics would reflect the synced state, right?

e2e/testcases/invalid_git_branch_test.go
nan-yu: For this particular test, it starts with the main branch as the generic reset function runs a `git init` and `git checkout -b main` in the beginning. It never checks out to a different branch within the test. Hence, it should be safe to not checkout to main explicitly.

e2e/testcases/invalid_git_branch_test.go
nan-yu: I think it is okay either way. No strong objection.

karlkfi: Prior to this PR, the test body was:
1. cause problem
2. verify problem is reflected in sync status
3. verify problem is reflected in metrics
4. fix problem
5. verify fix is reflected in sync status
6. verify fix is reflected in sync metrics

This is a common pattern. Should steps 4-6 always be in Cleanup? Or should all the steps be in the test body and the cleanup be just step 4?

karlkfi: Correction: the test is missing step 3. It should probably be added to validate the source error.

sdowell: I'd suggest having a combination of both. We can:
- Fix problem and make assertions in test body
- Reset branch in cleanup, without assertions

rquitales: @nan-yu @karlkfi Can we punt the test refactor into a separate PR where we can discuss the best pattern to use for test assertion and celanup? I'll add another commit to remove the cleanup changes made in `TestInvalidRootSyncBranchStatus` so that we can focus on fixing the broken `TestSyncFailureAfterSuccessfulSyncs` in the PR only.

nan-yu: I think we can maintain the test body with the 6 steps as Karl commented, but i'm just thinking if there is a need for an explicit cleanup?
The test doesn't switch to a different branch in the repo and no resources are added in the test. Even if it fails before step 4, the generic reset function will re-initialize the git repo.

nan-yu (approved):
/lgtm

karlkfi (approved):

PR #213, opened 2022-10-17
Use the public AR/GCR image in the stolos-dev project for testing

nan-yu:
https://testgrid.k8s.io/googleoss-kpt-config-sync runs all the periodic jobs in the oss-prow-build-kpt-config-sync project. It needs to get an exemption for the public access to the artifact registry and container registry. Before the exemption is approved, the test continues to use the public image in the stolos-dev project.


xinnywinne (approved):
/lgtm

PR #214, opened 2022-10-17
Remove the k8sallowedrepos.constraints.gatekeeper.sh CRD if needed

haiyanmeng:
If the k8sallowedrepos.constraints.gatekeeper.sh CRD exists on the cluster before TestConstraintTemplateAndConstraintInSameCommit starts, the test would fail since it expects UnknownKindError.

### DIFF
--- a/e2e/testcases/gatekeeper_test.go
+++ b/e2e/testcases/gatekeeper_test.go
@@ -30,6 +30,10 @@ func TestConstraintTemplateAndConstraintInSameCommit(t *testing.T) {
 	// TODO enable the test on autopilot clusters when GKE 1.21.3-gke.900 reaches regular/stable.
 	nt := nomostest.New(t, nomostesting.Reconciliation1, ntopts.Unstructured, ntopts.SkipAutopilotCluster)
 
+	crdName := "k8sallowedrepos.constraints.gatekeeper.sh"
+	nt.T.Logf("Delete the %q CRD if needed", crdName)
+	nt.MustKubectl("delete", "crd", crdName, "--ignore-not-found")
+
 	// Simulate install of Gatekeeper with just the ConstraintTemplate CRD
 	if err := nt.ApplyGatekeeperCRD("constraint-template-crd.yaml", "constrainttemplates.templates.gatekeeper.sh"); err != nil {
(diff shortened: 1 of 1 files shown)


e2e/testcases/gatekeeper_test.go
```diff
@@ -30,6 +30,10 @@ func TestConstraintTemplateAndConstraintInSameCommit(t *testing.T) {
 	// TODO enable the test on autopilot clusters when GKE 1.21.3-gke.900 reaches regular/stable.
```
sdowell, line 30: Out of scope for this change, but looks like we can address this TODO

e2e/testcases/gatekeeper_test.go
```diff
@@ -30,6 +30,10 @@ func TestConstraintTemplateAndConstraintInSameCommit(t *testing.T) {
...
 	nt := nomostest.New(t, nomostesting.Reconciliation1, ntopts.Unstructured, ntopts.SkipAutopilotCluster)
 
+	crdName := "k8sallowedrepos.constraints.gatekeeper.sh"
+	nt.T.Logf("Delete the %q CRD if needed", crdName)
+	nt.MustKubectl("delete", "crd", crdName, "--ignore-not-found")
```
sdowell, line 35: Seems like this logic belongs in the `ApplyGatekeeperCRD` function

e2e/testcases/gatekeeper_test.go
haiyanmeng: Good catch. Let us fix this in a separate PR.

e2e/testcases/gatekeeper_test.go
haiyanmeng: The `ApplyGatekeeperCRD` function applies a given CRD. 

There are two CRDs involved in this e2e test: `constrainttemplates.templates.gatekeeper.sh` and `k8sallowedrepos.constraints.gatekeeper.sh`.

This e2e test works fine if the `constrainttemplates.templates.gatekeeper.sh`  CRD already exists before the test starts.

This e2e test fails if the `k8sallowedrepos.constraints.gatekeeper.sh` CRD exists before the test starts. 

Therefore, we need to make sure the `k8sallowedrepos.constraints.gatekeeper.sh` CRD does not exist before the test starts.

PR #20, opened 2022-08-12
Load allow vertical scale flag in reconciler manager

victorpras:
Load --allow-vertical-scale flag from reconcilerManager deployment, keep current container resources when flag is enabled, and add e2e test.

--allow-vertical-scale flag is declared in the reconcilerManager deployment when allowVerticalScale is set to true in the ConfigManagement API. When the flag is enabled, the reconciler should load the flag, and copy over all containers' resources from the current reconciler Deployment to the declared manifest, so that the discrepancy will not cause any Deployment update. That implies any resource override applied to the declared Deployment will be ignored. Changes to anything other than the resource will not be maintained and will be reverted back

The e2e test will simulate the scenario by adding the --allow-vertical-scale flag to the reconciler-manager deployment, perform an update to the resource of a container in the reconciler deployment and another update to minReadySeconds field. The update to the resource should then be maintained and not reverted, while minReadySeconds update is reverted

PR #216, opened 2022-10-17
Revert VPA flag and e2e test

victorpras:
VPA isn't updating deployment and update the pods directly. We need to revert allowVerticalScale flag addition and revert the e2e test because they don't resemble how VPA works

https://github.com/GoogleContainerTools/kpt-config-sync/pull/20 was the original PR


nan-yu (approved):
/lgtm

PR #221, opened 2022-10-19
add new operator/oss build manifest targets

sdowell:
This change adds two new targets for building operator/oss manifests using kustomize. The move to kustomize is intended to improve readability of the manifest scaffolding.

A critical difference of the new build-manifests-operator target is that it only builds manifests for the multirepo mode. This is in preparation of migrating the legacy monorepo code to a separate repository. Since this will be a breaking change, it is implemented as a new make target. The old manifest related targets will be cleaned up and refactored once the monorepo migration is complete.

### DIFF
--- a/e2e/nomostest/config-sync.go
+++ b/e2e/nomostest/config-sync.go
@@ -351,6 +351,7 @@ func installationManifests(nt *NT, tmpManifestsDir string, nomos ntopts.Nomos) [
 		case "reconciler-manager.yaml":
 			// For the reconciler manager template, we want the latest image for the reconciler manager.
 			imgName = fmt.Sprintf("%s/reconciler-manager:%s", *e2e.ImagePrefix, *e2e.ImageTag)
+			replaced = strings.ReplaceAll(replaced, "RECONCILER_MANAGER_IMAGE_NAME", imgName)
 		case "reconciler-manager-configmap.yaml":
 			// For the reconciler deployment template, we want the latest image for the reconciler and hydration-controller.
 			reconcilerImgName := fmt.Sprintf("%s/reconciler:%s", *e2e.ImagePrefix, *e2e.ImageTag)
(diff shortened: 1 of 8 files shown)


manifests/oss/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resources:
+- ../base
+- ../test-resources/00-namespace.yaml
```
nan-yu, line 17: While we're here, it would be good if we can rename the `test-resources` directory as it is not test resources, but the namespaces and resource-group-controller that are needed for the multi-repo mode.

nan-yu (commented):
LGTM with one minor comment.

I had another 2 comments (rename `build-images-nomos` to `build-images-monorepo` and remove `NOMOS_IMAGE` from the build-manifests-oss target, but glad to see those were updated after refreshing the change :smiley:

manifests/oss/kustomization.yaml
sdowell: Agreed that the `test-resources` directory is a misnomer. Would you mind if I update that in a separate change? I'll have to update the naming in a few places so I'd like to isolate that as a separate change

e2e/nomostest/config-sync.go
```diff
@@ -363,12 +364,10 @@ func installationManifests(nt *NT, tmpManifestsDir string, nomos ntopts.Nomos) [
 			replaced = strings.ReplaceAll(replaced, "HELM_SYNC_IMAGE_NAME", helmSyncImgName)
 		case "admission-webhook.yaml":
 			imgName = fmt.Sprintf("%s/admission-webhook:%s", *e2e.ImagePrefix, *e2e.ImageTag)
+			replaced = strings.ReplaceAll(replaced, "WEBHOOK_IMAGE_NAME", imgName)
 		default:
 			// For any other template, we want the latest image for the nomos binary (mono-repo).
 			imgName = fmt.Sprintf("%s/nomos:%s", *e2e.ImagePrefix, *e2e.ImageTag)
-		}
-
-		if template != "reconciler-manager-configmap.yaml" {
```
sdowell, line 371: I think it would be better if we reused the make targets for constructing the manifests here
- to minimize duplicate logic
- to test our real `build-manifests` as part of our e2e testing

That's a bigger change though that can be done separately

manifests/oss/kustomization.yaml
nan-yu: SG

nan-yu (approved):
/lgtm

manifests/oss/kustomization.yaml
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/256

PR #222, opened 2022-10-19
Add retries to mitigate flakiness

haiyanmeng:


e2e/testcases/preserve_fields_test.go
```diff
@@ -66,7 +66,9 @@ func TestPreserveGeneratedServiceFields(t *testing.T) {
 	nt.WaitForRepoSyncs()
 
 	// Ensure the Service has the target port we set.
-	err := nt.Validate(serviceName, ns, &corev1.Service{}, hasTargetPort(targetPort1))
+	_, err := nomostest.Retry(30*time.Second, func() error {
```
tiffanny29631, line 69: What about using nt.DefaultReconcileTimeout, which should be 1 minute.

haiyanmeng: Good point. Fixed it.

tiffanny29631 (approved):
/lgtm

PR #224, opened 2022-10-20
Removing legacy_test.go, remaining bats test file, and bats linting

victorpras:
Removing bats test related file:

- apiservice.bats -> Converted to apiservice_test.go
- status_monitoring.bats -> Converted to policy_dir_test.go 
- resource_conditions.bats -> Converted to resource_conditions_test.go
- preserve_fields.bats -> Converted to preserve_fields_test.go 
- gatekeeper.bats -> Converted to gatekeeper_test.go 
- cluster_resources.bats -> Converted to cluster_resources_test.go 
- legacy_test.go -> removed since we don't have any bats test anymore

Removing bats related make target:
- manual-test-boostrap
- manual-test-clean
- e2e-staging
- __test-e2e-run
- test-e2e-nobuild
- test-e2e-ephemeral
- test-e2e
- test-e2e-dev

Also removed bats part of linting script


victorpras:
it looks like `setup.sh` and `e2e.sh` are still used for running e2e on Kind cluster, so I'm not sure if we can deprecate them just yet. Furthermore, `setup.sh` is using some function from `lib` folder, in particular: `wait`, `install`, and `resource`. The rest of the lib contents can be safely removed

sdowell:
My understanding is that `e2e.sh` and `setup.sh` is test scaffolding specifically for the bats tests. The go e2e tests use [this script](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/build/test-e2e-go/e2e.sh) as an entry point and the kind clusters are created [within the go test framework](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/ntopts/kind.go#L173).

nan-yu (changes requested):
Sam is correct.  `e2e.sh` and `setup.sh` are only used for bats. 

Can you also iterate through other folders and remove those unused ones, e.g. `e2e/lib`, `third_party/bats-core` and even some targets in the make file.

victorpras:
> e2e.sh and setup.sh is test scaffolding specifically for the bats tests. The go e2e tests use [this script](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/build/test-e2e-go/e2e.sh) as an entry point and the kind clusters are created [within the go test framework](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/ntopts/kind.go#L173). 

> Can you also iterate through other folders and remove those unused ones, e.g. e2e/lib, third_party/bats-core and even some targets in the make file.

Ack. Thanks for the info. `e2e/lib` and `third_party/bats-core` has been removed. Make target involving bats has also been updated.

sdowell:
I think we should turn off the bats prow jobs before removing all of this, otherwise this will be a breaking change

Makefile.e2e
```diff
@@ -21,7 +21,6 @@ e2e-staging: \
 	@rm -rf $(E2E_DIR)
 	@mkdir -p $(E2E_DIR)
 	@mkdir -p $(E2E_DIR)/third_party
```
nan-yu, line 23: Do we still need the `e2e-staging` target?

nan-yu: You can also check other targets that depend on `e2e-staging` and confirm if they're still needed.

nan-yu (changes requested):
Suggest searching for the following keywords and removing the references:
- third_party
- e2e.sh

Makefile.e2e
victorpras: it looks like `e2e-staging` is currently used for `make manual-test-boostrap` and `make manual-test-clean`. I'm not very familiar with the usage of these 2 make target, should they be removed or modified?

nan-yu: I would suggest looking into these two manual-test-xxx targets and figuring out whether they're needed.

victorpras: i did some searching and looks like they are intended for local nomos testing installation. Since we moved to OSS, I think they are no longer needed and can be removed. http://b/158858700

nan-yu: I don't understand why the local nomos tesitng installation is not needed after we moved to OSS.

But these two targets use the bats script to install Config Sync, that's why they can be removed.
You can find the details by looking at the `__test-e2e-run` target.

sdowell: +1 these look like convenience targets for developers to interact with the bats tests. They don't seem to be used by anything else, in which case they can be removed

victorpras: oh sorry, I mean because we are now maintaining another way to build and run from source as listed here https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/docs/installation.md#building-and-installing-from-source. Thanks for the clarification!

nan-yu (approved):
/lgtm

tiffanny29631 (approved):
/lgtm

PR #228, opened 2022-10-20
Increase the metrics wait timeout for Autopilot cluster

nan-yu:
The metrics validation hits timeout on Autopilot clusters, that is because it takes longer for the reconciler Pod to get ready. This commit increases the timeout from 1 minute to 3 minutes.


nan-yu:
@rquitales This PR is intended to fix [TestInvalidRepoSyncBranchStatus](https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/multi-repo-7-autopilot-stable/1582801225957511168) on autopilot stable, but it might fix the flaky [TestDeleteNamespaceReconcilerDeployment](https://testgrid.k8s.io/googleoss-kpt-config-sync-multi-repo-4#autopilot-stable) as well.

rquitales:
/lgtm
Thanks for getting around to this!

PR #229, opened 2022-10-21
Skip validating non-central root reconciler Pods in validateMultiRepoPods

nan-yu:
Some tests failed with a namespace reconciler Pod not found error, e.g. https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/multi-repo-7-standard-stable/1583109762550075392. The root cause is that the ValidateMultiRepoDeployments function validates the Pods after the Deployment objects are ready. When validating the Pods, it checks all Pods in the c-m-s namespace, but when validating the Deployments, it only checks the default root reconciler and other controllers. It is possible that the namespace reconciler Deployment is not available yet, which results in validation error for the corresponding Pod.

This commit adds the namespace reconciler Deployment to the validation list.

It also adds the error to the log when the git-server fails to be installed:
https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/stress-test/1581530245859643392.

### DIFF
--- a/e2e/nomostest/config-sync.go
+++ b/e2e/nomostest/config-sync.go
@@ -46,6 +46,7 @@ import (
 	"kpt.dev/configsync/pkg/importer/filesystem/cmpath"
 	"kpt.dev/configsync/pkg/importer/reader"
 	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/metrics"
 	"kpt.dev/configsync/pkg/monitor/state"
 	"kpt.dev/configsync/pkg/reconcilermanager"
@@ -462,13 +463,10 @@ func validateMonoRepoDeployments(nt *NT) error {
 
 // ValidateMultiRepoDeployments validates if all Config Sync Components are available.
 func ValidateMultiRepoDeployments(nt *NT) error {
-	for name := range nt.RootRepos {
-		// Create a RootSync to initialize the root reconciler.
-		rs := RootSyncObjectV1Beta1FromRootRepo(nt, name)
-		if err := nt.Create(rs); err != nil {
-			if !apierrors.IsAlreadyExists(err) {
-				nt.T.Fatal(err)
-			}
+	rs := RootSyncObjectV1Beta1FromRootRepo(nt, configsync.RootSyncName)
+	if err := nt.Create(rs); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			nt.T.Fatal(err)
 		}
 	}
 
@@ -532,6 +530,7 @@ func ValidateMultiRepoDeployments(nt *NT) error {
 
 // validateMultiRepoPods validates if all Config Sync Pods are healthy. It doesn't retry
 // because it is supposed to be invoked after validateMultiRepoDeployments succeeds.
+// It only checks the central root reconciler Pod.
 func validateMultiRepoPods(nt *NT) error {
 	for _, ns := range CSNamespaces {
 		pods := corev1.PodList{}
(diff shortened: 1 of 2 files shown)


nan-yu:
This PR should fix the `pods "ns-reconciler-bookinfo-xxxx" not found` failure in the following test:
- [TestOverrideGitSyncDepthV1Beta1](https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/multi-repo-1-standard-rapid-latest/1582347843740372992)
- [TestCACertSecretRefV1Alpha1](https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/multi-repo-7-standard-stable/1583109762550075392)
- TestCACertSecretWatch
-  [TestDeleteRepoSync_Centralized_AndRepoSyncV1Alpha1](https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/multi-repo-4-standard-stable/1582233086379167744)
- [TestSwitchFromGitToOci](https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/multi-repo-7-standard-regular-gitlab/1583169657689346048)
- ...

e2e/nomostest/new.go
```diff
@@ -424,6 +424,7 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 	if err := waitForConfigSync(nt, opts.Nomos); err != nil {
 		nt.T.Fatalf("waiting for ConfigSync Deployments to become available: %v", err)
 	}
+	createCentralRootSync(nt)
```
sdowell, line 427: Why do we need to create a central RootSync here and below?

e2e/nomostest/config-sync.go
```diff
@@ -527,18 +542,24 @@ func ValidateMultiRepoDeployments(nt *NT) error {
...
+// It checks the reconciler Pods depending on the includeReconcilers flag.
+func validateMultiRepoPods(nt *NT, includeReconcilers bool) error {
 	for _, ns := range CSNamespaces {
 		pods := corev1.PodList{}
 		if err := nt.List(&pods, client.InNamespace(ns)); err != nil {
 			return err
 		}
 		for _, pod := range pods.Items {
+			if app, found := pod.Labels["app"]; found && app == "reconciler" {
+				if !includeReconcilers {
```
sdowell, line 559: Why is it that the reconciler pods need to be filtered here? Since the PodList is coming from querying the cluster, I expected we are always iterating over valid pods in this loop

sdowell (commented):
The `ValidateMultiRepoDeployments` change mostly looks good to me, but I have a few questions

e2e/nomostest/config-sync.go
tiffanny29631: +1 that verifying all pods healthy seems to be sufficient in this case. Unless you are trying to match them with corresponding R*syncs?

e2e/nomostest/new.go
nan-yu: It was inside the `waitForConfigSync` function and then I extracted it out in the last patchset. 
Now, I reverted the change. 

The central RootSync is used to bootstrap other RSyncs in the central-control mode.

e2e/nomostest/config-sync.go
nan-yu: The intent was to use two flags to control whether to validate the reconciler Pods. The function was invoked both before and after RSync objects are created.

In the latest patchset, I updated the function to only validate the central root reconciler Pod, but skip other reconciler Pods because other reconcilers may not be reconciled yet.

The `nt.WaitForRepoSyncs()` also only checks the central root reconciler Pod and other non-reconciler Pods, but the RSync status should imply the corresponding Pod is running.

PR #234, opened 2022-10-21
Fix reconciler controllers to exit cleanly

karlkfi:


pkg/remediator/reconcile/worker_test.go
```diff
@@ -126,6 +127,55 @@ func TestWorker_ProcessNextObject(t *testing.T) {
 	}
 }
 
+func TestWorker_Run_Cancelled(t *testing.T) {
+	q := queue.New("test")
+
+	toProcess := []client.Object{}
```
sdowell, line 133: Seems like all of for loops are a noop since this is empty

sdowell (commented):
Functional changes LGTM but it seems like the unit testing is incomplete

pkg/remediator/reconcile/worker_test.go
karlkfi: Yeah, I copied most of this from another test. I wasn't sure if I should add more tests or simplify this one.

karlkfi: Alright, I cleaned this one up and added another to test a queue with items still left in it.

PR #238, opened 2022-10-24
make test KUBECONFIG thread safe

sdowell:
Setting environment variables in the test process is not thread safe and can lead to unpredictable behavior. This updates the usage of kubeconfig in the tests to pass the environment variable into the subprocess rather than setting the env var in the test process.

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -208,7 +208,7 @@ func NewSharedNT() *NT {
 		os.Exit(1)
 	}
 	fakeNTB := testing.NewShared(&testing.FakeNTB{})
-	opts := NewOptStruct(shared, tmpDir, fakeNTB)
+	opts := newOptStruct(shared, tmpDir, fakeNTB)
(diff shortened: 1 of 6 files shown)


tiffanny29631:
Thanks for the enhancement! 

/lgtm

e2e/nomostest/nt.go
```diff
@@ -808,6 +799,18 @@ func (nt *NT) Kubectl(args ...string) ([]byte, error) {
...
 }
 
+// Command is a convenience method for invoking a subprocess with the
+// KUBECONFIG environment variable set. Setting the environment variable
+// directly in the test process is not thread safe.
+func (nt *NT) Command(name string, args ...string) *exec.Cmd {
+	nt.T.Helper()
+
+	cmd := exec.Command(name, args...)
+	cmd.Env = os.Environ()
```
nan-yu, line 809: This is nice!

nan-yu (approved):
/lgtm

PR #240, opened 2022-10-24
Remove unused UpstreamRepoURL from e2e tests

karlkfi:
This feature of the Repository abstraction was designed to allow changing the upstream git repo. It's not clear it it was once used, but it's not used now and if you tried to use it, it probably wouldn't work, because the reset code doesn't handle changing it. So we can just delete it to clean up the code a bit. And if we need it later (unlikely) we can add it back after the reset code is rewritten to be simpler/cleaner.


nan-yu (approved):
/lgtm

PR #241, opened 2022-10-25
Increase the retry timeout and aggregate the validation errors in a multierr

nan-yu:
This is to fix the flakiness in https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/multi-repo-kind/1584975174380294144.

### DIFF
--- a/e2e/testcases/root_sync_test.go
+++ b/e2e/testcases/root_sync_test.go
@@ -19,7 +19,7 @@ import (
 	"testing"
 	"time"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"go.uber.org/multierr"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e/nomostest"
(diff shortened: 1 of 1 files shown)


e2e/testcases/root_sync_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	"testing"
 	"time"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"go.uber.org/multierr"
```
karlkfi, line 22: We have our own multierror in the status package. Might as well just use that. It's similar.

karlkfi: Actually, I'm surprised this didn't change vendor/... Do we already use this uber package elsewhere? Or is it maybe just used by a dependency?

nan-yu: The uber package is used in other places. 

Using the multierror in the status package will wrap the error with a KNV999 prefix, which seems to be weird in the e2e test.
```
2 error(s)

[1] KNV9999: fake error 1

For more information, see https://g.co/cloud/acm-errors#knv9999


[2] KNV9999: fake error 2

For more information, see https://g.co/cloud/acm-errors#knv9999
```

PR #243, opened 2022-10-26
run configsync-crds as part of presubmit

sdowell:
this ensures that the CRD yaml is up to date as part of presubmit checks. also consolidates the kustomize install make targets and runs addlicense after generating crds.


nan-yu (approved):
/lgtm

PR #246, opened 2022-10-27
override pointer instead of a struct instance.

mikebz:
This is another instance where we have something as "nullable" yet it's actually a struct instance.  I have changed the override to be a pointer.

```

	// override allows to override the settings for a root reconciler.
	// +nullable
	// +optional
	Override *OverrideSpec `json:"override,omitempty"`
```

Similar issue to this: https://github.com/GoogleContainerTools/kpt-config-sync/pull/231


pkg/api/configsync/v1alpha1/reposync_types.go
```diff
@@ -75,7 +75,17 @@ type RepoSyncSpec struct {
...
 	// +nullable
 	// +optional
-	Override OverrideSpec `json:"override,omitempty"`
+	Override *OverrideSpec `json:"override,omitempty"`
+}
+
+// GetOverride creates an override or returns an existing one
+// use it if you need to ensure that you are assigning
+// to an object, but not to test for nil (current existance)
+func (rs *RepoSyncSpec) GetOverride() *OverrideSpec {
```
karlkfi, line 84: It's a little unusual to have a Getter on an API struct like this. Are there any other examples you can point to that have set this precedence? The only ones I can think of are on Unstructured and are all top-level methods (since there are no lower level structs in Unstructured)

janetkuo (approved):
/lgtm

janetkuo:
Didn't see @karlkfi's comment before approving. Removed the `lgtm` tag for now.

pkg/api/configsync/v1alpha1/reposync_types.go
janetkuo: This avoid nil pointer dereference, so it seems fine to me. What's the concern?

karlkfi: For your other PR you used a package-level helper function, which I think I like better. Maybe `v1beta1.GetOverrideSpec(rs.Spec.Override)` instead of `rs.Spec.GetOverride()`. Or maybe `v1beta1.GetOverrideSpec(rs)`? I know it doesn't feel quite as nice to use, but IME most API structs are just the bare bones API and not full of a bunch of syntactic sugar.

Then again, if you can find some precedence that shows this isn't unusual, I don't really have any other compelling reason to make it harder to use.

karlkfi: > This avoids nil pointer dereference, so it seems fine to me. What's the concern?

Seems like bad practice to create a bunch of empty structs just to read the value. The only reason it was done this way this time is to make the refactor simpler.

The main reason not to do this is simply that I've never seen a help method on an inner struct like this before, for K8s API objects. But maybe I just haven't looked in the right places.

karlkfi: AFAICT, the only other API methods (e.g. not just a package function) in pkg/api are the ones generated for deepcopy.

karlkfi: If we add this helper, should all nullable fields have a similar getter method?

karlkfi: I did find some precedent for K8s API methods upstream. But not many.
AFAICT, none of them are just for protecting against nil pointers.

Screenshot from 2022-10-28 15-01-43

Also, from what I've seen in K8s Go, getter methods don't usually start with "Get", they just use the name of the return (ex: Cpu(), Memory(), Pods()). This is from Effective Go: https://go.dev/doc/effective_go#Getters

mikebz: Good question.  I also was curious about the precedent.  I think the question is whether this is a helper function for us or part of the API.  We could move it out of the file.  For inspiration I looked here:

```
// GetPeriodSecs returns the sync period defaulting to 15 if empty.
func GetPeriodSecs(g *Git) float64 {
	if g.Period.Duration == 0 {
		return configsync.DefaultPeriodSecs
	}
	return g.Period.Duration.Seconds()
}
```

It's also possible to move this out into another file.

I think the usage of the Go API is maybe a bigger discussion, right now this is exposed so technically someone could use it, but we didn't really productize it.

mikebz: Not sure if this is a best practice or a hack but here is a kustomize example:

https://github.com/kubernetes-sigs/kustomize/blob/daf81df437a2d875e92b83d970288db970be377a/api/types/kustomization.go#L180

mikebz: Somehow more chatter was collapsed and I didn't respond to that.  Sorry about that.  I am flex about moving this into another place. We do have our own precedent.  Other packages also include functions but they are not for getters.

I am flex on whatever you both want me to do here.  My only request is to keep the nullable types as pointers because generating YAML w/o that creates a bunch of empty structs.

karlkfi: I'm totally fine with the nullable/pointer part of this PR. The only potential breakage that could cause would be for people using these OSS API structs directly, which is probably zero people at this point.

That said, I do want to make it possible for people to use these API structs in the future to build integrations and controllers. We at Google probably will need to do this for higher level abstractions in the future.

I think I would feel more comfortable with this particular issue if the method was converted into a package-level function and put in a separate file. I really don't want to need to do this for every nullable pointer in the API. I'd perhaps be even more comfortable if this helper function was located closer to where it's being used. That way it's just a one-off helper func and not an official part of the API that people use as precedent later.

Anyway, sorry for blocking on this. It seems a little nit-picky. But I feel like when modifying the API, you really want to get it right the first time, if possible.

mikebz: no probs, I can make a change.

mikebz: updated, PTAL

karlkfi (approved):
/lgtm
/unhold

PR #247, opened 2022-10-27
update installation docs to install from url

sdowell:
this provides a bit more of a convenient UX to users to install the release manifests.


nan-yu (approved):
/lgtm

PR #250, opened 2022-10-28
fix: Avoid copying mutex/lock for Parser

rquitales:
Use the example from upstream Kubernetes to avoid copying mutexes.

### DIFF
--- a/pkg/declared/value.go
+++ b/pkg/declared/value.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/kube-openapi/pkg/schemaconv"
 	"k8s.io/kube-openapi/pkg/util/proto"
 	"k8s.io/kubectl/pkg/util/openapi"
+	"sigs.k8s.io/structured-merge-diff/v4/schema"
 	"sigs.k8s.io/structured-merge-diff/v4/typed"
 )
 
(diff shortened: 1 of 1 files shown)


sdowell:
LGTM but I'll leave approval to someone with more context
/lgtm

pkg/declared/value.go
```diff
@@ -105,7 +106,8 @@ func typedParser(doc *openapiv2.Document) (*typed.Parser, error) {
 	if err != nil {
 		return nil, fmt.Errorf("converting models to schema: %w", err)
 	}
-	// This fails the copylocks lint check, but this is exactly how the API server
-	// does it so I'm not sure what else to do.
-	return &typed.Parser{Schema: *typeSchema}, nil //nolint:govet
+
+	// We need to copy into a new schema.Schema to avoid copylock issue.
+	// See https://github.com/kubernetes/kubernetes/pull/109212/files
```
karlkfi, line 111: Do you have a non-upstream example of this biting us in Config Sync?

rquitales: AFAIK this is mostly a non-issue in production since the mutex/lock isn't really used before being copied so it shouldn't really affect Config Sync from having race conditions in theory. I believe upstream K8s developers are trying to update `typed.Parser` to use pointers for the Schema field in the future to avoid this issue.

This false positive was something that I came across when running `go vet` in a parallel PR to validate code. This is ignored by golangci-lint since we have the `nolint` tag.

karlkfi: Thanks for explaining! I had missed that it resolved the nolint issue.

karlkfi (approved):
/lgtm

PR #252, opened 2022-10-28
Split nt.go into multiple files

karlkfi:
Group methods by domain:
- wait
- wait-for-sync
- validate
- metrics


e2e/nomostest/wait-for-sync.go
```diff
@@ -0,0 +1,482 @@
+// Copyright 2022 Google LLC
```
sdowell, line 1: Naming convention for files is snake case, right? Ditto with `config-sync.go`

karlkfi: config-sync.go was already there. I just followed the package precedence, but I can rename them both.

karlkfi: done

PR #253, opened 2022-10-29
Only update condition timestamps on change

karlkfi:
- LastTransitionTime should only change if the Status changes.
- LastUpdateTime should only change if one of the condition field values changes.
- This should prevent unnecessary updates. Tho hopefully this behavior is redundant, due to the existing cmp.Equal check using IgnoreTimestampUpdates.
- Ideally this should alow us to stop using cmp.Equal in the future and consistently use semantic.DeepEqual instead.

### DIFF
--- a/pkg/reposync/conditions.go
+++ b/pkg/reposync/conditions.go
@@ -15,6 +15,7 @@
 package reposync
 
 import (
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 )
@@ -84,35 +85,46 @@ var singleErrorSummary = &v1beta1.ErrorSummary{
 }
 
 // SetReconciling sets the Reconciling condition to True.
-func SetReconciling(rs *v1beta1.RepoSync, reason, message string) {
-	if setCondition(rs, v1beta1.RepoSyncReconciling, metav1.ConditionTrue, reason, message, "", nil, &v1beta1.ErrorSummary{}, now()) {
+// Returns whether the condition was updated (any change) or transitioned
+// (status change).
+// Removes the Syncing condition if the Reconciling condition transitioned.
+func SetReconciling(rs *v1beta1.RepoSync, reason, message string) (updated, transitioned bool) {
+	updated, transitioned = setCondition(rs, v1beta1.RepoSyncReconciling, metav1.ConditionTrue, reason, message, "", nil, &v1beta1.ErrorSummary{}, now())
+	if transitioned {
 		removeCondition(rs, v1beta1.RepoSyncSyncing)
 	}
+	return updated, transitioned
 }
 
 // SetStalled sets the Stalled condition to True.
-func SetStalled(rs *v1beta1.RepoSync, reason string, err error) {
-	if setCondition(rs, v1beta1.RepoSyncStalled, metav1.ConditionTrue, reason, err.Error(), "", nil, singleErrorSummary, now()) {
+// Returns whether the condition was updated (any change) or transitioned
+// (status change).
+// Removes the Syncing condition if the Stalled condition transitioned.
+func SetStalled(rs *v1beta1.RepoSync, reason string, err error) (updated, transitioned bool) {
+	updated, transitioned = setCondition(rs, v1beta1.RepoSyncStalled, metav1.ConditionTrue, reason, err.Error(), "", nil, singleErrorSummary, now())
+	if transitioned {
 		removeCondition(rs, v1beta1.RepoSyncSyncing)
 	}
+	return updated, transitioned
 }
 
 // SetSyncing sets the Syncing condition.
-func SetSyncing(rs *v1beta1.RepoSync, status bool, reason, message, commit string, errorSources []v1beta1.ErrorSource, errorSummary *v1beta1.ErrorSummary, lastUpdate metav1.Time) {
+// Returns whether the condition was updated (any change) or transitioned
+// (status change).
+func SetSyncing(rs *v1beta1.RepoSync, status bool, reason, message, commit string, errorSources []v1beta1.ErrorSource, errorSummary *v1beta1.ErrorSummary, timestamp metav1.Time) (updated, transitioned bool) {
 	var conditionStatus metav1.ConditionStatus
 	if status {
 		conditionStatus = metav1.ConditionTrue
 	} else {
 		conditionStatus = metav1.ConditionFalse
 	}
-
-	setCondition(rs, v1beta1.RepoSyncSyncing, conditionStatus, reason, message, commit, errorSources, errorSummary, lastUpdate)
+	return setCondition(rs, v1beta1.RepoSyncSyncing, conditionStatus, reason, message, commit, errorSources, errorSummary, timestamp)
 }
 
 // setCondition adds or updates the specified condition with a True status.
-// It returns a boolean indicating if the condition status is transited.
-func setCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType, status metav1.ConditionStatus, reason, message, commit string, errorSources []v1beta1.ErrorSource, errorSummary *v1beta1.ErrorSummary, lastUpdate metav1.Time) bool {
-	conditionTransited := false
+// Returns whether the condition was updated (any change) or transitioned
+// (status change).
+func setCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType, status metav1.ConditionStatus, reason, message, commit string, errorSources []v1beta1.ErrorSource, errorSummary *v1beta1.ErrorSummary, timestamp metav1.Time) (updated, transitioned bool) {
 	condition := GetCondition(rs.Status.Conditions, condType)
 	if condition == nil {
 		i := len(rs.Status.Conditions)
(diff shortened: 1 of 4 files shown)


pkg/reposync/conditions.go
```diff
@@ -110,9 +111,8 @@ func SetSyncing(rs *v1beta1.RepoSync, status bool, reason, message, commit strin
 }
 
 // setCondition adds or updates the specified condition with a True status.
-// It returns a boolean indicating if the condition status is transited.
-func setCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType, status metav1.ConditionStatus, reason, message, commit string, errorSources []v1beta1.ErrorSource, errorSummary *v1beta1.ErrorSummary, lastUpdate metav1.Time) bool {
-	conditionTransited := false
+// It returns a boolean indicating if the condition status has transitioned.
+func setCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType, status metav1.ConditionStatus, reason, message, commit string, errorSources []v1beta1.ErrorSource, errorSummary *v1beta1.ErrorSummary, lastUpdate metav1.Time) (updated, transitioned bool) {
```
sdowell, line 115: Is it intentional that the `updated` return value is unused by the consumers of this function?

pkg/reposync/conditions.go
```diff
@@ -146,20 +158,23 @@ func GetCondition(conditions []v1beta1.RepoSyncCondition, condType v1beta1.RepoS
 }
 
 // removeCondition removes the RepoSync condition with the provided type.
-func removeCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType) {
-	rs.Status.Conditions = filterOutCondition(rs.Status.Conditions, condType)
+func removeCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType) (updated bool) {
```
sdowell, line 161: Is it intentional that the `updated` return value is unused by the consumers of this function?

pkg/reposync/conditions.go
```diff
@@ -110,9 +111,8 @@ func SetSyncing(rs *v1beta1.RepoSync, status bool, reason, message, commit strin
 }
 
 // setCondition adds or updates the specified condition with a True status.
-// It returns a boolean indicating if the condition status is transited.
-func setCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType, status metav1.ConditionStatus, reason, message, commit string, errorSources []v1beta1.ErrorSource, errorSummary *v1beta1.ErrorSummary, lastUpdate metav1.Time) bool {
-	conditionTransited := false
+// It returns a boolean indicating if the condition status has transitioned.
```
sdowell, line 114: Update docstring to reflect new signature (two booleans). Ditto for `rootsync`

pkg/reposync/conditions.go
```diff
@@ -123,16 +123,28 @@ func setCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType,
...
+		condition.Message = message
+		condition.Commit = commit
+		condition.ErrorSourceRefs = errorSources
+		condition.ErrorSummary = errorSummary
+		condition.LastUpdateTime = lastUpdate
+		transitioned = true
+		updated = true
+	} else {
+		copy := condition.DeepCopy()
+		copy.Reason = reason
```
sdowell, line 136: Add unit test case for update on each of these fields to verify the equality comparison. Ditto for `rootsync`

pkg/reposync/conditions.go
```diff
@@ -123,16 +123,28 @@ func setCondition(rs *v1beta1.RepoSync, condType v1beta1.RepoSyncConditionType,
...
-		conditionTransited = true
+		condition.Reason = reason
+		condition.Message = message
+		condition.Commit = commit
+		condition.ErrorSourceRefs = errorSources
+		condition.ErrorSummary = errorSummary
+		condition.LastUpdateTime = lastUpdate
+		transitioned = true
+		updated = true
+	} else {
```
sdowell, line 134: minor nit on code style:
```golang
	if condition.Status != status {
		condition.Status = status
		condition.LastTransitionTime = lastUpdate
		transitioned = true
		updated = true
	} else if !(condition.Reason == reason && condition.Message == message &&
		condition.Commit == commit && condition.ErrorSourceRefs == errorSources &&
		condition.ErrorSummary == errorSummary) {
		updated = true
	} else {
		return updated, transitioned
	}
	condition.Reason = reason
	condition.Message = message
	condition.Commit = commit
	condition.ErrorSourceRefs = errorSources
	condition.ErrorSummary = errorSummary
        condition.LastUpdateTime = lastUpdate

	return updated, transitioned
```
IMO this demonstrates the intent more clearly and avoids the need for `DeepCopy`/`DeepEqual`. Could alternatively extract the boolean logic to a helper function. WDYT?

pkg/reposync/conditions.go
karlkfi: The point of the updated return was to allow the caller not to have to do its own DeepEquals to check if anything changed, but I didn't want to get too distracted doing that refactor, since this was just something I was quickly fixing while doing the finalizer work. And it sounds like @xinnywinne wants to change the updates in the reconciler-manager to use SSA anyway, so she's going to be changing this code soon. When doing an SSA we can't really do a full DeepEquals on the RSync, because there might be other fields changed that we don't control. So having this update boolean might allow us to skip some SSA calls, if we know the condition hasn't changed.

pkg/reposync/conditions.go
karlkfi: Yeah, I was just feeling lazy and didn't want to write out all the comparisons... I'll paste yours in. But you could have used the github suggestion option to do it yourself and save me time ;).

karlkfi: Cant == slices... bleh PITA to compare in Go. Can't just add an Equal method without extracting a list struct or making a custom function.

karlkfi: Sigh. Can't == the timestamps either. They're pointers and we want to compare by value.

karlkfi: Is this better?

```
if condition.Status != status {
		condition.Status = status
		condition.LastTransitionTime = lastUpdate
		transitioned = true
		updated = true
	} else if condition.Reason != reason ||
		condition.Message != message ||
		condition.Commit != commit ||
		!equality.Semantic.DeepEqual(condition.ErrorSourceRefs, errorSources) ||
		!equality.Semantic.DeepEqual(condition.ErrorSummary, errorSummary) {
		updated = true
	} else {
		return updated, transitioned
	}
	condition.Reason = reason
	condition.Message = message
	condition.Commit = commit
	condition.ErrorSourceRefs = errorSources
	condition.ErrorSummary = errorSummary
	condition.LastUpdateTime = lastUpdate

	return updated, transitioned
}
```

sdowell: Yeah that LGTM

pkg/reposync/conditions.go
karlkfi: fixed

pkg/reposync/conditions.go
karlkfi: added

pkg/reposync/conditions.go
karlkfi: done

pkg/reposync/conditions.go
karlkfi: I routed the updates through the public Set calls so they can now be used by the reconciler and reonciler-manager, when we decide to do that refactor.

pkg/reposync/conditions_test.go
```diff
@@ -239,79 +274,263 @@ func TestClearCondition(t *testing.T) {
...
+		commit           string
+		errorSources     []v1beta1.ErrorSource
+		errorSummary     *v1beta1.ErrorSummary
+		timestamp        metav1.Time
+		want             []v1beta1.RepoSyncCondition
+		wantUpdated      bool
+		wantTransitioned bool
+	}{
+		{
+			"Set new syncing condition without error",
```
sdowell, line 409: Could you use named fields in the assignments? It's a bit hard to follow with so many parameters

pkg/reposync/conditions.go
sdowell: It still looks like consumers aren't using this

pkg/reposync/conditions.go
sdowell: I understand so I won't push too hard on this, but it would be nice to have more coverage here (e.g. to ensure that the equality comparisons work as expected and aren't broken in subsequent changes)

pkg/reposync/conditions.go
```diff
@@ -84,35 +85,46 @@ var singleErrorSummary = &v1beta1.ErrorSummary{
 }
 
 // SetReconciling sets the Reconciling condition to True.
-func SetReconciling(rs *v1beta1.RepoSync, reason, message string) {
-	if setCondition(rs, v1beta1.RepoSyncReconciling, metav1.ConditionTrue, reason, message, "", nil, &v1beta1.ErrorSummary{}, now()) {
+// Returns whether the condition was updated (any change) or transitioned
+// (status chnage).
```
sdowell, line 89: `status chnage` -> `status change` (ditto for other docstrings)

pkg/reposync/conditions.go
karlkfi: I added SetSyncing tests. They weren't being tested at all. :stuck_out_tongue: 

Feel free to add some expanded tests with the suggest feature or an added commit. I was initially using DeepEqual to avoid needing to test every field...

pkg/reposync/conditions_test.go
karlkfi: None of the others in the file use named fields... I was just being consistent. But I also hate this pattern.

pkg/reposync/conditions.go
sdowell: Thank you for adding the unit tests, more coverage is better :)

I think even if we were using DeepEqual then we should still unit test this to verify each field is getting correctly set and compared.

pkg/reposync/conditions_test.go
sdowell: I'm not suggesting you need to update everything else, just to use it for the new tests. But it's not a big deal, we can refactor it in a subsequent change.

pkg/reposync/conditions.go
karlkfi: SetReconciling & SetStalled use the `transitioned` bool to determine when to remove he Syncing condition. And I need the `updated` condition for the reconciler finalizer PR, where I add SetReconcilerFinalizing, RemoveReconcilerFinalizing, SetReconcilerFinalizerFailure, and RemoveReconcilerFinalizerFailure, all of which return the `updated` bool to decide whether to perform an UPDATE or not. Ideally the reconciler-manager and reconciler would use the `updated` bool for the same purpose, but I don't have time to do that refactor right now. So I just exposed both bools on all the Setter functions so we can use them later.

sdowell: Sounds good

pkg/reposync/conditions_test.go
karlkfi: Inserted them all with a neat vscode-go feature: https://github.com/golang/vscode-go/blob/master/docs/settings.md#goinlayhintscompositeliteralfields

PR #256, opened 2022-10-31
rename test-resources manifest directory

sdowell:
The test-resources name is a misnomer, since these manifests are no longer solely used for testing.


nan-yu (commented):
These resources are definitely for testing only, but they're not hack either.

They're required resources for Config Sync to functionally work.

What about renaming `00-namespace.yaml` to `cs-namespaces.yaml` and moving it along with `resourcegroup-manifest.yaml` to the `manifests` folder directly? They should also be included in `manifests/base/kustomization.yaml`.

sdowell:
> These resources are definitely for testing only, but they're not hack either.
> 
> They're required resources for Config Sync to functionally work.
> 
> What about renaming `00-namespace.yaml` to `cs-namespaces.yaml` and moving it along with `resourcegroup-manifest.yaml` to the `manifests` folder directly? They should also be included in `manifests/base/kustomization.yaml`.

They aren't testing only, they are used to build the multirepo manifests for OSS Config Sync installation. They also should not be added to `manifests/base/kustomization.yaml` since they are not used in ACM operator installation.

I'm fine with moving them directly to the manifests folder directly - However these are still kind of special resources since they are used to make Config Sync installable (e.g. in tests and in OSS) but not included in what we publish for ACM operator.

nan-yu:
> They aren't testing only, they are used to build the multirepo manifests for OSS Config Sync installation. 

Sorry for the confusion. I meant “These resources are definitely for **_NOT_** testing only, but they're not hack either.”

> They also should not be added to `manifests/base/kustomization.yaml` since they are not used in ACM operator installation.
> 
> I'm fine with moving them to the manifests folder directly - However these are still kind of special resources since they are used to make Config Sync installable (e.g. in tests and in OSS) but not included in what we publish for ACM operator.

That's right. They can't be in the base layer. It makes sense for them to be in the `oss` overlay, but not the `operator` overlay.

sdowell:
> What about renaming `00-namespace.yaml` to `cs-namespaces.yaml` and moving it along with `resourcegroup-manifest.yaml` to the `manifests` folder directly? 

I think conceptually it's reasonable to keep these in their own directory since they are special resources. The directory structure also seems to be designed around the old make targets, since they copy all files from the top level `manifests/` directory. I could work around this but I think it would be more straightforward to hold off on a flattening refactor until the old make targets are retired/removed (this will be done shortly after the monorepo migration is complete). Do you have a subdirectory name that you prefer to `hack`?

nan-yu:
> > What about renaming `00-namespace.yaml` to `cs-namespaces.yaml` and moving it along with `resourcegroup-manifest.yaml` to the `manifests` folder directly?
> 
> I think conceptually it's reasonable to keep these in their own directory since they are special resources. The directory structure also seems to be designed around the old make targets, since they copy all files from the top level `manifests/` directory. I could work around this but I think it would be more straightforward to hold off on a flattening refactor until the old make targets are retired/removed (this will be done shortly after the monorepo migration is complete). Do you have a subdirectory name that you prefer to `hack`?

I see. I'm okay with the separate directory, but still think it is not a `hack`, what about `standalone`, `dependencies`, or `cs-components`? `standalone` indicates these resources are needed for the standalone Config Sync installation. `dependencies` and `cs-components` mean that these resources are needed by Config Sync?

sdowell:
@nan-yu Actually now with the monorepo migration work done and the old make targets removed, I think we have a bit more flexibility now. I moved the namespaces file up to the manifests dir and renamed it `cs-namespaces.yaml` as you suggested. For the RG manifest I created a new `third_party` directory. WDYT?

nan-yu:
> Nan Yu Actually now with the monorepo migration work done and the old make targets removed, I think we have a bit more flexibility now. I moved the namespaces file up to the manifests dir and renamed it `cs-namespaces.yaml` as you suggested. For the RG manifest I created a new `third_party` directory. WDYT?

I like it.

nan-yu (approved):
/lgtm

PR #257, opened 2022-11-01
Fix vulnerability found in the git-sync image

nan-yu:
High: CVE-2022-43680 for expat/2.2.10-2+deb11u4 (debian) CVSS Score: 7.5
Fixed in 2.2.10-2+deb11u5

b/256594156

### DIFF
--- a/manifests/templates/git-importer.yaml
+++ b/manifests/templates/git-importer.yaml
@@ -99,7 +99,7 @@ spec:
             cpu: "150m"
             memory: "100Mi"
       - name: git-sync
-        image: gcr.io/config-management-release/git-sync:v3.6.1-gke.1__linux_amd64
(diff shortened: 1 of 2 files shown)


rquitales:
@nan-yu Should we also update `e2e/testdata/reconciler-manager-configmap-updated.yaml` just to ensure consistency with all our images?

nan-yu:
> Should we also update `e2e/testdata/reconciler-manager-configmap-updated.yaml` just to ensure consistency with all our images?

Not necessary. The  `e2e/testdata/reconciler-manager-configmap-updated.yaml` is used for testing only. It simulates an upgrade scenario, updating from `v1.9.1-rc.3` to the current version, so it is okay to use the old git-sync image.

PR #258, opened 2022-11-02
[WIP] Ignore updates while deleting RSync

karlkfi:
This new behavior allows the reconciler-manager to tolerate finalizers on RootSync & RepoSync.

The existing garbage collection behavior is preserved. It still deletes the Deployment, [Cluster]RoleBinding, and Secrets after the RSync is NotFound.

This is a temporary fix to pave the way for the reconciler finalizer without needing to first add the reconciler-manager finalizer.

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -124,6 +124,16 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		return controllerruntime.Result{}, status.APIServerError(err, "failed to get RepoSync")
 	}
 
+	if !rs.DeletionTimestamp.IsZero() {
+		// Deletion requested.
+		// Avoid making updates after deletion has started.
+		// Cleanup is handled after the RepoSync is NotFound.
+		log.Info("Ignoring update: deletion timestamp found",
+			logFieldObject, rsRef.String(),
+			logFieldKind, r.syncKind)
+		return controllerruntime.Result{}, nil
+	}
+
 	// The caching client sometimes returns an old R*Sync, because the watch
 	// hasn't received the update event yet. If so, error and retry.
 	// This is an optimization to avoid unnecessary API calls.
--- a/pkg/reconcilermanager/controllers/reposync_controller_test.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller_test.go
@@ -47,6 +47,7 @@ import (
 	"kpt.dev/configsync/pkg/validate/raw/validate"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -1646,7 +1647,7 @@ func TestMultipleRepoSyncs(t *testing.T) {
 		t.Error(err)
 	}
 
-	t.Log("Deployments, ServiceAccounts, and ClusterRoleBindings successfully created")
+	t.Log("Deployments, ServiceAccounts, and RoleBindings successfully created")
 
 	// Test updating Deployment resources for rs1: my-repo-sync
 	rs1.Spec.Git.Revision = gitUpdatedRevision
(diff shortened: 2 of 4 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -124,6 +124,16 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		return controllerruntime.Result{}, status.APIServerError(err, "failed to get RepoSync")
 	}
 
+	if !rs.DeletionTimestamp.IsZero() {
```
sdowell, line 127: I'm wondering whether we should check for the existence of a finalizer here. Was it intentional to only check deletion timestamp?

karlkfi: The finalizer doesn't really matter in this case, except that if there's no finalizers then the api server just deletes the object outright, rather than updating it with a deletionTimestamp. So I put the finalizer into the test to better simulate real behavior.

Once https://github.com/GoogleContainerTools/kpt-config-sync/pull/77 is merged, the fake.Client will handle adding the deletionTimestamp when a Delete is requested with a finalizer present, and then also deleting the object when the last finalizer is removed. So I'll have to update the test for that, but didn't want to block on that being merged first.

karlkfi: Perhaps more simply: The finalizer causes the deletionTimestamp. So we don't need to check for the finalizer. Without the finalizer, the deletionTimestamp is never added, so that code branch is never executed.

An example is the finalizer that k8s adds when you do a Foreground deletion. Then it looks for owned resource to delete before removing the finalizer. But since our fake.Client doesn't simulate that either yet, I just went with a "test-finalizer" to sort of imply that it doesn't matter what finalizer it is that causes the deletionTimestamp.

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -1806,6 +1807,129 @@ func validateGeneratedResourcesDeleted(t *testing.T, fakeClient *syncerFake.Clie
...
+		t.Fatalf("failed to update the repo sync, got error: %v, want error: nil", err)
+	}
+	if _, err := testReconciler.Reconcile(ctx, reqNamespacedName1); err != nil {
+		t.Fatalf("unexpected reconciliation error upon update, got error: %q, want error: nil", err)
+	}
+
+	// Validate nothing was updated
+	// TODO: Validate ResourceVersion has not been changed
+	// once the fake.Client has been updated to handle ResourceVersion simulation.
+	if err := validateServiceAccounts(wantServiceAccounts, fakeClient); err != nil {
```
sdowell, line 1893: Would anything have been updated even if the deletion timestamp wasn't set? I assume nothing would change here since neither the resources nor the RepoSync spec has changed since the previous reconcile. I'm wondering if we need to change something to verify the update is ignored. Or am I missing something?

pkg/reconcilermanager/controllers/reposync_controller.go
sdowell: Makes sense, thank you for explaining!

pkg/reconcilermanager/controllers/reposync_controller_test.go
karlkfi: My tests for ResourceVersion are in the fake.Client enhancement PR. Today we don't have any tests or way to prove that it hasn't actually updated because the fake.Client doesn't behave the same way the api server does. So I can't negative test it yet.

sdowell: Perhaps we could delete/modify an upserted resource or modify the RepoSync spec? I'm also fine with waiting until your other enhancement PR

sdowell:
/lgtm
/approve
/hold

LGTM but holding in case @nan-yu wants to take a look, feel free to unhold when you're ready to submit

pkg/reconcilermanager/controllers/reposync_controller_test.go
karlkfi: Fair, I could add a explicit update to ignore. Right now I guess it could pass just because of the DeepEqual skipping Updates when nothing changes. Tho ironically, we don't have a test for that yet either, because the fake.Client doesn't update ResourceVersion yet, so there's no way to prove it didn't happen yet.

karlkfi: Done.

karlkfi: I modified the RSync, to ignore the update, rather than modifying an updated resource, but both should be ignored.

sdowell: LGTM.

Minor nit is that these assertions on line 1893-1905 seem a bit redundant and could be rolled in with the assertions on line 1919-1931 with a single status update

karlkfi: > Minor nit is that these assertions on line 1893-1905 seem a bit redundant and could be rolled in with the assertions on line 1919-1931 with a single status update

I wanted to make sure both cases were covered. Yes, they could be merged, but I kind of like having failures from a minimal set of changes to make debugging easier.

karlkfi:
/hold

It's looking like this may not actually be required for the finalizer tests to pass. AND it prevents fixing config issues in the RSync and Secrets if the finalizer is erroring. So we might not need this at all.

I'm going to remove it from the finalizer PR and see if the tests still pass.

karlkfi:
This isn't needed for the Finalizer, and actually breaks some of the intended behavior. So I'm gonna close this. But I might pull out the SyncDelete tests just to have something as a baseline for expected behavior.

PR #260, opened 2022-11-03
[WIP] Update Syncing condition to include all errors

karlkfi:
- Update the Syncing condition to reflect errors from all phases
  with the current commit. This prevents hiding/dropping errors,
  especially non-blocking errors.
- Add/clarify some comments
- Set ErrorSummary and ErrorSources to nil when empty (no errors)

### DIFF
--- a/pkg/applier/kpt_applier.go
+++ b/pkg/applier/kpt_applier.go
@@ -278,6 +278,7 @@ func handleApplySkippedEvent(obj *unstructured.Unstructured, id core.ID, err err
 	return SkipErrorForResource(err, id, actuation.ActuationStrategyApply)
 }
 
+// processPruneEvent handles PruneEvents from the Applier
 func processPruneEvent(ctx context.Context, e event.PruneEvent, s *stats.PruneEventStats, objectStatusMap ObjectStatusMap, cs *clientSet) status.Error {
 	id := idFrom(e.Identifier)
 	klog.V(4).Infof("prune %v for object: %v", e.Status, id)
@@ -411,7 +412,7 @@ func (a *Applier) applyInner(ctx context.Context, objs []client.Object) (map[sch
 	// through annotation.
 	enabledObjs, disabledObjs := partitionObjs(objs)
 	if len(disabledObjs) > 0 {
-		klog.Infof("%v objects to be disabled: %v", len(disabledObjs), core.GKNNs(disabledObjs))
+		klog.Infof("Applier disabling %d objects: %v", len(disabledObjs), core.GKNNs(disabledObjs))
 		disabledCount, err := cs.handleDisabledObjects(ctx, a.inventory, disabledObjs)
 		if err != nil {
 			a.addError(err)
@@ -422,7 +423,7 @@ func (a *Applier) applyInner(ctx context.Context, objs []client.Object) (map[sch
 			Succeeded: disabledCount,
 		}
 	}
-	klog.Infof("%v objects to be applied: %v", len(enabledObjs), core.GKNNs(enabledObjs))
+	klog.Infof("Applier applying %d objects: %v", len(enabledObjs), core.GKNNs(enabledObjs))
 	resources, err := toUnstructured(enabledObjs)
 	if err != nil {
 		a.addError(err)
@@ -460,7 +461,7 @@ func (a *Applier) applyInner(ctx context.Context, objs []client.Object) (map[sch
 		switch e.Type {
 		case event.InitType:
 			for _, ag := range e.InitEvent.ActionGroups {
-				klog.Info("InitEvent", ag)
+				klog.Info("Applier InitEvent", ag)
 			}
 		case event.ActionGroupType:
 			klog.Info(e.ActionGroupEvent)
@@ -502,7 +503,7 @@ func (a *Applier) applyInner(ctx context.Context, objs []client.Object) (map[sch
 			klog.V(4).Info(logEvent)
 			a.addError(processPruneEvent(ctx, e.PruneEvent, s.PruneEvent, objStatusMap, cs))
 		default:
-			klog.V(4).Infof("skipped %v event", e.Type)
+			klog.V(4).Infof("Applier skipped %v event", e.Type)
 		}
 	}
 
@@ -517,12 +518,12 @@ func (a *Applier) applyInner(ctx context.Context, objs []client.Object) (map[sch
 
 	errs := a.Errors()
 	if errs == nil {
-		klog.V(4).Infof("all resources are up to date.")
+		klog.V(4).Infof("Applier completed without error")
 	}
 	if s.Empty() {
-		klog.V(4).Infof("The applier made no new progress")
+		klog.V(4).Infof("Applier made no new progress")
 	} else {
-		klog.Infof("The applier made new progress: %s", s.String())
+		klog.Infof("Applier made new progress: %s", s.String())
 		objStatusMap.Log(klog.V(0))
 	}
 	return gvks, errs
@@ -558,6 +559,7 @@ func (a *Applier) invalidateErrors() {
 	a.errs = nil
 }
 
+// Apply all managed resource objects and return any errors.
 // Apply implements Interface.
 func (a *Applier) Apply(ctx context.Context, desiredResource []client.Object) (map[schema.GroupVersionKind]struct{}, status.MultiError) {
 	a.applyMux.Lock()
--- a/pkg/parse/namespace.go
+++ b/pkg/parse/namespace.go
@@ -41,6 +41,7 @@ import (
 )
 
 // NewNamespaceRunner creates a new runnable parser for parsing a Namespace repo.
+// TODO: replace with builder pattern to avoid too many arguments.
 func NewNamespaceRunner(clusterName, syncName, reconcilerName string, scope declared.Scope, fileReader reader.Reader, c client.Client, pollingPeriod, resyncPeriod, retryPeriod, statusUpdatePeriod time.Duration, fs FileSource, dc discovery.DiscoveryInterface, resources *declared.Resources, app applier.KptApplier, rem remediator.Interface) (Parser, error) {
 	converter, err := declared.NewValueConverter(dc)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


pkg/applier/kpt_applier.go
```diff
@@ -508,55 +528,79 @@ func (a *Applier) sync(ctx context.Context, objs []client.Object) (map[schema.Gr
...
-	// TODO: Here and elsewhere, pass the MultiError as a parameter.
-	return a.sync(ctx, desiredResource)
+	a.syncLock.Lock()
+	defer a.syncLock.Unlock()
+
+	a.setSyncing(true)
+	defer a.setSyncing(false)
+
+	// Clear the errors at the start of a new apply.
+	// TODO: Keep track of old errors and new errors and merge them, until applyInner returns.
```
karlkfi, line 599: It turns out the parse already does a DeepEquals on the before and after MultiError to prevent unnecessary updates, but when the async status updates were added, they stomp all over the cache and the invalidation done by the parser...

This all needs a rewrite, but I don't have the time or energy to do it now.

pkg/parse/root.go
```diff
@@ -385,9 +385,6 @@ func (p *root) setSyncStatusWithRetries(ctx context.Context, errs status.MultiEr
 
 	currentRS := rs.DeepCopy()
 
-	// syncing indicates whether the applier is syncing.
-	syncing := p.applier.Syncing()
```
karlkfi, line 389: The reason I removed the Syncing call here is that the Syncing bool and Applier MultiError should come from the same time. It's already sketchy that they weren't locked, but having them be pulled at different times makes it hard to clean up. So now hopefully the next cleanup can somehow synchronize them a little better, because they're both being passed into the update function from the same source.

pkg/parse/run.go
```diff
@@ -368,12 +369,22 @@ func parseAndUpdate(ctx context.Context, p Parser, trigger string, state *reconc
...
 		}
+		// Prepend any remediator errors
+		// TODO: combine this with SetSyncStatus so it's a single update
+		remediatorErrs := p.RemediatorConflictErrors()
+		if len(remediatorErrs) > 0 {
+			if err := UpdateConflictManagerStatus(ctx, remediatorErrs, p.K8sClient()); err != nil {
+				syncErrs = status.Append(syncErrs, err)
+			}
+		}
+		// TODO: Am I suppossed to update state.syncStatus or state.syncingConditionLastUpdate again?
```
karlkfi, line 387: I really dont understand how this works.

Why is it called a `syncStatus` if it's a `sourceStatus`?

Why does it update the `state` on success to a hypothetical state that isn't actually passed into `SetSyncStatus`?

Why does it store a `now()` and then generate a new now timestamp when setting the sync status?

Why does it update the status and then modify the status? This should all be one big reconcile function, not a series of updates that drop and re-add errors....

e2e/testcases/multi_sync_test.go
```diff
@@ -399,10 +399,21 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
...
+	// RootSyncs, the manager annotation won't change, so the first RootSync
+	// won't get any management conflict errors.
+	// The first RootSync would only get an error if the webhook was disabled.
+
+	// TODO: Should adoption fights happen with the webhook enabled?
+	// The remediator doesn't return a KptManagementConflictError, so the
+	// following error validation will never succeed.
+
+	// nt.T.Logf("Root %s should also get surfaced with the conflict error", configsync.RootSyncName)
+	// nt.WaitForRootSyncSyncError(configsync.RootSyncName, status.ManagementConflictErrorCode, "declared in another repository")
```
karlkfi, line 415: AFAIK, the `declared in another repository` text only comes from the `KptManagementConflictError`, which only comes from the Applier, not the Remediator. If the manager is correct, the error won't be returned unless the object is adopted. But it can't be adopted with the webhook enabled, because the webhook rejects it, causing the `KptManagementConflictError` in the second RootSync, but not the first.

So the previous test validation seems impossible. Either the test is wrong and only worked because the status updates were flakey, OR the behavior is wrong and we need to change either the webhook or the remediator to make the test behavior reality.

karlkfi:
Extracted the duration period cleanup: https://github.com/GoogleContainerTools/kpt-config-sync/pull/273

karlkfi:
Extracted periodic status updates: https://github.com/GoogleContainerTools/kpt-config-sync/pull/279

karlkfi:
Extracted part of the logic to combine sync errors: https://github.com/GoogleContainerTools/kpt-config-sync/pull/281

The logic to preserve the non-applier errors from the updater will be done in a subsequent PR.

karlkfi:
Extracted thread-safety improvements to https://github.com/GoogleContainerTools/kpt-config-sync/pull/301

karlkfi:
Extracted the hydration e2e changes: https://github.com/GoogleContainerTools/kpt-config-sync/pull/310

pkg/parse/state.go
```diff
@@ -28,36 +28,40 @@ const (
...
-	message    string
-	errs       status.MultiError
-	lastUpdate metav1.Time
+	commonStatus
+	message string
 }
 
-func (rs renderingStatus) equal(other renderingStatus) bool {
-	return rs.commit == other.commit && rs.message == other.message && status.DeepEqual(rs.errs, other.errs)
+func (rs renderingStatus) Equal(other renderingStatus) bool {
```
haiyanmeng, line 54: Why do we need to convert `equal` to `Equal`? This method is only called in the same pacakge.

pkg/parse/state.go
```diff
@@ -28,36 +28,40 @@ const (
...
-	commit     string
-	errs       status.MultiError
-	lastUpdate metav1.Time
+	commonStatus
+	syncing bool
 }
 
-func (gs syncStatus) equal(other syncStatus) bool {
-	return gs.syncing == other.syncing && gs.commit == other.commit && status.DeepEqual(gs.errs, other.errs)
+func (ss syncStatus) Equal(other syncStatus) bool {
```
haiyanmeng, line 63: Why do we need to convert `equal` to `Equal`? This method is only called in the same package.

pkg/parse/state.go
```diff
@@ -28,36 +28,40 @@ const (
...
+type commonStatus struct {
 	commit     string
 	errs       status.MultiError
 	lastUpdate metav1.Time
 }
 
-func (gs sourceStatus) equal(other sourceStatus) bool {
-	return gs.commit == other.commit && status.DeepEqual(gs.errs, other.errs)
+func (cs commonStatus) Equal(other commonStatus) bool {
+	return cs.commit == other.commit && status.DeepEqual(cs.errs, other.errs)
```
haiyanmeng, line 38: Why do we need to convert `equal` to `Equal`? This method is only called in the same package.

pkg/parse/state.go
karlkfi: We don't _have_ to, but there's no reason to have private methods on a private struct. No one outside the package can access the struct methods anyway, unless they were on a public interface.

pkg/parse/run.go
```diff
@@ -354,20 +361,22 @@ func parseSource(ctx context.Context, p Parser, trigger string, state *reconcile
...
 func parseAndUpdate(ctx context.Context, p Parser, trigger string, state *reconcilerState) status.MultiError {
 	sourceErrs := parseSource(ctx, p, trigger, state)
-	newSourceStatus := sourceStatus{
-		commit:     state.cache.source.commit,
-		errs:       sourceErrs,
-		lastUpdate: metav1.Now(),
-	}
-	if state.needToSetSourceStatus(newSourceStatus) {
-		if err := p.setSourceStatus(ctx, newSourceStatus); err != nil {
+	parseStatus := sourceStatus{
```
sdowell, line 364: I think I'm a bit confused as to what a sourceStatus is supposed to represent. Looks like here it's related to issues with parsing (i.e. marshalling?) but above it's related to issues reading some config files. Is it supposed to represent any error with reading from the filesystem at any stage of the pipeline?

sdowell (commented):
Really like the refactoring. Left a few questions/comments but overall LGTM

pkg/parse/state.go
haiyanmeng: The reason I asked this is that when I added the `equal` method back then, I was told that I should avoid having a public method on a private struct.

karlkfi: That’s silly. It happens all the time. It’s common practice to make a private struct that satisfies a public interface and provide a constructor that returns the public interface as implemented by the private struct with public methods.

karlkfi: The think you want to avoid is having a public function or method that returns a private struct in its signature, because the caller won’t be able to create a variable to store the return value.

haiyanmeng: I see. Thanks for the explanation.

pkg/parse/namespace.go
```diff
@@ -161,13 +162,14 @@ func (p *namespace) setSourceStatusWithRetries(ctx context.Context, newStatus so
...
 
-	continueSyncing := (rs.Status.Source.ErrorSummary.TotalCount == 0)
-	var errorSource []v1beta1.ErrorSource
-	if len(rs.Status.Source.Errors) > 0 {
-		errorSource = []v1beta1.ErrorSource{v1beta1.SourceError}
-	}
-	reposync.SetSyncing(&rs, continueSyncing, "Source", "Source", newStatus.commit, errorSource, rs.Status.Source.ErrorSummary, newStatus.lastUpdate)
+	// Syncing is stopped/prevented if there are any errors from the same commit.
+	// TODO: Handle non-blocking errors
+	syncing := (syncErrorCount == 0)
```
nan-yu, line 170: Previously, `syncing` only depends on the source error. Now, it is changed to include all errors, including the former render or sync errors with the same commit.

It affects how the notification sends out alerts.

The current proposal is to leverage the Syncing condition to determine the sync status and send out alerts.

If the Syncing condition is `false`, it indicates the current apply loop is done. If there're no errors, then we send out `success` notification. Otherwise, it is `failed`.

In this case, if there're errors from the same commit, even if it might be fixed with the new loop, there will be a failure notification.

We can include previous errors with the same commit in the ErrorSummary to surface the errors earlier, but can we keep `syncing` only depend on the errors from the current phase (source, render, or sync)?

@sdowell FYI

karlkfi: > We can include previous errors with the same commit in the ErrorSummary to surface the errors earlier, but can we keep syncing only depend on the errors from the current phase (source, render, or sync)?

It's not super clear what the Syncing condition was/is supposed to represent. 

Internally, the phases are effectively Fetch/Read -> Render/Read -> Parse -> Update (Validate/Apply/Watch/Reconcile). 

It's confusing to have the `status.sync` field represent just the errors from the Updater (which wraps the Applier) and then have the Syncing conditions which is set to True by the previous source and rendering phases.

This PR effectively assumes that since the source and rendering phases where setting the Syncing condition to True that the Syncing condition must include the entire pipeline, and not just the Update/Apply phase.

If that is NOT what we want, then we need to document the proposed change and implement it consistency.

karlkfi: Also, it's worth noting that the inputs to a sync/apply are not just Git. They also can include Helm values in the spec and other config values, like Git dir, URL, cluster name, etc. So it doesn't make sense to have a unique pipeline execution depend on just a commit. Even just a generation would be technically insufficient, because the previous run may fail, perhaps due to some ephemeral network issue, triggering a retry. Today we don't indicate what generation or attempt/retry the Syncing condition represents, which makes it hard to use for triggering discrete alerts.

nan-yu: The Syncing condition represents the _**current/latest**_ fetch -> render -> parse -> update pipeline, while `status.source`, `status.rendering`, and `status.sync` represent the statuses of the last processed commit in each phase.

It is possible that the Syncing condition only includes status of **_partial_** pipeline while the loop is still in progress.

For example,
1. The first pipeline failed at commit A in the apply phase, `status.sync` will have errors for commit A.
2. The second pipeline failed to fetch commit A, `status.source` will have errors for commit A, and `status.sync` still contains the former apply errors.
3. The current pipeline succeeded in fetching and rendering commit A, source errors from the second pipeline will get cleared from `status.source`. Then the `parseAndUpdate` function will call `p.setSourceStatus`, which calls the `setSourceStatusWithRetries` function. The `setSourceStatusWithRetries` function should set Syncing condition to `true` because it will continue applying the parsed objects next. However, this PR sets the syncing condition to `false` because of the sync errors from the first pipeline.

mikebz:
curious if this is still WIP or if it's even relevant.  If it is consider closing the PR and keeping the private branch or converting to draft.

karlkfi:
The parser/applier code has been changed enough since this PR that it would need a rewrite. I'm just going to abandon it for now.

PR #262, opened 2022-11-05
set fake shared env as failed when a test fails

sdowell:
some of the cleanup functions check for Failed() when determining cleanup behavior. an example of this is the kind cluster cleanup, where the cluster is preserved if Failed() is true and the debug flag is provided. this change makes it so that such cleanup blocks can honor the --debug flag for shared kind environments.


karlkfi (approved):
nice, thanks!
/lgtm

PR #264, opened 2022-11-07
remove monorepo e2e test targets

sdowell:
These targets are no longer used by CI and can be removed. More thorough cleanup will be performed by a subsequent change.


nan-yu (approved):
/lgtm

PR #265, opened 2022-11-07
remove targets related to e2e-tests image

sdowell:
This image was used by the bats tests and is thus no longer used. This seems to have been missed in the bats test cleanup.


Makefile.e2e
```diff
@@ -5,36 +5,8 @@ E2E_PARAMS := \
...
-# The directory for e2e test related files
-E2E_DIR := $(OUTPUT_DIR)/nomos
-
-E2E_VERSION := v1.0.2
-E2E_TEST_IMAGE := gcr.io/$(GCP_PROJECT)/e2e-tests:$(E2E_VERSION)
-
-# Builds the e2e docker image and dependencies.
-image-e2e-tests: \
-		build/e2e-tests/Dockerfile \
-		build/e2e-tests/build.sh
```
nan-yu, line 17: These files may not be needed as well.

sdowell: Thank you for noticing this - Done!

PR #266, opened 2022-11-08
Add last sync timestamp into `nomos status`

haiyanmeng:


cmd/nomos/status/cluster_state.go
```diff
@@ -71,23 +71,28 @@ func unavailableCluster(ref string) *ClusterState {
...
+	errors            []string
 	// errorSummary summarizes the `errors` field.
 	errorSummary *v1beta1.ErrorSummary
 	resources    []resourceState
 }
 
 func (r *RepoState) printRows(writer io.Writer) {
 	fmt.Fprintf(writer, "%s%s:%s\t%s\t\n", util.Indent, r.scope, r.syncName, sourceString(r.sourceType, r.git, r.oci, r.helm))
-	fmt.Fprintf(writer, "%s%s\t%s\t\n", util.Indent, r.status, r.commit)
+	if r.status == syncedMsg {
```
nan-yu, line 91: Should we always print the lastSyncTimestamp if it was synced before?

The current sync status might be ERROR, but we can still track the timestamp of last successful sync. This requires adding a `LastSyncedTimestamp` field in the status. The new field can be set along with the `LastSyncedCommit` field: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/api/configsync/v1beta1/sync_types.go#L36.

haiyanmeng: If the sync status is not SYNCED, the users probably care more about why it is not synced. 

I don't feel comfortable to add new fields into the status APIs considering the feature complete date for 1.14 is coming soon. 

We can first deliver this PR to see whether this is good enough for the customers who want this feature. If not, we can explore the option of adding new status API fields.

nan-yu: sounds good

nan-yu (approved):
/lgtm

PR #269, opened 2022-11-08
Resolve the fight between reconciler manager and Autopilot

xinnywinne:
Create an allowlist in reconciler manager and add the fields that we
    allow others to modify to the allowlist.
    
    When compare the current reconciler deployment with the declared
    reconciler deployment, ignore the differences of the fields in the
    allowlist.
    
    Use Sever Side Apply to only manage the fields defined in
    reconciler-manager-configmap to prevent the reconciler manager from fighting with other controllers 
    on the fields we do not define

### DIFF
--- a/cmd/reconciler-manager/main.go
+++ b/cmd/reconciler-manager/main.go
@@ -24,7 +24,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2/klogr"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/core"
@@ -77,10 +76,14 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	dynamicClient, err := dynamic.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		setupLog.Error(err, "failed to build dynamic client")
+		os.Exit(1)
+	}
+	watchFleetMembership := fleetMembershipCRDExists(dynamicClient, mgr.GetRESTMapper())
 
-	watchFleetMembership := fleetMembershipCRDExists(mgr.GetConfig(), mgr.GetRESTMapper())
-
-	repoSync := controllers.NewRepoSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod, mgr.GetClient(),
+	repoSync := controllers.NewRepoSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod, mgr.GetClient(), dynamicClient,
 		ctrl.Log.WithName("controllers").WithName(configsync.RepoSyncKind),
 		mgr.GetScheme())
 	if err := repoSync.SetupWithManager(mgr, watchFleetMembership); err != nil {
@@ -88,7 +91,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	rootSync := controllers.NewRootSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod, mgr.GetClient(),
+	rootSync := controllers.NewRootSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod, mgr.GetClient(), dynamicClient,
 		ctrl.Log.WithName("controllers").WithName(configsync.RootSyncKind),
 		mgr.GetScheme())
 	if err := rootSync.SetupWithManager(mgr, watchFleetMembership); err != nil {
@@ -137,12 +140,8 @@ func main() {
 
 // fleetMembershipCRDExists checks if the fleet membership CRD exists.
 // It checks the CRD first so that the controller can watch the Membership resource in the startup time.
-func fleetMembershipCRDExists(config *rest.Config, mapper meta.RESTMapper) bool {
-	dc, err := dynamic.NewForConfig(config)
-	if err != nil {
-		setupLog.Error(err, "failed to build dynamic client")
-		os.Exit(1)
-	}
+func fleetMembershipCRDExists(dc dynamic.Interface, mapper meta.RESTMapper) bool {
+
 	crdRESTMapping, err := mapper.RESTMapping(kinds.CustomResourceDefinition())
 	if err != nil {
 		setupLog.Error(err, "failed to get mapping of CRD type")
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -20,6 +20,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/client-go/util/retry"
 	"kpt.dev/configsync/e2e/nomostest"
@@ -63,7 +64,7 @@ func TestManagingReconciler(t *testing.T) {
 	generation := reconcilerDeployment.Generation
 	managedImage := reconcilerDeployment.Spec.Template.Spec.Containers[0].Image
 	managedReplicas := *reconcilerDeployment.Spec.Replicas
-	managedMemoryLimit := reconcilerDeployment.Spec.Template.Spec.Containers[0].Resources.Limits.Memory().DeepCopy()
+	originalTolerations := reconcilerDeployment.Spec.Template.Spec.Tolerations
 
 	// test case 1: The reconciler-manager should manage most of the fields with one exception:
 	// - changes to the container resource requirements should be ignored when the autopilot annotation is set.
@@ -92,28 +93,39 @@ func TestManagingReconciler(t *testing.T) {
 			hasReplicas(managedReplicas))
 	})
 
-	// test case 3: override the memory limit has been moved to TestAutopilotReconcilerAdjustment
-
-	// test case 4: manually update the memory limit.
-	// The reconciler-manager manages the container resource requirements on both
-	// autopilot clusters and non-autopilot clusters. Any manual changes will be reverted.
-	rs := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
-	nt.T.Log("Manually update the first container CPU and Memory limit")
+	// test case 3:  the reconciler-manager should not revert the change to the fields that are not owned by reconciler-manager
+	nt.T.Log("Manually update fields that are not owned by reconciler-manager")
+	var modifiedTolerations []corev1.Toleration
 	mustUpdateRootReconciler(nt, func(d *appsv1.Deployment) {
-		d.Spec.Template.Spec.Containers[0].Resources.Limits = corev1.ResourceList{
-			"cpu":    resource.MustParse(updatedFirstContainerCPULimit),
-			"memory": resource.MustParse(updatedFirstContainerMemoryLimit),
-		}
-	})
-	nt.T.Log("Verify the reconciler-manager should revert the memory limit change")
-	generation += 2 // generation bumped by 2 because the change will be first applied then reverted by the reconciler-manager
-	nomostest.Wait(nt.T, "the memory limit to be reverted", nt.DefaultWaitTimeout, func() error {
-		return nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem,
-			&appsv1.Deployment{}, hasGeneration(generation),
-			firstContainerMemoryLimitIs(managedMemoryLimit.String()))
+		d.Spec.Template.Spec.Containers[0].TerminationMessagePath = "dev/termination-message"
+		d.Spec.Template.Spec.Containers[0].Stdin = true
+		d.Spec.Template.Spec.Tolerations = append(d.Spec.Template.Spec.Tolerations, corev1.Toleration{Key: "kubernetes.io/arch", Effect: "NoSchedule", Operator: "Exists"})
+		modifiedTolerations = d.Spec.Template.Spec.Tolerations
+		d.Spec.Template.Spec.PriorityClassName = "system-node-critical"
 	})
 
-	// test case 5: the reconciler-manager should update the reconciler Deployment if the manifest in the ConfigMap has been changed.
+	nt.T.Log("Verify the reconciler-manager does not revert the change")
+	generation++ // generation bumped by 1 because reconicler-manager should not revert this change
+	nomostest.Wait(nt.T, "the reconciler deployment to be updated", nt.DefaultWaitTimeout, func() error {
+		return nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem,
+			&appsv1.Deployment{}, hasGeneration(generation), firstContainerTerminationMessagePathIs("dev/termination-message"),
+			firstContainerStdinIs(true), hasTolerations(modifiedTolerations), hasPriorityClassName("system-node-critical"))
+	})
+	// change the fields back to default values
+	mustUpdateRootReconciler(nt, func(d *appsv1.Deployment) {
+		d.Spec.Template.Spec.Containers[0].TerminationMessagePath = "dev/termination-log"
+		d.Spec.Template.Spec.Containers[0].Stdin = false
+		d.Spec.Template.Spec.Tolerations = originalTolerations
+		d.Spec.Template.Spec.PriorityClassName = ""
+	})
+	generation++ // generation bumped by 1 because reconciler-manager should not revert this change
+	nomostest.Wait(nt.T, "the reconciler deployment to be updated", nt.DefaultWaitTimeout, func() error {
+		return nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem,
+			&appsv1.Deployment{}, hasGeneration(generation), firstContainerTerminationMessagePathIs("dev/termination-log"),
+			firstContainerStdinIs(false), hasTolerations(originalTolerations), hasPriorityClassName(""))
+	})
+
+	// test case 4: the reconciler-manager should update the reconciler Deployment if the manifest in the ConfigMap has been changed.
 	nt.T.Log("Update the Deployment manifest in the ConfigMap")
 	nt.MustKubectl("apply", "-f", "../testdata/reconciler-manager-configmap-updated.yaml")
 	nt.T.Log("Restart the reconciler-manager to pick up the manifests change")
@@ -131,8 +143,9 @@ func TestManagingReconciler(t *testing.T) {
 			firstContainerImageIsNot(managedImage))
 	})
 
-	// test case 6: the reconciler-manager should delete the git-creds volume if not needed
+	// test case 5: the reconciler-manager should delete the git-creds volume if not needed
 	currentVolumesCount := len(reconcilerDeployment.Spec.Template.Spec.Volumes)
+	rs := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
 	nt.T.Log("Switch the auth type from ssh to none")
 	nt.MustMergePatch(rs, `{"spec": {"git": {"auth": "none", "secretRef": {"name":""}}}}`)
 	nt.T.Log("Verify the git-creds volume is gone")
@@ -143,7 +156,7 @@ func TestManagingReconciler(t *testing.T) {
 			gitCredsVolumeDeleted(currentVolumesCount))
 	})
 
-	// test case 7: the reconciler-manager should add the gcenode-askpass-sidecar container when needed
+	// test case 6: the reconciler-manager should add the gcenode-askpass-sidecar container when needed
 	nt.T.Log("Switch the auth type from none to gcpserviceaccount")
 	nt.MustMergePatch(rs, `{"spec":{"git":{"auth":"gcpserviceaccount","secretRef":{"name":""},"gcpServiceAccountEmail":"test-gcp-sa-email@test-project.iam.gserviceaccount.com"}}}`)
 	nt.T.Log("Verify the gcenode-askpass-sidecar container should be added")
@@ -158,7 +171,7 @@ func TestManagingReconciler(t *testing.T) {
 			templateForGcpServiceAccountAuthType())
 	})
 
-	// test case 8: the reconciler-manager should mount the git-creds volumes again if the auth type requires a git secret
+	// test case 7: the reconciler-manager should mount the git-creds volumes again if the auth type requires a git secret
 	nt.T.Log("Switch the auth type gcpserviceaccount to ssh")
 	nt.MustMergePatch(rs, `{"spec":{"git":{"auth":"ssh","secretRef":{"name":"git-creds"}}}}`)
 	nt.T.Log("Verify the git-creds volume exists and the gcenode-askpass-sidecar container is gone")
@@ -215,7 +228,56 @@ func hasGeneration(generation int64) nomostest.Predicate {
 		return nil
 	}
 }
-
+func firstContainerTerminationMessagePathIs(terminationMessagePath string) nomostest.Predicate {
+	return func(o client.Object) error {
+		d, ok := o.(*appsv1.Deployment)
+		if !ok {
+			return nomostest.WrongTypeErr(d, &appsv1.Deployment{})
+		}
+		if d.Spec.Template.Spec.Containers[0].TerminationMessagePath != terminationMessagePath {
+			return fmt.Errorf("expected first container terminationMessagePath is: %s, got: %s", terminationMessagePath, d.Spec.Template.Spec.Containers[0].TerminationMessagePath)
+		}
+		return nil
+	}
+}
+func firstContainerStdinIs(stdin bool) nomostest.Predicate {
+	return func(o client.Object) error {
+		d, ok := o.(*appsv1.Deployment)
+		if !ok {
+			return nomostest.WrongTypeErr(d, &appsv1.Deployment{})
+		}
+		if d.Spec.Template.Spec.Containers[0].Stdin != stdin {
+			return fmt.Errorf("expected first container stdin is: %t, got: %t", stdin, d.Spec.Template.Spec.Containers[0].Stdin)
+		}
+		return nil
+	}
+}
+func hasTolerations(tolerations []corev1.Toleration) nomostest.Predicate {
+	return func(o client.Object) error {
+		d, ok := o.(*appsv1.Deployment)
+		if !ok {
+			return nomostest.WrongTypeErr(d, &appsv1.Deployment{})
+		}
+		for i, toleration := range d.Spec.Template.Spec.Tolerations {
+			if !equality.Semantic.DeepEqual(toleration, tolerations[i]) {
+				return fmt.Errorf("expected toleration is: %s, got: %s", tolerations[i].String(), toleration.String())
+			}
+		}
+		return nil
+	}
+}
+func hasPriorityClassName(priorityClassName string) nomostest.Predicate {
+	return func(o client.Object) error {
+		d, ok := o.(*appsv1.Deployment)
+		if !ok {
+			return nomostest.WrongTypeErr(d, &appsv1.Deployment{})
+		}
+		if d.Spec.Template.Spec.PriorityClassName != priorityClassName {
+			return fmt.Errorf("expected priorityClassName is: %s, got: %s", priorityClassName, d.Spec.Template.Spec.PriorityClassName)
+		}
+		return nil
+	}
+}
 func firstContainerImageIs(image string) nomostest.Predicate {
 	return func(o client.Object) error {
 		d, ok := o.(*appsv1.Deployment)
@@ -497,7 +559,7 @@ func TestAutopilotReconcilerAdjustment(t *testing.T) {
 	}
 	expectedTotalMemory += 100
 	// Waiting for reconciliation and validation
-	nomostest.Wait(nt.T, "Waiting for first reconciliation", nt.DefaultWaitTimeout, func() error {
+	nomostest.Wait(nt.T, "the first container resource requests to be increased", nt.DefaultWaitTimeout, func() error {
 		rd := &appsv1.Deployment{}
 		err := nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem, rd,
 			hasGeneration(generation), totalContainerCPURequestIs(expectedTotalCPU),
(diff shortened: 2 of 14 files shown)


pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -2867,11 +2888,17 @@ func validateClusterRoleBinding(want *rbacv1.ClusterRoleBinding, fakeClient *syn
 	return nil
 }
 
-// validateDeployments validates that important fields in the `wants` deployments match those same fields in the deployments found in the fakeClient
-func validateDeployments(wants map[core.ID]*appsv1.Deployment, fakeClient *syncerFake.Client) error {
+// validateDeployments validates that important fields in the `wants` deployments match those same fields in the current deployments found in the unstructured Map
+func validateDeployments(wants map[core.ID]*appsv1.Deployment, currents map[core.ID]*unstructured.Unstructured) error {
```
nan-yu, line 2892: Suggest passing the fake dynamic client as argument. From the caller's point of view, it doesn't need to know the underlying data of the fake client when calling the `validateDeployments` function.

pkg/reconcilermanager/controllers/jsonpath.go
```diff
@@ -0,0 +1,72 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/spyzhov/ajson"
+	"gopkg.in/yaml.v3"
+	"k8s.io/klog/v2"
+)
+
+// deleteFields evaluates the JSONPath expression to delete fields from the input map.
+func deleteFields(obj map[string]interface{}, expression string) error {
```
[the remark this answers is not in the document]
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -184,17 +188,26 @@ func (r *reconcilerBase) upsertDeployment(ctx context.Context, reconcilerRef typ
...
 			return controllerutil.OperationResultNone, err
 		}
 		r.log.V(3).Info("Managed object not found, creating",
 			logFieldObject, dRef.String(),
 			logFieldKind, kind)
-		if err := r.client.Create(ctx, declared); err != nil {
+		unObj, err := kinds.ToUnstructured(declared, r.scheme)
+		if err != nil {
+			return controllerutil.OperationResultNone, err
+		}
```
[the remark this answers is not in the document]
xinnywinne: Yes, changed to marshal declared object

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -184,17 +188,26 @@ func (r *reconcilerBase) upsertDeployment(ctx context.Context, reconcilerRef typ
 func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *appsv1.Deployment) (controllerutil.OperationResult, error) {
 	dRef := client.ObjectKeyFromObject(declared)
 	kind := "Deployment"
-
-	current := &appsv1.Deployment{}
-
-	if err := r.client.Get(ctx, dRef, current); err != nil {
+	forcePatch := true
+	fieldManager := "reconciler-manager"
```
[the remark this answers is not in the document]
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -237,17 +269,31 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
 			return controllerutil.OperationResultNone, err
 		}
-		if err := r.client.Create(ctx, declared); err != nil {
+		_, err = r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace).Patch(ctx, dRef.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: fieldManager, Force: &forcePatch})
+		if err != nil {
 			return controllerutil.OperationResultNone, err
 		}
 	}
 
 	return controllerutil.OperationResultUpdated, nil
```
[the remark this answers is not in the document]
xinnywinne: yes, add logic to compare the generation, only return OperationResultUpdated when generation after patch is different from the current generation.

pkg/reconcilermanager/controllers/reposync_controller_test.go
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -220,14 +239,27 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
 
 	r.log.V(3).Info("Managed object found, updating",
 		logFieldObject, dRef.String(),
 		logFieldKind, kind)
-	if err := r.client.Update(ctx, declared); err != nil {
+	data, err := json.Marshal(unObjDeclared)
+	if err != nil {
+		return controllerutil.OperationResultNone, fmt.Errorf("failed to marshal unstructured to byte array %w", err)
+	}
+	_, err = r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace).Patch(ctx, dRef.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: fieldManager, Force: &forcePatch})
```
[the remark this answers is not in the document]
xinnywinne: Done

e2e/testcases/reconciler_manager_test.go
```diff
@@ -92,25 +92,37 @@ func TestManagingReconciler(t *testing.T) {
...
+		d.Spec.Template.Spec.Containers[0].ImagePullPolicy = "IfNotPresent"
+		d.Spec.Template.Spec.Containers[0].Stdin = false
+		d.Spec.Template.Spec.Tolerations = originalTolerations
+		d.Spec.Template.Spec.PriorityClassName = ""
+		d.Spec.Paused = false
+	})
+	generation++ // generation bumped by 1 because reconciler-manager should not revert this change
+	nomostest.Wait(nt.T, "the reconciler deployment to be updated", nt.DefaultWaitTimeout, func() error {
+		return nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem,
+			&appsv1.Deployment{}, hasGeneration(generation), containerImagePullPolicy("IfNotPresent"))
```
nan-yu, line 125: The test updates a few fields, but it only verifies the imagePullPolicy. It would be good to see other fields are verified as well.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -515,6 +528,25 @@ func TestAutopilotReconcilerAdjustment(t *testing.T) {
 		return nil
 	})
 
+	// manually increase first container CPU and memory request to above current user override request
```
nan-yu, line 531: I feel this test can be combined with `e2e/testcases/override_resource_limits_test.go`: https://github.com/GoogleContainerTools/kpt-config-sync/pull/289.

This can be done in later PR.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -568,14 +636,43 @@ func TestAutopilotReconcilerAdjustment(t *testing.T) {
...
 		return nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem, &appsv1.Deployment{},
 			hasGeneration(generation), totalContainerCPURequestIs(expectedTotalCPU),
 			totalContainerMemoryRequestIs(expectedTotalMemory), firstContainerCPURequestIs(expectedFirstContainerCPU),
 			firstContainerMemoryRequestIs(expectedFirstContainerMemory))
 	})
 
 	// limit change
+	// Manually update resource limits
+	// The reconciler-manager does not manage the container resource limit if user does not override the resource limit through our API.
+	// In this case, we allow user manually update the memory limit.
```
nan-yu, line 649: Can you update the comment to make it clear that we only allow manual update on the standard cluster. For Autopilot, the limit is the same as request.

pkg/reconcilermanager/controllers/jsonpath_test.go
```diff
@@ -0,0 +1,426 @@
...
+        runAsUser: 1000
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+  revisionHistoryLimit: 15 # different from declared
+  progressDeadlineSeconds: 300 # different from declared
+  paused: true # different from declared
+`
+
+func TestIsDeploymentSame(t *testing.T) {
```
nan-yu, line 289: This test tests the `isDeploymentSame` function defined in `reconciler_base.go`. We should move it to `reconciler_base_test.go`.

pkg/reconcilermanager/controllers/jsonpath_test.go
```diff
@@ -0,0 +1,426 @@
...
+	}{
+		"non-Autopilot, no override limits, no manual change limits": {
+			declared:          declared,
+			unObjCurrent:      unObjCurrent,
+			isAutopilot:       false,
+			overrideLimit:     false,
+			manualChangeLimit: false,
+			isSame:            true,
+		},
+		"non-Autopilot, no override limits, manual change limits": {
```
nan-yu, line 308: For posterity, add a comment to explain why we allow manual limit change when limit is not specified.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -92,25 +92,37 @@ func TestManagingReconciler(t *testing.T) {
 			hasReplicas(managedReplicas))
 	})
 
-	// test case 3: override the memory limit has been moved to TestAutopilotReconcilerAdjustment
+	// test case 3: override the resource limits/requests and manually change the resource limits/requests has been moved to TestAutopilotReconcilerAdjustment
```
janetkuo, line 95: Do we still need to keep this comment here? The test is in a different function.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +245,95 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
+		"$.spec.template.spec.containers[*].*.failureThreshold",
+		"$.spec.template.spec.tolerations",
+		"$.spec.template.spec.restartPolicy",
+		"$.spec.template.spec.nodeSelector",
+		"$.spec.template.spec.terminationGracePeriodSeconds",
+		"$.spec.template.spec.dnsPolicy",
+		"$.spec.template.spec.schedulerName",
+		"$.spec.revisionHistoryLimit",
+		"$.spec.progressDeadlineSeconds",
+		"$.spec.paused",
```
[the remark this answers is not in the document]
xinnywinne: Yes, this means we allow users or other controllers to pause the reconciler, and since this field is not defined in the declared deployment, users and other controllers will be able to pause the reconciler. I add this field here is to make sure when that pause set to true by users / other controllers, reconciler manager does not keep patching the reconciler deployment. If we want to make sure others cannot pause reconciler, we should define this field in reconciler-manager-configmap.

e2e/testcases/reconciler_manager_test.go
xinnywinne: Do you suggest that we should combine the TestAutopilotReconcilerAdjustment test with TestOverrideReconcilerResourcesV1Alpha1? Currently we have TestOverrideReconcilerResourcesV1Alpha1 skip on Autopilot cluster. It looks like that override test is focusing on make sure the override function works and this  AutopilotReconcilerAdjustmentTest focusing on verify that adjustContainerResources works fine with resource override function. Combine these two tests will make a long complex test, we also need to manually check what's the resources result after autopilot adjustment for all the test cases in TestOverrideReconcilerResourcesV1Alpha1 to make a right prediction. So I think it is better to separate  these two tests.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +245,95 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
+	deploymentToPatch    []byte
+}
+
+// isDeploymentSame checks if current deployment is same with declared deployment when ignore the fields in allowlist.
+func isDeploymentSame(isAutopilot bool, declared *appsv1.Deployment, unObjCurrent *unstructured.Unstructured, scheme *runtime.Scheme) (*deploymentProcessResult, error) {
+	//The fields in allowlist are the fields we allow
+	// users or other controllers to modify.
+	allowList := []string{
+		"$.spec.template.spec.containers[*].terminationMessagePath",
+		"$.spec.template.spec.containers[*].imagePullPolicy",
```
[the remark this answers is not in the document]
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +245,95 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
+		"$.spec.template.spec.containers[*].terminationMessagePolicy",
+		"$.spec.template.spec.containers[*].resources.requests.ephemeral-storage",
+		"$.spec.template.spec.containers[*].*.timeoutSeconds",
+		"$.spec.template.spec.containers[*].*.periodSeconds",
+		"$.spec.template.spec.containers[*].*.successThreshold",
+		"$.spec.template.spec.containers[*].*.failureThreshold",
+		"$.spec.template.spec.tolerations",
+		"$.spec.template.spec.restartPolicy",
+		"$.spec.template.spec.nodeSelector",
```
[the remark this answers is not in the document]
xinnywinne: Added comment to remove nodeSelector after Config Sync support user-defined nodeSelector

pkg/reconcilermanager/controllers/reconciler_base.go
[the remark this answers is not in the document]
xinnywinne: Removed paused from allowList

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -329,13 +441,18 @@ func keepCurrentContainerResources(declared, current *appsv1.Deployment) bool {
...
+
+	if err != nil {
 		if apierrors.IsNotFound(err) {
 			return nil, errors.Errorf(
 				"Deployment %s not found in namespace: %s.", dRef.Name, dRef.Namespace)
 		}
 		return nil, errors.Wrapf(err, "error while retrieving deployment")
 	}
+	if err = runtime.DefaultUnstructuredConverter.FromUnstructured(depUnObj.Object, depObj); err != nil {
+		return nil, errors.Errorf("failed to convert unstructured object %s into Deployment object", depUnObj.GetName())
```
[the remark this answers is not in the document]
xinnywinne: Done.

e2e/testcases/reconciler_manager_test.go
xinnywinne: Done.

e2e/testcases/reconciler_manager_test.go
xinnywinne: Yes, updated.

pkg/reconcilermanager/controllers/jsonpath_test.go
xinnywinne: Done.

e2e/testcases/reconciler_manager_test.go
xinnywinne: Removed the comment.

pkg/reconcilermanager/controllers/jsonpath_test.go
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +245,95 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
+	isDeploymentSame     bool
+	isDeploymentAdjusted bool
+	deploymentToPatch    []byte
+}
+
+// isDeploymentSame checks if current deployment is same with declared deployment when ignore the fields in allowlist.
+func isDeploymentSame(isAutopilot bool, declared *appsv1.Deployment, unObjCurrent *unstructured.Unstructured, scheme *runtime.Scheme) (*deploymentProcessResult, error) {
+	//The fields in allowlist are the fields we allow
+	// users or other controllers to modify.
+	allowList := []string{
```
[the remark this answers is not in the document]
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -183,17 +188,23 @@ func (r *reconcilerBase) upsertDeployment(ctx context.Context, reconcilerRef typ
 func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *appsv1.Deployment) (controllerutil.OperationResult, error) {
 	dRef := client.ObjectKeyFromObject(declared)
 	kind := "Deployment"
-
-	current := &appsv1.Deployment{}
-
-	if err := r.client.Get(ctx, dRef, current); err != nil {
+	forcePatch := true
+	fieldManager := reconcilermanager.ManagerName
```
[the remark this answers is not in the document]
xinnywinne: Done

e2e/testcases/reconciler_manager_test.go
nan-yu: ACK

nan-yu (approved):
/lgtm

cmd/reconciler-manager/main.go
```diff
@@ -77,18 +76,22 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	dynamicClient, err := dynamic.NewForConfig(mgr.GetConfig())
```
karlkfi, line 79: I'm surprised we don't already have a dynamic client in the reconciler-manager...

I'm also wondering if we should be injecting the rest.Config, instead of letting the controller manager build it. I think this means we aren't using the same timeout and QPS settings used by the reconciler...

We might want to build the config with restconfig.NewRestConfig and pass it into the ctrl.Options, and then also use it to build the dynamic client (assuming the controller manager doesn't already build one we can use instead).

e2e/testcases/reconciler_manager_test.go
```diff
@@ -215,7 +228,56 @@ func hasGeneration(generation int64) nomostest.Predicate {
 		return nil
 	}
 }
-
+func firstContainerTerminationMessagePathIs(terminationMessagePath string) nomostest.Predicate {
```
karlkfi, line 231: How about renaming these predicates to end with -Equals/Equal? Like `containerTerminationMessagePathEquals` & `containerStdinEquals` (add a containerIndex int arg), and `tolerationsEqual`.

karlkfi: Maybe `hasGeneration` -> `hasGenerationEquals` too

pkg/kinds/resources.go
```diff
@@ -0,0 +1,25 @@
...
+
+package kinds
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// DeploymentResource returns the canonical Deployment GroupVersionResource.
+func DeploymentResource() schema.GroupVersionResource {
```
karlkfi, line 23: Is this a new file because it's a GVR and not a GVK?

pkg/reconcilermanager/controllers/jsonpath.go
```diff
@@ -0,0 +1,66 @@
...
```
karlkfi, line 27: Can you make a follow up to add a Delete function to the cli-utils jsonpath, which already has Get and Set funcs?

https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/jsonpath/jsonpath.go

It would be nice to keep them all together and just use that package here.

pkg/reconcilermanager/controllers/reconciler_base.go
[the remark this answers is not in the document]
karlkfi: Won't going yaml -> Deployment -> json cause a bunch of fields to be populated that are optional in the schema but don't have pointers? Is that a problem?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -183,17 +207,22 @@ func (r *reconcilerBase) upsertDeployment(ctx context.Context, reconcilerRef typ
 func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *appsv1.Deployment) (controllerutil.OperationResult, error) {
 	dRef := client.ObjectKeyFromObject(declared)
 	kind := "Deployment"
-
-	current := &appsv1.Deployment{}
-
-	if err := r.client.Get(ctx, dRef, current); err != nil {
+	forcePatch := true
+	deploymentResourceInterface := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace)
```
karlkfi, line 211: `deploymentResourceInterface` -> `deploymentClient` might make this easier to read.

It seems a little weird tho that we have both "deployment" hard-coded here and "deployments" hard-coded in the the kind package. This is what the RESTMapper is for. It maps from kind to resource based on the resources registered in the cluster. We probably shouldn't be hard-coding the resource name.

karlkfi: Usually when you inject a Dynamic Client, you also inject a Mapper to use with it.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +263,75 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
 		r.log.Error(err, "Managed object update failed, deleting and re-creating",
 			logFieldObject, dRef.String(),
 			logFieldKind, kind)
-		if err := r.client.Delete(ctx, declared); err != nil {
+		if err := deploymentResourceInterface.Delete(ctx, dRef.Name, metav1.DeleteOptions{}); err != nil {
```
karlkfi, line 266: Do we want to be using the Foreground PropagationPolicy here? That would ensure the ReplicaSets & Pods are deleted before we continue.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -183,17 +207,22 @@ func (r *reconcilerBase) upsertDeployment(ctx context.Context, reconcilerRef typ
 func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *appsv1.Deployment) (controllerutil.OperationResult, error) {
 	dRef := client.ObjectKeyFromObject(declared)
 	kind := "Deployment"
-
-	current := &appsv1.Deployment{}
-
-	if err := r.client.Get(ctx, dRef, current); err != nil {
+	forcePatch := true
```
karlkfi, line 210: What does this force patch actually do? Take over fields that are managed by another manager?

Should we instead just error if another owner has taken control of our fields?

karlkfi: @nan-yu do you have perspective on this?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +263,75 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
+		if err != nil {
+			return controllerutil.OperationResultNone, fmt.Errorf("failed to marshal declared deployment object to byte array: %w", err)
+		}
+		_, err = deploymentResourceInterface.Patch(ctx, dRef.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: reconcilermanager.ManagerName, Force: &forcePatch})
+		if err != nil {
 			return controllerutil.OperationResultNone, err
 		}
 	}
-
+	if deploymentAfterPatch.GetGeneration() == unObjCurrent.GetGeneration() {
```
karlkfi, line 278: Do we also need to be checking the UID? If the UID changed then the generation MIGHT change, but it also might not, if it was 1 or something low.

pkg/reconcilermanager/controllers/reconciler_base.go
xinnywinne: https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller 
It is strongly recommended for controllers to always "force" conflicts, since they might not be able to resolve or act on these conflicts.

xinnywinne: This means we make sure to own the fields we defined.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -206,27 +235,25 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
 		}
 		r.isAutopilotCluster = &isAutopilot
 	}
-
-	adjusted, err := adjustContainerResources(*r.isAutopilotCluster, declared, current)
+	dep, err := isDeploymentSame(*r.isAutopilotCluster, declared, unObjCurrent, reconcilerManagerAllowList, r.scheme)
```
karlkfi, line 238: Usually when you have a function named with the "Is-" prefix, it just returns a bool. Can this be named better to reflect what it's returning without me needing to go look at the impl?

pkg/reconcilermanager/controllers/jsonpath.go
xinnywinne: Yes, sounds good.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -206,27 +235,25 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
 	if err != nil {
 		return controllerutil.OperationResultNone, err
 	}
-	if adjusted {
+	if dep.isDeploymentAdjusted {
```
karlkfi, line 242: Given that the returned struct is always used for deployments, I don't think we need "deployment" in the field names. Also, usually in Go you forgo the "is-" prefix for boolean fields and only use them on methods/funcs that return a bool. So you could just make this `adjusted`.

That said, I don't really understand what "adjusted" means here without looking at the impl of `isDeploymentSame`. Does this mean it's the same or different?

Does `isDeploymentSame` really need to return a struct or can it just return the json patch data and a bool as two different return values?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +263,75 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
+	}
+	var resultDeployment appsv1.Deployment
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredDeployment.Object, &resultDeployment); err != nil {
+		return nil, fmt.Errorf("failed to convert from current reconciler unstructured object to deployment object: %w", err)
+	}
+	return &resultDeployment, nil
+}
+
+type deploymentProcessResult struct {
+	isDeploymentSame     bool
```
karlkfi, line 299: Are these two booleans just inverse of each other? What's the difference?

e2e/testcases/reconciler_manager_test.go
xinnywinne: Will adress this in follow up PR

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -236,17 +263,75 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
 	}
-
+	if deploymentAfterPatch.GetGeneration() == unObjCurrent.GetGeneration() {
+		return controllerutil.OperationResultNone, nil
+	}
 	return controllerutil.OperationResultUpdated, nil
 }
 
+// preprocessDeployment delete all the fields in allowlist from unstructured object and convert the unstructured object to Deployment object
+func preprocessDeployment(allowList []string, unstructuredDeployment *unstructured.Unstructured) (*appsv1.Deployment, error) {
```
karlkfi, line 285: Using "process" and "pre-process" in function names is really generic. All functions process things... Can they be named more clearly to say what they do? If it's not clear what the name should be, maybe they could be refactored to improve readability by having functions each do "one thing".

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -329,13 +436,18 @@ func keepCurrentContainerResources(declared, current *appsv1.Deployment) bool {
 // deployment returns the deployment from the server
 func (r *reconcilerBase) deployment(ctx context.Context, dRef client.ObjectKey) (*appsv1.Deployment, error) {
 	depObj := &appsv1.Deployment{}
-	if err := r.client.Get(ctx, dRef, depObj); err != nil {
+	depUnObj, err := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace).Get(ctx, dRef.Name, metav1.GetOptions{})
```
karlkfi, line 439: Don't we already build this deploymentResourceInterface? Can it be passed in? If not, like elsewhere, the resource should be retrieved from the rest mapper, not hard-coded.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -329,13 +436,18 @@ func keepCurrentContainerResources(declared, current *appsv1.Deployment) bool {
...
+
+	if err != nil {
 		if apierrors.IsNotFound(err) {
 			return nil, errors.Errorf(
 				"Deployment %s not found in namespace: %s.", dRef.Name, dRef.Namespace)
 		}
 		return nil, errors.Wrapf(err, "error while retrieving deployment")
 	}
+	if err = runtime.DefaultUnstructuredConverter.FromUnstructured(depUnObj.Object, depObj); err != nil {
```
karlkfi, line 448: Can `kinds.ToTypedObject` be used instead? I've been trying to route conversion through those helper utils, rather than using the converter directly.

pkg/reconcilermanager/controllers/reconciler_base.go
xinnywinne: I see, at the beginning I use the RESTMapper and then change it to be like this. I am wondering if I also need to make a fake RESTMapper for our unit test , do you have some suggestions?

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -261,19 +264,31 @@ func rolebinding(name, reconcilerName string, opts ...core.MetaMutator) *rbacv1.
 	return result
 }
 
-func setupNSReconciler(t *testing.T, objs ...client.Object) (*syncerFake.Client, *RepoSyncReconciler) {
+type repoReconcilerTest struct {
```
karlkfi, line 267: Until I found this impl, I wad _realy_ confused at why you'd replaced the client with the reconciler variable...

This makes it a lot more confusing to know what methods are being called on the client or the dynamic client or the reconciler itself...

Why not just have setupNSReconciler return 3 variables? 

This isn't java. Go supports multiple return values. You don't need to wrap everything in a class to return it.

xinnywinne: The reason I decided to wrap these three things is because the 3 variables are not always needed for all the times we call this function, sometime we only need 1 or 2,  the code changes to be like _,_,XX = setupNSReconciler() or XX,_XX = ..., makes me feel it is better to wrap the 3 variables.  And we use these 3 variables for our reconciler test, so I think it makes sense to put them together.

pkg/kinds/resources.go
xinnywinne: Yes, after we change to use the RESTMapper, this file can be removed

pkg/reconcilermanager/controllers/reconciler_base.go
xinnywinne: Sure. When I considered about how to name it, I think the main part for this function is to decide if we have "same" declared deployment with current deployment. If not we return a data array for PATCH. So I decided to use "isDeploymentSame".  I also do not want the function name to be really long, so do you have some better suggestions?

pkg/reconcilermanager/controllers/reconciler_base.go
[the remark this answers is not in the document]
xinnywinne: It's not a problem as I seen, I also added test to test the fields that we do not define and not a pointer.

pkg/reconcilermanager/controllers/reconciler_base.go
xinnywinne: isDeploymentAdjusted is using to detect if Autopilot make any adjustment to the resources requirement we defined.  Resources adjusted by Autopilot doesn't mean the declared and current deployments are not same. We can tell that if Autopilot did adjustment based on what we declared, if yes, we still consider the current deployment is same with declared deployment and no more PATCH needed. So the isDeployment same is used to decide if a PATCH is needed, the isDeploymentAdjusted is only used to notice the user that the resource defined by us or by them was adjusted by Autopilot.

PR #270, opened 2022-11-09
remove bats scaffolding for git-server

sdowell:
This scaffolding was used by the bats tests to start the in-cluster git server. The go tests use the scaffolding implemented in git-server.go


nan-yu (approved):
/lgtm

PR #272, opened 2022-11-10
use build-manifests for e2e testing

sdowell:
The e2e tests were using their own duplicated implementation of manifest generation. This duplication increases the potential for drift between the manifests used in testing and the real build artifacts.

This change updates the e2e tests to use the manifests generated by build-manifests, which should ensure closer parity to real build artifacts. This also should simplify:
- maintenance of the manifests
- the interface for running the tests


.dockerignore
```diff
@@ -1,2 +1 @@
 .idea/
-.output/
```
karlkfi, line 2: why remove output? a lot of stuff goes in here, doesn't it?

sdowell: I removed it so that the docker context used for the [GKE](https://github.com/karlkfi/kpt-config-sync/blob/main/build/test-e2e-go/gke/Dockerfile) and [kind](https://github.com/karlkfi/kpt-config-sync/blob/main/build/test-e2e-go/kind/Dockerfile) e2e images could access the built config sync manifests when creating the image. This is only really an issue when running things inside these e2e containers (e.g. in CI).

I'm also fine with mounting the manifests directory to the container at runtime, although this seemed like a simpler approach to me.

sdowell:
pulled the cluster role changes out to https://github.com/GoogleContainerTools/kpt-config-sync/pull/277

docs/development.md
```diff
@@ -37,8 +37,6 @@ These can be provided on the command line with `go test` or with program argumen
 
 - `--e2e` - If true, run end-to-end tests. (required to run the e2e tests)
 - `--debug` - If true, do not destroy cluster and clean up temporary directory after test.
-- `--image-prefix` - The prefix to use for Docker images. Defaults to the local Docker registry. Omit the trailing slash.
```
nan-yu, line 40: One concern about removing `--image-prefix` and `--image-tag` is that it is not that flexible to test with different images.

Previously, it was pretty easy to test with a particular image by setting  `--image-tag`. If we want to test with a different image, we just update the flag. Now, with this approach, we have to re-generate the .output directory. If we want to go back and forth between different images, we have to version the .output directory and rename it. Is there a better way to do so?

Makefile.e2e
```diff
@@ -85,5 +72,5 @@ test-e2e-kind-multi-repo-test-group3:
 # Runs the Go e2e tests against multi-repo Config Sync.
 # Rebuilds the nomos image, so use this when you're actively modifying the Nomos
 # code between test runs.
-test-e2e-go-multirepo: __push-local-image __install-nomos-local
-	@GO111MODULE=on go test ./e2e/... --e2e --timeout 60m --image-tag=$(IMAGE_TAG)
+test-e2e-go-multirepo: config-sync-manifest-local __install-nomos-local
```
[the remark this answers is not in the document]
sdowell: Agree it looks like this can be consolidated now. I'll update this

docs/development.md
sdowell: My intent for this use case is that the developer just runs `make build-manifests IMAGE_TAG=<TAG>` if they want to switch between manifests for test runs. This is fairly light weight and runs quickly for swapping back and forth.

Alternatively we could pass the flags into the tests and then the tests shell back to `make build-manifests` to construct the manifests, although I think I prefer the simplicity of just running `go test --e2e` and it will use whichever manifests are on the filesystem

docs/development.md
```diff
@@ -96,12 +94,10 @@ current context in your kubeconfig.
 gcloud config set project <PROJECT_ID>
 # Ensure kubectl context is set to correct cluster
 kubectl config set-context <CONTEXT>
-# Build images
-make build-images
-# Push images to GCR
-make push-images
+# Build images/manifests and push images
+make config-sync-manifest
 # Run the tests with image prefix/tag from previous step and desired test regex
```
nan-yu, line 99: Maybe add instructions here about how to test with a different image tag?

docs/development.md
nan-yu: Using `make build-manifests IMAGE_TAG=<TAG>` sounds good. Make sure we document the process for posterity.

sdowell: Added some documentation, PTAL

docs/development.md
sdowell: Done, PTAL

nan-yu (approved):
/lgtm

PR #273, opened 2022-11-10
Cleanup time duration options

karlkfi:
- Use the -Period suffix consistently. Technically, most of these should be "Delay", because they're a timer between attempts and not the duration of the event itself, but we're already using "period" for most of them. So this at least makes them all consistent without requiring rename of any external facing env vars or command flags.
- Standardize the many duplicate descriptions for duration options, at each layer of the stack.
- Add missing constants to replace hard-coded durations.
- Reafactor existing duration constants to make the names consistent.

### DIFF
--- a/cmd/hydration-controller/main.go
+++ b/cmd/hydration-controller/main.go
@@ -19,16 +19,17 @@ import (
 	"flag"
 	"os"
 	"strings"
-	"time"
 
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
+	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/hydrate"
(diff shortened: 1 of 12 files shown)


karlkfi:
The `TestInlineClusterSelectorFormat` failure is similar.

The validation that is failing is a `Validate` right after a `WaitForRepoSyncs` following a non-commit change. In this case, it was the cluster name that's been changed, which is stored in the `config-management-system/reconciler-manager` ConfigMap as the `CLUSTER_NAME` data field.

Even tho the reconciler-manage Pod is replaced, and the root-reconciler is updated with the new `CLUSTER_NAME` env var, there's no way to communicate that input data in the RootSync status. So the `WaitForRepoSyncs` call returns too early.

karlkfi:
Extracted some test flake "fixes" this change uncovered:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/276
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/275

nan-yu (approved):
/lgtm

PR #274, opened 2022-11-10
Fix flakey TestInvalidR*SyncBranchStatus

karlkfi:
- The test is waiting for the reconciler_errors metric to be recorded for both source and sync status, but if there is a source error, the sync status won't be updated and the metric won't be sent for the sync component.
- The only reason this works today is that the async calls to setSyncStatus aren't being deduped. Once they're deduped, this flakiness would become consistent failure. So this fix helps unblock  additional synchronization fixes.

### DIFF
--- a/e2e/nomostest/metrics/validate.go
+++ b/e2e/nomostest/metrics/validate.go
@@ -15,6 +15,7 @@
 package metrics
 
 import (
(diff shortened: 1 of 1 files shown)


nan-yu (approved):
/lgtm

PR #275, opened 2022-11-10
Disable Syncing condition validation on error

karlkfi:
When waiting for an error in the RSync status, ignore the Syncing condition, for now.

This is a temporary hack to make tests less flakey until the Syncing condition is fixed to reflect all errors and only the latest commit.

This problem should be fixed by https://github.com/GoogleContainerTools/kpt-config-sync/pull/260, which will be updated to re-add these disabled checks.


nan-yu (approved):
/lgtm

PR #276, opened 2022-11-10
Wait for non-commit changes to sync

karlkfi:
This patches multiple tests that wait for syncing after a non-commit change. They all now wait for their desired state, instead of assuming it already exists. This compensates for an API gap that doesn't track the generation of the RSync that was synced nor any changes to the reconciler config (namely cluster name). To fix this correctly, we will need to modify the API. So each case has a TODO and comment.

### DIFF
--- a/e2e/nomostest/wait.go
+++ b/e2e/nomostest/wait.go
@@ -75,6 +75,32 @@ func Wait(t testing.NTB, opName string, timeout time.Duration, condition func() 
 	t.Logf("took %v to wait for %s", took, opName)
 }
 
+// WaitForObject waits for the specified object. If predicates are specified,
+// they are waited for, otherwise existance is enough.
+// Predicates must accept the object as Unstructured.
+func WaitForObject(nt *NT, gvk schema.GroupVersionKind, name, namespace string, waitOptionsAndPredicates ...interface{}) {
+	nt.T.Helper()
+
+	var waitOpts []WaitOption
+	var predicates []Predicate
+	for _, opt := range waitOptionsAndPredicates {
+		switch tOpt := opt.(type) {
+		case WaitOption:
+			waitOpts = append(waitOpts, tOpt)
+		case Predicate:
+			predicates = append(predicates, tOpt)
+		default:
+			nt.T.Fatalf("invalid option passed to WaitForObject: expected WaitOption or Predicate, but received a %T", opt)
+		}
+	}
+
+	Wait(nt.T, fmt.Sprintf("wait for %q %v to meet expectations", name, gvk), nt.DefaultWaitTimeout, func() error {
+		u := &unstructured.Unstructured{}
+		u.SetGroupVersionKind(gvk)
+		return nt.Validate(name, namespace, u, predicates...)
+	}, waitOpts...)
+}
+
 // WaitForNotFound waits for the passed object to be fully deleted.
 // Immediately fails the test if the object is not deleted within the timeout.
 func WaitForNotFound(nt *NT, gvk schema.GroupVersionKind, name, namespace string, opts ...WaitOption) {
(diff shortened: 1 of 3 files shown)


e2e/nomostest/wait.go
```diff
@@ -75,6 +75,18 @@ func Wait(t testing.NTB, opName string, timeout time.Duration, condition func()
 	t.Logf("took %v to wait for %s", took, opName)
 }
 
+// WaitFor waits for the specified object to meet the specified predicates.
+// Predicates must accept the object as Unstructured.
+func WaitFor(nt *NT, gvk schema.GroupVersionKind, name, namespace string, predicates []Predicate, opts ...WaitOption) {
```
sdowell, line 80: Maybe something a bit more explicit like `WaitForObject` or `WaitForValidate` - WDYT?

e2e/testcases/cluster_selectors_test.go
```diff
@@ -170,30 +189,42 @@ func TestClusterSelectorOnObjects(t *testing.T) {
...
-	}
+	// TODO: Confirm that the change was Synced.
+	// This is not currently possible using the RootSync status API, because
+	// the commit didn't change, and the commit was already previously Synced.
+	// If sync state could be confirmed, the objects would already be updated,
+	// and we wouldn't need to wait for it.
+	// if err := nt.Validate(rb.Name, rb.Namespace, &rbacv1.RoleBinding{}); err != nil {
+	// 	nt.T.Fatal(err)
+	// }
+	nomostest.WaitForCurrentStatus(nt, kinds.RoleBinding(), rb.Name, rb.Namespace)
```
sdowell, line 227: What does `WaitForCurrentStatus` do to help in this context?

karlkfi: Wait for existence + validation of current status (reconciled and not terminating). There's a lot of things we could check for, like there's a Predicate for NotTerminating, but Current status is the generic "yes it exists" function that already exists in a Wait- form.

karlkfi: In this particular case, RoleBindings don't have status, so the current status means less, but the user shouldn't need to know that. "Current" reconciler status is what the applier waits for. So we can expect it to be true if it synced without errors.

e2e/nomostest/wait.go
karlkfi: It waits for the predicates passed in, and predicates are generic conditions. The only reason this is different from Wait() is that it does the object GET for you and passes it to the function so that the functions don't need to do the GET. 

But I can call it WaitForPredicates if you prefer.

karlkfi: Or WaitForObject?

karlkfi: WaitForObject would be great if the predicates were optional. But I dont know how to make them optional and also have optional wait options...

e2e/testcases/cluster_selectors_test.go
sdowell: Ah so is the key difference between this and the previous usage of `nt.Validate` that `nt.Validate` doesn't retry/wait?

e2e/nomostest/wait.go
sdowell: You could potentially nest the function, e.g.
```golang
func WaitForObject(nt *NT, gvk schema.GroupVersionKind, name, namespace string, predicates ...Predicate) func(opts ...WaitOption) {
    return func(opts ...WaitOption) {
    ...
    }
}
```
Although maybe that's overkill. I'm also good with `WaitForPredicates`

karlkfi: This is one of those times I wish a varargs could accept either one type or another type, but not any type. For now, I made it take any, which makes it a runtime problem, but avoids the problem of needing two optional arrays.

PR #277, opened 2022-11-10
use NS reconciler cluster role in e2e tests

sdowell:
The e2e tests were creating their own version of the ns reconciler cluster role with namespace admin permissions.

This updates the test scaffolding to more closely reflect real scenarios where the CS installation comes with a ClusterRole for the NS reconcilers and the user adds additional roles/bindings.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -64,6 +64,9 @@ const (
 
 	// e2e/raw-nomos/manifests/multi-repo-configmaps.yaml
 	multiConfigMapsName = "multi-repo-configmaps.yaml"
+
+	// clusterRoleName imitates a user-created (Cluster)Role for NS reconcilers.
+	clusterRoleName = "c2-e2e"
 )
 
 var (
@@ -78,9 +81,6 @@ var (
 
 	multiConfigMaps = filepath.Join(baseDir, "e2e", "raw-nomos", Manifests, multiConfigMapsName)
 
-	// clusterRoleName is the ClusterRole used by Namespace Reconciler.
-	clusterRoleName = fmt.Sprintf("%s:%s", configsync.GroupName, core.NsReconcilerPrefix)
-
 	templates = []string{
 		"admission-webhook.yaml",
 		"otel-collector.yaml",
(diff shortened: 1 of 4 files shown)


e2e/nomostest/config_sync.go
```diff
@@ -546,6 +547,7 @@ func waitForReconciler(nt *NT, name string) error {
 // RepoSyncClusterRole returns clusterrole with permissions to manage resources in the cluster.
 func RepoSyncClusterRole() *rbacv1.ClusterRole {
 	cr := fake.ClusterRoleObject(core.Name(clusterRoleName))
+	// TODO: make this more granular by test options
```
karlkfi, line 550: Are you planning to follow up on this yourself?
Can you make a bug to track it?

e2e/nomostest/config_sync.go
```diff
@@ -78,9 +81,6 @@ var (
 
 	multiConfigMaps = filepath.Join(baseDir, "e2e", "raw-nomos", Manifests, multiConfigMapsName)
 
-	// clusterRoleName is the ClusterRole used by Namespace Reconciler.
-	clusterRoleName = fmt.Sprintf("%s:%s", configsync.GroupName, core.NsReconcilerPrefix)
```
karlkfi, line 82: Will having multiple tests create & delete this role cause problems in a shared test env? I think that's why they had different name, right?

sdowell: The scaffolding already handles the lifecycle of this cluster role between tests (both for shared/ephemeral test environments). I'm basically just changing the name here so instead of creating the ns reconciler cluster role shipped with Config Sync we instead create one with a new name. Note this was essentially already a constant before (`configsync.gke.io:ns-reconciler`). I haven't tested it with a shared environment but I expect it should work

e2e/nomostest/config_sync.go
sdowell: I'm hoping I don't necessarily need to be the person to fix this, but I'll make an issue to track it

e2e/nomostest/config_sync.go
karlkfi: Nevermind. SharedTestEnv runs in serial, so this isn't a problem.

e2e/nomostest/config_sync.go
karlkfi: "put it on the backlog"

karlkfi (approved):
/lgtm

PR #278, opened 2022-11-10
remove prober-cred scaffolding

sdowell:
The prober-cred scaffolding has been replaced by workload identity in prow. For local dev usage users should just use their gcloud credentials.


nan-yu (approved):
/lgtm

PR #279, opened 2022-11-10
Move periodic status updates into the parser

karlkfi:
- This avoids needing to synchronize two goroutines. Instead, the
  periodic status updater will only run between parse.run calls.
- Disable sync status updates when a newer commit is being processed.

### DIFF
--- a/pkg/parse/run.go
+++ b/pkg/parse/run.go
@@ -67,6 +67,9 @@ func Run(ctx context.Context, p Parser) {
 	retryTimer := time.NewTimer(opts.retryPeriod)
 	defer retryTimer.Stop()
 
+	statusUpdateTimer := time.NewTimer(opts.statusUpdatePeriod)
+	defer statusUpdateTimer.Stop()
+
 	state := &reconcilerState{}
 	for {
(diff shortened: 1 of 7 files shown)


pkg/parse/run.go
```diff
@@ -111,17 +118,33 @@ func Run(ctx context.Context, p Parser) {
...
 			}
 			run(ctx, p, trigger, state)
-			retryTimer.Reset(opts.retryPeriod)
+
+			retryTimer.Reset(opts.retryPeriod)               // Schedule retry attempt
+			statusUpdateTimer.Reset(opts.statusUpdatePeriod) // Schedule status update attempt
+
+		// Update the sync status to report management conflicts (from the remediator).
+		case <-statusUpdateTimer.C:
+			// Skip sync status update if the .status.sync.commit is out of date.
```
nan-yu, line 127: I like the idea of using another timer to make sure the status update and parse-apply-watch loop are mutually exclusive.

nan-yu (approved):
/lgtm

Thanks for making the change in such a neat and simple way!

PR #280, opened 2022-11-11
Fix Otel Collector resource mapping

tiffanny29631:
Config Sync updated Otel Collector image in v1.12.2 and started having
[incompatibilities](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/breaking-changes.md) in resource label format. The issue was mitigated in 1.13.1 by reverting Otel Collector to legacy mode.

This change

- uses `resourcedetection` processor to correclty map resource labels from OpenCensus to OpenTelemetry
- modified the `googlecloud` exporter to remove `instrumentation_source` and `insrtumentation_version` labels when exporting to Cloud Monarch
- disables metricDescriptor creation in Monarch pipeline so it won't conflict with internal definitions

### DIFF
--- a/pkg/metrics/otel.go
+++ b/pkg/metrics/otel.go
@@ -54,13 +54,32 @@ exporters:
   googlecloud/kubernetes:
     metric:
       prefix: "kubernetes.io/internal/addons/config_sync/"
-      skip_create_descriptor: false
+      # skip_create_descriptor: Metrics start with 'kubernetes.io/' have already
+      # got descriptors defined internally. Skip sending dupeicated metric 
+      # descriptors here to prevent errors or conflicts.
+      skip_create_descriptor: true
+      # instrumentation_library_labels: Otel Collector by default attaches 
+      # 'instrumentation_version' and 'instrumentation_source' labels that are
+      # not specified in our Cloud Monarch definitions, thus skipping them here
+      instrumentation_library_labels: false
+      # create_service_timeseries: This is a recommended configuration for 
+      # 'service metrics' starts with 'kubernetes.io/' prefix. It uses 
+      # CreateTimeSeries API and has its own quotas, so that custom metric write
+      # will not break this ingestion pipeline
+      create_service_timeseries: true
+      service_resource_labels: false
     retry_on_failure:
       enabled: false
     sending_queue:
       enabled: false
 processors:
   batch:
+  # resourcedetection: This processor is needed to correctly mirror resource
+  # labels from OpenCensus to OpenTelemetry. We also want to keep this same
+  # processor in Otel Agent configuration as the resource labels are added from
+  # there
+  resourcedetection:
+    detectors: [env, gcp]
   filter/cloudmonitoring:
     metrics:
       include:
@@ -106,6 +125,11 @@ processors:
         match_type: strict
         metric_names:
           - rg_reconcile_duration_seconds
+          # TODO remove kcc_resource_count_total rule once Resource Group Controller
+          # 1.0.9 is updated into Config Sync. This metric was unintentionally
+          # included by the 'regex include' filter above and is not included in
+          # our Monarch metric definitions
+          - kcc_resource_count_total
   metricstransform/kubernetes:
     transforms:
       - include: declared_resources
@@ -151,7 +175,7 @@ service:
   pipelines:
     metrics/cloudmonitoring:
       receivers: [opencensus]
-      processors: [batch, filter/cloudmonitoring]
+      processors: [batch, filter/cloudmonitoring, resourcedetection]
       exporters: [googlecloud]
     metrics/prometheus:
       receivers: [opencensus]
@@ -159,6 +183,6 @@ service:
       exporters: [prometheus]
     metrics/kubernetes:
       receivers: [opencensus]
-      processors: [batch, filter/kubernetes, metricstransform/kubernetes]
+      processors: [batch, filter/kubernetes, metricstransform/kubernetes, resourcedetection]
       exporters: [googlecloud/kubernetes]`
 )
--- a/manifests/templates/otel-collector.yaml
+++ b/manifests/templates/otel-collector.yaml
@@ -135,9 +135,6 @@ spec:
         - /otelcol-contrib
         args:
         - "--config=/conf/otel-collector-config.yaml"
-        # TODO: Remove this feature gate when opentelemetry semantic conventions are used
-        # in the collector code.
-        - "--feature-gates=-exporter.googlecloud.OTLPDirect"
         resources:
           limits:
             cpu: 1
--- a/pkg/reconcilermanager/controllers/otel_controller_test.go
+++ b/pkg/reconcilermanager/controllers/otel_controller_test.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	depAnnotationGooglecloud = "6ba29d7cac4cdd7b23c0e1a6f355d0f5"
+	depAnnotationGooglecloud = "0e2c217290c9e809236c09007e961fe2"
 	depAnnotationCustom      = "9182661d55e260a55da649363c03c187"
 )
 


pkg/metrics/otel.go
```diff
@@ -151,14 +157,14 @@ service:
   pipelines:
     metrics/cloudmonitoring:
       receivers: [opencensus]
-      processors: [batch, filter/cloudmonitoring]
+      processors: [filter/cloudmonitoring, resourcedetection, batch]
```
haiyanmeng, line 160: Is it necessary for `batch` be the last processor? If yes, can you add a comment?

pkg/metrics/otel.go
```diff
@@ -106,6 +111,7 @@ processors:
         match_type: strict
         metric_names:
           - rg_reconcile_duration_seconds
+          - kcc_resource_count_total
```
haiyanmeng, line 114: Do you mean `kcc_resource_group_total`?

pkg/metrics/otel.go
```diff
@@ -54,13 +54,18 @@ exporters:
...
+      create_service_timeseries: true
+      service_resource_labels: false
     retry_on_failure:
       enabled: false
     sending_queue:
       enabled: false
 processors:
   batch:
+  resourcedetection:
+    detectors: [env, gcp]
```
haiyanmeng, line 68: Do we still need the `resourcedetection` processor in https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/otel-agent-cm.yaml#L38-L39?

https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/templates/otel-collector.yaml#L26-L45 does not have the `resourcedetection` processor.

Is it necessary to use the `resourcedetection` processor in `otel-collector` instead of `otel-agent`?

pkg/metrics/otel.go
```diff
@@ -54,13 +54,18 @@ exporters:
   googlecloud/kubernetes:
     metric:
       prefix: "kubernetes.io/internal/addons/config_sync/"
-      skip_create_descriptor: false
+      skip_create_descriptor: true
```
haiyanmeng, line 57: Instead of setting `skip_create_descriptor` to `true`, is it possible to make sure the metric descriptors sent from otel-collector and the internal metric descriptors are consistent with each other?

pkg/metrics/otel.go
```diff
@@ -54,13 +54,18 @@ exporters:
...
+      instrumentation_library_labels: false
```
haiyanmeng, line 58: Why do we need to set this field to false?
I don't think it increases cardinality.

pkg/metrics/otel.go
```diff
@@ -54,13 +54,18 @@ exporters:
...
+      create_service_timeseries: true
```
haiyanmeng, line 59: Why do we need to set this field to true?

pkg/metrics/otel.go
```diff
@@ -54,13 +54,18 @@ exporters:
...
+      service_resource_labels: false
```
haiyanmeng, line 60: I would recommend adding he CS-specific [resource attributes](https://github.com/GoogleContainerTools/kpt-config-sync/blob/208f34fde1d5cc3b05872319cdef89ff43a1dcb3/manifests/templates/reconciler-manager-configmap.yaml#L283-L292) into the [metric.resource_filters](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter) field, and test the new metrics in Prometheus and Cloud Monitoring to make sure things are working as expected.

pkg/metrics/otel.go
tiffanny29631: Confirmed that the Monarch metrics have already got descriptors defined (the internal codebase we discussed yesterday). If we are sending the descriptor again there will be conflicts and user will keep seeing permission errors of not being able to send metric descriptor through the `googlecloud/kubernetes` exporter configuration.

pkg/metrics/otel.go
tiffanny29631: Example error `Unable to send metric descriptor.       {"kind": "exporter", "data_type": "metrics", "name": "googlecloud", "error": "rpc error: code = InvalidArgument desc = Request was missing field metricDescriptor.valueType: The descriptor does not have the value type set.", "metric_descriptor": "name:\"internal_errors_total\"  type:\"custom.googleapis.com/opencensus/config_sync/internal_errors_total\"  labels:{key:\"instrumentation_source\"}  labels:{key:\"instrumentation_version\"}  metric_kind:CUMULATIVE  unit:\"1\"  description:\"The total number of internal errors triggered by Config Sync\"  display_name:\"opencensus/config_sync/internal_errors_total\""}`. By default the exporter will try to attach these two labels: `instrumentation_source` and `instrumentation_version`, using this flag to disable the labels.

pkg/metrics/otel.go
tiffanny29631: Confirmed that this is a recommended configuration for 'system metrics' that starts with `kubernetes.io/`. It uses CreateTimeSeries API and has its own quota so the writes to custom API from user won't break this ingestion pipeline. 

From the [exporter doc](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter#configuration-reference) it also auto skips creating metric descriptor, but it's still worthy to list `skip_create_descriptor` out to reduce confusion.

pkg/metrics/otel.go
tiffanny29631: I'm thinking maintaining a list of CS resource label might be similar to excluding the exporter's special labels?

Testing ...

pkg/metrics/otel.go
tiffanny29631: From the test I've been doing so far this `resourcedetection` in Otel Collector (the shared deployment) is necessary to correctly mirror the resource labels from OpenCensus to OpenTelemetry style. Haven't tried modifying the Otel Agent, but since the resource labels were originally added there, the `resourcedetection` should be kept.

https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/templates/otel-collector.yaml#L26-L45 This one is the default configuration that does not connect with Google Cloud APIs. When this configuration is in use, it usually mean user is not on a GKE cluster / has no access to Google Cloud API. In this case the `otel-collector` deployment only exports to the Prometheus endpoint, and with `resource_to_telemetry_conversion` it could carry CS resource labels as is.

pkg/metrics/otel.go
tiffanny29631: This one is accidentally included by the above `regex include` filer with the name `kcc_resource_count.*`. Excluding this as `kcc_resource_count_total` is not defined in `kubernetes.io/` metrics definition. 

This metric has been removed from Resource Group Controller, we can erase this rule once the RGC is updated in Config Sync.

pkg/metrics/otel.go
tiffanny29631: Just retried and no it is not, moving it back.

pkg/metrics/otel.go
haiyanmeng: I see. Please add this as a comment into this file.

pkg/metrics/otel.go
haiyanmeng: Add a comment to explain this in the file.

pkg/metrics/otel.go
haiyanmeng: Add a comment to explain this in the file.

pkg/metrics/otel.go
haiyanmeng: > From the test I've been doing so far this resourcedetection in Otel Collector (the shared deployment) is necessary to correctly mirror the resource labels from OpenCensus to OpenTelemetry style. Haven't tried modifying the Otel Agent, but since the resource labels were originally added there, the resourcedetection should be kept.

Can you add this a comment in the file?

pkg/metrics/otel.go
haiyanmeng: Sounds good. 
Can you add a comment regarding this TODO item?

pkg/metrics/otel.go
tiffanny29631: Done!

pkg/metrics/otel.go
tiffanny29631: Done!

pkg/metrics/otel.go
tiffanny29631: Done!

pkg/metrics/otel.go
tiffanny29631: Done!

pkg/metrics/otel.go
tiffanny29631: Done!

pkg/metrics/otel.go
tiffanny29631: The resource_filter is not working as expected on my end at this point. I would like to split this work to a different PR - just not sure if we still have enough time. 

Also looking at the spec in their doc:  If provided, `resource attributes` matching any filter will be included in `metric labels`, seems like there is a conversion from resource label to a custom label

PR #281, opened 2022-11-11
Simplify status reporting of sync errors

karlkfi:
- Add a new setSyncStatus that handles:
  - Updating the cached sync status in `state`
  - Checking the cached sync status to avoid unnecessary updates
  - Calling SetSyncStatus with the new sync errors
  - Calling reportRootSyncConflicts to update remote RootSyncs, if
    they are causing conflicts with the current RSync.
- Rename UpdateConflictManagerStatus -> reportRootSyncConflicts
- Rename parse.setXStatus -> setXStatusFields to allow the name
  setSyncStatus to be used by the wrapper that handles state
- Call the new parse.setSyncStatus to make all the updates
  consistent, instead of calling parser.SetSyncStatus directly.
- Rename updateSyncStatus -> updateSyncStatusPeriodically to
  indicate it handles retrying until cancelled.
- Rename Parser.ApplierErrors() -> SyncErrors() and have it
  delegate to the new updater.Errors(), which centralizes
  aggregation of the following errors that are all reported as
  sync errors:
  - Remediator conflict errors
  - Validation errors from updating the object set
  - Applier errors
  - Watch update errors
- Remove Parser.RemediatorConflictErrors() because it's obsolete
  with the new SyncErrors. The conflict errors are a subset of the
  sync errors and are extracted as-needed in setSyncErrors.
  This extraction probably won't be necessary in the future, as we
  plan to stop reporting errors across RSyncs.
- Enhance TestConflictingDefinitions_RootToRoot to handle catching
  conflict errors in parallel, for speed and consistently.
- Fix linter copylock error

### DIFF
--- a/e2e/testcases/multi_sync_test.go
+++ b/e2e/testcases/multi_sync_test.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	"fmt"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
@@ -415,10 +416,33 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
 	nt.RootRepos[rootSync2].Add(podRoleFilePath, rootPodRole())
 	nt.RootRepos[rootSync2].CommitAndPush("add conflicting pod owner role")
 
-	nt.T.Logf("The admission webhook should deny the update request in Root %s", rootSync2)
-	nt.WaitForRootSyncSyncError(rootSync2, applier.ApplierErrorCode, "denied the request")
-	nt.T.Logf("Root %s should also get surfaced with the conflict error", configsync.RootSyncName)
-	nt.WaitForRootSyncSyncError(configsync.RootSyncName, status.ManagementConflictErrorCode, "declared in another repository")
+	// When the webhook is enabled, it will block adoption of managed objects.
+	nt.T.Logf("Both RootSyncs should still report conflicts with the webhook enabled")
+	waitInParallel(
+		// Reconciler conflict, detected by the second reconciler & reported to the first reconciler's RootSync
+		func() {
+			nomostest.WatchForObject(nt, kinds.RootSyncV1Beta1(), configsync.RootSyncName, configsync.ControllerNamespace,
+				[]nomostest.Predicate{
+					nomostest.RootSyncHasSyncError(nt, status.ManagementConflictErrorCode, "declared in another repository"),
+				})
+		},
+		// Reconciler conflict, detected by the second reconciler
+		func() {
+			nomostest.WatchForObject(nt, kinds.RootSyncV1Beta1(), rootSync2, configsync.ControllerNamespace,
+				[]nomostest.Predicate{
+					nomostest.RootSyncHasSyncError(nt, status.ManagementConflictErrorCode, "declared in another repository"),
+				})
+		},
+		// Webhook rejection detected by the second reconciler's applier
+		// https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/errors/statuserror.go#L29
+		func() {
+			nomostest.WatchForObject(nt, kinds.RootSyncV1Beta1(), rootSync2, configsync.ControllerNamespace,
+				[]nomostest.Predicate{
+					nomostest.RootSyncHasSyncError(nt, applier.ApplierErrorCode, "denied the request"),
+				})
+		},
+	)
+
 	nt.T.Logf("The Role resource version should not be changed")
 	if err := nt.Validate("pods", testNs, &rbacv1.Role{},
 		nomostest.ResourceVersionEquals(nt, roleResourceVersion)); err != nil {
@@ -434,9 +458,25 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
 	}); err != nil {
 		nt.T.Fatal(err)
 	}
-	nt.T.Logf("Both of the two RootSyncs still report problems because the remediators detect the conflicts")
-	nt.WaitForRootSyncSyncError(configsync.RootSyncName, status.ManagementConflictErrorCode, "declared in another repository")
-	nt.WaitForRootSyncSyncError(rootSync2, status.ManagementConflictErrorCode, "declared in another repository")
+
+	// When the webhook is disabled, both RootSyncs will repeatedly try to adopt the object.
+	nt.T.Logf("Both RootSyncs should still report conflicts with the webhook disabled")
+	waitInParallel(
+		// Reconciler conflict, detected by the first reconciler's applier OR reported by the second reconciler
+		func() {
+			nomostest.WatchForObject(nt, kinds.RootSyncV1Beta1(), configsync.RootSyncName, configsync.ControllerNamespace,
+				[]nomostest.Predicate{
+					nomostest.RootSyncHasSyncError(nt, status.ManagementConflictErrorCode, "declared in another repository"),
+				})
+		},
+		// Reconciler conflict, detected by the second reconciler's applier OR reported by the first reconciler
+		func() {
+			nomostest.WatchForObject(nt, kinds.RootSyncV1Beta1(), rootSync2, configsync.ControllerNamespace,
+				[]nomostest.Predicate{
+					nomostest.RootSyncHasSyncError(nt, status.ManagementConflictErrorCode, "declared in another repository"),
+				})
+		},
+	)
 
 	nt.T.Logf("Remove the declaration from one Root repo %s", configsync.RootSyncName)
 	nt.RootRepos[configsync.RootSyncName].Remove(podRoleFilePath)
@@ -678,4 +718,21 @@ func roleHasRules(wantRules []rbacv1.PolicyRule) nomostest.Predicate {
 		}
 		return nil
 	}
+}
+
+// waitInParallel executes multiple functions in parallel until they all return,
+// which allows waiting for multiple conditions simultaniously, without
+// requiring them all to be true at the same time.
+// This improves testing speed and consistency.
+func waitInParallel(fns ...func()) {
+	var wg sync.WaitGroup
+	for i := range fns {
+		fn := fns[i]
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			fn()
+		}()
+	}
+	wg.Wait()
 }
--- a/pkg/applier/kpt_applier.go
+++ b/pkg/applier/kpt_applier.go
@@ -525,7 +525,9 @@ func (a *Applier) sync(ctx context.Context, objs []client.Object) (map[schema.Gr
 // Errors implements Interface.
 // Errors returns the errors encountered during apply.
 func (a *Applier) Errors() status.MultiError {
-	return a.errs
+	// TODO: Make read/write of a.errs thread-safe
+	// Return a copy
+	return status.Append(nil, a.errs)
 }
 
 // Syncing implements Interface.
--- a/pkg/parse/root.go
+++ b/pkg/parse/root.go
@@ -57,28 +57,30 @@ func NewRootRunner(clusterName, syncName, reconcilerName string, format filesyst
 		return nil, err
 	}
 
-	opts := opts{
-		clusterName:        clusterName,
-		syncName:           syncName,
-		reconcilerName:     reconcilerName,
-		client:             c,
-		pollingPeriod:      pollingPeriod,
-		resyncPeriod:       resyncPeriod,
-		retryPeriod:        retryPeriod,
-		statusUpdatePeriod: statusUpdatePeriod,
-		files:              files{FileSource: fs},
-		parser:             filesystem.NewParser(fileReader),
-		updater: updater{
-			scope:      declared.RootReconciler,
-			resources:  resources,
-			applier:    app,
-			remediator: rem,
+	return &root{
+		opts: opts{
+			clusterName:        clusterName,
+			syncName:           syncName,
+			reconcilerName:     reconcilerName,
+			client:             c,
+			pollingPeriod:      pollingPeriod,
+			resyncPeriod:       resyncPeriod,
+			retryPeriod:        retryPeriod,
+			statusUpdatePeriod: statusUpdatePeriod,
+			files:              files{FileSource: fs},
+			parser:             filesystem.NewParser(fileReader),
+			updater: updater{
+				scope:      declared.RootReconciler,
+				resources:  resources,
+				applier:    app,
+				remediator: rem,
+			},
+			discoveryInterface: dc,
+			converter:          converter,
+			mux:                &sync.Mutex{},
 		},
-		discoveryInterface: dc,
-		converter:          converter,
-		mux:                &sync.Mutex{},
-	}
-	return &root{opts: opts, sourceFormat: format}, nil
+		sourceFormat: format,
+	}, nil
 }
 
 type root struct {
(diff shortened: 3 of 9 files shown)


e2e/testcases/multi_sync_test.go
```diff
@@ -407,10 +406,14 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
 	nt.RootRepos[rootSync2].Add(podRoleFilePath, rootPodRole())
 	nt.RootRepos[rootSync2].CommitAndPush("add conflicting pod owner role")
 
-	nt.T.Logf("The admission webhook should deny the update request in Root %s", rootSync2)
-	nt.WaitForRootSyncSyncError(rootSync2, applier.ApplierErrorCode, "denied the request")
-	nt.T.Logf("Root %s should also get surfaced with the conflict error", configsync.RootSyncName)
-	nt.WaitForRootSyncSyncError(configsync.RootSyncName, status.ManagementConflictErrorCode, "declared in another repository")
+	// When the webhook is enabled, it will stop adoption of managed objects.
+	// So only the second RootSync will show the conflict error.
```
nan-yu, line 410: The second RootSync will show the conflict error because the admission webhook denied the update. Then, the remediator of the second RootSync will detect the drift and surface the conflict to both RootSyncs. The reason for reporting to both is that the object can be removed from either of the repos.

Here is the logic to report the conflict error for the second RootSync:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L422.

Here is how to surface a similar error to the conflicting RootSync: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L427.

pkg/parse/root.go
```diff
@@ -173,7 +173,7 @@ func (p *root) setSourceStatusWithRetries(ctx context.Context, newStatus sourceS
 
 	currentRS := rs.DeepCopy()
 
-	setSourceStatus(&rs.Status.Source, p, newStatus, denominator)
+	setSourceStatusFields(&rs.Status.Source, p, newStatus, denominator)
```
nan-yu, line 176: Just curious why do we want to rename the function. Is the intent to make it clear this is different from uppercase one `SetSourceStatus`?

pkg/parse/run.go
```diff
@@ -129,14 +127,13 @@ func Run(ctx context.Context, p Parser) {
...
+
 				klog.V(3).Info("Updating sync status (periodic while not syncing)")
-				if err := p.SetSyncStatus(ctx, p.ApplierErrors()); err != nil {
+				if err := p.SetSyncStatus(ctx, syncErrs); err != nil {
 					klog.Warningf("failed to update remediator errors: %v", err)
 				}
-				remediatorErrs := p.RemediatorConflictErrors()
-				if len(remediatorErrs) > 0 {
-					UpdateConflictManagerStatus(ctx, remediatorErrs, p.K8sClient())
-				}
```
nan-yu, line 139: This is to report a similar conflict error to the other conflicting RootSync. Do we want to remove that?

pkg/parse/root.go
```diff
@@ -365,14 +365,7 @@ func setRenderingStatus(rendering *v1beta1.RenderingStatus, p Parser, newStatus
 func (p *root) SetSyncStatus(ctx context.Context, errs status.MultiError) error {
 	p.mux.Lock()
 	defer p.mux.Unlock()
-	var allErrs status.MultiError
-	remediatorErrs := p.remediator.ConflictErrors()
-	for _, e := range remediatorErrs {
-		allErrs = status.Append(allErrs, e)
-	}
-	// Add conflicting errors before other apply errors.
-	allErrs = status.Append(allErrs, errs)
```
nan-yu, line 374: Why do we move this block outside of this function?

pkg/parse/run.go
```diff
@@ -418,37 +428,15 @@ func updateSyncStatus(ctx context.Context, p Parser) {
...
+			// Combine remediator and applier errors
+			syncErrs := status.Append(p.RemediatorConflictErrors(), p.ApplierErrors())
+
 			klog.V(3).Info("Updating sync status (periodic while syncing)")
-			if err := p.SetSyncStatus(ctx, p.options().updater.applier.Errors()); err != nil {
+			if err := setSyncStatus(ctx, p, state, syncErrs); err != nil {
 				klog.Warningf("failed to update sync status: %v", err)
 			}
-			remediatorErrs := p.RemediatorConflictErrors()
-			if len(remediatorErrs) > 0 {
```
nan-yu, line 426: Here as well. This is to surface a similar conflict error to the other conflicting RootSync. I feel like it would be good to keep the behavior.

When the webhook is enabled, it is okay to only surface the error to the second RootSync.   But if the webhook is disabled (default behavior), two RootSync will fight with each other. It would be good to notify both RootSyncs when the conflict happens.

e2e/testcases/multi_sync_test.go
karlkfi: I see. So by removing `UpdateConflictManagerStatus` and just merging `p.RemediatorConflictErrors()`  I broke the "logic to report the conflict error for the second RootSync".

It wasn't super clear what that logic was doing. But I see now it's calling `conflictError.ConflictingManagerError()` on every element of the `[]ManagementConflictError`, which I guess inverts the error.

Why isn't that logic all just in the Remediator? It seems pretty convoluted to need to process the ManagementConflictErrors into a MultiError in the parser.

pkg/parse/root.go
karlkfi: I wanted the lowercase name for `setSyncStatus` in run.go, which acts as a wrapper to `SetSyncStatus` that handles updating the state cache, and I wanted to make sure all the setXStatus functions stayed consistently named.

It's kind of a mess that these are all package functions and not struct methods. It makes it hard to really see the layering.

karlkfi:
/hold

Turns out I unintentionally deleted the cross-RootSync conflict reporting in this PR. While this functionality is of debatable value, it does help compensate for the flakiness of errors from the applier, which are deleted before each apply run.

So after talking to @nan-yu and @janetkuo, we decided to remove cross-RootSync conflict reporting, but only after adding error merging to the applier. Error merging will hold onto apply errors until we see an event like the one that caused it. This will avoid loss of errors (including conflict errors) while the applier is running. Afterwards, we can continue this PR to remove the cross-RootSync conflict reporting.

karlkfi:
/unhold

I reverted the UpdateConflictManagerStatus removal and went with a refactor instead, to make sure it was being called consistently.

We can remove it later, but it doesn't need to block the rest of these status consistency improvements.

e2e/testcases/multi_sync_test.go
karlkfi: I've reverted this change and instead migrated both `WaitForRootSyncSyncError` calls to use WaitForObject in parallel instead. This should evaluate faster and more consistently, without actually changing the existing behavior, except to make it less flakey.

pkg/parse/run.go
karlkfi: Reverted this removal and moved it into `setSyncStatus`.

pkg/parse/root.go
karlkfi: I've moved the aggregation of sync errors into the updater's new `Errors()` method, wrapped by `Parser.SyncErrors()`. This way it includes all the different sync errors consistency and in once place, instead of having the merges spread across the code.

pkg/parse/run.go
karlkfi: Reverted this removal and moved it into `setSyncStatus`.

pkg/parse/root.go
```diff
@@ -57,28 +57,30 @@ func NewRootRunner(clusterName, syncName, reconcilerName string, format filesyst
...
-		retryPeriod:        retryPeriod,
-		statusUpdatePeriod: statusUpdatePeriod,
-		files:              files{FileSource: fs},
-		parser:             filesystem.NewParser(fileReader),
-		updater: updater{
-			scope:      declared.RootReconciler,
-			resources:  resources,
-			applier:    app,
-			remediator: rem,
+	return &root{
```
karlkfi, line 60: This refactor just moves the `opts` construction into the `root` to avoid a copy of the mutex pointer, which fixes a linter error.

pkg/parse/run.go
```diff
@@ -419,21 +442,21 @@ func updateSyncStatus(ctx context.Context, p Parser) {
...
 			updateTimer.Reset(updatePeriod) // Schedule status update attempt
 		}
 	}
 }
 
-// UpdateConflictManagerStatus reports the conflict in the conflicting manager.
-func UpdateConflictManagerStatus(ctx context.Context, conflictErrs []status.ManagementConflictError, k8sClient client.Client) {
+// reportRootSyncConflicts reports conflicts to the RootSync that manages the
+// conflicting resources.
+func reportRootSyncConflicts(ctx context.Context, k8sClient client.Client, conflictErrs []status.ManagementConflictError) error {
```
karlkfi, line 456: I don't know if this name change is critical. I can revert it if you like the old name better. But the old name made me think THIS RootSync was being updated and not some other RootSync. I used "report" instead of "update" because it seemed more like something you might do to a remote object. I also made it lowercase because now that the async status update is here, it's only used in this file, so it doesn't need to be public any more.

pkg/parse/run.go
```diff
@@ -443,12 +466,19 @@ func UpdateConflictManagerStatus(ctx context.Context, conflictErrs []status.Mana
 
 	for conflictingManager, conflictErrors := range conflictingManagerErrors {
 		scope, name := declared.ManagerScopeAndName(conflictingManager)
-		if scope != declared.RootReconciler {
-			klog.Infof("No need to notify namespace reconciler for the conflict")
-			continue
-		}
-		if err := prependRootSyncRemediatorStatus(ctx, k8sClient, name, conflictErrors, defaultDenominator); err != nil {
-			klog.Warningf("failed to add the management conflict error to RootSync %s: %v", name, err)
+		if scope == declared.RootReconciler {
```
karlkfi, line 469: This little refactor hopefully only changes readability, comments, and logging. Otherwise it's functionally the same.

pkg/applier/kpt_applier.go
```diff
@@ -525,7 +525,9 @@ func (a *Applier) sync(ctx context.Context, objs []client.Object) (map[schema.Gr
 // Errors implements Interface.
 // Errors returns the errors encountered during apply.
 func (a *Applier) Errors() status.MultiError {
-	return a.errs
+	// TODO: Make read/write of a.errs thread-safe
```
karlkfi, line 528: Unfortunately, we can't just add a mutex around the errors here, because they're modified in multiple places. But fixing that is a larger change I want to do in a separate PR.

pkg/parse/root.go
karlkfi: If you can think of a better name for the wrapper func, I can rename these back.

e2e/testcases/multi_sync_test.go
```diff
@@ -415,10 +415,27 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
...
+		func() {
+			nomostest.WatchForObject(nt, kinds.RootSyncV1Beta1(), configsync.RootSyncName, configsync.ControllerNamespace,
+				[]nomostest.Predicate{
+					nomostest.RootSyncHasSyncError(nt, status.ManagementConflictErrorCode, "declared in another repository"),
+				})
+		},
+		func() {
+			nomostest.WatchForObject(nt, kinds.RootSyncV1Beta1(), rootSync2, configsync.ControllerNamespace,
+				[]nomostest.Predicate{
+					nomostest.RootSyncHasSyncError(nt, status.ManagementConflictErrorCode, "declared in another repository"),
```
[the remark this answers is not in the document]
nan-yu: For the first RootSync, it should show the `denied the request` error.

pkg/parse/root.go
nan-yu: I see. I missed the new package function `setSyncStatus` defined in another file. It sounds good to have them consistent.

pkg/parse/root.go
nan-yu: ACK

pkg/parse/run.go
nan-yu: sounds good

pkg/parse/run.go
```diff
@@ -443,12 +466,19 @@ func UpdateConflictManagerStatus(ctx context.Context, conflictErrs []status.Mana
 
 	for conflictingManager, conflictErrors := range conflictingManagerErrors {
 		scope, name := declared.ManagerScopeAndName(conflictingManager)
-		if scope != declared.RootReconciler {
-			klog.Infof("No need to notify namespace reconciler for the conflict")
-			continue
-		}
-		if err := prependRootSyncRemediatorStatus(ctx, k8sClient, name, conflictErrors, defaultDenominator); err != nil {
```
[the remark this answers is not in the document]
nan-yu: The original intent was to make the conflict easy to be discovered by both RootSyncs. It is possible that the syncing process takes 10-20 minutes and the RootSync remains syncing without any issue. The cross-posted conflict error reports the conflicts earlier.

After talking with Karl, he will remove the cross-posted error in a separate PR after adding error merging to the applier.

nan-yu (commented):
looks good with one minor error message. When the webhook is disabled, the first RootSync should expect a `denied the request` error.

pkg/parse/run.go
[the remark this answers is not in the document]
karlkfi: This code causes reconciler-B to update the rsync status of reconciler-A. This makes it possible for the owner of rsync-A to know there's a conflict.

Since reconciler-A applied the object first, the object has the manager set to reconciler-A. So when reconciler-B tries to update that object the apply fails, rejected by the webhook. Without this code, when the webhook is enabled, only rsync-B shows the conflict.

When the webhook is disabled, any two RootSyncs trying to manage the same object with actively fight, each taking ownership of the object with every apply. In this case, both will show the error from both the applier and the remediator.

RepoSyncs only adopt if unmanaged. So they won't fight, but can conflict. As with the webhook being enabled, this code reports the conflict on both sides.

Personally I think this behavior is confusing, harmful because it causes status flakiness, and technically incorrect because it makes the metrics not match the status API. But @nan-yu thinks we should keep it for now to make it easier to detect conflicts. Our compromise is for us to remove it later, after we improve the applier errors to not invalidate all errors when starting a new apply until we receive the event that created the error, thus making it easier to detect conflict errors from the applier.

e2e/testcases/multi_sync_test.go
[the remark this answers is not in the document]
karlkfi: AFAICT, "denied the request" doesn't exist in the code. 

"declared in another repository" is from the `KptManagementConflictError` which is returned by both the applier and the remediator.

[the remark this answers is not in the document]
nan-yu: `denied the request` is from the admission webhook: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/errors/statuserror.go#L29.

The error will first show up because the admission webhook denied the request. The `declared in another repository` error is surfaced by the conflicting remediator. The RootSync status will flip.

Technically, we should expect the `denied the request` error for the first RootSync, especially the cross-RootSync update is going to be removed.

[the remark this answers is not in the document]
karlkfi: Thanks for finding the `denied the request` code. Not sure why it didn't come up in my search.

Should we expect both the `denied the request` and the `declared in another repository` then? The tests are passing as-is, which seems to indicate that the conflict error from the remediator is definitely being propagated from rootSync2 to rootSync1. Adding `denied the request` would confirm that it also was detected via the apply being rejected by the webhook. But we probably wouldn't expect to see them at the same time...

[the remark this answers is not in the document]
nan-yu: Yes, we can expect both errors. Both RootSync1's applier and RootSync2's remediator updates the RootSync1's status. The RootSync1's status was expected to bounce back and forth to report the conflict in different ways. For this test, we care more about the `denied the request` error.

As we're planning to remove the cross-RootSync update, we can validate the `denied the request` error for now.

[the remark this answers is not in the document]
karlkfi: I've updated the test to check for all the possible errors.

When we disable cross-RootSync update, some of these will fail and we can remove them.

nan-yu (approved):
/lgtm

PR #284, opened 2022-11-11
Several enhancements to metrics

haiyanmeng:
Several enhancements to metrics
    
    1) Enhance the `reconciler_errors` metric
       * stop using the `parsing` `component` label value to be consistent
         with R*Syncs status APIs
       * support the `errorclass` label correctly. Currently, this label is
         always set to an empty string. With this PR, the value of the
         errorclass label can be “1xxx”, “2xxx”, “9xxx”. For every
         component, three time series will be recorded, one for each
         errorclass.
    2) Enhance the `apply_operations` metric
        * add a new label named `controller` to track whether the operation
          is from the applier or the remediator
        * record this metric when an ApplyFailed/PruneFailed event is
          observed on the event channel returned from the cli-utils library.
    3) Enhance the `last_sync_timestamp` metric
        * only send this metric when the `commit` label is not empty.
          Currently, the metric may be sent with an empty commit, since we
          support asynchronous status updates.
    4) Enahnce the `parser_duration_seconds` metric
        * Currently, the bounds for the histogram distribution is
          []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}, and
          all the timeseries whose values are greater than 10 are put into
          the same bucket, [10, +Infinite]. This PR uses the bounds
          []float64{1, 5, 10, 30, 60, 300, 600, 1200, 1800} to better align
          with the common use cases.
    5) Enhance the `apply_duration_seconds` metric
        * Currently, the bounds for the histogram distribution is
          []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}, and
          all the timeseries whose values are greater than 10 are put into
          the same bucket, [10, +Infinite]. This PR uses the bounds
          []float64{1, 5, 10, 30, 60, 300, 600, 1200, 1800} to better align
          with the common use cases.

### DIFF
--- a/pkg/metrics/tagkeys.go
+++ b/pkg/metrics/tagkeys.go
@@ -31,7 +31,10 @@ var (
 	// KeyOperation groups metrics by their operation. Possible values: create, patch, update, delete.
 	KeyOperation, _ = tag.NewKey("operation")
 
-	// KeyComponent groups metrics by their component. Possible values: parsing, source, sync, rendering, readiness(from Resource Group Controller).
+	// KeyController groups metrics by their controller. Possible values: applier, remediator, syncer.
+	KeyController, _ = tag.NewKey("controller")
+
+	// KeyComponent groups metrics by their component. Possible values: source, sync, rendering, readiness(from Resource Group Controller).
 	KeyComponent, _ = tag.NewKey("component")
 
 	// KeyErrorClass groups metrics by their error code.
@@ -94,6 +97,10 @@ const (
 	// CommitNone is the string value for the commit key indicating that no
 	// commit has been synced.
 	CommitNone = "NONE"
+	// ApplierController is the string value for the applier controller in the multi-repo mode
+	ApplierController = "applier"
+	// RemediatorController is the string value for the remediator controller in the multi-repo mode
+	RemediatorController = "remediator"
 )
 
 // StatusTagKey returns a string representation of the error, if it exists, otherwise success.
--- a/pkg/parse/root.go
+++ b/pkg/parse/root.go
@@ -139,8 +139,6 @@ func (p *root) parseSource(ctx context.Context, state sourceState) ([]ast.FileOb
 		objs, err = validate.Hierarchical(objs, options)
 	}
 
-	metrics.RecordReconcilerErrors(ctx, "parsing", status.NonBlockingErrors(err))
-
 	if status.HasBlockingErrors(err) {
 		return nil, err
 	}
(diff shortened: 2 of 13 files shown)


pkg/metrics/tagkeys.go
```diff
@@ -94,6 +97,12 @@ const (
 	// CommitNone is the string value for the commit key indicating that no
 	// commit has been synced.
 	CommitNone = "NONE"
+	// ApplierController is the string value for the applier controller in the multi-repo mode
+	ApplierController = "applier"
+	// RemediatorController is the string value for the remediator controller in the multi-repo mode
+	RemediatorController = "remediator"
+	// SyncerController is the string value for the syncer controller in the mono-repo mode
+	SyncerController = "syncer"
```
nan-yu, line 105: Do we still need this? The monorepo code has been moved out of this project.

pkg/parse/root.go
```diff
@@ -139,8 +139,6 @@ func (p *root) parseSource(ctx context.Context, state sourceState) ([]ast.FileOb
 		objs, err = validate.Hierarchical(objs, options)
 	}
 
-	metrics.RecordReconcilerErrors(ctx, "parsing", status.NonBlockingErrors(err))
```
nan-yu, line 142: Instead of removing the reconciler_errors caused by parsing errors, should we replace the component with `source`?

pkg/metrics/tagkeys.go
janetkuo: It doesn't seem to be referenced anywhere?

pkg/parse/root.go
haiyanmeng: The parsing errors is a part of the source errors. This metric record is duplicate and does not align with the R*Sync status APIs.

pkg/metrics/tagkeys.go
haiyanmeng: Good point. Let me remove it.

haiyanmeng: Removed.

haiyanmeng:
@nan-yu , @janetkuo , I split this PR into 4 smaller PRs so that it'd be easier to debug or rollback if things go wrong.

https://github.com/GoogleContainerTools/kpt-config-sync/pull/292 Enhance the `parser_duration_seconds` and `apply_duration_seconds` metrics 
https://github.com/GoogleContainerTools/kpt-config-sync/pull/293 Enhance the `last_sync_timestamp` metric
https://github.com/GoogleContainerTools/kpt-config-sync/pull/294 Enhance the `apply_operations` metric
https://github.com/GoogleContainerTools/kpt-config-sync/pull/295 Enhance the `reconciler_errors` metric

pkg/parse/root.go
nan-yu: ACK

PR #285, opened 2022-11-12
test: fine grained permissions for ns reconciler

sdowell:
This change updates the test scaffolding to grant NS reconciler permissions based on test options rather than granting namespace admin by default. This should make the tests more closely reflect real usage, and make it possible to verify that the default NS reconciler ClusterRole is sufficient for core Config Sync functionality.


e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -104,3 +110,54 @@ func WithReconcileTimeout(timeout time.Duration) func(opt *New) {
...
+	policy := rbacv1.PolicyRule{
+		APIGroups: []string{rbacv1.APIGroupAll},
+		Resources: []string{rbacv1.ResourceAll},
+		Verbs:     []string{rbacv1.VerbAll},
+	}
+	NSReconcilerCustomPolicy(policy)(opt)
+}
+
+// NSReconcilerCustomPolicy specifies a custom policy to use for the NS reconciler ClusterRole
+func NSReconcilerCustomPolicy(policy rbacv1.PolicyRule) Opt {
```
karlkfi, line 159: How replacing all this with a func that accepts a vararg of rules? Then you would call `RepoSyncPolicy(policy.NamespaceAdmin(), policy.RepoSyncAdmin())` to return a `func(opt *New)`, and you could put the policy helpers in a sub-package. The nomostest packag eis already too cluttered. And I suspect it might be easier to find RepoSync- than NSReconciler- names.

e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -47,6 +50,9 @@ type MultiRepo struct {
 
 	// SkipNonLocalGitProvider will skip the test if run with a GitProvider type other than local.
 	SkipNonLocalGitProvider bool
+
+	// NSReconcilerPolicyRules will grant a list of PolicyRules to NS reconcilers
+	NSReconcilerPolicyRules []rbacv1.PolicyRule
```
karlkfi, line 55: I don't live having NS- here and also Namespace- in the same struct. Maybe this could be NamespaceAccess or RepoSyncPermissions or something?

e2e/nomostest/ntopts/multi_repo.go
sdowell: That sounds reasonable from an organization/discovery perspective. I think my current solution would be a bit more concise but I like your proposal. I'll try refactoring accordingly to see how it looks

e2e/nomostest/ntopts/multi_repo.go
sdowell: Updated, PTAL

e2e/nomostest/ntopts/multi_repo.go
sdowell: Updated, PTAL

e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -104,3 +108,10 @@ func WithReconcileTimeout(timeout time.Duration) func(opt *New) {
 		opt.ReconcileTimeout = &timeoutCopy
 	}
 }
+
+// RepoSyncPolicy specifies PolicyRule(s) to grant NS reconcilers
+func RepoSyncPolicy(policy ...rbacv1.PolicyRule) Opt {
```
karlkfi, line 113: How about just RepoSyncPermissions to make it consistent with the struct field?

sdowell: Done

karlkfi (approved):
/lgtm

Thanks for getting this cleaned up!
I'm glad it was easy to validate with least privilege.

PR #78, opened 2022-09-01
Avoid object updates if semantically equal

karlkfi:
- This avoids generation & resourceVersion bloat, which can cause
  unnecessary watch updates and ectd memory leaks.

`equality.Semantic.DeepEqual` is used by `CreateOrUpdate` in controller-runtime:  
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.12.3/pkg/controller/controllerutil/controllerutil.go#L215

PR #289, opened 2022-11-14
Add e2e and unit test coverage to make sure a root|ns-reconciler Deployment is not updated unnecessarily

haiyanmeng:
The quantity of CPU resources can be specified in multiple formats: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/. For example, '1' and '1000m' are semantically equal to each other.

There is a bug in ACM 1.11.0 - ACM 1.13.1, which causes a root|ns-reconciler Deployment to be updated constantly if the CPU resource request/limit are in the format of '1000m' instead of '1'. The issue is fixed in
https://github.com/GoogleContainerTools/kpt-config-sync/pull/78.

This PR add e2e and unit test coverage for this.

### DIFF
--- a/e2e/nomostest/predicates.go
+++ b/e2e/nomostest/predicates.go
@@ -24,6 +24,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"kpt.dev/configsync/pkg/core"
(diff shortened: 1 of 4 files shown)


e2e/nomostest/predicates.go
```diff
@@ -208,16 +209,16 @@ func HasCorrectResourceRequestsLimits(containerName string, cpuRequest, cpuLimit
 		}
 		for _, container := range dep.Spec.Template.Spec.Containers {
 			if containerName == container.Name {
-				if container.Resources.Requests[corev1.ResourceCPU] != cpuRequest {
+				if !equality.Semantic.DeepEqual(container.Resources.Requests[corev1.ResourceCPU], cpuRequest) {
```
sdowell, line 212: This looks like something that should be unit-testable. Can you add some unit tests to cover this bug fix?

sdowell, line 212: I'm a bit confused, is this a bug in our code or e2e tests? Also, this seems like something that should be unit-testable. Do we have coverage for this in our e2e tests?

sdowell: Just saw the description so I see this was fixed in another PR. I'm still curious about unit testing though

haiyanmeng: > I'm still curious about unit testing though

We can add some unit test coverage into pkg/reconcilermanager/controller.
Are you recommending to avoid testing this in the e2e test completely?

sdowell: A validation such as this seems like a better fit for lightweight unit tests. It doesn't seem like we should need e2e tests for this IMO, but I haven't looked too closely at the related code. We can also get more thorough scenario coverage through unit tests

haiyanmeng: This PR does not add new test cases. It just adds more validations into two existing e2e testcases.

nan-yu (approved):
/lgtm

PR #291, opened 2022-11-15
chore: Update resourcegroup-controller to v1.0.9

rquitales:


manifests/third_party/resourcegroup-manifest.yaml
```diff
@@ -357,6 +357,7 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
+  creationTimestamp: null
```
sdowell, line 360: Remove this (cause of presubmit failure)

rquitales: Yup, I'm just waiting on the e2e statuses first before pushing a new update.

rquitales: Done. Prior e2e tests also passed fyi.

PR #292, opened 2022-11-15
Enhance the `parser_duration_seconds` and `apply_duration_seconds` metrics

haiyanmeng:
Currently, the bounds for the histogram distribution is `[]float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}`, and all the timeseries whose values are greater than 10 are put into the same bucket, [10, +Infinite]. This PR uses the bounds `[]float64{1, 5, 10, 30, 60, 300, 600, 1200, 1800, 3600, 5400}` to better align with the common use cases.

### DIFF
--- a/pkg/metrics/views.go
+++ b/pkg/metrics/views.go
@@ -19,7 +19,11 @@ import (
 	"go.opencensus.io/tag"
 )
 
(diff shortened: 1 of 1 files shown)


nan-yu (approved):
/lgtm

PR #293, opened 2022-11-15
Enhance the `last_sync_timestamp` metric

haiyanmeng:
Currently, the metric may be sent with an empty commit, since we support asynchronous status updates.  This PR ensures this metric is only recorded when the `commit` label is not empty.


nan-yu (approved):
/lgtm

PR #295, opened 2022-11-15
Enhance the `reconciler_errors` metric

haiyanmeng:
* stop using the `parsing` `component` label value to be consistent with R*Syncs status APIs
   * support the `errorclass` label correctly. Currently, this label is always set to an empty string. With this PR, the value of the errorclass label can be “1xxx”, “2xxx”, “9xxx”. For every component, three time series will be recorded, one for each errorclass.


nan-yu (approved):
/lgtm

nan-yu (approved):
/lgtm

PR #297, opened 2022-11-15
Update the kubevirt version from v0.46.1 to v0.58.0

haiyanmeng:
Currently, the version of kubevirt used in CS is v0.46.1, which uses the policy/v1beta1 API version of PodDisruptionBudget.  However, the policy/v1beta1 API version of PodDisruptionBudget is no longer served in v1.25.
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125

This PR also removes TestApplyScopedResourcesHierarchicalMode and     renames     TestApplyScopedResourcesUnstructuredMode to TestApplyScopedResources.     The reason for removing TestApplyScopedResourcesHierarchicalMode is that     it is very complex to reconstruct the hierachical directory after     updating kubevirt version.


nan-yu (approved):
/lgtm

PR #298, opened 2022-11-15
Removing PodSecurityPolicy validation

victorpras:
PodSecurityPolicy is deprecated in Kubernetes 1.25+, so we will be removing PodSecurityPolicy validation from our e2e test


nan-yu (approved):
/lgtm

PR #299, opened 2022-11-15
Add e2e and unit test coverage to make sure a root|ns-reconciler Deployment is not updated unnecessarily

haiyanmeng:
The quantity of CPU resources can be specified in multiple formats: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/. For example, '1' and '1000m' are semantically equal to each other.

There is a bug in ACM 1.11.0 - ACM 1.13.1, which causes a root|ns-reconciler Deployment to be updated constantly if the CPU resource request/limit are in the format of '1000m' instead of '1'. The issue is fixed in
https://github.com/GoogleContainerTools/kpt-config-sync/pull/78.

This PR add e2e and unit test coverage for this.

This PR is the second attempt to fix this, the first attempt (https://github.com/GoogleContainerTools/kpt-config-sync/pull/289) was reverted because it broken GKE CI jobs which runs e2e tests on a shared environment. The first attempt assumes the generation of a root|ns reconciler deployment is 1 at the start of a e2e test. This PR removes this assumption.

### DIFF
--- a/e2e/testcases/override_resource_limits_test.go
+++ b/e2e/testcases/override_resource_limits_test.go
@@ -90,7 +90,8 @@ func TestOverrideReconcilerResourcesV1Alpha1(t *testing.T) {
 	defaultGitSyncCPULimits, defaultGitSyncMemLimits := gitSyncResourceLimits[corev1.ResourceCPU], gitSyncResourceLimits[corev1.ResourceMemory]
 
 	// Verify root-reconciler uses the default resource requests and limits
-	err := nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem, &appsv1.Deployment{},
+	rootReconcilerDeployment := &appsv1.Deployment{}
+	err := nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem, rootReconcilerDeployment,
 		nomostest.HasCorrectResourceRequestsLimits(reconcilermanager.Reconciler, defaultReconcilerCPURequest, defaultReconcilerCPULimits, defaultReconcilerMemRequest, defaultReconcilerMemLimits),
 		nomostest.HasCorrectResourceRequestsLimits(reconcilermanager.GitSync, defaultGitSyncCPURequest, defaultGitSyncCPULimits, defaultGitSyncMemRequest, defaultGitSyncMemLimits))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


e2e/testcases/override_resource_limits_test.go
```diff
@@ -144,6 +149,7 @@ func TestOverrideReconcilerResourcesV1Alpha1(t *testing.T) {
 
 	// Verify ns-reconciler-frontend uses the default resource requests and limits
 	err = nt.Validate(core.NsReconcilerName(frontendNamespace, configsync.RepoSyncName), v1.NSConfigManagementSystem, &appsv1.Deployment{},
+		hasGeneration(nsReconcilerBackendDeployment.Generation),
```
nan-yu, line 152: nsReconcilerBackendDeployment -> nsReconcilerFrontendDeployment

e2e/testcases/override_resource_limits_test.go
```diff
@@ -233,16 +243,18 @@ func TestOverrideReconcilerResourcesV1Alpha1(t *testing.T) {
 
 	// Verify the resource limits of ns-reconciler-backend are not affected by the resource limit change of root-reconciler
 	err = nt.Validate(core.NsReconcilerName(backendNamespace, configsync.RepoSyncName), v1.NSConfigManagementSystem, &appsv1.Deployment{},
-		nomostest.HasCorrectResourceRequestsLimits(reconcilermanager.Reconciler, resource.MustParse("500m"), resource.MustParse("555m"), resource.MustParse("500Mi"), resource.MustParse("555Mi")),
-		nomostest.HasCorrectResourceRequestsLimits(reconcilermanager.GitSync, resource.MustParse("600m"), resource.MustParse("666m"), resource.MustParse("600Mi"), resource.MustParse("666Mi")))
+		hasGeneration(nsReconcilerBackendDeployment.Generation+1),
```
nan-yu, line 246: I would suggest declaring variables for these generations to show incremental changes. For example, we can have a `nsBackendDeploymentGeneration` and increase the value when needed. In this case, we could clearly see the generation remains the same.

haiyanmeng: Great suggestion! Fixed it. Thanks!

e2e/testcases/override_resource_limits_test.go
haiyanmeng: Fixed it

nan-yu (approved):
/lgtm

PR #301, opened 2022-11-16
Improve thread-safety of sync status updates

karlkfi:
- Wrap applier errors in their own mutex
- Move applier.Syncing() to Parser.Syncing() with updater.Updating() doing the heavy lifting. The Syncing condition status should reflect the whole updater, not just the applier.
- Rename applier.Interface -> KptApplier to make room for adding KptDestroyer in the future.
- Rename KptApplier.sync -> applyInner to make room for adding destroyInner in the future.
- Refactor state.syncStatus and Parser.SetSyncStatus to use a new syncStatus struct, which is like the sourceStatus struct but with an added sync bool. This should also be less confusing, because state.syncStatus no longer uses a sourceStatus struct as its value.
- Add a new setSyncStatusErrors func to use in prependRootSyncRemediatorStatus, to avoid needing to to re-construct the initial syncStatus when prepending errors. (Note: This cross-rsync update changes the `.status.sync.errs` and `.status.sync.lastUpdate` without updating the Syncing condition or `reconciler_errors` metric, which means they will continue to be out of sync when there's a conflict. But this is not a new problem.)

### DIFF
--- a/pkg/applier/kpt_applier.go
+++ b/pkg/applier/kpt_applier.go
@@ -77,38 +77,37 @@ type Applier struct {
 	// errs tracks all the errors the applier encounters.
 	// This field is cleared at the start of the `Applier.Apply` method
 	errs status.MultiError
-	// syncing indicates whether the applier is syncing.
-	syncing bool
 	// syncKind is the Kind of the RSync object: RootSync or RepoSync
 	syncKind string
 	// syncName is the name of RSync object
 	syncName string
 	// syncNamespace is the namespace of RSync object
 	syncNamespace string
-	// mux is an Applier-level mutext to prevent concurrent Apply() and Refresh()
-	mux sync.Mutex
 	// statusMode controls if the applier injects the acutation status into the
 	// ResourceGroup object
 	statusMode string
 	// reconcileTimeout controls the reconcile and prune timeout
 	reconcileTimeout time.Duration
+
+	// applyMux prevents concurrent Apply() calls
+	applyMux sync.Mutex
+	// errorMux prevents concurrent modifications to the cached set of errors
+	errorMux sync.RWMutex
 }
 
-// Interface is a fake-able subset of the interface Applier implements.
+// KptApplier is a fake-able subset of the interface Applier implements.
 //
 // Placed here to make discovering the production implementation (above) easier.
-type Interface interface {
+type KptApplier interface {
 	// Apply updates the resource API server with the latest parsed git resource.
 	// This is called when a new change in the git resource is detected. It also
 	// returns a map of the GVKs which were successfully applied by the Applier.
 	Apply(ctx context.Context, desiredResources []client.Object) (map[schema.GroupVersionKind]struct{}, status.MultiError)
 	// Errors returns the errors encountered during apply.
 	Errors() status.MultiError
-	// Syncing indicates whether the applier is syncing.
-	Syncing() bool
 }
 
-var _ Interface = &Applier{}
+var _ KptApplier = &Applier{}
 
 // NewNamespaceApplier initializes an applier that fetches a certain namespace's resources from
 // the API server.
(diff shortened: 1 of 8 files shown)


pkg/applier/kpt_applier.go
```diff
@@ -512,57 +514,60 @@ func (a *Applier) sync(ctx context.Context, objs []client.Object) (map[schema.Gr
...
 // Errors implements Interface.
-// Errors returns the errors encountered during apply.
 func (a *Applier) Errors() status.MultiError {
-	// TODO: Make read/write of a.errs thread-safe
-	// Return a copy
+	a.errorMux.RLock()
+	defer a.errorMux.RUnlock()
+
+	// Return a copy to avoid persisting caller modifications
 	return status.Append(nil, a.errs)
```
haiyanmeng, line 539: why not use `return a.errs`?

karlkfi: to avoid persisting modifications by any caller that might append additional errors to the MultiError.

pkg/applier/kpt_applier.go
```diff
@@ -77,38 +77,37 @@ type Applier struct {
 	// errs tracks all the errors the applier encounters.
 	// This field is cleared at the start of the `Applier.Apply` method
 	errs status.MultiError
-	// syncing indicates whether the applier is syncing.
-	syncing bool
```
sdowell, line 81: oh man was this another bool used for synchronization?

pkg/applier/kpt_applier.go
```diff
@@ -512,57 +514,60 @@ func (a *Applier) sync(ctx context.Context, objs []client.Object) (map[schema.Gr
...
-	// CRD, then a future cycle is able to apply the CR).
-	// TODO: Here and elsewhere, pass the MultiError as a parameter.
-	return a.sync(ctx, desiredResource)
+	a.applyMux.Lock()
+	defer a.applyMux.Unlock()
+
+	// Ideally we want to avoid invalidating errors that will continue to happen,
+	// but for now, invalidate all errors until they recur.
+	// TODO: improve error cache invalidation to make rsync status more stable
+	a.invalidateErrors()
```
sdowell, line 569: What's the plan for how we want to handle this in the long run?

pkg/applier/kpt_applier.go
karlkfi: yup. and set outside of the lock.

pkg/parse/updater.go
```diff
@@ -69,17 +72,22 @@ func (u *updater) Errors() status.MultiError {
...
 func (u *updater) setWatchErrs(errs status.MultiError) {
-	u.mux.Lock()
-	defer u.mux.Unlock()
+	u.errorMux.Lock()
+	defer u.errorMux.Unlock()
 	u.watchErrs = errs
 }
 
+// Updating returns true if the Update method is running.
+func (u *updater) Updating() bool {
```
sdowell, line 87: This smells a bit like the other bool. Is the idea that by setting this at a higher level (rather than reaching into the applier to get the value) it is less prone to race conditions?

pkg/applier/kpt_applier.go
karlkfi: I'm writing a doc. But the TLDR is to do more careful invalidation by caching what event type and object added/owns the error, and not invalidating that error until we see that event again, or the apply ends.

pkg/parse/updater.go
karlkfi: Moving it up a level allows including the other actions performed by `updater.Update()`.

Setting the bool inside the newly added mutex ensures that the bool value is not affected by race conditions.

But yes, it's a lazy fix, not an ideal fix. Ideally the control flow would be modified to use channels, but I decided to start with the quick fix, because I've had trouble getting large rewrites tested and approved. And I'd like to get back to finishing the finalizer.

karlkfi: The errors and syncing bool could be bundled together in a Status() result or something, to be atomic, and the current status could be the result of a for loop that reads from a channel accepting errors and start/stop signals.

But once you go down that road, you realize that the applier, updater, parser, and reconciler layers ALL have the same problematic pattern and ALL need to be rewritten to use a top level reconcile loop. Then you wouldn't need to synchronize any of these lower layers.

nan-yu (approved):
/lgtm

karlkfi:
/retest

Looks like TestInvalidRepoSyncBranchStatus might be flakey with a missing `reconciler_errors`

PR #303, opened 2022-11-16
Clean up REST client construction

karlkfi:
- Split out NewFromConfigFile & NewFromInClusterConfig for direct use
- Add a new WatchConfig and WatchCLient to the e2e test config. **This fixes a bug where the WatchForObject calls would timeout in 5s on GKE.**
- Use NewRestConfig in e2e for both GKE and Kind. **This uses UpdateQPS in the e2e tests for faster API calls.**
- Remove the unused MonoRepo client config (slower QPS)

### DIFF
--- a/e2e/nomostest/client.go
+++ b/e2e/nomostest/client.go
@@ -35,6 +35,7 @@ import (
 	configsyncv1alpha1 "kpt.dev/configsync/pkg/api/configsync/v1alpha1"
 	configsyncv1beta1 "kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/client/restconfig"
+	"kpt.dev/configsync/pkg/remediator/watch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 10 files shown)


e2e/nomostest/client.go
```diff
@@ -93,13 +94,23 @@ func newScheme(t testing.NTB) *runtime.Scheme {
...
+
+	restConfig, err := restconfig.NewFromConfigFile(opts.KubeconfigPath)
+	if err != nil {
+		t.Fatalf("building rest.Config: %v", err)
+	}
+	opts.RESTConfig = restConfig
+
+	// Copy the RESTConfig to create the WatchConfig, but use a longer timeout.
+	opts.WatchConfig = restconfig.DeepCopy(restConfig)
+	opts.WatchConfig.Timeout = watch.RESTConfigTimeout
```
sdowell, line 115: Is it important we use the same timeout defined in remediator, or can we just define our own timeout here?

karlkfi: It's not important, but the constant existed already. It's basically just "a really long time".

When using an Informer, it handles restarting cancelled requests, so it's not really a problem as long as it's long enough to not be creating new requests all the time.

However, when using a raw Watch, like in WatchForObject, the timeout just needs to be longer than the timeout configured in the WaitOption (DefaultWaitTimeout by default).

pkg/client/restconfig/restconfig.go
```diff
@@ -21,108 +21,100 @@ import (
...
 // throttling is disabled, vice versa.
-// It is used by the multi-repo mode.
 func NewRestConfig(timeout time.Duration) (*rest.Config, error) {
-	return newRestConfig(timeout, false)
-}
-
-func newRestConfig(timeout time.Duration, isMonoRepo bool) (*rest.Config, error) {
 	var errorStrs []string
 
-	for _, source := range configSources {
```
sdowell, line 74: I think I may prefer the previous syntax. What was your motivation for this refactor?

pkg/client/restconfig/config.go
```diff
@@ -55,16 +55,6 @@ func newConfigPath() (string, error) {
 	return path, nil
 }
 
-// newConfigFromPath creates a rest.Config from a configuration file at the
-// supplied path.
-func newConfigFromPath(path string) (*rest.Config, error) {
```
sdowell, line 60: What's the motivation for moving these functions to the other file rather than renaming them here?

karlkfi: The other file has the public constructors. But it doesn’t really matter.

pkg/client/restconfig/restconfig.go
karlkfi: I needed to expose a constructor that just used the specified path, and I always found this for loop over a package variable of structs to be a convoluted and hard to read way to save a few lines of repeated text. 

The new way does have a slightly deeper cyclomatic complexity but as trade off it shows the fallback options all in one place on the screen at the same time with less indirection and clearer user facing error messages.

pkg/client/restconfig/config.go
karlkfi: Having them next to where they’re used makes for less flipping between files when reading the code. 

If you prefer them here, tho. I can move them back. I’m not really sure why these are two separate files.

PR #304, opened 2022-11-17
ci: remove mono repo clusters

sdowell:
These clusters are no longer in use since the mono repo jobs have been deleted


nan-yu (approved):
/lgtm

PR #305, opened 2022-11-17
chore: bump default k8s version for kind

sdowell:


e2e/flags.go
```diff
@@ -56,7 +56,7 @@ var Debug = flag.Bool("debug", false,
 
 // KubernetesVersion is the version of Kubernetes to test against. Only has effect
 // when testing against test-created Kind clusters.
-var KubernetesVersion = flag.String("kubernetes-version", "1.21",
+var KubernetesVersion = flag.String("kubernetes-version", "1.23",
```
karlkfi, line 59: Is this the only place this is configured now?
Does the kind version need to be upgraded?

sdowell: This doesn't change the kind version, just the default k8s version that is used when creating the kind clusters. Each version of kind supports a range of kind versions
e.g. https://github.com/kubernetes-sigs/kind/releases/tag/v0.14.0

I imagine we probably want to bump this at some cadence so that our presubmits are testing on one of the k8s versions that we support

sdowell: I do want to also simplify the way we define the kind version to make future upgrades easier, but that's a different change

karlkfi (approved):
/lgtm

PR #307, opened 2022-11-17
ci: consolidate gke cluster args

sdowell:
There was a combination of a required kubeconfig and env vars that needed to be set to run tests on GKE. This change consolidates the setup to only require the env vars to be set. This simplifies some of the scaffolding and interface for running GKE tests.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -104,10 +104,14 @@ current context in your kubeconfig.
 ```shell
 # Ensure gcloud context is set to correct project
 gcloud config set project <PROJECT_ID>
-# Ensure kubectl context is set to correct cluster
-kubectl config set-context <CONTEXT>
 # Build images/manifests and push images
 make config-sync-manifest
+# Set env vars for GKE cluster
+export GCP_PROJECT=<PROJECT_ID>
+export GCP_CLUSTER=<CLUSTER_NAME>
+# One of GCP_REGION and GCP_ZONE must be set (but not both)
+export GCP_REGION=<REGION>
+export GCP_ZONE=<ZONE>
 # Run the tests with image prefix/tag from previous step and desired test regex
 go test ./e2e/... --e2e --debug --test.v --share-test-env=true --test.parallel=1 --test-cluster=gke --test.run (test name regexp)
 ```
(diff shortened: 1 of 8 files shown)


docs/development.md
```diff
@@ -104,10 +104,14 @@ current context in your kubeconfig.
...
 # Ensure gcloud context is set to correct project
 gcloud config set project <PROJECT_ID>
-# Ensure kubectl context is set to correct cluster
-kubectl config set-context <CONTEXT>
 # Build images/manifests and push images
 make config-sync-manifest
+# Set env vars for GKE cluster
+export GCP_PROJECT=<PROJECT_ID>
+export GCP_CLUSTER=<CLUSTER_NAME>
+# One of GCP_REGION and GCP_ZONE must be set
```
karlkfi, line 112: but not both?

e2e/nomostest/new.go
```diff
@@ -96,6 +96,18 @@ func newOptStruct(testName, tmpDir string, t nomostesting.NTB, ntOptions ...ntop
...
 	}
 
+	if *e2e.TestCluster == e2e.GKE { // required env vars for GKE
+		if nomostesting.GCPProjectIDFromEnv == "" {
+			t.Fatal("Environment variable 'GCP_PROJECT' is required for GKE clusters")
+		}
+		if nomostesting.GCPClusterFromEnv == "" {
+			t.Fatal("Environment variable 'GCP_CLUSTER' is required for GKE clusters")
+		}
+		if nomostesting.GCPRegionFromEnv == "" && nomostesting.GCPZoneFromEnv == "" {
```
karlkfi, line 106: Should we still error if both are set, like the shell script did?

sdowell: Yeah I think so, I'll update this

docs/development.md
sdowell: Updated

e2e/nomostest/new.go
sdowell: Done

karlkfi (approved):
/lgtm

PR #311, opened 2022-11-18
Make sure users can reset the resources.limits override

xinnywinne:
Remove resources.limits field from reconciler manager allowList:

Previous reconciler manager allowList related logic added resources.limits to allowList to ignore the differences between declared and current deployments on this field when the declared deployment does not have resources.limits defined. However, this did not consider the case when user want to reset the resources.limits override, the declared deployment will also not have resources.limits fields defined.

Rename some functions in reconciler_base

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_base.go
+++ b/pkg/reconcilermanager/controllers/reconciler_base.go
@@ -208,8 +208,8 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
 	dRef := client.ObjectKeyFromObject(declared)
 	kind := "Deployment"
 	forcePatch := true
-	deploymentResourceInterface := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace)
-	unObjCurrent, err := deploymentResourceInterface.Get(ctx, dRef.Name, metav1.GetOptions{})
+	deploymentClient := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace)
+	currentDeploymentUnstructured, err := deploymentClient.Get(ctx, dRef.Name, metav1.GetOptions{})
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return controllerutil.OperationResultNone, err
@@ -221,12 +221,14 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
 		if err != nil {
 			return controllerutil.OperationResultNone, fmt.Errorf("failed to marshal declared deployment object to byte array: %w", err)
 		}
-		_, err = deploymentResourceInterface.Patch(ctx, dRef.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: reconcilermanager.ManagerName, Force: &forcePatch})
+		_, err = deploymentClient.Patch(ctx, dRef.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: reconcilermanager.ManagerName, Force: &forcePatch})
 		if err != nil {
 			return controllerutil.OperationResultNone, err
 		}
 		return controllerutil.OperationResultCreated, nil
 	}
+	currentGeneration := currentDeploymentUnstructured.GetGeneration()
+	currentUID := currentDeploymentUnstructured.GetUID()
 
 	if r.isAutopilotCluster == nil {
 		isAutopilot, err := util.IsGKEAutopilotCluster(r.client)
@@ -235,24 +237,24 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
 		}
 		r.isAutopilotCluster = &isAutopilot
 	}
-	dep, err := isDeploymentSame(*r.isAutopilotCluster, declared, unObjCurrent, reconcilerManagerAllowList, r.scheme)
+	dep, err := compareDeploymentsToCreatePatchData(*r.isAutopilotCluster, declared, currentDeploymentUnstructured, reconcilerManagerAllowList, r.scheme)
 	if err != nil {
 		return controllerutil.OperationResultNone, err
 	}
-	if dep.isDeploymentAdjusted {
+	if dep.adjusted {
 		mutator := "Autopilot"
 		r.log.V(3).Info("Managed object container resources updated",
 			logFieldObject, dRef.String(),
 			logFieldKind, kind,
 			"mutator", mutator)
 	}
-	if dep.isDeploymentSame {
+	if dep.same {
 		return controllerutil.OperationResultNone, nil
 	}
 	r.log.V(3).Info("Managed object found, patching",
 		logFieldObject, dRef.String(),
 		logFieldKind, kind)
-	deploymentAfterPatch, err := deploymentResourceInterface.Patch(ctx, dRef.Name, types.ApplyPatchType, dep.deploymentToPatch, metav1.PatchOptions{FieldManager: reconcilermanager.ManagerName, Force: &forcePatch})
+	deploymentAfterPatch, err := deploymentClient.Patch(ctx, dRef.Name, types.ApplyPatchType, dep.dataToPatch, metav1.PatchOptions{FieldManager: reconcilermanager.ManagerName, Force: &forcePatch})
 	if err != nil {
 		// Let the next reconciliation retry the patch operation for valid request.
 		if !apierrors.IsInvalid(err) {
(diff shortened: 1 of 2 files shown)


pkg/reconcilermanager/controllers/reconciler_base_test.go
```diff
@@ -513,13 +514,14 @@ func TestIsDeploymentSame(t *testing.T) {
...
-			// so we will allow the manual limits change.
-			isSame: true,
+			// We do not define some containers' resource limits to make the resource burstable
+			// Config Sync only manage the resource limits when user
+			// define the resource limits through our resource override API.
+			// When user try to reset the override limits, the declared deployment will not have
+			// the limits defined. To make sure user can always successfully
+			// reset the resource limits override, we should keep tracking the resource limits differences
+			// between declared and current deployment, even there is no resource limits defined in declared deployment.
+			isSame: false,
```
sdowell, line 524: `isSame` doesn't make it clear this is the expected value we are asserting for in testcases. I'd suggest something like `expected`, `expectedSame`, etc

pkg/reconcilermanager/controllers/reconciler_base_test.go
```diff
@@ -488,7 +489,7 @@ spec:
   progressDeadlineSeconds: 300 # different from declared
 `
 
-func TestIsDeploymentSame(t *testing.T) {
+func TestCompareDeploymentsToCreatePatchData(t *testing.T) {
 	declared := yamlToDeployment(t, declaredDeployment)
 	unObjCurrent := yamlToUnstructured(t, currentDeployment)
```
sdowell, line 494: ditto

pkg/reconcilermanager/controllers/reconciler_base_test.go
```diff
@@ -513,13 +514,14 @@ func TestIsDeploymentSame(t *testing.T) {
...
 			overrideLimit:     false,
 			manualChangeLimit: true,
-			// We do not define some container resource limits to make the resource burstable
-			// For these container resource limits, Config Sync only manage them when user
-			// define the resources limits through our resource override API.
-			// In this test case, the resource limits are not defined in original declared deployment,
-			// and user also not define the resource limits through our resource override API,
-			// so we will allow the manual limits change.
-			isSame: true,
+			// We do not define some containers' resource limits to make the resource burstable
```
sdowell, line 517: nit: grammar

suggested fix:
```golang
			// We do not define some containers' resource limits to make the resource burstable.
			// Config Sync only manages resource limits when a user
			// defines resource limits through the resource override API.
			// When a user resets resource limit overrides, the declared deployment will not have
			// resource limits defined. To make sure a user can always successfully
			// reset the resource limits override we must track differences in resource limits
			// between the declared and current deployment, even if no resource limits are defined
     			// in the declared deployment.
```

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -391,30 +391,6 @@ func adjustContainerResources(isAutopilot bool, declared, current *appsv1.Deploy
 	return resourcesChanged, nil
 }
 
-// processAllowlist add resources.limits to allowlist when renconciler-manager does not own this field
-func processAllowlist(isAutopilot bool, declared *appsv1.Deployment, allowList []string) []string {
```
sdowell, line 395: Why was this originally put in and why is it being removed now? Also, was this not covered by any unit tests?

xinnywinne: This is covered by unit test TestCompareDeploymentsToCreatePatchData. Since we change to SSA, we can only manage the fields that defined in our declared reconciler deployment. The previous assumption is that if we do not declare the resource limit fields, we should ignore the differences on this field, to avoid unnecessary PATCH. However this miss the case I mentioned in the description, when user reset the resource limits override, there is no resource limit fields defined in declared deployment, and we should not ignore the difference for this case. So this optimization logic should be removed.

sdowell: The reason I asked about unit tests is it seems like this was able to be removed without any testcases being changed/removed, but maybe I'm missing something. I see the `manualChangeResourceLimits` helper function was changed - is that related?

Autopilot manages resource `requests` but not `limits`, right?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -208,8 +208,8 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
 	dRef := client.ObjectKeyFromObject(declared)
 	kind := "Deployment"
 	forcePatch := true
-	deploymentResourceInterface := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace)
-	unObjCurrent, err := deploymentResourceInterface.Get(ctx, dRef.Name, metav1.GetOptions{})
+	deploymentClient := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace)
+	unObjCurrent, err := deploymentClient.Get(ctx, dRef.Name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
xinnywinne: Sure, thanks for the suggestion, changed it to currentDeploymentUnstructured

pkg/reconcilermanager/controllers/reconciler_base_test.go
xinnywinne: Done, changed to expectedSame

pkg/reconcilermanager/controllers/reconciler_base_test.go
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base_test.go
xinnywinne: Done

pkg/reconcilermanager/controllers/reconciler_base.go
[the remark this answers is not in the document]
sdowell: Thank you!
My point stands regarding using a consistent convention - if using `currentDeploymentUnstructured` then I'd suggest renaming the variables below to `currentGeneration` and `currentUID`

pkg/reconcilermanager/controllers/reconciler_base_test.go
```diff
@@ -601,8 +604,21 @@ func manualChangeResourceLimits(unDep *unstructured.Unstructured, isManaulChange
 	if !isManaulChange {
```
sdowell, line 604: I'd suggest moving this `if` statement up to the unit test to make it clear what this does when reading the test. It's generally a best practice to minimize logic in helper functions for unit tests. Ditto for the `isOverride` if statement in the above function

pkg/reconcilermanager/controllers/reconciler_base_test.go
```diff
@@ -601,8 +604,21 @@ func manualChangeResourceLimits(unDep *unstructured.Unstructured, isManaulChange
...
+	}
+	resources := &dep.Spec.Template.Spec.Containers[0].Resources
+	resources.Limits = corev1.ResourceList{}
+	resources.Limits[corev1.ResourceMemory] = resource.MustParse("400Mi")
+	resources.Limits[corev1.ResourceCPU] = resource.MustParse("400m")
+	resultUnDep, err = kinds.ToUnstructured(dep, core.Scheme)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert Deployment Object to Unstructured: %v", err)
 	}
 	return resultUnDep, nil
```
sdowell, line 623: I'd suggest renaming `resultUnDep` to something a bit more intuitive. Maybe `updatedDeployment`?

pkg/reconcilermanager/controllers/reconciler_base_test.go
```diff
@@ -488,97 +489,99 @@ spec:
...
-			isSame:            true,
+			expectedSame:      true,
 		},
 	}
 
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			declared := overrideResourceLimits(tc.declared, tc.overrideLimit)
-			unObjCurrent, err := manualChangeResourceLimits(tc.unObjCurrent, tc.manualChangeLimit)
+			currentDeploymentUnstructured, err := manualChangeResourceLimits(tc.current, tc.manualChangeLimit)
```
sdowell, line 580: nit: define a new variable inside this loop instead of reusing `currentDeploymentUnstructured` variable that is scoped above. Maybe just
```golang
if tc.overrideLimit {
  tc.declared = overrideResourceLimits(tc.declared)
}
if tc.manualChangeLimit {
  tc.current, err := manualChangeResourceLimits(tc.current)
  require.NoError(t, err)
}
```

pkg/reconcilermanager/controllers/reconciler_base.go
xinnywinne: There is one test case "non-Autopilot, no override limits, manual change limits" expectation changed. This optimization used to be standard cluster only, Autopilot cluster forces the resource limits to be same with resource requests, so the processAllowlist function directly return if it is Autopilot cluster.

sdowell: Oh I didn't notice that flipped from `true` -> `false`, my mistake. What was the original motivation for allowing drift on these fields for non-autopilot clusters?

pkg/reconcilermanager/controllers/reconciler_base.go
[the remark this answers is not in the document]
xinnywinne: changed to currentGeneration/UID

pkg/reconcilermanager/controllers/reconciler_base.go
xinnywinne: We did not define the resource limits for some containers to make the resource burstable, so if we do not define a field, we are not able to manage it.

pkg/reconcilermanager/controllers/reconciler_base_test.go
xinnywinne: Sure, moved the logic out of the function.

pkg/reconcilermanager/controllers/reconciler_base_test.go
xinnywinne: the idea is to keep the tc.declared(*Deployment) and tc.current(*Unstructured) same for each testcase. So I need to define new variables instead of reuse the tc.declared.

pkg/reconcilermanager/controllers/reconciler_base_test.go
xinnywinne: Done

PR #313, opened 2022-11-21
Update reconciler manager deployment status compute related logic to …

xinnywinne:
…align with the recent client switch on reconciler deployment upsert

We used to use typed client for reconciler deployment create and update, which can updates the deployment in-place to the latest value. After switching to dynamic client, it returns a new unstrcutured deployment object instead. Modify the status compute related logic to make sure reconciler-manager can still get the up-to-date deployment condition after upsert reconciler deployment.

Update rsync_controller_test setup*Reconciler function signature, return 3 variables instead of a struct.


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -329,7 +328,7 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		}
 	}
 
-	result, err := ComputeDeploymentStatus(deployObj)
+	result, err := kstatus.Compute(unstructuredDeployment)
```
karlkfi, line 331: Is ComputeDeploymentStatus still used, if not here?

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -258,10 +258,9 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 	}
 	rs.Status.Reconciler = reconcilerRef.Name
 
-	// Get the latest deployment to check the status.
-	// For other operations, upsertDeployment will have returned the latest already.
-	if op == controllerutil.OperationResultNone {
-		deployObj, err = r.deployment(ctx, reconcilerRef)
+	// Get the latest deployment to check the status if current deployment pointer is nil.
+	if unstructuredDeployment == nil {
```
karlkfi, line 262: Why would the return from upsertDeployment be nil?

pkg/reconcilermanager/controllers/reposync_controller.go
xinnywinne: We are only using it in nomostest git-server after this change

pkg/reconcilermanager/controllers/rootsync_controller.go
xinnywinne: After we compare the the current deployment and declared deployment, if we find out no PATCH is needed, current logic will return nil and no error. We are converting the unstructured deployment we get at the beginning to delete several fields and convert back. I think it is better/ safer to just get a deployment directly for the status compute in this case.

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: Does nomostest git-server need it? or should it just use kstatus.Compute too?

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: It's confusing (hard to read) to have a nil object returned mean there was no diff. That doesn't follow any of the existing client method conventions. What if you keep returning the operation too?

xinnywinne: Sure!

pkg/reconcilermanager/controllers/reposync_controller.go
xinnywinne: kstatus.Compute can be used directly in  git-server, so removed the ComputeDeploymentStatus and related unit test

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -288,7 +288,7 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 	mut := r.mutationsFor(ctx, rs, containerEnvs)
 
 	// Upsert Namespace reconciler deployment.
-	deployObj, op, err := r.upsertDeployment(ctx, reconcilerRef, labelMap, mut)
+	unstructuredDeployment, op, err := r.upsertDeployment(ctx, reconcilerRef, labelMap, mut)
```
karlkfi, line 291: Do we need to rename this or can we just keep using deployObj? I don't think we need to say it's unstructured in the name if it's always unstructured in this method.

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -75,6 +75,7 @@ const (
 var filesystemPollingPeriod time.Duration
 var hydrationPollingPeriod time.Duration
 var nsReconcilerName = core.NsReconcilerName(reposyncNs, reposyncName)
+var reconcilerDeploymentReplicaCount int32 = 1
```
karlkfi, line 78: What's this for? I don't see it used.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -258,29 +256,29 @@ func (r *reconcilerBase) createOrPatchDeployment(ctx context.Context, declared *
...
-			return controllerutil.OperationResultNone, err
+			return nil, controllerutil.OperationResultNone, err
 		}
 		data, err := json.Marshal(declared)
 		if err != nil {
-			return controllerutil.OperationResultNone, fmt.Errorf("failed to marshal declared deployment object to byte array: %w", err)
+			return nil, controllerutil.OperationResultNone, fmt.Errorf("failed to marshal declared deployment object to byte array: %w", err)
 		}
-		_, err = deploymentClient.Patch(ctx, dRef.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: reconcilermanager.ManagerName, Force: &forcePatch})
+		deploymentAfterPatch, err = deploymentClient.Patch(ctx, dRef.Name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: reconcilermanager.ManagerName, Force: &forcePatch})
```
karlkfi, line 273: Since this is an Apply and not just a Patch, and this method only deals with deployments, how about `appliedObj` instead of `deploymentAfterPatch`. It's shorter and less confusing when the apply performs a create, not just a patch.

pkg/reconcilermanager/controllers/reposync_controller_test.go
xinnywinne: It is used in reposync_controller_test(this test), it was originally defined in the deployment condition unit test that I just deleted

pkg/reconcilermanager/controllers/reconciler_base.go
xinnywinne: Sure, Done!

pkg/reconcilermanager/controllers/reposync_controller.go
xinnywinne: Sure, keep it as deployObj.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -410,9 +408,8 @@ func keepCurrentContainerResources(declared, current *appsv1.Deployment) bool {
 }
 
 // deployment returns the deployment from the server
-func (r *reconcilerBase) deployment(ctx context.Context, dRef client.ObjectKey) (*appsv1.Deployment, error) {
-	depObj := &appsv1.Deployment{}
-	depUnObj, err := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace).Get(ctx, dRef.Name, metav1.GetOptions{})
+func (r *reconcilerBase) deployment(ctx context.Context, dRef client.ObjectKey) (*unstructured.Unstructured, error) {
+	unstructuredDeployment, err := r.dynamicClient.Resource(kinds.DeploymentResource()).Namespace(dRef.Namespace).Get(ctx, dRef.Name, metav1.GetOptions{})
```
karlkfi, line 412: nit: This could remain depObj or be deployObj to be consistent with the other method. But it's not super important.

karlkfi (approved):
/lgtm

PR #314, opened 2022-11-22
Add retries to reduce flakiness

haiyanmeng:


nan-yu (approved):
/lgtm

e2e/testcases/multi_sync_test.go
```diff
@@ -484,10 +484,13 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
 	nt.WaitForRepoSyncs()
 
 	nt.T.Logf("Ensure the Role is managed by the other Root repo %s", rootSync2)
-	err = nt.Validate("pods", testNs, &rbacv1.Role{},
-		roleHasRules(rootPodRole().Rules),
-		nomostest.IsManagedBy(nt, declared.RootReconciler, rootSync2))
-	if err != nil {
+	// The pod role may be deleted from the cluster after it was removed from the `root-sync` Root repo.
```
karlkfi, line 487: What if we add the abandon annotation instead of just deleting the object?
That's how we should recommend customers do hand-off to resolve conflicts without deleting the object.
/cc @nan-yu @haiyanmeng

haiyanmeng: @karlkfi , which abandon annotation you are referring to?

nan-yu: I think Karl meant the `configmanagement.gke.io/managed:disabled` annotation. Yes, that sounds a better solution to resolve the conflict. We should add it to the documentation.

PR #315, opened 2022-11-22
[WIP] Add test for GIT_SSL_CAINFO env var removal

karlkfi:
This confirms that the new SSA used to update the reconciler Deployment correctly deletes the GIT_SSL_CAINFO env var (as long as no one else has modified it).

### DIFF
--- a/e2e/testcases/private_cert_secret_test.go
+++ b/e2e/testcases/private_cert_secret_test.go
@@ -280,6 +280,25 @@ func TestCACertSecretRefV1Beta1(t *testing.T) {
 	if err != nil {
 		nt.T.Fatal(err)
 	}
+
+	// Set RepoSync to use HTTPS with no auth and SSL verification disabled.
+	// This validates that the GIT_SSL_CAINFO env var can be removed.
+	// TODO: switch to HTTP instead of HTTPS, when supported by the local git server nginx proxy. That way we're not testing NoSSLVerify in the CACertSecretRef test.
+	repoSyncBackend.Spec.Git.Repo = repoSyncHTTPS
+	repoSyncBackend.Spec.Git.Auth = "none"
+	repoSyncBackend.Spec.Git.NoSSLVerify = true
+	repoSyncBackend.Spec.Git.SecretRef = nil
+	repoSyncBackend.Spec.Git.CACertSecretRef = nil
+	nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(backendNamespace, configsync.RepoSyncName), repoSyncBackend)
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update backend RepoSync use HTTPS with no auth")
+	nt.WaitForRepoSyncs()
+	err = nt.Validate(reconcilerName, v1.NSConfigManagementSystem, &appsv1.Deployment{},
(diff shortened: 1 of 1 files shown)


karlkfi:
Gonna wait to merge this until @sdowell adds support for http access to git. That way we won't pollute the CACertSecretRef test with needing to use NoSSLVerify (which has its own test).

e2e/testcases/private_cert_secret_test.go
```diff
@@ -280,6 +280,25 @@ func TestCACertSecretRefV1Beta1(t *testing.T) {
 	if err != nil {
 		nt.T.Fatal(err)
 	}
+
+	// Set RepoSync to use HTTPS with no auth and SSL verification disabled.
+	// This validates that the GIT_SSL_CAINFO env var can be removed.
```
sdowell, line 285: I believe line 266 should already be verifying this. It unsets `caCertSecretRef` and then checks that the env var has been removed

karlkfi: Technically you're correct, but it also doesn't sync. I wanted one where it actually worked.

sdowell: Seems like you could add a `DeploymentMissingEnvVar` predicate to line 279 then, right?

karlkfi: Fair enough.

That said, I don't understand how 278 works without the CACertSecretRef. The SecretRef is a private key to authenticate the user, right? How does it validate the server cert without a CA file?

karlkfi: Does git@ scheme not validate the server cert by default??

sdowell: caCertSecretRef is only used for verifying the authenticity of the server certificate over HTTPS. SSH is a different protocol

karlkfi: Does SSH not validate the server cert? I thought it did.

Or is the git server configured to use a known CA when using SSH?

sdowell:
@karlkfi This should add support for HTTP, although I'm still not sure whether it's needed https://github.com/GoogleContainerTools/kpt-config-sync/pull/316

e2e/testcases/private_cert_secret_test.go
sdowell: Certificate authorities aren't involved with SSH protocol. They are used to sign certificates used in HTTPS. Authenticity over SSH is established using the public/private key pair

karlkfi: > In the connection setup phase, the SSH server authenticates itself to the client by providing its public key. This allows the SSH client to verify that it is actually communicating with the correct SSH server (instead of an attacker that could be posing as the server).

Weird... How can you trust that the SSH server is who they say they are if you get the public key from them instead of a trusted third party? In TLS, the server gives you its public key, but it has to have been signed by a CA you trust. It sounds like maybe SSH (or at least git?) doesn't have that layer of security by default, which is... really surprising.

mikebz:
curious if this is still WIP or if it's even relevant.  If it is consider closing the PR and keeping the private branch or converting to draft.

karlkfi:
Obsolete

PR #319, opened 2022-12-01
Fix over-agressive role binding subject deletion

karlkfi:
- The old code for subject deletion used a List of RSyncs to decide what all the subjects should be in the [Cluster]RoleBinding for the reconciler ServiceAccount. This aggresively removes subjects when their RSync is deleted. However, this caused the upsert to remove subjects when it only needed to add subjects, which can cause reconcilers to be de-authorized before they're deleted.
- This change splits the update code into add and remove code. So the subject will only ever be removed if the RSync is already NotFound, and only by the Reconcile loop specific to that RSync.
- This should unblock the Finalizer to delete managed resources even if the deletion timestamp is already set on the RSync (which is the only time the finalizer runs anyway).

### DIFF
--- a/pkg/reconcilermanager/controllers/garbage_collector.go
+++ b/pkg/reconcilermanager/controllers/garbage_collector.go
@@ -16,7 +16,6 @@ package controllers
 
 import (
 	"context"
-	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -27,7 +26,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
-	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/reconcilermanager"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -64,7 +62,7 @@ func (r *RepoSyncReconciler) cleanupNSControllerResources(ctx context.Context, r
 		return err
 	}
 	// rolebinding
-	if err := r.deleteRoleBinding(ctx, rsList, rsKey.Namespace, reconcilerRef.Namespace); err != nil {
+	if err := r.deleteRoleBinding(ctx, reconcilerRef, rsKey); err != nil {
 		return err
 	}
 	// secret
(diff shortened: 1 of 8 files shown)


pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -131,17 +129,16 @@ func (r *RepoSyncReconciler) deleteServiceAccount(ctx context.Context, reconcile
...
-		return errors.Wrapf(err, "failed to get the RoleBinding object %s/%s", rsNamespace, rbName)
+	if err := r.client.Get(ctx, rbKey, rb); err != nil {
+		return errors.Wrapf(err, "failed to get the RoleBinding object %s", rbKey)
 	}
-	if err := r.updateRoleBindingSubjects(rb, rsList, reconcilerNamespace); err != nil {
-		return errors.Wrapf(err, "failed to update subjects of the RoleBinding object %s/%s", rsNamespace, rbName)
+	rb.Subjects = removeSubject(rb.Subjects, r.serviceAccountSubject(reconcilerRef))
+	if len(rb.Subjects) == 0 {
+		// Delete the whole RB
+		return r.cleanup(ctx, rbKey, kinds.RoleBinding())
```
sdowell, line 141: Is it necessary to delete the RoleBinding?

pkg/reconcilermanager/controllers/permissions.go
```diff
@@ -25,11 +29,43 @@ func rolereference(name, kind string) rbacv1.RoleRef {
...
+	}
+	return -1
+}
+
+// addSubject adds a subject to the subject list.
+// Subjects are sorted, if modified.
+func addSubject(subjects []rbacv1.Subject, subject rbacv1.Subject) []rbacv1.Subject {
+	if i := findSubjectIndex(subjects, subject); i < 0 {
+		subjects = append(subjects, subject)
+		if len(subjects) > 0 {
```
sdowell, line 56: Did you mean to do `> 1`? I assume after an `append` the slice should always have nonzero length

pkg/reconcilermanager/controllers/permissions.go
```diff
@@ -25,11 +29,43 @@ func rolereference(name, kind string) rbacv1.RoleRef {
...
+			})
+		}
+	}
+	return subjects
+}
+
+// removeSubject removes the subject from the subject list.
+func removeSubject(subjects []rbacv1.Subject, subject rbacv1.Subject) []rbacv1.Subject {
+	if i := findSubjectIndex(subjects, subject); i >= 0 {
+		subjects = append(subjects[:i], subjects[i+1:]...)
```
sdowell, line 68: Does this error if the element is the last in the slice (from the `i+1` index)?

sdowell (commented):
Mostly LGTM with a few comments - I like the refactor with the helper functions

pkg/reconcilermanager/controllers/garbage_collector.go
karlkfi: This is the existing behavior when all the subjects are removed. Previously it was just implemented after the LIST of RSyncs, instead of after the GET of the RoleBinding.

pkg/reconcilermanager/controllers/permissions.go
karlkfi: Good catch. There's no point sorting an array of size 1.

pkg/reconcilermanager/controllers/permissions.go
karlkfi: unfortunately, yes... I'll have to fix that.

sdowell: on that note, it wouldn't hurt to have some unit tests for these helper functions

pkg/reconcilermanager/controllers/permissions.go
karlkfi: fixed.

pkg/reconcilermanager/controllers/permissions.go
karlkfi: fixed and added unit tests for all the new subject list functions. Found another bug with sorting and fixed it.

PR #320, opened 2022-12-01
fix the override reconcile timeout test

xinnywinne:
TestOverrideReconcileTimeout is failing on Autopilot cluster, add logic to scale up the Autopilot cluster before creating the pod for test.

Pull in resourcegroup package and update klog version from v2.60.1 to
v2.70.1


e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -40,6 +41,19 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 	// Only actuation should succeed, reconcile should time out.
 	nt.MustMergePatch(rootSync, `{"spec": {"override": {"reconcileTimeout": "30s"}}}`)
 	nt.WaitForRepoSyncs()
+	// Scale up the Autopilot cluster to prevent the pod used for test from failing to schedule
+	// TODO: remove this after resource group controller updated to update the reconcile status
```
karlkfi, line 45: Won't this still be needed when the reconcile status is fixed?

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -40,6 +41,19 @@ func TestOverrideReconcileTimeout(t *testing.T) {
...
+	if nt.IsGKEAutopilot {
+		nt.MustKubectl("apply", "-f", "../testdata/deployment-helloworld.yaml")
+		_, err := nomostest.Retry(nt.DefaultWaitTimeout, func() error {
```
karlkfi, line 48: You can use the new WatchForObject with the StatusEquals Current Precondition.

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -122,7 +131,7 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 	} else {
 		expectActuationStatus := "Succeeded"
 		expectReconcileStatus := "Succeeded"
-		err = verifyObjectStatus(u, idToVerify, expectActuationStatus, expectReconcileStatus)
+		_, err := nomostest.Retry(nt.DefaultWaitTimeout, func() error { return verifyObjectStatus(u, idToVerify, expectActuationStatus, expectReconcileStatus) })
```
karlkfi, line 134: This retry doesn't make any sense. The input is always the same object that was retrieved in the previous Get.

If you want to do this right, I suggest converting it to use WatchForObject and then make a new Precondition factory function that returns a Precondition that checks the ResourceGroup status.resourceStatuses contents. Re-writing  verifyObjectStatus to use a ResourceGroup instead of an Unstructured will make it simpler to read too.

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -85,7 +99,7 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 	} else {
 		expectActuationStatus := "Succeeded"
 		expectReconcileStatus := "Timeout"
```
karlkfi, line 101: This expectation will pass today, but after the RGC bug is fixed, the reconcile status may no still be Timeout. Inc fact, the longer you wait, the less likely it will still be Timeout. So this might be a race condition.

karlkfi (changes requested):
Left some comments.

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: Sounds good, changed to use the new WatchForObject.

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: Sure, changed to use WatchForObject and use ResourGroup in Precondition function

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: What wait time are you discussing here?  The reconcile status is updated as timeout by the Config Sync reconciler at the end of sync operation. WaitforRepoSyncs() only waits until the sync operation finishes, and we check the reconcile status right after it. After the RGC bug is fixed, the resourcegroup controller will also update the reconcile status. For this "timeout" case, when the resourcegroup controller changes the reconcile status from timeout to current, the reconcile status check is already finished in the test.

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -40,6 +41,14 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 	// Only actuation should succeed, reconcile should time out.
 	nt.MustMergePatch(rootSync, `{"spec": {"override": {"reconcileTimeout": "30s"}}}`)
 	nt.WaitForRepoSyncs()
+	// Scale up the Autopilot cluster to prevent the pod used for test from failing to schedule
+	if nt.IsGKEAutopilot {
+		nt.MustKubectl("apply", "-f", "../testdata/deployment-helloworld.yaml")
+		nomostest.WatchForObject(nt, kinds.Deployment(), "hello-world", "default",
+			[]nomostest.Predicate{nomostest.StatusEquals(nt, kstatus.CurrentStatus)})
```
karlkfi, line 48: nit: for style consistency, I have previously wrapped the lines so each precondition gets its own line. it makes it a little easier to read.
ex: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/namespaces_test.go#L720-L723

karlkfi (approved):
/lgtm

e2e/testcases/override_reconcile_timeout_test.go
karlkfi: Ok, I think if the RGC preserves the Timeout status instead of reverting to InProgress/Pending, we should be ok for up to 5s. If it's flakey afterwards tho, we may need to fix it later.

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -40,6 +41,15 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 	// Only actuation should succeed, reconcile should time out.
 	nt.MustMergePatch(rootSync, `{"spec": {"override": {"reconcileTimeout": "30s"}}}`)
 	nt.WaitForRepoSyncs()
+	// Scale up the Autopilot cluster to prevent the pod used for test from failing to schedule
```
karlkfi, line 44: Maybe explain how this works a little more:

// Pre-provision a low priority workload to force the cluster to scale up. Later, when the real workload is being scheduled, if there's no more resources available (common on auto-pilot clusters, which are optimized for utilization), this low priority workload will be evicted to make room for the new normal priority workload.

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -40,6 +41,15 @@ func TestOverrideReconcileTimeout(t *testing.T) {
...
+	if nt.IsGKEAutopilot {
+		nt.MustKubectl("apply", "-f", "../testdata/low-priority-pause-deployment.yaml")
+		nomostest.WatchForObject(nt, kinds.Deployment(), "pause-deployment", "default",
+			[]nomostest.Predicate{
+				nomostest.StatusEquals(nt, kstatus.CurrentStatus),
+			})
+	}
```
karlkfi, line 51: Should we add a Cleanup step here to delete the pause-deployment?

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -54,10 +64,10 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 		ReadinessProbe: &corev1.Probe{
 			ProbeHandler: corev1.ProbeHandler{
 				TCPSocket: &corev1.TCPSocketAction{
-					Port: intstr.FromInt(8081), // wrong port will never be ready
+					Port: intstr.FromInt(8080),
 				},
 			},
-			InitialDelaySeconds: 5,
+			InitialDelaySeconds: 60, // make initial delay longer than current reconcile timeout 30s.
```
karlkfi, line 70: Explain why:

// This will cause the applier to exit with a timeout, but the workload will still become available afterwards.

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -68,88 +78,55 @@ func TestOverrideReconcileTimeout(t *testing.T) {
...
-		expectReconcileStatus := "Timeout"
-		err = verifyObjectStatus(u, idToVerify, expectActuationStatus, expectReconcileStatus)
-		if err != nil {
-			nt.T.Fatal(err)
-		} else {
-			nt.T.Logf("object %q has actuation and reconcile status as expected", namespaceName)
-		}
-	}
+	// Override reconcileTimeout to 5m, namespace actuation should succeed, namespace reconcile should succeed.
+	nt.MustMergePatch(rootSync, `{"spec": {"override": {"reconcileTimeout": "5m"}}}`)
```
karlkfi, line 89: This is likely going to be flakey.

Unfortunately, we don't actually have any signal that the spec change has taken affect. This will be fixed when we add `syncedGeneration` or similar to the RSync status.

In the mean time, you might want to add a WatchForObject with a predicate that ensures the `observedGeneration` has been updated to match the new value, before calling WaitForRepoSyncs. That way we know the reconciler-manager at least saw the new reconcileTimeout.

To do so, you'll also need to Patch the object directly, instead of using MustMergePatch. That way you can read the new generation out of the object from the result of the Patch call. Maybe make a local function to make this readable.

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -68,88 +78,55 @@ func TestOverrideReconcileTimeout(t *testing.T) {
...
+	nt.WaitForRepoSyncs()
 
 	nt.RootRepos[configsync.RootSyncName].Remove("acme/pod-1.yaml")
-	nt.RootRepos[configsync.RootSyncName].CommitAndPush(fmt.Sprintf("Remove pod/%s (never ready)", pod1Name))
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush(fmt.Sprintf("Remove pod/%s", pod1Name))
 	nt.WaitForRepoSyncs()
 
 	// Verify pod is deleted.
 	// Otherwise deletion of a managed resource will be denied by the webhook.
 	nt.MustKubectl("delete", "pods", pod1Name, "-n", namespaceName, "--ignore-not-found")
```
karlkfi, line 98: This seems like a confusing way to validate deletion, especially if the webhook isn't enabled by default.

Don't we have a ValidateNotFound or similar?

e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -68,88 +78,55 @@ func TestOverrideReconcileTimeout(t *testing.T) {
...
-		err = verifyObjectStatus(u, idToVerify, expectActuationStatus, expectReconcileStatus)
-		if err != nil {
-			nt.T.Fatal(err)
-		} else {
-			nt.T.Logf("object %q has actuation and reconcile status as expected", namespaceName)
-		}
-	}
+	expectActuationStatus = "Succeeded"
+	expectReconcileStatus = "Succeeded"
+	nomostest.WatchForObject(nt, kinds.ResourceGroup(), "root-sync", "config-management-system",
```
karlkfi, line 106: Do we need to Watch or Wait here?

If the sync is complete, the pod should already be scheduled and available, right?

You should be able to just use Validate (instead of WatchForObject) with the same predicates.

e2e/testdata/low-priority-pause-deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+globalDefault: false
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: pause-deployment
+  labels:
+    app: pause
+spec:
+  replicas: 3
```
karlkfi, line 29: Do we need 3 replicas? Is 1 good enough?

xinnywinne: Yes, 1 is good enough.

e2e/testdata/low-priority-pause-deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+    metadata:
+      labels:
+        app: pause
+    spec:
+      containers:
+      - name: pause
+        image: registry.k8s.io/pause:3.7
+        ports:
+        - containerPort: 80
+      priorityClassName: low-priority
```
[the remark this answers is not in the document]
xinnywinne: Done

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: Yes, Validate is good enough here, we do not need extra wait for resource status check here since we have WaitForRepoSyncs

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: Changed to ValidateNotFound

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: Done.

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: yes, added a cleanup step

e2e/testcases/override_reconcile_timeout_test.go
xinnywinne: Once we change the field spec.override.reconcileTimeout in Rsync, a new reconcile process start, the reconciler pod will be recreated and the Rsync status will change to be:       
      message: 'Available: 0/1'
      reason: Deployment
      status: "True"
      type: Reconciling
And the WaitForRepoSyncs() will detect the the deployment reconciling status is true, it will wait until the deployment is ready again to return. So this Patch should not be flakey.

karlkfi: You're assuming that the Reconciling condition will be set to True by the reconciler-manager before  WaitForRepoSyncs does its first GET on the RSync. But it's possible the reconciler-manager is slow or delayed, which could cause the test GEt to happen before the reconicler-manager has updated the reconciler and RSync. That's a race condition.

The way to prove that the reconciler-manager has seen the desired RSync spec and that it's updated the Reconciling condition is to check that the RSync's status.observedGeneration has been updated to be greater than or equal to the desired generation (the new generation returned from the RSync UPDATE call).

karlkfi (approved):
/lgtm

Thanks!

PR #322, opened 2022-12-06
Switch to docker buildx for faster builds

karlkfi:
- buildx is demonstratably faster than the current build behavior, and is even faster than just seperating out the multi-stage builds into thier own Dockerfiles.


sdowell:
LGTM, but let me confirm this is supported by our build environments
/lgtm
/hold

sdowell:
Looks like it should work, unholding
/unhold
/approve

PR #325, opened 2022-12-12
Add formatted otel configmap to filter doc

tiffanny29631:
This change adds a section of a sample ConfigMap for users reference.

### DIFF
--- a/docs/custom-metric-filter.md
+++ b/docs/custom-metric-filter.md
@@ -24,10 +24,10 @@ when configuring pipelines:
 1. `otel-collector`
 
    This is the default configuration that Config Sync deploys in a non-GKE environment, which only exports the metrics through Prometheus exporter
-2. `otel-collector-googlecloud` (or `otel-collector-stackdriver` before Config Sync v1.12.0)
+1. `otel-collector-googlecloud` (or `otel-collector-stackdriver` before Config Sync v1.12.0)
 
    Config Sync automatically deploys this configuration when Workload Identity is configured, usually on a GKE cluster. 
-3. `otel-collector-custom`
+1. `otel-collector-custom`
 
     Users can apply this ConfigMap and configure their own metric pipelines.
     Use this ConfigMap instead of modifying the others in-place, to ensure modifications don't get reverted by Config Sync.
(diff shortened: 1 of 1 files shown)


docs/custom-metric-filter.md
```diff
@@ -83,11 +83,16 @@ in Config Sync.
     ```
```
haiyanmeng, line 83: Please mention that in 1.14.0, the output of `kubectl get cm otel-collector-googlecloud -n config-management-monitoring -o yaml` has formatting issue.

docs/custom-metric-filter.md
```diff
@@ -147,20 +152,168 @@ in Config Sync.
...
 
-1. Check everything is up an running
+6. Check everything is up an running
     
     - `kubectl get all -n config-management-monitoring` should show all objects running with no error.
     - `kubectl get cm -n config-management-monitoring` should include a new ConfigMap named `otel-collector-custom`
+
+## Sample Configmap
+
+```
```
sdowell, line 175: \`\`\`yaml

docs/custom-metric-filter.md
```diff
@@ -147,20 +152,168 @@ in Config Sync.
            exporters: [googlecloud/kubernetes]
          ```
 
-1. Apply the config map
+4. Apply the config map
```
sdowell, line 155: Why did the indexing change?

tiffanny29631: They were from Intellij MD auto save, it converted them in-place. Fixed.

docs/custom-metric-filter.md
tiffanny29631: Done

docs/custom-metric-filter.md
tiffanny29631: Done, added note before the command to separate 1.14.0

docs/custom-metric-filter.md
```diff
@@ -164,3 +171,151 @@ This can be done by deleting the Otel Collector deployment.
     
     - `kubectl get all -n config-management-monitoring` should show all objects running with no error.
     - `kubectl get cm -n config-management-monitoring` should include a new ConfigMap named `otel-collector-custom`
+
+## Sample Configmap
```
karlkfi, line 175: Won't having this here mean it needs to be updated every time the real one is updated? Won't that cause it to quickly become out of date? Would it be better to just link to the real one?

tiffanny29631: Forgot about this PR.. The ConfigMap won't go stale, once one with name `otel-collector-custom` exists the otel collector deployment will always load the custom one. I'm not sure what the `real one` is implying, could you elaborate?

karlkfi: I mean, this "sample" will drift from what's in the code as the default config. So if people copy this one, it might not match what's in the code.

tiffanny29631: Not sure if this change is necessary anymore, it was for users with version v1.14.1, when they want to modify their ConfigMap they will find the format is all off. This problem is fixed in later versions and I concur adding a copy of config here would cause drift. The link to the actual config is given on line70 as well.

If @haiyanmeng also agrees I'll close this PR.

tiffanny29631:
This change only adds a copy of Otel Collector configuration to the doc which brings extra work to maintain the alignment with code, what we should do is split Otel Collector with GoogleCloud exporter to a separate yaml file. Closing this PR.

PR #326, opened 2022-12-14
Export SLI-related metrics to Google Cloud Monitoring

haiyanmeng:
Config Sync SLIs:
https://cloud.google.com/anthos-config-management/docs/how-to/config-sync-slis

### DIFF
--- a/pkg/metrics/otel.go
+++ b/pkg/metrics/otel.go
@@ -90,6 +90,10 @@ processors:
         match_type: regexp
         metric_names:
           - reconciler_errors
+          - apply_duration_seconds
+          - reconcile_duration_seconds
+          - rg_reconcile_duration_seconds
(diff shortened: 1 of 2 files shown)


tiffanny29631 (commented):
So these metrics are not yet available in 1.14.0, but the how-to doc is published?

haiyanmeng:
> So these metrics are not yet available in 1.14.0, but the how-to doc is published?

`how-to doc`: are you referring to https://cloud.google.com/anthos-config-management/docs/how-to/config-sync-slis? This only includes Prometheus examples for now. You can add the cloud monitoring examples after you finish converting the resource attributes to labels.

tiffanny29631:
@haiyanmeng yes that is the one I'm referring to, thanks for the clarification!

/lgtm
/approve

PR #327, opened 2022-12-14
Update go-junt-report version and config

nan-yu:
go-junit-report v2.0.0 added support for -subtest-mode flag to exclude or ignore results of subtest parent tests:
https://g3doc.corp.google.com/third_party/golang/github_com/jstemmer/go_junit_report/v/v2/README.md#changelog.

### DIFF
--- a/build/test-e2e-go/e2e.sh
+++ b/build/test-e2e-go/e2e.sh
@@ -30,7 +30,7 @@ echo "Tests took $(( end_time - start_time )) seconds"
 
 if [ -d "/logs/artifacts" ]; then
   echo "Creating junit xml report"
-  cat test_results.txt | go-junit-report > /logs/artifacts/junit_report.xml
+  cat test_results.txt | go-junit-report --subtest-mode=exclude-parents > /logs/artifacts/junit_report.xml
 fi
 
 exit $exit_code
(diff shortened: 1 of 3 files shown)


build/test-e2e-go/gke/Dockerfile
```diff
@@ -56,7 +56,7 @@ COPY --from=gcloud-install /usr/bin/kubectl /opt/gcloud/google-cloud-sdk/bin/kub
 ENV PATH /opt/gcloud/google-cloud-sdk/bin:$PATH
 
 # Get go-junit-report
-RUN go install github.com/jstemmer/go-junit-report@v1.0.0
+RUN go install github.com/jstemmer/go-junit-report/v2@latest
```
sdowell, line 59: Are we not able to pin to a semver?

build/test-e2e-go/e2e.sh
```diff
@@ -30,7 +30,7 @@ echo "Tests took $(( end_time - start_time )) seconds"
 
 if [ -d "/logs/artifacts" ]; then
   echo "Creating junit xml report"
-  cat test_results.txt | go-junit-report > /logs/artifacts/junit_report.xml
+  cat test_results.txt | go-junit-report --subtest-mode=exclude-parents > /logs/artifacts/junit_report.xml
```
sdowell, line 33: This seems to make sense to me. I expect this means the parent results will disappear from the testgrid and we will ignore past failures, correct? Whereas if we used the `ignore-parent-results` option the parent result would show up as green?

build/test-e2e-go/gke/Dockerfile
nan-yu: changed to v2.0.0

PR #328, opened 2022-12-15
Refactor the Applier to make it easier to test

karlkfi:
- Move ClientSet construction out of the Applier. This removes the need to inject the clientSetFunc, configFlags, and statusMode.
- Renamed KptApplier to Applier, to allow KptApplier to be used for the interface that matches the actual kpt/cli-utils applier.
- Rename the Applier to applier to allow Applier to be used for the publicly exposed interface.
- Move handling of protected namespace deletion from processPruneEvent into handlePruneSkippedEvent, to make it easier to read.
- Moved the methods for handling disabled objects from the ClientSet into the applier itself, to simplify the ClientSet. Now the ClientSet only wraps clients and configs, and doesn't have any methods of its own.
- Remove the resourceClient. It was just a thin wrapper that was only being used in one place. So I moved the code directly into the disableObject function.


nan-yu (approved):
/lgtm

PR #331, opened 2022-12-21
Fake client fixes

karlkfi:
- Fix ID lookup to work for typed objects without GVK set. Use the scheme to look up the GVK by struct type.
- Add support for auto-incrementing Generation when the spec changes (technically any change that isn't status or metadata, not just the spec field)
- Fix status update to log the diff correctly
- Fix Update to copy UID, ResourceVersion, and Generation changes back to the input object.
- Make fake.Client.Check ignore Timestamps, so conditions can be easily compared.
- Fix Update to error if the specified UID or ResourceVersion is no longer current.
- Update impacted test expectations.

### DIFF
--- a/pkg/reconcilermanager/controllers/otel_controller_test.go
+++ b/pkg/reconcilermanager/controllers/otel_controller_test.go
@@ -34,9 +34,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// These constants hard-code the hash of the ConfigMaps that configure the
+// otel-collector. This value is used to check the `configsync.gke.io/configmap`
+// annotation value, which is populated by the reconciler-manager to ensure that
+// the otel-collector deployment is updated (and pods replaced) any time the
+// ConfigMap changes. The hash will change any time the ConfigMap changes. So
+// you'll need to update these values any time you make a deliberate change to
+// the config or its ConfigMap metadata.
 const (
-	depAnnotationGooglecloud = "c7420e47ee5c361e221cbe0896d1bea0"
-	depAnnotationCustom      = "7021ec2418429875d452e1cd5455724a"
+	// depAnnotationGooglecloud is the expected hash of the GCP/GKE-specific
+	// otel-collector ConfigMap.
+	// See `CollectorConfigGooglecloud` in `pkg/metrics/otel.go`
+	// Used by TestOtelReconcilerGooglecloud.
+	depAnnotationGooglecloud = "48a5b088ea2af4e9bc259230f88cafd3"
+	// depAnnotationGooglecloud is the expected hash of the custom
+	// otel-collector ConfigMap test artifact.
+	// Used by TestOtelReconcilerCustom.
+	depAnnotationCustom = "d166bfb4bea41bdc98b5b718e6c34b44"
 )
 
 func setupOtelReconciler(t *testing.T, objs ...client.Object) (*syncerFake.Client, *OtelReconciler) {
@@ -56,7 +70,7 @@ func TestOtelReconciler(t *testing.T) {
 		metrics.MonitoringNamespace,
 		metrics.OtelCollectorName,
 		map[string]string{"otel-collector-config.yaml": ""},
-		core.ResourceVersion("1"),
+		core.UID("1"), core.ResourceVersion("1"), core.Generation(1),
 	)
 	reqNamespacedName := namespacedName(metrics.OtelCollectorName, metrics.MonitoringNamespace)
 	fakeClient, testReconciler := setupOtelReconciler(t, cm, fake.DeploymentObject(core.Name(metrics.OtelCollectorName), core.Namespace(metrics.MonitoringNamespace)))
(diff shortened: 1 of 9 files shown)


pkg/syncer/syncertest/fake/client.go
```diff
@@ -356,15 +366,27 @@ func (c *Client) Delete(_ context.Context, obj client.Object, opts ...client.Del
...
 		return newNotFound(id)
 	}
 
-	klog.V(5).Infof("Deleting %T %s (ResourceVersion: %q): %s",
-		cachedObj, client.ObjectKeyFromObject(cachedObj), cachedObj.GetResourceVersion(), log.AsJSON(cachedObj))
+	if obj.GetUID() != "" && obj.GetUID() != cachedObj.GetUID() {
+		return newConflictingUID(id, obj.GetResourceVersion(), cachedObj.GetResourceVersion())
+	}
+	// Copy cached UID to updated object
+	obj.SetUID(cachedObj.GetUID())
```
xinnywinne, line 380: Why do we need to set the obj UID here?

karlkfi: The fake client is simulating the apiserver, and the apiserver sets the UID on newly created objects. That way you can confirm that it's the same instance of the object you expected. If it's changed, that means the object was deleted and re-created.

So for creates we initialize it, and for updates we make sure the UID stays the same as what was previously created (from the in-memory cache, in this case).

karlkfi: Oh, and when you Create, Update, or Delete an object with the normal client, it copies the updated values back to the input object, so you can read them, since there's no return object. That's what it's doing here.

pkg/reconcilermanager/controllers/otel_controller_test.go
```diff
@@ -35,8 +35,8 @@ import (
 )
 
 const (
-	depAnnotationGooglecloud = "c7420e47ee5c361e221cbe0896d1bea0"
-	depAnnotationCustom      = "7021ec2418429875d452e1cd5455724a"
+	depAnnotationGooglecloud = "48a5b088ea2af4e9bc259230f88cafd3"
```
sdowell, line 38: Can we document the meaning of these consts?

pkg/syncer/syncertest/fake/dynamic_client.go
```diff
@@ -219,6 +225,13 @@ func (dc *DynamicClient) patch(action clienttesting.Action) (bool, runtime.Objec
 	if err := incrementResourceVersion(patchedObj); err != nil {
 		return true, nil, fmt.Errorf("failed to increment resourceVersion: %w", err)
 	}
+	if found {
+		if err = updateGeneration(cachedObj, patchedObj, dc.scheme); err != nil {
+			return true, nil, fmt.Errorf("failed to update generation: %w", err)
+		}
+	} else {
+		patchedObj.SetGeneration(1)
```
sdowell, line 233: `initGeneration`?

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -492,6 +498,9 @@ func TestUpdateNamespaceReconcilerWithOverride(t *testing.T) {
 		},
 	}
 
+	if err := fakeClient.Get(ctx, client.ObjectKeyFromObject(rs), rs); err != nil {
```
sdowell, line 501: What's the reason for the `Get` calls that are sprinkled in?

pkg/syncer/syncertest/fake/dynamic_client.go
karlkfi: initGeneration allows the user to specify a generation, while just setting it overrides any value specified by the caller.

I deliberately only used initGeneration in the Create call (which the dynamic client doesn't have implemented ATM).

I'm not really sure if Update should allow creation or not, but I assume it should at least for SSA.

Normally, I think the apiserver would always set Generation, and ignore what the user specified on Create, but I wanted to allow users of the fake client to be able to set their own Generation for testing purposes, but only on Create.

We should probably add Create for the dynamic client, with similar behavior, but we haven't needed it for tests yet.

If you think Updates that create a new object should allow the user to set the Generation tho, let me know. It kinda depends on whether we want to follow the apiserver behavior or allow different behavior because we need it for testing.

pkg/reconcilermanager/controllers/otel_controller_test.go
karlkfi: It's the hash of the otel collector configmap object. I don't know if it really needs to be tested, but it does help detect changes. In this case it changed because the fake client is setting UID now.

pkg/reconcilermanager/controllers/reposync_controller_test.go
karlkfi: The fake client Update now validates the UID and RV, if specified.

Since these tests specify both, they need to GET the latest values in order to do an update. Alternatively, we could remove both values, but that might erase changes done by the controller. Another option would be to switch to SSA, but that requires using Unstructured and the dynamic client, which would have been a lot more change. Adding Gets was the easy way to get the latest UID & RV and reflects how you might use the real client.

pkg/reconcilermanager/controllers/otel_controller_test.go
sdowell: I was thinking we could leave a comment to document it for future readability

pkg/syncer/syncertest/fake/client.go
xinnywinne: It looks like Delete is different from Create and Update, the typed client does not update the input object https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/vendor/sigs.k8s.io/controller-runtime/pkg/client/typed_client.go#L74.  And I did not find any test in this PR using this UID after Delete.

karlkfi:
/hold
minor change inc

karlkfi:
/unhold

Pushed some minor changes to comment the config map hash as requested, fix a few comments, and make Delete more like the other methods (copies UID, RV, & Gen from the cached obj back to the input object).

pkg/reconcilermanager/controllers/otel_controller_test.go
karlkfi: added comments to explain

PR #333, opened 2022-12-21
Modify the Delete method in fake client

xinnywinne:
To align with the real typed client, remove the logic to copy latest values back to input object from the Delete method.

### DIFF
--- a/pkg/syncer/syncertest/fake/client.go
+++ b/pkg/syncer/syncertest/fake/client.go
@@ -380,10 +380,8 @@ func (c *Client) Delete(_ context.Context, obj client.Object, opts ...client.Del
 		return newConflictingResourceVersion(id, obj.GetResourceVersion(), cachedObj.GetResourceVersion())
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/syncer/syncertest/fake/client.go
```diff
@@ -380,11 +380,6 @@ func (c *Client) Delete(_ context.Context, obj client.Object, opts ...client.Del
 		return newConflictingResourceVersion(id, obj.GetResourceVersion(), cachedObj.GetResourceVersion())
 	}
 
-	// Copy latest values back to input object
```
karlkfi, line 383: Please add a comment with a link to the client code for delete to show that it doesn't use the `Into` like the other methods.  Pick a tag version to pin to when linking to github.

xinnywinne: Sounds good, done.

karlkfi (approved):
/lgtm

PR #334, opened 2023-01-03
Re-enable the validation for the resource_conflicts metric

nan-yu:
b/236191762


karlkfi:
Is this safe to re-enable now? Or will it still be flakey?
Was the root cause fixed?

nan-yu:
> Is this safe to re-enable now? Or will it still be flakey? Was the root cause fixed?

Per http://b/236013416#comment3, it seems commit b12faedc5d1229b5186e828fafca6f0862a17a9e delete the constraint template and constraint before deleting the CRD, which should not emit the resource conflict metric anymore. I also ran the test group 5 locally and all test cases passed. 

I think we can enable this metric and monitor the CI. If it brings new flake, we can revert the change.

karlkfi (approved):
/lgtm

PR #335, opened 2023-01-03
Update license header to fix the presubmit failure

nan-yu:


manifests/reposync-crd.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2023 Google LLC
```
sdowell, line 1: Do all of the headers need to be updated?

nan-yu: Nope. Only the CRD files because they're regenerated in the presubmit job.

PR #338, opened 2023-01-05
update provisioning scripts for multi-project

sdowell:
- move make targets to Makefile.oss.prow for discoverability
- update WI script to support provisioning in separate project


nan-yu (approved):
/lgtm

PR #342, opened 2023-01-09
add log/metric writing roles to compute account

sdowell:
this enables the compute account to write cloud logs and cloud monitoring metrics


nan-yu (approved):
/lgtm

PR #343, opened 2023-01-10
Update git-sync version to 3.6.2 to clear vulnerabilities

tiffanny29631:
Bumped from https://github.com/kubernetes/git-sync/issues/644

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -110,7 +110,7 @@ data:
                - NET_RAW
            imagePullPolicy: IfNotPresent
          - name: git-sync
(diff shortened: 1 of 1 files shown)


nan-yu (approved):
/lgtm

Just FYI, it seems there are no vulnerabilities if you rebuild the v3.6.1 git-sync image. 
Bumping the git-sync version may introduce new changes, which requires release note update. 

It is okay for this case as there is no major change between v3.6.1 and v3.6.2: https://github.com/kubernetes/git-sync/compare/v3.6.1...v3.6.2.

tiffanny29631:
@nan-yu Thanks for the reminder! Should we still note this upgrade in the release note if to use 3.6.2?

nan-yu:
> Should we still note this upgrade in the release note if to use 3.6.2?

There is no functional change or bug fixes. The version update for git-sync is considered as implementation details. So we are good.

PR #345, opened 2023-01-11
add nomos image to build/publish

sdowell:
this image is not intended for use with Config Sync backend components. This image simply contains the nomos binary, and is intended to support use cases such as containerized CI/CD.

PR #346, opened 2023-01-11
add nomos image to build/publish (#345)

sdowell:
this image is not intended for use with Config Sync backend components. This image simply contains the nomos binary, and is intended to support use cases such as containerized CI/CD.


nan-yu (approved):
/lgtm

PR #350, opened 2023-01-13
fixup: dont use buildx for nomos image

sdowell:
main switched all of the docker image builds to buildx, which happened after the 1.14 release branch was cut. We are seeing an issue with the build of this image in CI. This is a fixup to the cherry pick to use docker build instead.


nan-yu (approved):
/lgtm

PR #353, opened 2023-01-13
make default VERSION and IMAGE_TAG the same

sdowell:
This passes the same options for the default value of VERSION and IMAGE_TAG. This ensures that the version the image is tagged with is the same as the version that the nomos binary is built with. This disparity was leading to a test failure when the HEAD commit has a tag.

This change does not alter the release functionality, since the release flows pass in a value rather than relying on the default.


nan-yu (approved):
/lgtm

PR #355, opened 2023-01-14
RoleBinding -> ClusterRoleBinding in e2e tests

karlkfi:
Using CRB, instead of RB, makes it easier to avoid deadlocks when deleting namespaces containing RepoSyncs with deletion propagation enabled. When using RBs, the RB and RepoSync are deleted at the same time, violating depends-on deletion ordering, because the namespace finalizer doesn't respect or know about dependencies.

Users can still use RoleBindings with RepoSyncs, and we still test that in a couple tests, but we may want to recommend using ClusterRoleBindings once the deletion propagation feature is promoted to a GA feature.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -429,24 +429,6 @@ func noOOMKilledContainer(nt *NT) Predicate {
 	}
 }
 
-func setupRootSync(nt *NT, rsName string) {
-	// create RootSync to initialize the root reconciler.
-	rs := RootSyncObjectV1Beta1FromRootRepo(nt, rsName)
-	if err := nt.Create(rs); err != nil {
-		if !apierrors.IsAlreadyExists(err) {
-			nt.T.Fatal(err)
-		}
-	}
-}
-
-func setupRepoSync(nt *NT, nn types.NamespacedName) {
-	// create RepoSync to initialize the Namespace reconciler.
-	rs := RepoSyncObjectV1Beta1FromNonRootRepo(nt, nn)
-	if err := nt.Create(rs); err != nil {
-		nt.T.Fatal(err)
-	}
-}
-
 func waitForReconciler(nt *NT, name string) error {
 	took, err := Retry(60*time.Second, func() error {
 		return nt.Validate(name, configmanagement.ControllerNamespace,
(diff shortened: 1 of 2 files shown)


e2e/nomostest/config_sync.go
```diff
@@ -892,34 +875,30 @@ func setupCentralizedControl(nt *NT, opts *ntopts.New) {
...
-			// it will only grant usage for pods being run in the same namespace as the binding.
-			// TODO: Remove the psp related change when Kubernetes 1.25 is
-			// available on GKE.
-			crb := repoSyncClusterRoleBinding(nn)
-			dependencies = append(dependencies, crb)
-			nt.RootRepos[configsync.RootSyncName].Add(fmt.Sprintf("acme/cluster/crb-%s-%s.yaml", ns, nn.Name), crb)
-		}
 
+		// Add a ClusterRoleBinding to grant the RepoSync reconciler permission
+		// to manage resources in the RepoSync namespace.
```
sdowell, line 883: I can see how this makes the dependencies/cleanup easier to manage, however I don't see this as being realistic. Switching to a ClusterRoleBinding means that the RepoSync permissions are being granted for the whole cluster rather than being restricted to the RepoSync's namespace.

It doesn't seem to me like this is something we would recommend to users, since RepoSyncs are intended to be limited in scope to a single namespace. I feel that we should come up with a solution to make this work with RoleBindings - WDYT?

karlkfi: Hmm. For some reason I thought ClusterRoleBinding could apply to a namespace, but you're right, they apply to all namespaces... That means our existing PSP workaround is a big security risk, and so is this PR...

karlkfi: The only way I can think of to avoid deadlock when deleting a namespace is to make a controller that injects finalizers on objects with an incoming dependency to block their deletion until the objects that depend on them are deleted. This would enforce deletion ordering on all clients.

Sounds like a fair amount of work tho...

karlkfi:
/hold

Security concerns in comments.

e2e/nomostest/config_sync.go
karlkfi: bug filed with k8s (tho we'll want a faster fix locally): https://github.com/kubernetes/kubernetes/issues/115070

karlkfi:
This is a bad idea and a security risk. I'm withdrawing this PR.

PR #361, opened 2023-01-18
Replace Wait with Watch (continued)

karlkfi:
- Rename WatchForObject to WatchObject
- Update WatchObject to List before Watch, to get the current state,
  instead of needing to wait for the next change.
- Update Watch funcs to return an error, instead of calling Error or
  Fatal. This makes it possible to use in parallel and in Clean/Reset
  where test failure is undesirable.
- Update Watch to use log.AsYAMLDiff, for non-truncating diffs that are
  only computed if debug is enabled.
- Replace WaitForNotFound with WatchForNotFound. This makes it faster
  and easier to debug (diffs logged for each state change)
- Update log.AsYAMLDiff to work with nil objects

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/362

### DIFF
--- a/e2e/nomostest/watch.go
+++ b/e2e/nomostest/watch.go
@@ -19,82 +19,173 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	"go.uber.org/multierr"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/util/log"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// WatchForObject watches the specified object util all predicates return nil,
-// or the timeout is reached.
-func WatchForObject(nt *NT, gvk schema.GroupVersionKind, name, namespace string, predicates []Predicate, waitOpts ...WaitOption) {
-	nt.T.Helper()
+// WatchOption is an optional parameter for Watch
+type WatchOption func(watch *watchSpec)
 
-	startTime := time.Now()
-
-	wait := waitSpec{
-		timeout:         nt.DefaultWaitTimeout,
-		failureStrategy: WaitFailureStrategyFatal,
-	}
-	for _, opt := range waitOpts {
-		opt(&wait)
-	}
-
-	err := watchForObjectInner(nt, gvk, name, namespace, predicates, wait)
-	if err != nil {
-		took := time.Since(startTime)
-		nt.T.Logf("failed after %v watching for predicates to be satisfied", took)
-		switch wait.failureStrategy {
-		case WaitFailureStrategyFatal:
-			nt.T.Fatal(err)
-		case WaitFailureStrategyError:
-			nt.T.Error(err)
-		}
-		return
-	}
-	took := time.Since(startTime)
-	nt.T.Logf("took %v watching for predicates to be satisfied", took)
+type watchSpec struct {
+	timeout         time.Duration
+	errorOnNotFound bool
 }
 
-func watchForObjectInner(nt *NT, gvk schema.GroupVersionKind, name, namespace string, predicates []Predicate, wait waitSpec) error {
-	nt.T.Helper()
+// WatchTimeout provides the timeout option to Watch.
+func WatchTimeout(timeout time.Duration) WatchOption {
+	return func(watch *watchSpec) {
+		watch.timeout = timeout
+	}
+}
 
-	listGVK := gvk.GroupVersion().WithKind(gvk.Kind + "List")
-	rObj, err := nt.WatchClient.Scheme().New(listGVK)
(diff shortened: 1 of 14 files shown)


e2e/nomostest/watch.go
```diff
@@ -19,95 +19,198 @@ import (
...
 	if namespace != "" {
 		listOpts = append(listOpts, client.InNamespace(namespace))
 	}
 
-	ctx, cancel := context.WithTimeout(nt.Context, wait.timeout)
+	ctx, cancel := context.WithTimeout(nt.Context, spec.timeout)
 	defer cancel()
+
+	// Use LIST with a single-object selector, instead of GET, because we want
+	// the resource version of the LIST response to start the watch with, so we
```
nan-yu, line 108: If we use GET, does the watch not capture the resource version change?

e2e/nomostest/watch.go
```diff
@@ -121,15 +224,67 @@ func watchForObjectInner(nt *NT, gvk schema.GroupVersionKind, name, namespace st
...
+				errPrefix, event)
 		}
 	}
 
 	if len(prevErrs) > 0 {
 		// watch exited with predicate errors
 		return multierr.Combine(prevErrs...)
 	}
-	return errors.Errorf("watch exited before any add/update events were recieved")
+	return errors.Errorf("%s exited before any add/update events were recieved",
```
nan-yu, line 239: type: received

e2e/nomostest/watch.go
```diff
@@ -121,15 +224,67 @@ func watchForObjectInner(nt *NT, gvk schema.GroupVersionKind, name, namespace st
...
+// WatchForNotFound waits for the passed object to be fully deleted or not found.
+// Returns an error if the object is not deleted within the timeout.
+func WatchForNotFound(nt *NT, gvk schema.GroupVersionKind, name, namespace string, opts ...WatchOption) error {
+	opts = append(opts, watchErrorOnNotFound())
+	err := WatchObject(nt, gvk, name, namespace,
+		[]Predicate{objectNotFoundPredicate},
+		opts...)
+	if err == nil {
+		// Should never happen!
+		panic("WatchForNotFound exited without error or timeout")
```
nan-yu, line 259: It will happen if `errorOnNotFound` is false and a message will be logged. Do we still need to panic?

e2e/testcases/multi_sync_test.go
```diff
@@ -771,17 +778,31 @@ func roleHasRules(wantRules []rbacv1.PolicyRule) nomostest.Predicate {
...
+			defer func() {
+				wg.Done()
+			}()
+			errCh <- fn()
 		}()
 	}
-	wg.Wait()
+	go func() {
+		wg.Wait()
+		close(errCh)
```
sdowell, line 800: minor nit: is it safer to `defer close(errCh)` before the call to `wg.Wait()`?

e2e/nomostest/watch.go
karlkfi: errorOnNotFound is a private field on a private struct modified by a private function. No caller outside of this package can use it. It's only enabled by this function, and it's enabled by appending to the caller's options, overriding what the caller specified. So it will always be true in this context.

e2e/nomostest/watch.go
karlkfi: List and Get have different resource versions. It's the difference between the Object and the ObjectList. How the list versions work, I don't really understand, except to know that they're different things...

Unfortunately the controller-runtime plays fast and loose with this idea (it doesn't pass all the fields through the client.ListOptions to the metav1.ListOptions), so I'm not 100% sure this actually works the same way it does with client-go clients, which use the same strategy for the ListAndWatch that backs the Informer library.

e2e/nomostest/watch.go
karlkfi: fixed the 3 occurances.

e2e/testcases/multi_sync_test.go
karlkfi: I honestly have no idea. It depends if Wait panics.
That said, we're probably more at risk of a deadlock than a panic from the stdlib.

karlkfi: Added a defer to be defensive, but i dont actually think it matters. 

I guess if a panic is caught by the test runner you'd want the other goroutines to exit cleanly too...

karlkfi:
/retest

Known flakey error:
`Error: cannot build pipelines: failt to create "opencensus" receiver, in pipeline "metrics": failed to bind to address "0.0.0.0:55678": listen tcp 0.0.0.0:55678: bind: address already in use`

e2e/testcases/multi_sync_test.go
```diff
@@ -771,17 +778,33 @@ func roleHasRules(wantRules []rbacv1.PolicyRule) nomostest.Predicate {
...
 	var wg sync.WaitGroup
+	var errs []error
+	errCh := make(chan error)
 	for i := range fns {
 		fn := fns[i]
 		wg.Add(1)
 		go func() {
-			defer wg.Done()
-			fn()
+			defer func() {
```
sdowell, line 792: What's the difference between this and `wg.Done()`?

sdowell:
/lgtm
I'll leave approve to @nan-yu so she can take another look

e2e/nomostest/watch.go
nan-yu: ACK. Thanks for the explanation.

e2e/nomostest/watch.go
nan-yu: ACK

nan-yu (approved):
/lgtm

PR #362, opened 2023-01-19
Add log.AsYAML and log.AsYAMLDiff

karlkfi:
- Use the Stringer pattern (like the existing log.AsJSON) to avoid expensive diffing and serializing of objects, for logging with configurable verbosity.
- Use github.com/kylelemons/godebug/diff for diffing, because it doesn't truncate the way cmp.Diff does.
- Use the (configurable) scheme to lookup object GKV from its type and encode YAML, if possible. Fallback to yaml.Marshal for other Go types.

### DIFF
--- a/pkg/util/log/stringer_test.go
+++ b/pkg/util/log/stringer_test.go
@@ -1,0 +1,280 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


pkg/util/log/stringer_test.go
```diff
@@ -0,0 +1,280 @@
...
+ kind: Namespace
+ metadata:
+   labels:
+-    key1: value1
++    key2: value2
+   name: example
+ `,
+		},
+		{
+			name: "core.ID (non-object compund struct) with no diff",
```
nan-yu, line 219: typo: compound

nan-yu (commented):
one typo, otherwise lgtm

pkg/util/log/stringer_test.go
karlkfi: fixed

nan-yu: line 219 and 243 also need to be fixed

karlkfi: oops. thanks. fixed.

nan-yu (approved):
/lgtm

PR #363, opened 2023-01-20
Add TaskGroup to exec tasks in parallel

karlkfi:


e2e/nomostest/taskgroup/taskgroup.go
```diff
@@ -0,0 +1,93 @@
...
+// functions and collect their errors, if any.
+func New() *TaskGroup {
+	return &TaskGroup{
+		errCh: make(chan error),
+	}
+}
+
+// NewWithFilter returns a new TaskGroup, like `New`, but with a filter for
+// optionally ignoring/muting errors using a `Filter` function.
+func NewWithFilter(f Filter) *TaskGroup {
```
sdowell, line 56: I'm thinking a variable arg for the Filter would offer more flexibility here. That way we could have some narrowly scoped Filter functions that can be reused rather than having to create a new Filter function for combinations of errors to ignore. This would also give the option of a single `New` function that serves all purposes (no Filter, single Filter, multiple Filters, etc). WDYT?

karlkfi: Don't have any use cases yet, so i'm not inclined to add them yet. 

I copied the idea from https://github.com/creachadair/taskgroup because it was quick and easy, but my impl waits until all are complete and batches the errors in a MultiError, instead of just returning the first error.

karlkfi: meh... i'll add it. what the hell

karlkfi: lol too late

PR #365, opened 2023-01-24
Clean RepoSyncs with deletion-propagation enabled

karlkfi:
Just disable deletion-propagation and the finalizer and let namespace deletion handle deleting the objects without ordering requirements.


nan-yu (approved):
/lgtm

PR #367, opened 2023-01-25
Prevent declaring the reconciler's own RSync object in its repo

nan-yu:
This further prevents the reconciler updates its own RSync object in either the admission webhook or remediator.


pkg/validate/raw/raw.go
```diff
@@ -101,6 +102,7 @@ func Unstructured(objs *objects.Raw) status.MultiError {
 		objects.VisitAllRaw(validate.CRDName),
 		objects.VisitAllRaw(validate.RootSync),
 		objects.VisitAllRaw(validate.RepoSync),
+		objects.VisitAllRaw(validate.SelfReconcile(objs.ReconcilerName)),
```
karlkfi, line 105: wow, i didn't know we had all this validation! nifty...

do we need an e2e test? 

I don't know where this validation is being called.

pkg/validate/validate_test.go
```diff
@@ -871,6 +899,42 @@ func TestHierarchical(t *testing.T) {
 			},
 			wantErrs: fake.Errors(validate.InvalidSyncCode),
 		},
+		{
+			name: "RepoSync manages itself",
```
sdowell, line 903: Do we need test cases for nested RootSyncs in hierarchical mode as well?

pkg/validate/raw/raw.go
nan-yu: It is called in the `parseSource` function of both `pkg/parse/root.go` and `pkg/parse/namespace.go`: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/root.go#L139 and https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/namespace.go#L122.

I added two e2e tests, one for RootSync and one for RepoSync.

pkg/validate/validate_test.go
nan-yu: Nested RootSyncs are not supported in hierarchical mode because it requires declaring the `config-management-system` namespace explicitly in the path of `namespaces/config-management-system/ns.yaml`. However, cms should not be a user managed namespace.

sdowell: Ack, thanks for the explanation. I'm not very familiar with hierarchical mode

PR #369, opened 2023-01-26
Stream e2e log output

karlkfi:
Streaming requires test verbosity and one package at a time, but still allows any number of parallel tests.


nan-yu (approved):
/lgtm

PR #370, opened 2023-01-26
Escape the newline character in the apply/reconcile log

nan-yu:
The applier logs each status with a new line, which makes it an Error log in cloud logging: http://screen/sk9CAhpH7e4YqxA.

This commit escapes the newline character, so it will be a single line in cloud logging. It can also be easily converted to multiple lines.


karlkfi (approved):
/lgtm

PR #371, opened 2023-01-27
Cleanup Service Firewalls in e2e tests

karlkfi:
The TestPublicHelm produces a Service of type LoadBalancer, which
causes a firewall rule and target pool to be created by the K8s GCP
cloud provider. But there seems to be a bug that leaves the firewall
behind when the service is deleted.

This change deletes the firewall rule with gcloud in test cleanup.


tiffanny29631 (commented):
/lgtm

karlkfi:
/hold

still testing

e2e/testcases/helm_sync_test.go
```diff
@@ -47,6 +53,29 @@ const (
...
+	if err != nil {
+		nt.T.Log(out)
+		return err
+	}
+	return nil
+}
+
 // TestPublicHelm can run on both Kind and GKE clusters.
 // It tests Config Sync can pull from public Helm repo without any authentication.
 func TestPublicHelm(t *testing.T) {
```
nan-yu, line 81: `TestHydrateHelmComponents` also uses the remote ingress-nginx helm chart for testing: http://go/cs-oss/blob/main/e2e/testdata/hydration/helm-components/kustomization.yaml#L19.

We will need to add the cleanup in that test as well.

An alternative to bypass the cloud-provider-gce issue is to use a simple helm chart without ingress.

sdowell: I also would prefer to use a simpler helm chart

nan-yu: Agreed. Then we don't need to worry about removing the code snippet when the cloud-provider-gce issue is fixed in the future.

karlkfi: If y’all want to make the PR to a different helm chart, be my guest. I’m just trying to unblock CI. 

This is looking like a legitimate GKE bug. I’m following up on internal channels.

nan-yu: Sure. I can make a PR that changes the ingress chart to wordpress.

karlkfi: It’s being used in a half dozen places, including the nomos hydrate tests and helm tests. I’m not sure if those all need to be updated or not, because I haven’t narrowed down which Services are leaving GCP resources behind.

nan-yu: https://github.com/GoogleContainerTools/kpt-config-sync/pull/372

karlkfi:
Obsolete. Replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/372

PR #373, opened 2023-01-30
Use deletion propagation for e2e test cleanup

karlkfi:
- Rewrite test cleanup to use the RSync finalizer to delete all
  managed resources by default (for e2e tests only)
- Fix TestDeleteRootSyncAndRootSyncV1Alpha1 to handle longer RSync
  deletion timeout
- Add depends-on to RepoSyncs for [Cluster]RoleBindings to ensure
  the reconciler finalizer has permission to delete before the
  binding is deleted.
- Remove test cleanup blocks that are no longer necessary.
- Delete and re-create all RootSyncs & RepoSyncs between each test,
  to ensure better test isolation, to make it easier to debug test
  failure.

This is a repeat of https://github.com/GoogleContainerTools/kpt-config-sync/pull/205, which was reverted, with modifications.

The recent fixes to `Clean` should handle forcibly deleting any RSyncs which get stuck finalizing due to race conditions in deletion ordering. So even if the reconciler gets wedged, it shouldn't break the cluster like it did last time. This way we can narrow down which tests are causing test failures more easily, if any.

### DIFF
--- a/e2e/nomostest/reset.go
+++ b/e2e/nomostest/reset.go
@@ -1,0 +1,513 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nomostest
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
+	"kpt.dev/configsync/pkg/api/configmanagement"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/importer/filesystem"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/metadata"
+	"kpt.dev/configsync/pkg/metrics"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+	"kpt.dev/configsync/pkg/syncer/differ"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Reset performs multi-repo test reset:
+// - Delete unmanaged RootSyncs & RepoSyncs (with deletion propagation)
+// - Validate managed RepoSyncs & RootSyncs are deleted
+// - Delete all test namespaces not containing config-sync itself
+// - Clear Repository-to-RSync assignments
+//
+// This should cleanly delete or reset all registered RSyncs.
+// Any managed RSyncs must have deletion propagation enabled by the test that
+// created them, otherwise their managed resources will not be deleted when the
+// RSync is deleted.
+// Any unregistered Repos must be reset by individual test Cleanup.
+func Reset(nt *NT) error {
+	start := time.Now()
+	defer func() {
+		elapsed := time.Since(start)
+		nt.T.Logf("[RESET] Test environment reset took %v", elapsed)
+	}()
+
+	// Delete all existing RootSyncs with the test label.
+	// Enable deletion propagation first, to clean up managed resources.
+	rootSyncList, err := listRootSyncs(nt)
+	if err != nil {
+		return err
+	}
+	if err := ResetRootSyncs(nt, rootSyncList.Items); err != nil {
+		return err
+	}
+
+	// Delete all existing RepoSyncs with the test label.
+	// Enable deletion propagation first, to clean up managed resources.
+	repoSyncList, err := listRepoSyncs(nt)
+	if err != nil {
+		return err
+	}
+	if err := ResetRepoSyncs(nt, repoSyncList.Items); err != nil {
+		return err
+	}
+
+	// Delete all Namespaces with the test label (except shared).
+	nsList, err := listNamespaces(nt)
+	if err != nil {
+		return err
+	}
+	// Error if any protected namespace was modified by a test (test label added)
+	// and not reverted by the test.
+	protectedNamespaces, nsListItems := filterNamespaces(nsList.Items,
+		protectedNamespaceList()...)
+	if len(protectedNamespaces) > 0 {
+		return errors.Errorf("protected namespace(s) modified by test: %+v",
+			protectedNamespaces)
+	}
+	// Skip deleting namespaces that contain test infra or Config Sync itself.
+	_, nsListItems = filterNamespaces(nsListItems,
+		configsync.ControllerNamespace,
+		configmanagement.RGControllerNamespace,
+		metrics.MonitoringNamespace,
+		testGitNamespace)
+	if err := ResetNamespaces(nt, nsListItems); err != nil {
+		return err
+	}
+
+	// NOTE: These git repos are not actually being deleted here, just
+	// unassigned to a specific RSync. All remote repos are cached in
+	// nt.RemoteRepositories and then reassigned in `resetRepository`.
+	// Repos are actually deleted by `Clean` in environment setup and teardown.
+	nt.NonRootRepos = make(map[types.NamespacedName]*Repository)
+	nt.RootRepos = make(map[string]*Repository)
+
+	return nil
+}
+
+func protectedNamespaceList() []string {
+	list := make([]string, 0, len(differ.SpecialNamespaces))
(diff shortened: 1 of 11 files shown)


karlkfi:
/hold

Failure in TestConflictingDefinitions_NamespaceToNamespace when testing with GKE:
```
RootSync: ... "errors":[{"code":"2004","errorMessage":"KNV2004: unable to sync repo\nError in the git-sync container: {\"Msg\":\"error syncing repo, will retry\",\"Err\":\"Run(git clone -v --no-checkout -b main --depth 1 git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync /repo/source): exit status 128: { stdout: \\\"\\\", stderr: \\\"Cloning into '/repo/source'...\\\\nfatal: '/git-server/repos/config-management-system/root-sync' does not appear to be a git repository\\\\nfatal: Could not read from remote repository.\\\\n\\\\nPlease make sure you have the correct access rights\\\\nand the repository exists.\\\" }\",\"Args\":{\"failCount\":27,\"waitTime\":15000000000}}\n\nFor more information, see https://g.co/cloud/acm-errors#knv2004"}]
```

And then later:
```
    new.go:349: [SETUP] New test case
    nt.go:534: kubectl --kubeconfig /tmp/nomos-e2e/cs-e2e-1675120845-0/KUBECONFIG exec -n config-management-system-test test-git-server-6c6cdbb685-g2qlj -- git init --bare --shared /git-server/repos/config-management-system/root-sync
    nt.go:534: kubectl --kubeconfig /tmp/nomos-e2e/cs-e2e-1675120845-0/KUBECONFIG exec -n config-management-system-test test-git-server-6c6cdbb685-g2qlj -- git -C /git-server/repos/config-management-system/root-sync config receive.denyNonFastforwards false
[repo root-sync] Setting repo format to hierarchy
[repo root-sync] committing "initial commit" (ca9335d64121cc400acd4d087108d8c264194cd5)
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

ERROR: exit status 128
```

I suspect it has to do with the git repo reset not working correctly.

Before TestConflictingDefinitions_NamespaceToNamespace was TestConflictingDefinitions_NamespaceToRoot and TestConflictingDefinitions_RootToRoot.

karlkfi:
The access rights errors seen above seem flakey and unrelated to the reset code. I suspect they're from using a local git server that isn't highly available.

The same tests pass when re-run.

karlkfi:
Testing progress on GKE:

# [TODO] group1 (a few local errors with nomos CLI diffs from dependency mismatch)
# [PASS] group2
# [PASS] group3
# [PASS] group4
# [PASS] group5
# [TODO] group6
# [TODO] group7
# [TODO] group8

karlkfi:
All e2e tests are now passing on GKE.

But now I've got a bunch of small changes to extract:
- nomostest field-manager name
- Use gcloud in OCI tests to avoid requiring local application credentials (user credentials are more secure on dev workstations because they expire)
- Add more setup and teardown logging with time elapsed

karlkfi:
Extracted:
- #384
- #383
- #382

And rebased.

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,513 @@
...
+			if err := WatchObject(nt, kinds.RepoSyncV1Beta1(), rs.Name, rs.Namespace, []Predicate{
+				HasFinalizer(metadata.ReconcilerFinalizer),
+			}); err != nil {
+				return err
+			}
+		}
+
+		// Grant the reconcile the permissions specified by this test.
+		rsCRB := RepoSyncRoleBinding(rsNN)
+		if err := nt.Apply(rsCRB); err != nil {
```
sdowell, line 259: Maybe instead of having the for loops below to clean these up, we could add a `defer` statement here to delete the `RoleBinding`. WDYT?

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,513 @@
...
+		return deleteRepoSyncClusterRole(nt)
+	}
+
+	nt.T.Logf("[RESET] Deleting RepoSyncs (%d)", len(rsList))
+
+	record := newResetRecord()
+
+	// Apply ClusterRole with the permissions specified by this test.
+	rsCR := nt.RepoSyncClusterRole()
+	if err := nt.Apply(rsCR); err != nil {
```
sdowell, line 222: ditto here for your thoughts to `defer` deleting the `ClusterRole`

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,513 @@
...
+			nt.T.Logf("[RESET] Namespace %s managed by %q", nn, manager)
+			continue
+		}
+
+		nt.T.Logf("[RESET] Deleting Namespace %s", nn)
+		if err := nt.Delete(obj, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+			return err
+		}
+	}
+	for _, item := range nsList {
```
sdowell, line 357: hmm.. If one of the namespaces was managed, won't it not get deleted in the above loop? I assume that means we would fail in this loop

e2e/nomostest/reset.go
karlkfi: We could put `deleteRepoSyncClusterRole` in a defer. Tho I'm skeptical about trying to clean up after failure in the clean up function... I kinda prefer cleanup that fails hard and stops, so you know where it broke and can debug without the state changing afterwards.

e2e/nomostest/reset.go
karlkfi: Defer could work. Or I could build up the Delete & Wait task group now and defer the tg.Wait()...

But do we want to defer in Reset/Clean (see other comment)?

e2e/nomostest/reset.go
```diff
@@ -0,0 +1,513 @@
...
+		// and dependenencies will be garbage collected before the finalizer
+		// can delete the managed resources.
+		// TODO: Remove explicit Background policy after the reconciler-manager finalizer is added.
+		nt.T.Logf("[RESET] Deleting RootSync %s", rsNN)
+		if err := nt.Delete(rs, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
+			return err
+		}
+	}
+	tg := taskgroup.New()
+	for _, item := range rsList {
```
sdowell, line 193: I think generally this should result in all of the `RootSync` objects being deleted, but I'm trying to think of unexpected corner cases. For example:
- a `RootSync` in the input is managed by a `RootSync` that is not included in the input
- a `RootSync` in the input is managed by a `RepoSync`

It might be the case that neither of these scenarios ever happen, but just wanted to bring up this consideration

e2e/nomostest/reset.go
karlkfi: > If one of the namespaces was managed, won't it not get deleted in the above loop?
Correct

> I assume that means we would fail in this loop.
Maybe. It depends on how ResetRepoSyncs is used.

For example, when called after ResetRootSyncs, you can assume all the RootSyncs are deleted, and RepoSyncs can't manage namespaces, so your assertion would be true. But if you call ResetRepoSyncs without ResetRootSyncs, then it's possible that other RootSyncs are being torn down asynchronously... 

To really be clean about it, we could wait here for the unmanaged ones, but then we'd have to duplicate the code to wait for all the rest at the end of `Reset`.

Is it worth duplicating the code to keep this method "pure"?

sdowell: Ah right, on second thought I don't think there should ever be any managed namespaces at this point since the `RootSync`/`RepoSync` cleanup happens before this function is invoked. I suppose maybe that means we don't need the special handling of managed namespaces? I assume all of the `RootSync` managed resources should be cleaned up at this point since we wait at the end of `ResetRootSyncs`. Is that right?

e2e/nomostest/reset.go
karlkfi: > a RootSync in the input is managed by a RootSync that is not included in the input

Mmm... Yeah... decoupling the List did make it risky...

The List in Reset pulls all the RootSyncs that have the test label. So it should be all of them. But it's possible that someone could call ResetRootSyncs with some other set of RootSyncs and it might break.

My goal with the reset code wasn't to cover 100% of possible edge cases tho, only the ones the tests _should_ be using. If a test violates the expectations and Reset breaks, the Clean should fix it because it's more aggressive and thorough (and slower).

> a RootSync in the input is managed by a RepoSync

I think the webhook rejects requests to create RepoSyncs in the config-management-system namespace.  You can have RootSyncs manage RootSyncs, but not RepoSyncs managing RootSyncs.

karlkfi: If you can think of a quick way to detect or mitigate those tho, I can add them in...

sdowell: I think it would be sufficient to just include a cautionary warning in the docstring to inform future adventurers. e.g. something like
```
this function assumes all RootSyncs in the cluster are passed in as input
```

e2e/nomostest/reset.go
sdowell: Yeah that seems reasonable, I don't have any real opposition here just seemed like a viable alternative

nan-yu (approved):
/lgtm

e2e/nomostest/reset.go
karlkfi: >  I suppose maybe that means we don't need the special handling of managed namespaces?

I exposed these functions as public so I could call ResetRepoSyncs from the composition test, to replace the cleanup it was doing there, which is before all the other RootSyncs are deleted. It's used there because half the RSyncs are managed and half are unmanaged, but some of the unmanaged RepoSyncs depend on namespaces and rolebindings managed by the RootSync...

So there's a little bit of extra complexity in here to allow that usage that isn't strictly necessary for just the full `Reset` use case.

e2e/nomostest/reset.go
karlkfi: Assuming CI is happy later, I'll make some follow up PRs to do minor improvements to the reset code. But I don't want to do them now in case we need to revert again.

PR #374, opened 2023-01-30
Update the public helm chart used by the namespace repo

nan-yu:
`TestHelmNamespaceRepo` validates that a namespace reconciler gets a source error when syncing from a public helm chart that includes cluster-scoped resources. The wordpress chart doesn't include any cluster-scoped resources, so this commit updates the chart to pass the validation.


nan-yu:
@tiffanny29631 FYI, the should prevent a new autobug from your onduty list.

PR #376, opened 2023-01-31
Disable MariaDB persistence to avoid leaking PVCs

nan-yu:
Mariadb persistent storage is enabled by default, which creates a PVC in the namespace. When wordpress is deleted, the PVC is left behind after the test.

This commit updates the chart values to disable mariadb persistent storage.

### DIFF
--- a/e2e/testdata/root-sync-helm-chart-cr.yaml
+++ b/e2e/testdata/root-sync-helm-chart-cr.yaml
@@ -42,6 +42,10 @@ spec:
         limits:
           cpu: 1
           memory: 300Mi
+      mariadb:
+        primary:
+          persistence:
+            enabled: false
     releaseName: my-wordpress
     namespace: "wordpress"
     auth: none


e2e/testcases/helm_sync_test.go
```diff
@@ -328,3 +337,15 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
 		return chartVersion, nil
 	}
 }
+
+func deleteWordpressPVC(nt *nomostest.NT) error {
```
karlkfi, line 341: Is this created indirectly? Why doesn't the normal cleanup delete it? Are we missing Cleanup that removes the helm chart and deletes all the bits in the inventory?

nan-yu: It is not an object declared in the chart and managed by Config Sync. I believe MariaDB creates it on the cluster to persist data, but never deletes it even when MariaDB is deleted.

e2e/testcases/helm_sync_test.go
```diff
@@ -98,6 +100,13 @@ func TestPublicHelm(t *testing.T) {
 
 	nt.T.Log("Update RootSync to sync from a public Helm Chart without specified release namespace")
 	nt.MustMergePatch(rs, `{"spec": {"helm": {"namespace": ""}}}`)
+	nt.T.Cleanup(func() {
+		// WordPress MariaDB creates a PVC in the namespace, which is left after the test.
+		// The cleanup function deletes the wordpress-mariadb PVC so that MariaDB can be reinstalled in the namespace in the next test.
+		if err := deleteWordpressPVC(nt); err != nil {
```
sdowell, line 106: Is there a `PVC` created in the default namespace as well by the updated RootSync on line `102`?

e2e/testcases/helm_sync_test.go
```diff
@@ -98,6 +100,13 @@ func TestPublicHelm(t *testing.T) {
 
 	nt.T.Log("Update RootSync to sync from a public Helm Chart without specified release namespace")
 	nt.MustMergePatch(rs, `{"spec": {"helm": {"namespace": ""}}}`)
+	nt.T.Cleanup(func() {
```
sdowell, line 103: Is this `Cleanup` associated with the change in namespace on the line above? From reading this it sounds like it's associated with the initial apply on line `61`. In which case, I'm worried we won't reach this `Cleanup` block before the validations fail. Am I missing something? If not, can this `Cleanup` be moved up to around line `62` to ensure it runs?

e2e/testcases/helm_sync_test.go
nan-yu: Yes, and the `deleteWordpressPVC` function deletes the PVC in the `default` namespace. There is also a PVC created in the `wordpress` namespace from `root-sync-helm-chart-cr.yaml` that is applied in line 61. That is deleted as a side-effect of pruning the `wordpress` namespace.

e2e/testcases/helm_sync_test.go
karlkfi: Is this a generic problem users will have? Do we need to document it?

How is the PVC created? Is it the result of a Deployment field?

Is there a way to configure the chart not to use persistent storage?

e2e/testcases/helm_sync_test.go
sdowell: Ah I didn't realize `DefaultHelmReleaseNamespace` was defined as `default`. Ack, thank you for explaining

sdowell:
LGTM, but I'll give @karlkfi a chance to finish his review

e2e/testcases/helm_sync_test.go
nan-yu: If user's workload creates new K8s resources and doesn't have a finalizer to remove them, then it is a problem.
The PVC seems to be created by MariaDB, which is not a field specified in the Deployment.
Let me check the chart values to see whether we can disable persistence.

nan-yu:
/hold

Exploring the option to disable persistence.

e2e/testcases/helm_sync_test.go
nan-yu: Done. Updated the chart value to disable mariadb persistent storage.

karlkfi: New config looks good!

e2e/testdata/root-sync-helm-chart-cr.yaml
```diff
@@ -42,6 +42,10 @@ spec:
         limits:
           cpu: 1
           memory: 300Mi
+      mariadb:
+        primary:
+          persistence:
+            enabled: false
```
sdowell, line 48: Nice! Will we still need to manually clean up the lingering `PVC` resources on the clusters from previous runs?

nan-yu: The cleanup function is no longer needed and removed.

sdowell: I mean specifically for the resources that were created before this change, i.e. test runs on commits before this change is submitted. Will we need to manually clean up those?

karlkfi (approved):
/lgtm

e2e/testdata/root-sync-helm-chart-cr.yaml
nan-yu: That should not be needed as the mariadb doesn't check the persisted data, but it is good to delete the test data anyway.
I'll remove them later once the first CI job passes after this PR is merged.

PR #378, opened 2023-02-01
Bump the default CPU request for the helm-sync container to avoid helm failures

nan-yu:
When the helm-sync container pulls the wordpress chart on the autopilot clusters, it fails with a `signal: killed` error. It turns out it consumes more CPU than requested.

This commit bumps the default CPU request from 10m to 50m.


nan-yu:
I'll create a separate PR to support overriding the helm-sync's resource requests and limits.

PR #379, opened 2023-02-01
Bump the default CPU request for the helm-sync container to avoid helm failures (#378)

nan-yu:
When the helm-sync container pulls the wordpress chart on the autopilot clusters, it fails with a `signal: killed` error. It turns out it consumes more CPU than requested.

This commit bumps the default CPU request from 10m to 50m.

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -167,7 +167,7 @@ data:
              runAsUser: 65533
            resources:
              requests:
-               cpu: "10m"
+               cpu: "50m"
                memory: "200Mi"
          - name: otel-agent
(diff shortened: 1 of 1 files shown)


sdowell:
Just to be clear the wordpress chart change hasn't been cherry picked to the release branch, so those tests are still using the prior chart. Is this change intended as a functional change for 1.14.2 rather than just for test stability?

nan-yu:
Correct, this change is intended as functional change for 1.14.2.

For the wordpress change, I'm inclined to not cherry pick to the release branch. The downside is that it will continue leaking Firewall rules in the release CI project, but once we get to 1.15.0, it should be fixed.

sdowell:
SGTM
/lgtm
/approve

PR #380, opened 2023-02-01
Support overriding the helm-sync container's resource requirements

nan-yu:


pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -426,7 +426,7 @@ func (r *reconcilerBase) deployment(ctx context.Context, dRef client.ObjectKey)
 	return deployObj, nil
 }
 
-func mutateContainerResource(ctx context.Context, c *corev1.Container, override *v1beta1.OverrideSpec, reconcilerType string) {
```
sdowell, line 429: So these variables were unused? I'm surprised the linter didn't complain

pkg/reconcilermanager/controllers/rootsync_controller_test.go
```diff
@@ -3141,7 +3180,9 @@ func containerResourcesMutator(overrides []v1beta1.ContainerResourcesSpec) depMu
 	return func(dep *appsv1.Deployment) {
 		for _, container := range dep.Spec.Template.Spec.Containers {
 			switch container.Name {
-			case reconcilermanager.Reconciler, reconcilermanager.GitSync, reconcilermanager.HydrationController, reconcilermanager.OciSync:
+			case reconcilermanager.Reconciler, reconcilermanager.GitSync,
```
sdowell, line 3183: Having this much logic to build out test assertions (this function and the surrounding functions included) seems like a bit of a code smell, but not a big deal for this change since it's operating within the existing framework

sdowell (approved):
/lgtm
/approve

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: Yes, those are unused. Reporting unused function parameters is done via the `unparam` option, which is disabled by default.

I tried to turn it on locally, however, the `always receives` check is annoying: https://github.com/mvdan/unparam/issues/31. Once we're able to disable the `always receives` check, we can turn on the `unparam` check.

pkg/reconcilermanager/controllers/rootsync_controller_test.go
nan-yu: agreed. The test is also growing huge, and needs a refactor, but should go to a separate PR.

PR #382, opened 2023-02-02
Fix OCI e2e tests to run locally

karlkfi:
Use gcloud to check image digests, instead of the OCI SDK.
- Work with user credentials. No longer require app credentials.
- Automatically authenticate with GCR and GAR.
- Read the digest without pulling the whole image.
- Cache the digest to avoid excessive API calls to the registry.

### DIFF
--- a/e2e/testcases/oci_sync_test.go
+++ b/e2e/testcases/oci_sync_test.go
@@ -15,20 +15,20 @@
 package e2e
 
 import (
-	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
 	"testing"
 
-	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/google/go-containerregistry/pkg/v1/google"
-	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/e2e/nomostest/policy"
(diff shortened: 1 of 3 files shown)


e2e/testcases/oci_sync_test.go
```diff
@@ -522,33 +521,49 @@ func testDigestUpdate(nt *nomostest.NT, image string) {
...
-	return func(*nomostest.NT, types.NamespacedName) (string, error) {
-		return getImageDigest(imageName, publicImage)
+// imageDigestFunc wraps getImageDigest to return a Sha1Func that caches the
+// image digest, to avoid polling the image registry unnecessarily.
+func imageDigestFunc(imageName string) nomostest.Sha1Func {
+	var cached bool
+	var digest string
+	var err error
+	return func(nt *nomostest.NT, _ types.NamespacedName) (string, error) {
+		if cached {
```
nan-yu, line 561: `cached` is always false. Do you want to make it a parameter of `imageDigestFunc`?

karlkfi: why would it always be false?
It's set to true a few lines down.

karlkfi: It should only be false the first time. Then it looks up the digest and sets it to true to skip digest lookup on subsequent calls.

e2e/testcases/oci_sync_test.go
```diff
@@ -522,33 +521,49 @@ func testDigestUpdate(nt *nomostest.NT, image string) {
 }
 */
 
-func getImageDigest(imageName string, publicImage bool) (string, error) {
-	var auth authn.Authenticator
-	if publicImage {
-		auth = authn.Anonymous
-	} else {
-		a, err := google.NewEnvAuthenticator()
```
sdowell, line 530: What is the material difference of using `gcloud` rather than oci with this provided authentication method?

e2e/testcases/oci_sync_test.go
sdowell: Nice optimization 👍

e2e/testcases/oci_sync_test.go
karlkfi: gcloud supports user auth, not just app credentials.

Before this PR, you would need to run `gcloud auth application-default login` to generate app creds and persist them on disk. Not only is this another required step, but it's also less secure.

After this PR, you can now just use `gcloud auth login`, which users need to have done anyway in order to run the rest of the e2e tests on GKE. gcloud is already used elsewhere in the nomostest framework.

e2e/testcases/oci_sync_test.go
nan-yu: Oh, right. The function is called in a retry loop and `cache` is set to true in the subsequent calls.

nan-yu (approved):
/lgtm

e2e/testcases/oci_sync_test.go
sdowell: SG. Was it generating app creds in CI?

karlkfi: CI uses Workload Identity to inject app creds.

karlkfi:
/unhold

Comments adressed.

PR #383, opened 2023-02-02
Add a nomostest field-manager

karlkfi:
Use Apply instead of Create when ignoring IsAlreadyExists

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -150,10 +150,8 @@ func installConfigSync(nt *NT, nomos ntopts.Nomos) error {
 			cm := o.(*corev1.ConfigMap)
 			cm.Data[filesystem.SourceFormatKey] = string(nomos.SourceFormat)
 		}
-		if err := nt.Create(o); err != nil {
-			if !apierrors.IsAlreadyExists(err) {
-				return err
-			}
+		if err := nt.Apply(o); err != nil {
+			return err
 		}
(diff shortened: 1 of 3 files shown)


nan-yu (approved):
/lgtm

karlkfi:
The fact that this breaks tests now is scary.
It passed tests without the Create -> Apply changes, in another PR.
So that means one or more of those Apply calls is breaking many of the tests... I wonder if it's causing conflicts between managed RootSyncs and the config in nt.RootRepos...

karlkfi:
Ah, I think I know what the problem is.

The `Apply` in `ValidateMultiRepoDeployments` is reverting RootSync changes made by individial tests after test setup.

I think we may need to remove that Apply/Create and replace it with one in test setup, but for now, I'll just undo the Clean->Apply there, since it worked before.

karlkfi:
/retest

I think this is a flakey failure in TestDeclareImplicitNamespace. The Deployment didn't come up.

karlkfi:
/retest 

```
I0202 22:16:58.049312   18143 root.go:196] New source errors for RootSync config-management-system/my-rs: [{Code:2002 ErrorMessage:KNV2002: API discovery failed: APIServer error: unable to retrieve the complete list of server APIs: rbac.authorization.k8s.io/v1: context deadline exceeded, v1: context deadline exceeded
```

I think the kind cluster died in the presubmit...

PR #387, opened 2023-02-03
Reset in parallel

karlkfi:
- Refactor Reset code to delete and wait in parallel. It should be a little faster.
- Handle async RSync & Namespace deletion in Reset
- Remove the now unused ResetRecords


karlkfi:
I don't think we actually want to do this. It might be faster but it also might be a lot harder to debug.

PR #390, opened 2023-02-07
Update git-sync version to fix vulnerabilities

victorpras:


victorpras:
Looks like this is not needed to fix the vulnerabilities

nan-yu (approved):
/lgtm

PR #391, opened 2023-02-07
Rewrite WatchObject to use client-go

karlkfi:
- Switch from using the controller-runtime client to client-go. This allows us access to the following features, which aren't exposed in the controller-runtime client.
- Use ResourceVersion between LIST and WATCH to ensure we don't miss events, especially DELETE events which were causing timeouts when missed.
- Use watch.Until with cache.NewListWatchFromClient to handle retrying watches that failed due to recoverable errors.
- Remove unused NT.WatchConfig. We can just use nt.Config with a custom timeout based on the input to WatchObject.
- Log watch events when watching, to show progress, but only log
  the diffs if --debug is enabled.

Fixes b/267941443

### DIFF
--- a/e2e/nomostest/watch.go
+++ b/e2e/nomostest/watch.go
@@ -21,15 +21,22 @@ import (
 
 	"github.com/pkg/errors"
 	"go.uber.org/multierr"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/util/log"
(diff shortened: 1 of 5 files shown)


e2e/nomostest/watch.go
```diff
@@ -88,166 +95,230 @@ func WatchObject(nt *NT, gvk schema.GroupVersionKind, name, namespace string, pr
...
-	if err != nil {
-		return err
-	}
+	// Don't need to specify RV in the options. watchtools.Until does it for us.
+	// listOpts.ResourceVersion = cObjList.GetResourceVersion()
 
-	// Specify ResourceVersion to ensure the Watch starts where the List stopped.
-	cObjList.SetResourceVersion(rv)
+	// ResourceVersionMatch is forbidden for WATCH.
+	// listOpts.ResourceVersionMatch = metav1.ResourceVersionMatchNotOlderThan
```
sdowell, line 175: did you intend to leave the commented code here and above?

e2e/nomostest/watch.go
```diff
@@ -88,166 +95,230 @@ func WatchObject(nt *NT, gvk schema.GroupVersionKind, name, namespace string, pr
...
-	result, err := nt.WatchClient.Watch(ctx, cObjList, listOpts...)
-	defer result.Stop()
-	if err != nil {
-		return err
-	}
+	// Enable bookmarks to ensure retries start from the latest ResourceVersion,
+	// even if there haven't been any updates to the object being watched.
+	listOpts.AllowWatchBookmarks = true
+
+	condition := func(event watch.Event) (bool, error) {
```
sdowell, line 181: Could you leave a brief comment about the meaning behind this function signature

nan-yu (commented):
/lgtm

I'll let Sam approve this PR in case you want to address his comments.

e2e/nomostest/watch.go
karlkfi: Removed.

e2e/nomostest/watch.go
karlkfi: Done.

PR #394, opened 2023-02-07
Modify Predicates to recieve nil for NotFound

karlkfi:
- Update all Predicate impls to tollerate nil objects without panic.
- Remove hacky workaround previously used to exit early when the object was not found. Now a normal predicate can handle it.


nan-yu (approved):
/lgtm

PR #395, opened 2023-02-08
Use WatchObject instead of Retry+Validate

karlkfi:
- Update most e2e usages for Retry & Valdiate to use Watch functions
  instead. This should make tests less flakey, slightly faster, and
  easier to debug.
- Use 30s as a minimum timeout to account for occationally slow CI.
- Refactor a few tests to wait in parallel with a TaskGroup
- Update more Predicates to support nil object as NotFound

### DIFF
--- a/e2e/testcases/namespace_repo_test.go
+++ b/e2e/testcases/namespace_repo_test.go
@@ -26,11 +26,13 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/metrics"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/e2e/nomostest/policy"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
 	"kpt.dev/configsync/pkg/testing/fake"
 	"kpt.dev/configsync/pkg/validate/raw/validate"
@@ -54,10 +56,12 @@ func TestNamespaceRepo_Centralized(t *testing.T) {
 	// namespace reconciler deployment.
 	// Log error if the Reconciling condition does not progress to False before
 	// the timeout expires.
-	_, err := nomostest.Retry(15*time.Second, func() error {
-		return nt.Validate("repo-sync", bsNamespace, &v1beta1.RepoSync{},
-			hasReconcilingStatus(metav1.ConditionFalse), hasStalledStatus(metav1.ConditionFalse))
-	})
+	err := nomostest.WatchObject(nt, kinds.RepoSyncV1Beta1(), "repo-sync", bsNamespace,
+		[]nomostest.Predicate{
+			hasReconcilingStatus(metav1.ConditionFalse),
+			hasStalledStatus(metav1.ConditionFalse),
+		},
+		nomostest.WatchTimeout(30*time.Second))
 	if err != nil {
 		nt.T.Errorf("RepoSync did not finish reconciling: %v", err)
 	}
@@ -79,10 +83,9 @@ func TestNamespaceRepo_Centralized(t *testing.T) {
 	repo.CommitAndPush("Adding service account")
 	nt.WaitForRepoSyncs()
 
-	// Validate service account 'store' is present.
-	_, err = nomostest.Retry(15*time.Second, func() error {
-		return nt.Validate("store", bsNamespace, &corev1.ServiceAccount{})
-	})
+	// Validate service account 'store' is current.
+	err = nomostest.WatchForCurrentStatus(nt, kinds.ServiceAccount(), "store", bsNamespace,
+		nomostest.WatchTimeout(30*time.Second))
 	if err != nil {
 		nt.T.Fatalf("service account store not found: %v", err)
 	}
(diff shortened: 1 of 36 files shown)


e2e/testcases/preserve_fields_test.go
```diff
@@ -100,19 +100,17 @@ func TestPreserveGeneratedServiceFields(t *testing.T) {
...
-		// to change; copied from bats.
-		return nt.Validate(serviceName, ns, &corev1.Service{},
-			hasDifferentNodePortOrClusterIP(generatedNodePort, generatedClusterIP))
-	})
+	// This can only return nil if the NodePort/ClusterIP was updated.
+	// Potentially flaky check since other things can cause NodePort/ClusterIP
+	// to change; copied from bats.
+	err = nomostest.WatchObject(nt, kinds.Service(), serviceName, ns,
+		[]nomostest.Predicate{hasDifferentNodePortOrClusterIP(generatedNodePort, generatedClusterIP)},
+		nomostest.WatchTimeout(30*time.Second))
```
sdowell, line 108: hmm -> why the bump from 5 to 30 seconds?

e2e/testcases/preserve_fields_test.go
```diff
@@ -198,12 +195,12 @@ aggregationRule:
 	nt.WaitForRepoSyncs()
 
 	// Ensure the aggregate rule is actually aggregated.
-	duration, err := nomostest.Retry(20*time.Second, func() error {
-		return nt.Validate(aggregateRoleName, "", &rbacv1.ClusterRole{}, clusterRoleHasRules([]rbacv1.PolicyRule{
-			nsViewer.Rules[0], rbacViewer.Rules[0],
-		}))
-	})
-	nt.T.Logf("took %v to wait for aggregate ClusterRole", duration)
+	err := nomostest.WatchObject(nt, kinds.ClusterRole(), aggregateRoleName, "",
```
sdowell, line 198: should this specify the same 20 second timeout?

e2e/testcases/status_enablement_test.go
```diff
@@ -63,16 +66,47 @@ func TestStatusEnabledAndDisabled(t *testing.T) {
...
 	}
 }
+
+func resourceGroupHasNoStatus(obj client.Object) error {
+	if obj == nil {
+		return nomostest.ErrObjectNotFound
+	}
+	rg, ok := obj.(*resourcegroupv1alpha1.ResourceGroup)
+	if !ok {
+		return nil
```
sdowell, line 87: Could this return a false positive? I assume this predicate should only be used on the ResourceGroup kind and thus could return an error if it's the wrong type

e2e/testcases/sync_ordering_test.go
```diff
@@ -442,18 +442,14 @@ func TestDependencyWithReconciliation(t *testing.T) {
 	nt.WaitForRepoSyncs()
 
 	var pod1, pod2 *corev1.Pod
-	_, err := nomostest.Retry(nt.DefaultWaitTimeout, func() error {
+	_, err := nomostest.Retry(30*time.Second, func() error {
```
sdowell, line 445: What's the reason for `Retry` rather than switching to watch for these?

e2e/testcases/namespace_repo_test.go
```diff
@@ -54,10 +56,12 @@ func TestNamespaceRepo_Centralized(t *testing.T) {
...
-	_, err := nomostest.Retry(15*time.Second, func() error {
-		return nt.Validate("repo-sync", bsNamespace, &v1beta1.RepoSync{},
-			hasReconcilingStatus(metav1.ConditionFalse), hasStalledStatus(metav1.ConditionFalse))
-	})
+	err := nomostest.WatchObject(nt, kinds.RepoSyncV1Beta1(), "repo-sync", bsNamespace,
+		[]nomostest.Predicate{
+			hasReconcilingStatus(metav1.ConditionFalse),
+			hasStalledStatus(metav1.ConditionFalse),
+		},
+		nomostest.WatchTimeout(30*time.Second))
```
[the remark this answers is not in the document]
karlkfi: Yeah, I used 30s as a minimum. Short timeouts tend to lead to flakey failures on contested CI machines running multiple test suites. In the future I’d like to make a constant for it AND reduce the 6m default timeout to 2m. But that seems like a different PR.

e2e/testcases/namespace_repo_test.go
```diff
@@ -295,73 +304,36 @@ func getNsReconcilerSecrets(nt *nomostest.NT, ns string) []string {
...
 	})
-	if err != nil {
-		nt.T.Errorf("RepoSync present after deletion: %v", err)
-	}
-
-	// Verify Namespace Reconciler deployment no longer present.
-	_, err = nomostest.Retry(5*time.Second, func() error {
-		return nt.ValidateNotFound(core.NsReconcilerName(namespace, configsync.RepoSyncName), v1.NSConfigManagementSystem, fake.DeploymentObject())
+	tg.Go(func() error {
+		return nomostest.WatchForNotFound(nt, kinds.Deployment(), core.NsReconcilerName(namespace, configsync.RepoSyncName), v1.NSConfigManagementSystem)
```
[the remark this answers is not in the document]
karlkfi: They were serial before, with a long first timeout. Now they’re parallel, so I just used the default timeout.

e2e/testcases/preserve_fields_test.go
karlkfi: Any change that should happen in 5s could be much longer if the apiserver is congested or the local test machine is slow.

e2e/testcases/status_enablement_test.go
karlkfi: Good catch. This a bug. I’ll fix it.

e2e/testcases/sync_ordering_test.go
karlkfi: These do some weird logic that uses the Validate to get the Pod objects and use them for additional validation. So I just left them alone. 

It’s possible there’s a way to clean them up but it wasn’t immediately obvious. They use one object to check another object, so a normal Predicate wouldn’t work unless you do them in serial.

sdowell: Ack

e2e/testcases/preserve_fields_test.go
karlkfi: I can add a 30s timeout. It seemed like the default would be ok after WaitForRepoSyncs, but the default is pretty long.

e2e/testcases/status_enablement_test.go
karlkfi: Fixed.

e2e/testcases/preserve_fields_test.go
karlkfi: Added.

PR #396, opened 2023-02-08
Add apt-get upgrade for vulnerability fix

victorpras:
Add apt-get upgrade to hydration-controller-with-shell image to address vulnerability


nan-yu (approved):
/lgtm

karlkfi:
I don't know if we really want to upgrade to the latest packages every time we build. Not only is it slow, but it's not reproducable.

Why not just update the base image?

karlkfi:
You can see that the bullseye-v1.4.2-gke.7 image is the latest and has vulnerabilities here:
https://console.cloud.google.com/gcr/images/gke-release/GLOBAL/debian-base

We should get that image updated and then update our image to point to it, if possible.

If not possible, I would accept explicitly apt-get updating the problematic package to the desired version with the known fix, but I don't think our build should be always updating to the latest version. That would make the vulnerability whack-a-mole even worse, slow down our builds, and make them unreproducable.

PR #400, opened 2023-02-09
Don't allow namespace for cluster-scope watch

karlkfi:
This validation will prevent accidentally specifying a namespace for a cluster-scoped object when using WatchObject in e2e tests.


nan-yu (approved):
/lgtm

PR #402, opened 2023-02-10
Add liveness & readiness probes to e2e git-server

karlkfi:
This should restart the container if git stops responding to new connections and delay availability until ready.


nan-yu (approved):
/lgtm

PR #403, opened 2023-02-11
use WatchForCurrentStatus for composition_test

sdowell:
The validateStatusCurrent function is called after waitForSync in the test. This assumes that all managed objects are reconciled after waitForSync, which may not be true.


nan-yu (approved):
/lgtm

PR #404, opened 2023-02-11
Fix TestDependencyWithReconciliation

karlkfi:
- Add validation for deletion ordering
- Enhance apply ordering validtion by using asynchronous watches
- Increase reconcile timeout to avoid needing to retry all the reconciliation validations. This should fix general flakiness when using cluster autoscaling, like on GKE Autopilot.
- Use multierr to perform multiple validations before failing the test.

### DIFF
--- a/e2e/testcases/sync_ordering_test.go
+++ b/e2e/testcases/sync_ordering_test.go
@@ -15,16 +15,17 @@
 package e2e
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
+	"github.com/pkg/errors"
 	"go.uber.org/multierr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/applier"
@@ -32,6 +33,7 @@ import (
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/testing/fake"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/object/dependson"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -406,7 +408,9 @@ func TestExternalDependencyError(t *testing.T) {
 }
 
 func TestDependencyWithReconciliation(t *testing.T) {
-	nt := nomostest.New(t, nomostesting.Lifecycle, ntopts.Unstructured)
+	// Increase reconcile timeout to account for slow pod scheduling due to cluster autoscaling.
+	nt := nomostest.New(t, nomostesting.Lifecycle, ntopts.Unstructured,
+		ntopts.WithReconcileTimeout(5*time.Minute))
 
 	namespaceName := "bookstore"
 	nt.T.Logf("Remove the namespace %q if it already exists", namespaceName)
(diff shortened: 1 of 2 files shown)


tiffanny29631 (commented):
LGTM, were you able to test this against an Autopilot cluster locally?

e2e/testcases/sync_ordering_test.go
```diff
@@ -438,44 +442,145 @@ func TestDependencyWithReconciliation(t *testing.T) {
...
+	var pod1DeletionTimestamp time.Time   // when pod1 is first MODIFIED to have a DeletionTimestamp
+	var pod2LastUpdateTimestamp time.Time // when pod2 is last MODIFIED before being DELETED
+	pod1DeletionPredicate := func(obj client.Object) error {
+		if obj != nil && pod1DeletionTimestamp.IsZero() && !obj.GetDeletionTimestamp().IsZero() {
+			pod1DeletionTimestamp = obj.GetDeletionTimestamp().Time
+		}
+		return nil
+	}
+	pod2LastUpdatedPredicate := func(obj client.Object) error {
+		if obj != nil && pod2LastUpdateTimestamp.IsZero() {
```
sdowell, line 522: Why do we only enter this if `pod2LastUpdateTimestamp.IsZero()` is true? I assume this means we only update this the first time we get a nonzero `LastUpdateTimestamp`, but my understanding is that we want to update this until we see the final `LastUpdateTimestamp`

e2e/testcases/sync_ordering_test.go
```diff
@@ -487,9 +592,11 @@ func TestDependencyWithReconciliation(t *testing.T) {
 	nt.WaitForRootSyncSyncError(configsync.RootSyncName, applier.ApplierErrorCode,
 		"skipped apply of Pod, bookstore/pod4: dependency apply reconcile timeout: bookstore_pod3__Pod")
 
-	_, err = nomostest.Retry(30*time.Second, func() error {
+	nt.T.Logf("Verify that pod3 is created but not ready and pod4 is not found")
+	_, err := nomostest.Retry(60*time.Second, func() error {
```
sdowell, line 596: Maybe we can use `nt.DefaultWaitTimeout` instead? Is there a reason we should use a one-off value here?

e2e/testcases/sync_ordering_test.go
```diff
@@ -516,12 +630,11 @@ func TestDependencyWithReconciliation(t *testing.T) {
...
 	nt.WaitForRepoSyncs()
 
-	_, err = nomostest.Retry(30*time.Second, func() error {
-		var errs error
-		errs = multierr.Append(errs, nt.Validate("pod5", namespaceName, &corev1.Pod{}, isPodReady))
-		errs = multierr.Append(errs, nt.Validate("pod6", namespaceName, &corev1.Pod{}, isPodReady))
-		return errs
-	})
+	nt.T.Logf("Verify that pod5 and pod6 are ready")
+	err = multierr.Append(err, nt.Validate("pod5", namespaceName, &corev1.Pod{},
```
sdowell, line 634: Do we not need any sort of retry here? I assume it may not be `CurrentStatus` immediately after `WaitForRepoSyncs`

karlkfi: With the explicit 5m reconcile timeout on this test, the pods _should_ all be current.

I increased the reconcile timeout for the whole test here to avoid needing to retry all the validations. 5m is the default timeout customers get anyway. 1m is just the e2e default.

e2e/testcases/sync_ordering_test.go
karlkfi: Fixed

karlkfi:
> were you able to test this against an Autopilot cluster locally?

Trying now

e2e/testcases/sync_ordering_test.go
sdowell: Where does the test wait for reconcile after adding `pod5`/`pod6`? My understanding is that `WaitForRepoSyncs` doesn't wait for the managed resources to reconcile

sdowell: Ah I just dug a bit. The applier in cli-utils waits for the specified ReconcileTimeout for the resources to be current?

karlkfi: The applier DOES wait for reconcile, until the reconicle timeout, there's just no RSync status field that exposes the reconcile status, only sync status. And it won't update the Synced status until the applier has returned, after it's waited for the last apply wave to reconcile or timeout.

So WaitForRepoSyncs waits for reconcile until timeout, it just doesn't fail if there are reconcile failures, unless those are in dependencies that block syncing dependents.

sdowell:
LGTM. Did you want to wait to submit until you validate?
/lgtm

e2e/testcases/sync_ordering_test.go
karlkfi: To really fix this UX, we need to add a Reconcile status condition to aggregate the status in ResourceGroup, and fix some bugs in the ResourceGroup reconcile status itself, which isn't being correctly updated by the resource-group-controller. Unfortunately, there's already a Reconciling condition that means something else (reconciler-manager updating the reconciler). So it might be a big change. I suspect we'll need it for upcoming UI enhancements and rollouts, but it hasn't been prioritized yet.

PR #405, opened 2023-02-11
Fix usages of TaskGroup in a for loop

karlkfi:
Range values are re-used in a for loop.
Use indirection to avoid using the last value for every task.

### DIFF
--- a/e2e/nomostest/reset.go
+++ b/e2e/nomostest/reset.go
@@ -195,7 +195,7 @@ func ResetRootSyncs(nt *NT, rsList []v1beta1.RootSync) error {
 		rsNN := client.ObjectKeyFromObject(rs)
 		nt.T.Logf("[RESET] Waiting for deletion of RootSync %s ...", rsNN)
 		tg.Go(func() error {
-			return WatchForNotFound(nt, kinds.RootSyncV1Beta1(), rs.Name, rs.Namespace)
+			return WatchForNotFound(nt, kinds.RootSyncV1Beta1(), rsNN.Name, rsNN.Namespace)
 		})
 	}
 	return tg.Wait()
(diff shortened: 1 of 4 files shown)


e2e/testcases/composition_test.go
```diff
@@ -373,10 +373,10 @@ func waitForSync(nt *nomostest.NT, sha1Func nomostest.Sha1Func, objs ...client.O
 func validateStatusCurrent(nt *nomostest.NT, objs ...client.Object) {
 	tg := taskgroup.New()
 	for _, obj := range objs {
+		nn := client.ObjectKeyFromObject(obj)
 		tg.Go(func() error {
 			return nomostest.WatchForCurrentStatus(
-				nt, obj.GetObjectKind().GroupVersionKind(), obj.GetName(), obj.GetNamespace(),
-			)
+				nt, obj.GetObjectKind().GroupVersionKind(), nn.Name, nn.Namespace)
```
sdowell, line 379: Does `obj.GetObjectKind().GroupVersionKind()` yield the same result as `kinds.Lookup(obj, nt.scheme)`?

karlkfi: Only if the GVK is populated on the object. For Unstructured objects GVK is always populated, but for typed objects it’s only sometimes populated, depending on source. The lookup uses the GVK if set, otherwise looks up the GVK using the Scheme’s type map.

e2e/nomostest/reset.go
```diff
@@ -282,7 +282,7 @@ func ResetRepoSyncs(nt *NT, rsList []v1beta1.RepoSync) error {
 		nn := client.ObjectKeyFromObject(obj)
 		nt.T.Logf("[RESET] Waiting for deletion of RepoSync %s ...", nn)
 		tg.Go(func() error {
-			return WatchForNotFound(nt, kinds.RepoSyncV1Beta1(), obj.Name, obj.Namespace)
+			return WatchForNotFound(nt, kinds.RepoSyncV1Beta1(), nn.Name, nn.Namespace)
```
nan-yu, line 285: line 361 also needs to be updated.

karlkfi: I'll update it for consistency, but technically it doesn't need it, because the WatchForNotFound is synchronous. The range item won't be overwritten until after it exits.

karlkfi: Actually, I'll change that loop to use the TaskGroup too

e2e/testcases/composition_test.go
karlkfi: It looks like in the case of validateStatusCurrent, it's always being called with objects from a Repository.GetAll, which delegates to Repository.GetAll, which parses to Unstructured and then converts to typed before returning, but the method it uses to convert to typed (runtime.DefaultUnstructuredConverter.FromUnstructured) leaves the GVK on the typed object.

I'll change the func to use Lookup, just in case that changes in the future, but there's a lot of tests relying on that implementation detail today.

PR #406, opened 2023-02-14
Improve clean/reset of namespaces

karlkfi:
- Use server-side filter to avoid deleting shared namespaces.
- Make it more clear which namespaces are protected and which are
  shared between tests.
- Use client-go to delete namespaces in clean, instead fo kubectl

### DIFF
--- a/e2e/nomostest/clean.go
+++ b/e2e/nomostest/clean.go
@@ -25,11 +25,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"kpt.dev/configsync/e2e"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	"kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/api/configmanagement"
 	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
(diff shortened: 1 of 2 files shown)


e2e/nomostest/reset.go
```diff
@@ -441,6 +454,19 @@ func withLabelListOption(key, value string) client.MatchingLabelsSelector {
 	return client.MatchingLabelsSelector{Selector: labelSelector}
 }
 
+func withNameNotInListOption(values ...string) client.MatchingFieldsSelector {
```
sdowell, line 457: Looks like the input arg is unused here and it's always using `sharedTestNamespaces`. Should update this to use the input arg

e2e/nomostest/reset.go
```diff
@@ -39,6 +40,32 @@ import (
...
+	testGitNamespace,
+}
+
+// protectedNamespaces is a list of namespaces that should never be deleted.
+//
+// Reset will error if these namespaces exist and have the test label.
+//
+// Individual test cleanup MUST revert these namespaces, if modified.
+// See `checkpointProtectedNamespace` for an example.
+var protectedNamespaces = func() []string {
```
sdowell, line 60: I didn't mind having these arrays returned by a function as it ensures you always get a pseudo-immutable value (even if it's a bit suboptimal), but I don't feel too strongly about it

e2e/nomostest/clean.go
```diff
@@ -218,15 +220,66 @@ func Clean(nt *NT, failOnError FailOnError) {
...
+	return DeleteNamespaces(nt, nsListItems)
+}
+
+// DeleteNamespaces cleans up one or more Namespaces and all their namespaced objects.
+// Use this for bulk deletion of Namespaces.
+func DeleteNamespaces(nt *NT, nsList []corev1.Namespace) error {
+	if len(nsList) == 0 {
+		return nil
+	}
+	nt.T.Logf("[CLEANUP] Deleting Namespaces (%d)", len(nsList))
```
sdowell, line 264: Consider moving this line above the zero length check so we know if the array was zero length from the test logs

e2e/nomostest/reset.go
karlkfi: I suspect it's less important for a private variable where you control all usages. But yeah, it's sort of a language failing that there's no immutable collections you can use as constant.

Java users fixed this problem by adding libraries for ImmutableList and such, but that's hard to do in Go where slices and arrays aren't structs and you can't override their methods or provide alternate impls for their interfaces.

e2e/nomostest/reset.go
karlkfi: good catch

e2e/nomostest/clean.go
karlkfi: Done

e2e/nomostest/reset.go
karlkfi: Fixed

PR #408, opened 2023-02-14
fix edge case in namespace safegaurd logic

harpratap:
The current logic for namespace safeguard deletion only checks for the length of the namespaces to see if all the namespaces are being deleted. But it should check for the contents instead. Reason being, there is an edge case where if you simply replace the namespaces with new ones it would still end up deleting all the previous namespaces. 

I think the test case I added would explain the bug more eloquently

### DIFF
--- a/cmd/nomoserrors/examples/examples.go
+++ b/cmd/nomoserrors/examples/examples.go
@@ -339,7 +339,7 @@ func Generate() AllExamples {
 
 	// 2006
 	result.add(status.EmptySourceError(10, "namespaces"))
-	result.add(declared.DeleteAllNamespacesError([]string{"shipping", "billing"}))
+	result.add(declared.DeleteAllNamespacesError(map[string]struct{}{"shipping": {}, "billing": {}}))
 
 	// 2007 is Deprecated.
 	result.markDeprecated("2007")
--- a/pkg/declared/namespace_safeguard.go
+++ b/pkg/declared/namespace_safeguard.go
@@ -26,25 +26,28 @@ import (
 //
 // Returns an error if:
 // 1) previous contains more than 1 Namespace.
-// 2) current contains zero Namespaces.
+// 2) intersection of current and previous namespaces is zero.
 //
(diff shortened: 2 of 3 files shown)


karlkfi:
What use case is this supposed to protect the use from?

The existing namespace deletion protection isn't really something we want to expand or emulate, because it violates the principle of least surprise, and makes it a two-step process to delete all your resources.

harpratap:
@karlkfi We recently hit an edge case where we suspect because of partial read of configs from git it ended up wiping out 75% namespaces from the cluster. We are already in talks with GCP support and the preventive measures for that happening are supposed to land in 1.14.3. However, there is still an edge case of that edge case where because of race condition it ends up picking up only the new namespaces to be created partially. This is supposed to prevent that from happening. 

Also, there is another bug which we haven't been able to reproduce or debug. The log for `KNV2006` listed out `default` namespace but we never created `default` using config-sync, so ideally it shouldn't have this namespace in the list of `previousNamespaces`. This will prevent it from happening (see the second test case added in this PR)

karlkfi:
Sorry to be slow getting back to this. I'm also sorry you ran into issues. 

We think that the issue you ran into on an old version of Config Sync (race between git-sync and reconciler) has been already fixed in more recent versions.

We also found a similar race condition between the git-sync and hydration-controller sidecars and are working on a fix: https://github.com/GoogleContainerTools/kpt-config-sync/pull/415

As for this change... I'm generally against adding more show stopping validation that requires two-steps to make a change. I think adding the existing namespace deletion validation was probably a mistake and would rather not use it as precedent to add more. They both interrupt the normal declarative config management workflow by adding an imperative step, which can be both surprising and annoying to automate around. 

I think in general, using any GitOps workflow requires validating the behavior of the automation. The typical way to do that is to sync to an ephemeral or staging environment first, validate the result, and then promote the change to the production environment. This would help insulate you from accidental config change mistakes as well as any previously unknown automation bugs, reducing the severity of both.

PR #409, opened 2023-02-15
E2E Clean enhancements

karlkfi:
- Make Clean return an error, so the caller can decide to error or fatal.
- Delete serially. Watch for Not Found in parallel.
- Error if trying to delete protected namespaces in deleteManagedNamespaces or deleteImplicitNamespaces.
- Skip deleting objects that we already know are terminating. Just wait for them to be not found.
- Avoid using kubectl where possible. Using client-go gives better error messages and allows us to use common functions to avoid common mistakes. It should also be slightly faster.
- Always delete KubeVirt resources, even if the namespace is already deleted.
- Remove legacy references to mono-repo Syncs
- Add a simple kubevirt package to avoid needing to import the SDK
- Move nt.MustKubectl logging into nt.KubectlContext to ensure all calls get logged.
- Use exec.CommandContext to ensure commands are killed when the test is cancelled.
- Update k8s.io/kube-aggregator to use the latest APIService object


e2e/nomostest/clean.go
```diff
@@ -372,198 +313,207 @@ func resetSystemNamespaces(nt *NT, failOnError FailOnError) {
...
-	err := nt.Client.List(nt.Context, list, opts...)
-	if err != nil && !meta.IsNoMatchError(err) && !apierrors.IsNotFound(err) {
-		// Ignore cases where the type doesn't exist on the cluster. Obviously
-		// there isn't anything to clean in that case.
+	if err := nt.List(list, withLabelListOption(TestLabel, TestLabelValue)); err != nil {
+		if meta.IsNoMatchError(err) {
+			// Resource not registered. So none can exist.
+			return nil, nil
+		}
 		return nil, errors.Wrapf(err, "unable to clean objects of type %v from cluster", gvk)
```
sdowell, line 465: Granted this error message was already here, but is this a precise error message? Seems like this error is related to listing rather than cleaning

e2e/nomostest/clean.go
```diff
@@ -372,198 +313,207 @@ func resetSystemNamespaces(nt *NT, failOnError FailOnError) {
...
 		}
-		if failOnError {
-			nt.T.Fatalf("failed to get the Kubevirt namespace", err)
-		} else {
-			nt.T.Logf("failed to get the Kubevirt namespace", err)
+	} else {
+		// Remove finalizers on KubeVirt object to skip cleanup, if it exists.
+		// WARNING: This may leave behind GCP resources on GKE clusters that
+		// support nested virtualization, like those with N2 nodes.
+		// TODO: Find a safer way to clean up KubeVirt...
```
sdowell, line 357: Maybe a better solution would be to use a different resource type in our test case 🤷

e2e/nomostest/clean.go
```diff
@@ -244,42 +171,58 @@ func DeleteManagedNamespaces(nt *NT) error {
...
 	if len(protectedNamespacesWithTestLabel) > 0 {
-		return errors.Errorf("protected namespace(s) managed by config sync and cannot be cleaned: %+v",
+		return errors.Errorf("protected namespace(s) still have config sync labels and cannot be cleaned: %+v",
+			protectedNamespacesWithTestLabel)
+	}
+	return deleteNamespaces(nt, nsListItems)
+}
+
+// deleteImplicitNamespaces deletes the namespaces with the PreventDeletion annotation.
+func deleteImplicitNamespaces(nt *NT) error {
```
sdowell: nit: looks like the order of this function was changed which makes the diff larger than it needs to be

e2e/nomostest/clean.go
```diff
@@ -372,198 +313,207 @@ func resetSystemNamespaces(nt *NT, failOnError FailOnError) {
...
+		if !obj.GetDeletionTimestamp().IsZero() {
+			nt.T.Logf("[CLEANUP] already terminating %s object %s ...", gvk.Kind, nn)
+		} else {
+			nt.T.Logf("[CLEANUP] deleting %s object %s ...", gvk.Kind, nn)
+			if err := nt.Delete(obj, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+				if apierrors.IsNotFound(err) {
+					// skip waiting
+					continue
+				}
+				return errors.Wrapf(err, "unable to delete %s object %s",
```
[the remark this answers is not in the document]
karlkfi: All of the clean deletion functions have this same problem. Some were trying to continue and some were exiting early. But trying to make it continue is pointless, because we don't want to run the tests if the test environment is dirty. If the environment is dirty we need to manually clean it up enough for Clean to finish the job.

So instead of continuing after some but not all of them, I made them all terminal. 

If Clean errors before the environment setup, it's fatal.
If Clean errors after the tests run, it's an error which will fail the suite after other registered Cleanups execute.

e2e/nomostest/clean.go
karlkfi: I'll just make it return err.

e2e/nomostest/clean.go
```diff
@@ -372,198 +313,207 @@ func resetSystemNamespaces(nt *NT, failOnError FailOnError) {
...
-		nt.MustKubectl("delete", "crd", "-l", "app.kubernetes.io/component=kubevirt")
+		elapsed := time.Since(start)
+		nt.T.Logf("[CLEANUP] Cleaning up kubevirt resources took %v", elapsed)
 	}()
+
 	kubevirtNS := corev1.Namespace{}
 	if err := nt.Get("kubevirt", "", &kubevirtNS); err != nil {
-		if apierrors.IsNotFound(err) {
-			return
+		if !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
karlkfi: The problem is that this is only checking for the namespace. If the namespace were gone, it wouldn't delete the other cluster-scoped resources. So I made it continue instead.

e2e/nomostest/clean.go
karlkfi: All I did was refactor what we have here and add a warning. If you want to take a stab at not using KubeVirt in our tests, be my guest.

e2e/nomostest/clean.go
karlkfi: Yeah, I moved it up with the other delete and delete namespace functions.

e2e/nomostest/clean.go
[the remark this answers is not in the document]
karlkfi: The short version: Clean always needs to pass for a test to run. If Clean fails, it always runs again before tests can run. So Clean exiting early is fine.

e2e/nomostest/clean.go
karlkfi: Done.

e2e/nomostest/clean.go
karlkfi: Perhaps that was snarky. Sorry. What I meant was: I agree with you but I don't want to do it right now. If you feel inspired, feel free.

karlkfi:
/retest

Seemingly unrelated test error:
```
=== CONT  TestReconcilerFinalizer_Foreground
    watch.go:66: WatchObject(Deployment config-management-system/admission-webhook) watched for 12m0.000753461s
    new.go:392: ERROR: waiting for ConfigSync Deployments to become available: WatchObject(Deployment config-management-system/admission-webhook) timed out waiting for predicates: expected Deployment config-management-system/admission-webhook to have status "Current", but got "InProgress"
```

e2e/nomostest/clean.go
sdowell: I wasn't saying you need to do this for this PR, I still see your change as an improvement. I was just being a bit conversational with your TODO

e2e/nomostest/clean.go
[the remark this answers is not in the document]
sdowell: To be clear, I'm not saying to continue - just to wait for the already deleted resources to be NotFound. But perhaps it doesn't make much difference assuming we exit anyway after seeing this error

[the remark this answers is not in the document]
karlkfi: I think I prefer exiting early/fast so we get the error faster and can triage faster. If the environment is dirty, it doesn't really matter how dirty.

PR #411, opened 2023-02-15
Include command output when metrics curl errors

karlkfi:


e2e/nomostest/metrics/parse.go
```diff
@@ -27,9 +27,12 @@ import (
...
 func ParseMetrics(port int) (ConfigSyncMetrics, error) {
 	var out []byte
-	out, err := exec.Command("curl", "-s", fmt.Sprintf("localhost:%d/metrics", port)).CombinedOutput()
+	address := fmt.Sprintf("localhost:%d/metrics", port)
+	out, err := exec.Command("curl", "-s", address).CombinedOutput()
 	if err != nil {
-		return nil, errors.Errorf("error parsing metrics from port %d: %v", port, err)
+		return nil, errors.Errorf(
+			"error fetching metrics (%s): %v: combined output:\n%s",
+			address, err, string(out))
```
sdowell, line 35: IIRC I tried testing something similar locally and found that `curl` didn't provide any meaningful output to stdout/stderr alongside the exit code

karlkfi: Fair. I realized the exit code 7 means connection issue. So it probably doesn't print much more without verbose output enabled, which would break our parsing.

karlkfi: I suspect the real solution here is to actually use prometheus, instead of polling the metrics endpoint in a loop.

tiffanny29631 (approved):
/lgtm

PR #413, opened 2023-02-22
Skip .spec.ports check for ExternalName Service

tiffanny29631:
ExternalName Service is not required to have ports specified.
This change explicitly filters this type out, while maintaining
the check for other Services.

### DIFF
--- a/pkg/validate/raw/hydrate/declared_field_hydrator.go
+++ b/pkg/validate/raw/hydrate/declared_field_hydrator.go
@@ -117,8 +117,9 @@ func encodeDeclaredFields(converter *declared.ValueConverter, obj runtime.Object
 }
 
 // setDefaultProtocol sets the nested protocol field in anything containing
-// an array of Ports.
-// TODO: This should be deleted once we've upgraded to k8s 1.21 libraries.
+// an array of Ports. This function is required in OpenAPI v2 to fulfill the
+// missing defaults.
+// TODO: This should be deleted once OpenAPI v3 is available
 func setDefaultProtocol(u *unstructured.Unstructured) status.MultiError {
 	var errs []error
 	switch u.GroupVersionKind().GroupKind() {
@@ -134,7 +135,7 @@ func setDefaultProtocol(u *unstructured.Unstructured) status.MultiError {
 	case kinds.CronJob().GroupKind():
 		errs = setDefaultProtocolInNestedPodSpec(u.Object, "spec", "jobTemplate", "spec", "template", "spec")
 	case kinds.Service().GroupKind():
-		errs = setDefaultProtocolInNestedPorts(u.Object, true, "spec", "ports")
+		errs = setDefaultProtocolInNestedPorts(u.Object, "spec", "ports")
 	}
 
 	if len(errs) > 0 {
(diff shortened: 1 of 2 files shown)


e2e/testinfra/terraform/prow/.terraform.lock.hcl
```diff
@@ -0,0 +1,40 @@
+# This file is maintained automatically by "terraform init".
```
karlkfi, line 1: Is this supposed to be committed?

log
```diff
@@ -0,0 +1,41 @@
+I0213 22:23:53.464215       1 logr.go:261] controller-runtime/metrics "msg"="Metrics server is starting to listen" "addr"=":8080"
```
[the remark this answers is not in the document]
tiffanny29631: Please ignore .. I included an unexpected commit

e2e/testinfra/terraform/prow/.terraform.lock.hcl
tiffanny29631: No

manifests/templates/otel-collector.yaml
```diff
@@ -40,9 +41,205 @@ data:
       extensions: [health_check]
       pipelines:
         metrics:
+          receivers: [opencensus]
+          processors: [batch]
+          exporters: [logging]
+---
+apiVersion: v1
```
karlkfi, line 48: Why is this ConfigMap here if it's already in a constant in `otel_collectore.go`?

Isn't it supposed to be optional, only added for GCP clusters with workload identity enabled?

tiffanny29631: Ignore this file too :]

pkg/validate/raw/hydrate/declared_field_hydrator.go
```diff
@@ -134,6 +134,13 @@ func setDefaultProtocol(u *unstructured.Unstructured) status.MultiError {
 	case kinds.CronJob().GroupKind():
 		errs = setDefaultProtocolInNestedPodSpec(u.Object, "spec", "jobTemplate", "spec", "template", "spec")
 	case kinds.Service().GroupKind():
+		serviceType, found, err := unstructured.NestedString(u.Object, "spec", "type")
+		if err != nil {
+			errs = append(errs, err)
```
janetkuo, line 139: Will the error here be kept?

tiffanny29631: It should get returned together with the MultiError

pkg/validate/raw/hydrate/declared_field_hydrator.go
```diff
@@ -134,6 +134,13 @@ func setDefaultProtocol(u *unstructured.Unstructured) status.MultiError {
...
+		}
+		if found && serviceType == "ExternalName" {
+			break
+		}
```
janetkuo, line 143: Actually, Kubernetes doesn't say `Ports` is a required fields for `Service`: https://github.com/kubernetes/kubernetes/blob/fb57080e069fe2efe11734a4570a92873e673c35/pkg/apis/core/v1/defaults.go#L114-L122 

Therefore, please fix `func setDefaultProtocolInNestedPorts` instead. The part that assumes `ports` must be set in `Service` isn't correct.

janetkuo (changes requested):

pkg/validate/raw/hydrate/declared_field_hydrator.go
tiffanny29631: So instead we don't need to check ports here for services? Can make the change

janetkuo: Yes. The logic should be similar to the LOC I linked. If `Ports` isn't specified it will just return.

tiffanny29631: Okie removed the port check for Service, updated the commit message and the TODO comment as well.

janetkuo (approved):
/lgtm

pkg/validate/raw/hydrate/declared_field_hydrator.go
janetkuo: Don't forget to update PR description as well

Issue #226, opened 2022-10-20
Is `nomos` supposed to work when using OSS CS?


mathieu-benoit:
If I install CS like this:
```
CONFIG_SYNC_VERSION=1.13.0
kubectl apply \
    -f https://github.com/GoogleContainerTools/kpt-config-sync/releases/download/v${CONFIG_SYNC_VERSION}/config-sync-manifest.yaml \
    --wait
```
And then deploy my `RootSyncs` and `RepoSyncs`, my resources are successfully synced, etc. no problem.

But if I run `nomos status --contexts $(k config current-context)`, I'm getting this error:
```
NOT INSTALLED   The ACM operator is neither installed in the "kube-system" namespace nor the "config-management-system" namespace
```
Is `nomos` supposed to work when installing/using OSS CS?

I'm locally with `nomos --version`: `v1.13.0-rc.7`, that I installed like this:
```
curl -L https://github.com/GoogleContainerTools/kpt-config-sync/releases/download/v${CONFIG_SYNC_VERSION}/nomos_linux_amd64 > nomos
chmod +x nomos
```

sdowell:
Thank you for your post!

It appears that `nomos status` is currently designed to use the ACM operator. I think it would be a reasonable feature request to make it work with OSS Config Sync

PR #414, opened 2023-02-23
Enable nomos status command in OSS

victorpras:
From Github issue https://github.com/GoogleContainerTools/kpt-config-sync/issues/226

Currently, nomos status will check for Operator pods and ConfigManagement resource, so it wouldn't work in OSS Config Sync installation.

To get nomos status to work in OSS, the function IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and RSync CRD.  If RSync CRD exist but ConfigManagement and Operator doesn't, it indicates an OSS installation

### DIFF
--- a/cmd/nomos/status/client.go
+++ b/cmd/nomos/status/client.go
@@ -56,7 +56,6 @@ const (
 	ACMOperatorDeployment            = "config-management-operator"
 	syncingConditionSupportedVersion = "v1.10.0-rc.1"
 	rootSyncCRDName                  = "rootsyncs.configsync.gke.io"
-	repoSyncCRDName                  = "reposyncs.configsync.gke.io"
 )
 
 // ClusterClient is the client that talks to the cluster.
@@ -378,7 +377,7 @@ func (c *ClusterClient) IsOssInstallation(ctx context.Context, cs *ClusterState)
 	_, operatorDepErr := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, ACMOperatorDeployment, metav1.GetOptions{})
 	if operatorDepErr != nil && !apierrors.IsNotFound(operatorDepErr) {
 		err := fmt.Errorf("Failed to get the Operator Deployment: %v", operatorDepErr)
-		cs.Error = err.Error() 
+		cs.Error = err.Error()
 		return false, err
 	}
 
@@ -390,7 +389,7 @@ func (c *ClusterClient) IsOssInstallation(ctx context.Context, cs *ClusterState)
 	if rootSyncCRDErr == nil {
 		return true, nil
 	}
-	err := fmt.Errorf("Failed to get the RootSync CRD: %v", rootSyncCRDErr) 
+	err := fmt.Errorf("Failed to get the RootSync CRD: %v", rootSyncCRDErr)
 	cs.Error = err.Error()
 	return false, err
 }
--- a/cmd/nomos/status/client.go
+++ b/cmd/nomos/status/client.go
@@ -28,6 +28,7 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -54,6 +55,8 @@ const (
 	// ACMOperatorDeployment is the name of the ACM operator Deployment.
 	ACMOperatorDeployment            = "config-management-operator"
 	syncingConditionSupportedVersion = "v1.10.0-rc.1"
+	rootSyncCRDName                  = "rootsyncs.configsync.gke.io"
+	repoSyncCRDName                  = "reposyncs.configsync.gke.io"
 )
 
 // ClusterClient is the client that talks to the cluster.
@@ -139,7 +142,16 @@ func (c *ClusterClient) resourceGroups(ctx context.Context, ns string, nsAndName
 // clusterStatus returns the ClusterState for the cluster this client is connected to.
 func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace string) *ClusterState {
 	cs := &ClusterState{Ref: cluster}
-	var err error
+	isOss, err := c.IsOssInstallation(ctx, cs)
+	if err != nil {
+		return cs
+	}
+
+	if isOss {
+		c.multiRepoClusterStatus(ctx, cs)
+		return cs
+	}
+
 	cs.isMulti, err = c.ConfigManagement.IsMultiRepo(ctx)
 
 	if !c.IsInstalled(ctx, cs) {
@@ -352,6 +364,35 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
 		cs.status = util.UnknownMsg
 		cs.Error = "No RepoSync resources found"
 	}
+}
+
+// IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and RootSync CRD
+// If RootSync CRD exist but ConfigManagement and Operator doesn't, it indicates an OSS installation
+func (c *ClusterClient) IsOssInstallation(ctx context.Context, cs *ClusterState) (bool, error) {
+	v, cmErr := c.ConfigManagement.Version(ctx)
+	if cmErr != nil {
+		err := fmt.Errorf("Failed to get the ConfigManagment version: %v", cmErr)
+		cs.Error = err.Error()
+		return false, err
+	}
+	_, operatorDepErr := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, ACMOperatorDeployment, metav1.GetOptions{})
+	if operatorDepErr != nil && !apierrors.IsNotFound(operatorDepErr) {
+		err := fmt.Errorf("Failed to get the Operator Deployment: %v", operatorDepErr)
+		cs.Error = err.Error() 
+		return false, err
+	}
+
+	if v != util.NotInstalledMsg && operatorDepErr == nil {
+		return false, nil
+	}
+
+	rootSyncCRDErr := c.Client.Get(ctx, client.ObjectKey{Name: rootSyncCRDName}, &apiextensionsv1.CustomResourceDefinition{})
+	if rootSyncCRDErr == nil {
+		return true, nil
+	}
+	err := fmt.Errorf("Failed to get the RootSync CRD: %v", rootSyncCRDErr) 
+	cs.Error = err.Error()
+	return false, err
 }
 
 // IsInstalled returns true if the ClusterClient is connected to a cluster where
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -1034,3 +1034,19 @@ func checkFileExists(prefix string, targetFiles, allFiles []string) status.Multi
 	}
 	return err
 }
+
+func TestNomosStatus(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI)
+
+	// get status
+	cmd := nt.Command("nomos", "status")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		nt.T.Log(string(out))
+		nt.T.Fatal(err)
+	}
+
+	if !strings.Contains(string(out), "git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync/acme@main") {
+		nt.T.Fatalf("Expected to find Git provider string in output:\n%s\n", string(out))
+	}
+}


cmd/nomos/status/client.go
```diff
@@ -354,6 +358,18 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
 	}
 }
 
+// IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and root-reconciler deployment
+// If root-reconciler exist but ConfigManagement and Operator doesn't, it indicates an OSS installation
+func (c *ClusterClient) IsOssInstallation(ctx context.Context) bool {
+	v, _ := c.ConfigManagement.Version(ctx)
+	_, operatorDeployment := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, ACMOperatorDeployment, metav1.GetOptions{})
+	_, rootDeployment := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, "root-reconciler", metav1.GetOptions{})
```
sdowell, line 366: I don't think `root-reconciler` necessarily needs to exist, I would suggest checking either for the RSync CRDs or for the existence of `reconciler-manager`

cmd/nomos/status/client.go
```diff
@@ -139,6 +139,10 @@ func (c *ClusterClient) resourceGroups(ctx context.Context, ns string, nsAndName
 // clusterStatus returns the ClusterState for the cluster this client is connected to.
 func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace string) *ClusterState {
 	cs := &ClusterState{Ref: cluster}
+	if c.IsOssInstallation(ctx) {
```
sdowell, line 142: Please add e2e testcase(s) to verify/document this behavior

sdowell (commented):
Left some comments. Also you can link to the github issue in the PR description

cmd/nomos/status/client.go
```diff
@@ -354,6 +358,18 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
 	}
 }
 
+// IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and root-reconciler deployment
+// If root-reconciler exist but ConfigManagement and Operator doesn't, it indicates an OSS installation
+func (c *ClusterClient) IsOssInstallation(ctx context.Context) bool {
+	v, _ := c.ConfigManagement.Version(ctx)
```
nan-yu, line 364: The error returned by `c.ConfigManagement.Version` is swallowed. You need to better handle the error here.

cmd/nomos/status/client.go
```diff
@@ -354,6 +358,18 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
...
+	_, operatorDeployment := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, ACMOperatorDeployment, metav1.GetOptions{})
```
nan-yu, line 365: `operatorDeployment` is actually the get error. Can you rename the variable to make it clear?

cmd/nomos/status/client.go
nan-yu: +1. We need to check the RSync CRD as you proposed, not the reconciler.

cmd/nomos/status/client.go
```diff
@@ -354,6 +358,18 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
...
+	_, rootDeployment := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, "root-reconciler", metav1.GetOptions{})
+	if operatorDeployment != nil && v == util.NotInstalledMsg && rootDeployment == nil {
```
nan-yu, line 367: `operatorDeployment != nil`: it returns true only when it is not-found error. For other types of error, like connection error, we would need to handle it differently.

cmd/nomos/status/client.go
```diff
@@ -139,6 +141,11 @@ func (c *ClusterClient) resourceGroups(ctx context.Context, ns string, nsAndName
 // clusterStatus returns the ClusterState for the cluster this client is connected to.
 func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace string) *ClusterState {
 	cs := &ClusterState{Ref: cluster}
+	fmt.Println(cluster)
```
victorpras, line 144: need to remove debugging log

cmd/nomos/status/client.go
victorpras: Updated validation to check for RootSync CRD

cmd/nomos/status/client.go
victorpras: renamed variable to be clearer

cmd/nomos/status/client.go
victorpras: updated to check that the error that we want for oss is NotFound error.

cmd/nomos/status/client.go
victorpras: added error handling to ConfigManagement and made sure that the message that we need for OSS is `NOT INSTALLED`

cmd/nomos/status/client.go
```diff
@@ -354,6 +361,29 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
...
+// If RootSync CRD exist but ConfigManagement and Operator doesn't, it indicates an OSS installation
+func (c *ClusterClient) IsOssInstallation(ctx context.Context, cs *ClusterState) bool {
+	rootSyncCRD := &apiextensionsv1.CustomResourceDefinition{}
+	repoSyncCRD := &apiextensionsv1.CustomResourceDefinition{}
+	rootSyncCRDErr := c.Client.Get(ctx, client.ObjectKey{Name: rootSyncCRDName}, rootSyncCRD)
+	repoSyncCRDErr := c.Client.Get(ctx, client.ObjectKey{Name: repoSyncCRDName}, repoSyncCRD)
+	if apierrors.IsNotFound(rootSyncCRDErr) && apierrors.IsNotFound(repoSyncCRDErr) {
+		cs.Error = "RootSync CRD and RepoSync CRD are not found"
+		return false
+	}
```
nan-yu, line 374: It is wrong to check the R*Sync CRD first as it might use the mono-repo mode.

You need to check the CM object and the operator Deployment first.

- If CM and operator exist => return false
- If CM or operator doesn't exist
   - If R*Sync CRD exist => return true to indicate OSS installation
   - else => return installation error
- If failed to get CM, or operator, or R*Sync CRD => return get error

victorpras: thanks Nan! The logic has been updated to follow your suggestion

cmd/nomos/status/client.go
```diff
@@ -354,6 +361,34 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
...
 
+// IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and RootSync CRD
+// If RootSync CRD exist but ConfigManagement and Operator doesn't, it indicates an OSS installation
+func (c *ClusterClient) IsOssInstallation(ctx context.Context, cs *ClusterState) bool {
+	v, cmErr := c.ConfigManagement.Version(ctx)
+	_, operatorDepErr := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, ACMOperatorDeployment, metav1.GetOptions{})
+
+	if cmErr != nil || (operatorDepErr != nil && !apierrors.IsNotFound(operatorDepErr)) {
+		cs.Error = "Failed to check for operator installation status"
+		return false
```
nan-yu, line 372: Here you return `false`, but the caller will continue executing, which is not necessary because you've already hit some unexpected error.
I would suggest update the `IsOssInstallation` function to return a boolean and an error.
If the error is nil, the caller can return directly.

cmd/nomos/status/client.go
```diff
@@ -354,6 +361,34 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
...
 }
 
+// IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and RootSync CRD
+// If RootSync CRD exist but ConfigManagement and Operator doesn't, it indicates an OSS installation
+func (c *ClusterClient) IsOssInstallation(ctx context.Context, cs *ClusterState) bool {
+	v, cmErr := c.ConfigManagement.Version(ctx)
+	_, operatorDepErr := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, ACMOperatorDeployment, metav1.GetOptions{})
+
+	if cmErr != nil || (operatorDepErr != nil && !apierrors.IsNotFound(operatorDepErr)) {
+		cs.Error = "Failed to check for operator installation status"
```
nan-yu, line 371: Suggest including the error in `cs.Error` to provide more details.

cmd/nomos/status/client.go
```diff
@@ -354,6 +361,34 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
...
+		return false
+	}
+
+	if v == util.NotInstalledMsg || apierrors.IsNotFound(operatorDepErr) {
+		rootSyncCRDErr := c.Client.Get(ctx, client.ObjectKey{Name: rootSyncCRDName}, &apiextensionsv1.CustomResourceDefinition{})
+		repoSyncCRDErr := c.Client.Get(ctx, client.ObjectKey{Name: repoSyncCRDName}, &apiextensionsv1.CustomResourceDefinition{})
+		if rootSyncCRDErr == nil || repoSyncCRDErr == nil {
+			return true
+		}
+		cs.Error = "Failed to check for Rootsync and Reposync"
```
nan-yu, line 385: The error message doesn't make sense if `rootSyncCRDErr` or `repoSyncCRDErr` is not_found error.

cmd/nomos/status/client.go
victorpras: Updated error message to include more details

cmd/nomos/status/client.go
victorpras: Updated error message to include more details

cmd/nomos/status/client.go
victorpras: Updated isOss to return bool and error, updated the return condition

cmd/nomos/status/client.go
```diff
@@ -140,6 +143,16 @@ func (c *ClusterClient) resourceGroups(ctx context.Context, ns string, nsAndName
 func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace string) *ClusterState {
 	cs := &ClusterState{Ref: cluster}
 	var err error
```
nan-yu, line 145: nit: `var err error` is not needed as you declared `isOss, err :=`.

cmd/nomos/status/client.go
```diff
@@ -354,6 +367,38 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
 	}
 }
 
+// IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and R*Sync CRD
```
nan-yu, line 370:
```

// IsOssInstallation will check for the existence of ConfigManagement object, Operator deployment, and R*Sync CRD
// If R*Sync CRD exist but ConfigManagement and Operator doesn't, it indicates an OSS installation
func (c *ClusterClient) IsOssInstallation(ctx context.Context, cs *ClusterState) (bool, error) {
	v, cmErr := c.ConfigManagement.Version(ctx)
	if cmErr != nil {
		err := fmt.Errorf("failed to get the ConfigManagment version: %v", cmErr)
		cs.Error = err.Error() // Include cmErr in cs.Error to display more details
		return false, err      // return here when cmErr != nil so that we don't need other API calls
	}

	_, operatorDepErr := c.K8sClient.AppsV1().Deployments(configmanagement.ControllerNamespace).Get(ctx, ACMOperatorDeployment, metav1.GetOptions{})
	if operatorDepErr != nil && !apierrors.IsNotFound(operatorDepErr) {
		err := fmt.Errorf("failed to get the Operator Deployment: %v", operatorDepErr)
		cs.Error = err.Error() // Include operatorDepErr in cs.Error to display more details
		return false, err
	}

	if v != util.NotInstalledMsg && operatorDepErr == nil {
		return false, nil
	}

	// You've handled the error cases, and `v != util.NotInstalledMsg && operatorDepErr == nil`,
	// So the if condition is not needed anymore.
	// if v == util.NotInstalledMsg || apierrors.IsNotFound(operatorDepErr) {
	rootSyncCRDErr := c.Client.Get(ctx, client.ObjectKey{Name: rootSyncCRDName}, &apiextensionsv1.CustomResourceDefinition{})
	// I feel like checking either RootSync CRD or RepoSync CRD should be sufficient
	//repoSyncCRDErr := c.Client.Get(ctx, client.ObjectKey{Name: repoSyncCRDName}, &apiextensionsv1.CustomResourceDefinition{})
	if rootSyncCRDErr == nil {
		return true, nil
	}
	// The above if condition has a return statement, so you don't need `else` and the indentation.
	err := fmt.Errorf("failed to get the RootSync CRD: %v", rootSyncCRDErr) // If you include the original error in cs.Error, you don't need to distinguish between the not_found error and other errors.
	cs.Error = err.Error()
	return false, err
}
```

cmd/nomos/status/client.go
victorpras: removed!

cmd/nomos/status/client.go
victorpras: updated to follow this feedback

nan-yu (approved):
/lgtm

e2e/testcases/cli_test.go
```diff
@@ -1034,3 +1034,19 @@ func checkFileExists(prefix string, targetFiles, allFiles []string) status.Multi
...
+
+	// get status
+	cmd := nt.Command("nomos", "status")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		nt.T.Log(string(out))
+		nt.T.Fatal(err)
+	}
+
+	if !strings.Contains(string(out), "git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync/acme@main") {
```
nan-yu, line 1049: This breaks the CI jobs running against Bitbucket and Gitlab as the URLs are not the test-git-server. 
Link: https://oss.gprow.dev/view/gcs/oss-prow-build-kpt-config-sync/logs/multi-repo-1-standard-regular-bitbucket/1630776205252759552.

@victorpras Please file a RP to fix it.

PR #415, opened 2023-02-24
Fix race condition

tiffanny29631:
In a rare condition hydration-controller and parser may proceed with file hydration or listing when git-sync sees and updates the commit hash in status, which causes resource mismatch and accidental deletion.

This change resolves the risk by double checking if commit has changed during hydration or listing.

go/cs-race-condition


pkg/parse/source.go
```diff
@@ -96,6 +96,9 @@ func (o *files) readConfigFiles(state *sourceState) status.Error {
 	if err != nil {
 		return status.PathWrapError(errors.Wrap(err, "listing files in the configs directory"), syncDir.OSPath())
 	}
+	if state.commit != commit {
```
sdowell, line 99: AFAICT this is still a race condition. The suggested fix was to pass the symlink rather than the real path into this function, then check if the real path changed after listing the files

pkg/hydrate/controller.go
```diff
@@ -115,6 +126,10 @@ func (h *Hydrator) runHydrate(sourceCommit, syncDir string) HydrationError {
 	if err := kustomizeBuild(syncDir, dest, true); err != nil {
 		return err
 	}
+	if changed, commit := h.checkCommitChanged(sourceCommit); changed {
+		return status.ObsoleteFileError(fmt.Sprintf("Source commit has changed when running Kustomize build, was %s, now %s", sourceCommit, commit))
```
karlkfi, line 130: "Source commit changed while running kustomize build"

pkg/hydrate/controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+				PollingPeriod:   1 * time.Minute,
+				RehydratePeriod: 1 * time.Minute,
+				ReconcilerName:  "root-reconciler",
+			}
+
+			path, err := os.Getwd()
+			if err != nil {
+				log.Println(err)
+			}
+			fmt.Println(path)
```
nan-yu, line 93: This can be deleted.

pkg/parse/source.go
```diff
@@ -96,6 +96,9 @@ func (o *files) readConfigFiles(state *sourceState) status.Error {
...
+		return status.ObsoleteFileErrorf("Commit/image digest/chart version has changed when listing the files, files not updated")
```
karlkfi, line 100: "Source commit changed while listing files, was %s, now %s"

pkg/hydrate/controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+	ft "kpt.dev/configsync/pkg/importer/filesystem/filesystemtest"
+)
+
+var commitHash = "1234567890abcdef"
+var differentHash = "abcdef1234567890"
+
+func TestRunHydrate(t *testing.T) {
+	testCases := []struct {
+		name         string
+		wantedRes    bool
```
nan-yu, line 37: Suggest renaming `wantedRes` to `commitChanged`.

pkg/status/obsolete_file_error.go
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package status
+
+// PathErrorCode is the error code for a generic PathError.
+const ObsoleteFileErrorCode = "2016"
```
karlkfi, line 18: Will this error be exposed to users as a source error?
Is there any way we can catch it and suppress it, since it's likely transitory?
Maybe just log it and retry at a higher level?

pkg/status/obsolete_file_error.go
```diff
@@ -0,0 +1,30 @@
...
+
+var ObsoleteFileErrorBuilder = NewErrorBuilder(ObsoleteFileErrorCode)
+
+// ObsoleteFileError returns a ObsoleteFileError wrapping an error one or more relative paths.
+func ObsoleteFileError(message string) Error {
```
sdowell, line 23: I think something like `NewCommitDetectedError` or similar might be a bit more precise

pkg/parse/source.go
nan-yu, line 99: The commit passed in the `readConfigFiles` function is always the same as `state.commit`. You need to recompute the symlink.

pkg/parse/source.go
nan-yu, line 100: The error message will be visible to the user. Suggest making it more user-friendly.

pkg/hydrate/controller.go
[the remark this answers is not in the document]
tiffanny29631: Suggest waiting for the next sync?

pkg/status/obsolete_file_error.go
tiffanny29631: It is exposed to users as a source error. Seems like parse is updating it to the status but not the hydration controller when encountered, but both will retry at a higher level IIUC.

pkg/hydrate/controller.go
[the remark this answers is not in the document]
nan-yu: Yes, something like `Source commit changed while running Kustomize build, was %s, now %s. It will be fixed in the next sync`.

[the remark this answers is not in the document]
karlkfi: Sounds too sure of itself. Maybe just "Retry in N seconds."

karlkfi:
I don't think this is an error we need to expose to users. It should just be temporary and retried quickly. Its fine to log as an error, but I wouldn't expect the user to need to know it's happened unless it's happening a lot due to frequent commits, and even then having it in the log as a warning should be fine.

I suggest still stopping the parse/render by returning an error, but then catching it with an `errors.Is` further up and logging it, before it gets reported as as a source/rendering error. It should still trigger retry tho.

nan-yu:
> I don't think this is an error we need to expose to users. It should just be temporary and retried quickly. Its fine to log as an error, but I wouldn't expect the user to need to know it's happened unless it's happening a lot due to frequent commits, and even then having it in the log as a warning should be fine.
> 
> I suggest still stopping the parse/render by returning an error, but then catching it with an `errors.Is` further up and logging it, before it gets reported as as a source/rendering error. It should still trigger retry tho.

Yes, it makes sense. This type of error doesn't require any actions from the user side. Having them in the log sounds good.

pkg/parse/source.go
tiffanny29631: Good catch. Now passing the parser with its hydratedRoot and hydratedLink, the commit is recomputed after running listFiles then compared. Test was modified to match the behavior.

pkg/hydrate/controller.go
tiffanny29631: Done

pkg/status/obsolete_file_error.go
tiffanny29631: CommitChangedError?

pkg/hydrate/controller.go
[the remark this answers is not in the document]
tiffanny29631: Added, using verb "retry" so no hardcoded time diff has to be promised.

pkg/hydrate/controller_test.go
tiffanny29631: Deleted :p

pkg/parse/source.go
tiffanny29631: Done, addressed in an earlier reply, now passing the parser with its hydratedRoot and hydratedLink to compute the new commit if changed.

pkg/parse/source.go
tiffanny29631: Aligned with hydration controller.

pkg/hydrate/controller_test.go
tiffanny29631: Yeah

pkg/parse/source.go
tiffanny29631: Done

pkg/status/obsolete_file_error.go
tiffanny29631: Removed, using internal error instead

pkg/status/obsolete_file_error.go
tiffanny29631: Changed to use InternalError instead, the error should still get returned, logged and trigger retry

tiffanny29631:
> I don't think this is an error we need to expose to users. It should just be temporary and retried quickly. Its fine to log as an error, but I wouldn't expect the user to need to know it's happened unless it's happening a lot due to frequent commits, and even then having it in the log as a warning should be fine.

>I suggest still stopping the parse/render by returning an error, but then catching it with an errors.Is further up and logging it, before it gets reported as as a source/rendering error. It should still trigger retry tho.

Changed to use InternalError.

pkg/hydrate/controller.go
```diff
@@ -115,13 +115,28 @@ func (h *Hydrator) runHydrate(sourceCommit, syncDir string) HydrationError {
 	if err := kustomizeBuild(syncDir, dest, true); err != nil {
 		return err
 	}
+	if changed, commit := h.checkCommitChanged(sourceCommit); changed {
+		return NewInternalError(fmt.Errorf("source commit changed while running Kustomize build, was %s, now %s. It will be retried in the next sync", commit, sourceCommit))
```
nan-yu, line 119: You need to swap the order of commit and sourceCommit. `commit` represents the new commit, while sourceCommit represents the old commit.

pkg/hydrate/controller.go
```diff
@@ -115,13 +115,28 @@ func (h *Hydrator) runHydrate(sourceCommit, syncDir string) HydrationError {
...
 	return nil
 }
 
+// checkCommitChange returns true if the computed commit is different from given commit
+func (h *Hydrator) checkCommitChanged(commit string) (bool, string) {
+	absSourceDir := h.SourceRoot.Join(cmpath.RelativeSlash(h.SourceLink))
+	newCommit, _, err := SourceCommitAndDir(h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
+	if err != nil {
+		klog.Errorf("failed to get the commit hash and sync directory from the source directory %s: %v", absSourceDir.OSPath(), err)
+		return false, ""
```
nan-yu, line 135: In the error case, you return false for commit unchanged, which might result in proceeding with a new commit.
Instead of logging the error, I would suggest returning the error.

pkg/parse/source.go
```diff
@@ -96,10 +96,39 @@ func (o *files) readConfigFiles(state *sourceState) status.Error {
...
 	return nil
 }
 
+func (o *files) checkCommitChanged(p Parser, state *sourceState) (bool, string) {
+	// compute the source commit and compare with the given sourceCommit
+	opts := p.options()
+	absHydratedRoot, err := cmpath.AbsoluteOS(opts.HydratedRoot)
+	if err != nil {
+		klog.Errorf("not able to get absolute HydratedRoot while checking commit changed after listing files: %v", err)
+		return false, ""
```
nan-yu, line 115: Same here. If it failed to get the absolute path, it might proceed with the new commit.

pkg/hydrate/controller.go
tiffanny29631: Done

pkg/hydrate/controller.go
tiffanny29631: Added error in return and check

pkg/parse/source.go
tiffanny29631: Added error in return and check

Makefile.e2e
```diff
@@ -42,6 +42,7 @@ test-e2e-kind-multi-repo: config-sync-manifest-local
 	kind delete clusters --all
 	GCP_PROJECT=$(GCP_PROJECT) ./scripts/e2e-kind.sh \
 		--timeout 60m \
+		--test.run TestAcmeCorpRepo \
```
sdowell, line 45: remove

tiffanny29631: : P

pkg/parse/source.go
```diff
@@ -95,10 +95,33 @@ func (o *files) readConfigFiles(state *sourceState) status.Error {
...
 	state.files = fileList
 	return nil
 }
 
+// checkCommitChanged returns true and the updated commit if the computed commit
+// is different from the given commit in sourceState, or returns error if found
+// during computation
+func (o *files) checkCommitChanged(p Parser, state *sourceState) (bool, string, error) {
+	gs := sourceStatus{}
+	gs.commit, _, gs.errs = hydrate.SourceCommitAndDir(p.options().SourceType, p.options().SourceDir, p.options().SyncDir, p.options().reconcilerName)
```
karlkfi, line 115: Seems like this could be significantly simplified. The commit is just this:
```
gitDir, err := sourceRevDir.EvalSymlinks()
	if err != nil {
		return "", "", status.SourceError.Sprintf("unable to evaluate the source link %s: %v", sourceRevDir, err).Build()
	}

	commit := filepath.Base(gitDir.OSPath())
```

Maybe make a smaller `hydrate.SourceCommit` function.

karlkfi: Or do we care what the error file says and want that to interrupt this?

nan-yu: I like the idea. We don't need to recheck the existence of the mounted directory and the error file. All we need is to re-evaluate the symlink.

karlkfi: Anyway, my motivation here is to not pass the full parser into checkCommitChanged. You only needs the "options" to find the commit, right?

nan-yu: > Or do we care what the error file says and want that to interrupt this?

I don't think we care. It might be the error from the previous commit. If it is the error for the current commit, it will be returned in the next retry.

pkg/hydrate/controller.go
```diff
@@ -115,13 +115,32 @@ func (h *Hydrator) runHydrate(sourceCommit, syncDir string) HydrationError {
...
 	if err := updateSymlink(h.HydratedRoot.OSPath(), h.HydratedLink, newHydratedDir.OSPath()); err != nil {
 		return NewInternalError(errors.Wrapf(err, "unable to update the symbolic link to %s", newHydratedDir.OSPath()))
 	}
 	klog.Infof("Successfully rendered %s for commit %s", syncDir, sourceCommit)
 	return nil
 }
 
+// checkCommitChange returns true and the updated commit if the computed commit
+// is different from the given commit, or returns error if found during computation
+func (h *Hydrator) checkCommitChanged(commit string) (bool, string, error) {
```
karlkfi, line 135: Why is this func in both files and Hydrator? Can they be consolidated?

pkg/parse/source.go
```diff
@@ -95,10 +95,33 @@ func (o *files) readConfigFiles(state *sourceState) status.Error {
 	if err != nil {
 		return status.PathWrapError(errors.Wrap(err, "listing files in the configs directory"), syncDir.OSPath())
 	}
+
+	commitChanged, newCommit, err := o.checkCommitChanged(p, state)
+	if err != nil {
+		return status.TransientError(err)
```
sdowell, line 101: Is this necessarily a transient error? This scenario seems like an unexpected error which could potentially not be transient

pkg/hydrate/controller.go
sdowell, line 135: The boolean return value seems unnecessary here, I'd suggest just returning the commit `string` and doing the comparison in the caller. e.g.
```golang
	commit, err := h.checkCommitChanged() // potentially rename the function
	if err != nil {
		return NewTransientError(err)
	} else if commit != sourceCommit {
		return NewTransientError(fmt.Errorf("source commit changed while running Kustomize build, was %s, now %s. It will be retried in the next sync", sourceCommit, commit))
	}
```

pkg/hydrate/controller.go
```diff
@@ -115,13 +115,32 @@ func (h *Hydrator) runHydrate(sourceCommit, syncDir string) HydrationError {
 	if err := kustomizeBuild(syncDir, dest, true); err != nil {
 		return err
 	}
+
+	changed, commit, err := h.checkCommitChanged(sourceCommit)
+	if err != nil {
+		return NewTransientError(err)
```
sdowell, line 121: ditto here regarding whether this is a transient error

pkg/hydrate/controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+			commitDir := filepath.Join(tempDir, originCommit)
+			err = os.Mkdir(commitDir, os.ModePerm)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// create a symlink to point to the temporary directory
+			dir := ft.NewTestDir(t)
+			symDir := dir.Root().Join(cmpath.RelativeSlash("test-hydrate-symlink"))
+			err = os.Symlink(commitDir, symDir.OSPath())
```
sdowell, line 74: add a `defer` to clean up the symlink after the test

pkg/parse/source_test.go
```diff
@@ -0,0 +1,114 @@
...
+			commitDir := filepath.Join(tempRoot, tc.commit)
+			err := os.Mkdir(commitDir, os.ModePerm)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// create a symlink to point to the temporary directory
+			dir := ft.NewTestDir(t)
+			symDir := dir.Root().Join(cmpath.RelativeSlash("test-list-file-symlink"))
+			err = os.Symlink(commitDir, symDir.OSPath())
```
sdowell, line 75: same here, clean up the symlink

karlkfi:
```
pkg/hydrate/controller.go:134:1: exported: comment on exported function ComputeCommit should be of the form "ComputeCommit ..." (revive)
// computeCommit returns the computed commit from given sourceDir, or error
// if the sourceDir fails symbolic link evaluation
make: *** [Makefile:248: lint-go] Error 1
```

pkg/parse/source.go
tiffanny29631: Now passing the sourceRevDir only

pkg/hydrate/controller.go
tiffanny29631: Consolidated to ComputeCommit in Hydrator.

pkg/parse/source.go
```diff
@@ -95,10 +95,33 @@ func (o *files) readConfigFiles(state *sourceState) status.Error {
 	if err != nil {
 		return status.PathWrapError(errors.Wrap(err, "listing files in the configs directory"), syncDir.OSPath())
 	}
+
+	commitChanged, newCommit, err := o.checkCommitChanged(p, state)
```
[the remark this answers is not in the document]
tiffanny29631: Simplified

pkg/parse/source.go
tiffanny29631: No longer calling hydrate.sourceCommitAndDir in checkCommitChanged, so error should be transient.

pkg/hydrate/controller.go
tiffanny29631: Simplified the return values

pkg/hydrate/controller_test.go
tiffanny29631: Done

pkg/parse/source_test.go
tiffanny29631: Done

pkg/hydrate/controller.go
tiffanny29631: Simplified the nested func call so error should be transient.

pkg/hydrate/controller.go
```diff
@@ -115,13 +115,33 @@ func (h *Hydrator) runHydrate(sourceCommit, syncDir string) HydrationError {
 	if err := kustomizeBuild(syncDir, dest, true); err != nil {
 		return err
 	}
+
+	absSourceDir := h.SourceRoot.Join(cmpath.RelativeSlash(h.SourceLink))
```
sdowell, line 119: This same combination of function calls is also used on line 79 to evaluate the absolute path. Perhaps we could we move this to a helper function?

e.g.
```golang
func (h *Hydrator) absSourceDir() string {
  return h.SourceRoot.Join(cmpath.RelativeSlash(h.SourceLink))
}
```
Just in case this changes at some point in the future, we don't accidentally only change it in one place

pkg/hydrate/controller_test.go
```diff
@@ -0,0 +1,170 @@
...
+var originCommit = "1234567890abcdef"
+var differentCommit = "abcdef1234567890"
+
+func TestRunHydrate(t *testing.T) {
+	testCases := []struct {
+		name      string
+		commit    string // the commit in the hydrator object that might differ from origin commit
+		wantedErr error
+	}{
+		{
```
sdowell, line 40: Can you add a happy path test case as well like in `TestReadConfigFiles`?

pkg/hydrate/controller_test.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+			defer func() {
+				err := os.Remove(symDir.OSPath())
+				if err != nil {
+					t.Fatal(err)
+				}
+			}()
+
+			absSourceDir := absTempDir.Join(cmpath.RelativeSlash(tc.sourceCommit))
+			computed, err := ComputeCommit(absSourceDir)
```
sdowell, line 162: I think the input to this function should be `symDir`, right?

sdowell (commented):
mostly LGTM with a few suggestions

pkg/hydrate/controller.go
tiffanny29631: Added helper function

pkg/hydrate/controller_test.go
tiffanny29631: Added happy path

pkg/hydrate/controller_test.go
tiffanny29631: Fixed!

karlkfi (approved):
/lgtm

PR #417, opened 2023-02-25
Pause the remediator while the Applier is running

karlkfi:
- Fix a bug that might cause objects to be applied or deleted
  in the wrong order. This was being caused by the remediator
  detecting new changes as drift, when an object was updated
  asynchronously, before the applier had a chance to apply
  the new intended state.
  The reconciler now pauses the remediator before syncing,
  and resumes remediating after syncing is successful.
  The remediator will continue watching and enqueuing
  objects to remediate when changed while paused. This
  way no drift is missed while other objects are being applied.
- Fix a bug where sync retries might use unsanitized objects.
  This following fields are silently ignored:
  - status
  - metadata.creationTimestamp
  - metadata.finalizers
  This helps when applying objects from 3rd party sources
  which can't be easily sanitized.
- Fix RootSync vs RootSync conflicts when the webhook is enabled.
  Previously, the remediator would detect and report the conflict
  to the other RootSync, but now that the remediator is paused
  during apply, there is no conflict. The webhook prevents it.
  So only the 2nd RootSync reports the error, when attempting to
  apply. This is how RepoSyncs already worked.
- Fix e2e metrics validation to tolerate missing RemediateDuration
  metric. Now that the remediator is paused while applying, it's
  possible that no objects are remediated during the course of a test.
- Move ObjectQueue shutdown into the Remediator and
  modify ObjectQueue.Get to take a context. This allows
  Workers to be stopped without shutting down the queue.
- Add additional logging that was useful for debugging

Fixes: b/235358559

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/431
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/432
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/433
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/434
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/435
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/437
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/442
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/447
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/449

### DIFF
--- a/pkg/parse/updater.go
+++ b/pkg/parse/updater.go
@@ -30,6 +30,7 @@ import (
 	"kpt.dev/configsync/pkg/remediator"
 	"kpt.dev/configsync/pkg/status"
 	"kpt.dev/configsync/pkg/util/clusterconfig"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // updater mutates the most-recently-seen versions of objects stored in memory.
@@ -62,12 +63,20 @@ func (u *updater) Errors() status.MultiError {
 	defer u.errorMux.RUnlock()
 
 	var errs status.MultiError
-	for _, conflictErr := range u.remediator.ConflictErrors() {
-		errs = status.Append(errs, conflictErr)
-	}
+	errs = status.Append(errs, u.conflictErrors())
 	errs = status.Append(errs, u.validationErrs)
 	errs = status.Append(errs, u.applier.Errors())
 	errs = status.Append(errs, u.watchErrs)
+	return errs
+}
+
+// conflictErrors converts []ManagementConflictError into []MultiErrors.
+// This method is safe to call while Update is running.
+func (u *updater) conflictErrors() status.MultiError {
+	var errs status.MultiError
+	for _, conflictErr := range u.remediator.ConflictErrors() {
+		errs = status.Append(errs, conflictErr)
+	}
 	return errs
 }
 
@@ -92,7 +101,7 @@ func (u *updater) Updating() bool {
 // declared resources.
 func (u *updater) declaredCRDs() ([]*v1beta1.CustomResourceDefinition, status.MultiError) {
 	var crds []*v1beta1.CustomResourceDefinition
-	for _, obj := range u.resources.Declarations() {
+	for _, obj := range u.resources.DeclaredUnstructureds() {
 		if obj.GroupVersionKind().GroupKind() != kinds.CustomResourceDefinition() {
 			continue
 		}
@@ -105,8 +114,17 @@ func (u *updater) declaredCRDs() ([]*v1beta1.CustomResourceDefinition, status.Mu
 	return crds, nil
 }
 
-// Update updates the declared resources in memory, applies the resources, and sets
-// up the watches.
+// Update does the following:
+// 1. Pauses the remediator
+// 2. Validates and sterilizes the objects
+// 3. Updates the declared resource objects in memory
+// 4. Applies the objects
+// 5. Updates the remediator watches
+// 6. Restarts the remediator
+//
+// Any errors returned will be prepended with any known conflict errors from the
+// remediator. This is required to preserve errors that have been reported by
+// another reconciler.
 func (u *updater) Update(ctx context.Context, cache *cacheForCommit) status.MultiError {
 	u.updateMux.Lock()
 	u.updating = true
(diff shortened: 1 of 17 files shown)


pkg/parse/updater.go
```diff
@@ -105,8 +114,17 @@ func (u *updater) declaredCRDs() ([]*v1beta1.CustomResourceDefinition, status.Mu
...
 
-// Update updates the declared resources in memory, applies the resources, and sets
-// up the watches.
+// Update does the following:
+// 1. Pauses the remediator
+// 2. Validates and sterilizes the objects
+// 3. Updates the declared resource objects in memory
+// 4. Applies the objects
+// 5. Updates the remediator watches
+// 1. Restarts the remediator
```
sdowell, line 123: nit: index

pkg/parse/updater.go
```diff
@@ -115,54 +133,119 @@ func (u *updater) Update(ctx context.Context, cache *cacheForCommit) status.Mult
...
+			cache.watchesUpdated = true
+		}
+	}
+
+	// Restart remediator workers.
+	// Queue will probably include all the declared objects, but they should
+	// show no diff, unless they've been updated asynchronously.
+	// Only resume after validation & apply & watch update are successful,
+	// otherwise the objects may be updated in the wrong order (dependencies).
+	u.remediator.Resume()
```
sdowell, line 209: Is it intentional that the remediator doesn't resume if we return an error earlier on in the function? Wondering if we should `defer` the Resume call

pkg/remediator/remediator.go
```diff
@@ -108,7 +160,26 @@ func (r *Remediator) Start(ctx context.Context) <-chan struct{} {
...
 	}()
-	return doneCh
+
+	r.doneCh = doneCh
+	r.stopFn = cancel
+}
+
+// Pause the Remediator by stopping the workers and waiting for them to be done.
+func (r *Remediator) Pause() {
+	klog.V(1).Info("Remediator pausing...")
```
sdowell, line 170: Does this need to be synchronized with lifecycleMux? Seems that way since we are using `stopFn` and `doneCh`

pkg/parse/updater.go
karlkfi: It's intentional. If the changes are half applied and we re-enable the remediator, it may try to apply the rest of the changes the next time any event occurs for that object. For things like Roles that have no status, that may not happen without a 3rd party, but for things like Deployments that get continuous status updates from a controller and are affected by auto-scaling, it could cause changes to be made out of order, which is the problem we're trying to mitigate by pausing.

pkg/remediator/remediator.go
karlkfi: I was thinking that the channel and cancel func were thread-safe, but you're right, the variables aren't. We probably don't want Start or Resume to be called while Pause is being called. They won't with the current usage, but I'll add the lock to be defensive.

pkg/parse/updater.go
karlkfi: fixed

pkg/parse/updater.go
```diff
@@ -115,54 +133,119 @@ func (u *updater) Update(ctx context.Context, cache *cacheForCommit) status.Mult
...
-		applyStart := time.Now()
-		gvks, applyErrs = u.applier.Apply(ctx, objs)
-		metrics.RecordApplyDuration(ctx, metrics.StatusTagKey(applyErrs), cache.source.commit, applyStart)
-		if applyErrs != nil {
-			klog.Warningf("Failed to apply declared resources: %v", applyErrs)
-		} else if cache.parserErrs == nil {
+		var err status.MultiError
+		gvks, err = u.apply(ctx, objs, cache.source.commit)
+		if err != nil {
+			return err
```
nan-yu, line 183: This is a behavioral change. Before this PR, it continues to update the remediator's watches even if there're apply errors.
Do we want to stop updating the watches when there're apply errors, especially for the non-blocking errors (missing CRD)?

karlkfi: Should we update the watches before applying?
The declared resources are already updated beforehand.
And we're effectively disregarding any drift that happens to objects intended for deletion.
So it doesn't really matter if we stop watching them before they're deleted.

haiyanmeng:
> The reconciler now pauses the remediator before syncing,
> and resumes remediating after syncing is successful.
> The remediator will continue watching and enqueuing
> objects to remediate when changed while paused. This
> way no drift is missed while other objects are being applied.

Is there a size limit of the queue? I wonder what would happen if a Git commit cannot be synced successfully for a long time.

karlkfi:
> Is there a size limit of the queue? I wonder what would happen if a Git commit cannot be synced successfully for a long time.

The queue uses a slice and a map and dedupes objects. So it wean only ever grow as large as the number of objects on the cluster, which will never be larger than max size for map/slice. This could increase memory usage, but only when failing to apply. 

The bigger risk is that the queue is only in memory, so it will get purged if the app crashes or is rescheduled. To mitigate that we’d have to use a persistent disk, which would be a big change. And it seems like less of a problem than we currently have with apply races causing dependency order violations.

pkg/parse/updater.go
nan-yu: The watches are always updated after applying.
My question is about whether we should update the watches when there're apply errors?

After checking the `apply` function, if error is not nil, the gvks will be nil. It basically won't update any watches anyway.

nan-yu (approved):
/lgtm
/hold

The change looks good to me. Hold it for now in case @haiyanmeng or @sdowell wants to take another look.

haiyanmeng:
> This could increase memory usage, but only when failing to apply.

Should we document this potential risk in the code or in the git commit to help people understand the potential memory usage hike?

karlkfi:
Looks like stopping the remediator watches before applying has uncovered a pre-existing bug in the applier that the remediator was... remediating.

```
=== CONT  TestPreventDeletionRole
    wait_for_sync.go:199: took 15.089212485s to wait for config-management-system/root-sync configsync.gke.io/v1beta1, Kind=RootSync to be synced
    lifecycle_directives_test.go:163: ERROR: object "shipping-admin" shouldn't have configsync metadata (labels: map[app.kubernetes.io/managed-by:configmanagement.gke.io configsync.gke.io/declared-version:v1 nomos-test:enabled], annotations: map[client.lifecycle.config.k8s.io/deletion:detach configmanagement.gke.io/cluster-name:e2e-test-cluster configmanagement.gke.io/managed:enabled configmanagement.gke.io/source-path:acme/namespaces/shipping/role.yaml configmanagement.gke.io/token:a9a6056fc561c26c2d8f6fa5d5b4a91f66b52117 configsync.gke.io/declared-fields:{"f:metadata":{"f:annotations":{"f:client.lifecycle.config.k8s.io/deletion":{}},"f:labels":{"f:nomos-test":{}}},"f:rules":{}} configsync.gke.io/git-context:{"repo":"git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync","branch":"main","rev":"HEAD"} configsync.gke.io/manager::root configsync.gke.io/resource-id:rbac.authorization.k8s.io_role_shipping_shipping-admin])
```

The TestPreventDeletionRole test applies the `shipping-admin` Role with the deletion prevention lifecycle annotation and then removes from the source, but the test is failing because the `NoConfigSyncMetadata()` predicate is failing afterwards.

I think the Config Sync metadata was previously being removed by the remediator, but now it's not, which implies that the code in the applier to do the same isn't working.

karlkfi:
/hold

Too many changes. need to extract a few to new PRs for independent review.

But yay, the tests pass!

karlkfi:
> Should we document this potential risk in the code or in the git commit to help people understand the potential memory usage hike?

@haiyanmeng It can only grow to the total size of all the objects managed by this reconciler, which we already store in memory for syncing. So it's at most a 2x memory increase, and more likely significantly less due to other memory usage. The growth is also linear with the size of the source repo, same as today. Anyone using a large amount of objects will already have to increase the amount of memory specified for the reconciler, and we already have that documented.

pkg/parse/updater.go
```diff
@@ -115,54 +133,111 @@ func (u *updater) Update(ctx context.Context, cache *cacheForCommit) status.Mult
...
+	// After this, any objects removed from the declared resources will no
+	// longer be remediated, if they drift.
+	if !cache.declaredResourcesUpdated {
+		objs := filesystem.AsCoreObjects(cache.objsToApply)
+		_, err := u.declare(ctx, objs)
+		if err != nil {
+			return err
 		}
+		// Only mark the declared resources as updated if there were no (non-blocking) parse errors.
+		// This ensures the update will be retried until parsing fully succeeds.
```
sdowell, line 164: Thanks for adding all of these comments and breaking out the helper methods, it helps with following the control flow here

pkg/remediator/queue/queue.go
```diff
@@ -44,11 +46,15 @@ func GVKNNOf(obj client.Object) GVKNN {
 	}
 }
 
+// ErrShutdown is returned by `ObjectQueue.Get` if the queue is shut down
+// before an object is available/added.
+var ErrShutdown = errors.New("object queue shutting down")
```
sdowell, line 51: Maybe define this as a function rather than a `var` to enforce that this doesn't get changed?

pkg/remediator/queue/queue.go
```diff
@@ -251,7 +295,8 @@ func (g *genericWrapper) Add(item interface{}) {
 
 // Get implements workqueue.Interface.
 func (g *genericWrapper) Get() (item interface{}, shutdown bool) {
-	return g.oq.Get()
+	obj, err := g.oq.Get(context.Background())
+	return obj, (err != nil)
```
sdowell, line 299: Could we have this return the error rather than a bool? Seems like the bool just buries information

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -134,10 +135,10 @@ func (w *filteredWatcher) pruneErrors() {
 
 // addError checks whether an error identified by the errorID has been tracked,
 // and handles it in one of the following ways:
-//   * tracks it if it has not yet been tracked;
-//   * updates the time for this error to time.Now() if `errorLoggingInterval` has passed
+//   - tracks it if it has not yet been tracked;
```
sdowell, line 138: Duel of the IDEs

pkg/remediator/reconcile/worker_test.go
```diff
@@ -44,6 +45,184 @@ import (
...
+		// Role change should be reverted
+		fake.ClusterRoleObject(syncertest.ManagementEnabled,
+			core.UID("1"), core.ResourceVersion("3"), core.Generation(1),
+		),
+	}
+
+	q := queue.New("test")
+	defer q.ShutDown()
+
+	c := testingfake.NewClient(t, core.Scheme)
```
sdowell, line 81: nit: use more descriptive variable names. (had to go back up here to find out what `c` was)

pkg/remediator/queue/queue.go
karlkfi: Yeah, that's potentially a concern, but this method has precedent in both the std libs and k8s libs and makes it easy to do an == comparison.

pkg/remediator/queue/queue.go
karlkfi: Nope. It has to satisfy the upstream interface, which returns a bool.
That's why it's wrapping the more expressive ObjectQueue with a wrapper.

karlkfi: In this case, the context is never cancelled, so the error will always be the ErrShutdown, if not nil.

pkg/remediator/watch/filteredwatcher.go
karlkfi: I think this is actually a GoFmt version issue. 
I have Go 1.20 installed locally, but the project is compatible back to 1.17.

pkg/remediator/reconcile/worker_test.go
karlkfi: This code is copied from the existing tests. Do you want me to refactor them all in this PR?

sdowell: no that's not necessary

karlkfi: Variable naming is a bit of a holy war. I also prefer more descriptive names, but some go authors have said they prefer short ones. `c` in particular gets used a lot in the K8s libs, because `client` is a common package name.

If you feel strongly, I suggest doing some refactoring afterwards. I'd be happy to approve.

PR #418, opened 2023-02-28
Avoid exposing transient errors to the R*Sync API

nan-yu:
If the *-sync container is not ready, the hydration controller and the reconciler will fail because the source is not available yet. However, a source KNV2004 error is exposed to the user with a message: unable to check the error file...

This commit makes it a transient error and avoids exposing transient errors to the R*Sync API.


pkg/parse/run.go
```diff
@@ -234,6 +234,12 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
 // It also updates the .status.rendering and .status.source fields.
 func read(ctx context.Context, p Parser, trigger string, state *reconcilerState, sourceState sourceState) status.MultiError {
 	hydrationStatus, sourceStatus := readFromSource(ctx, p, trigger, state, sourceState)
+	if status.HasTransientErrors(hydrationStatus.errs) {
+		return hydrationStatus.errs
```
tiffanny29631, line 238: Will the other non-transient errors still get updated into the source status?

nan-yu: If `hydrationStatus.errs` has other non-transient errors, it won't get updated into the source status in the current loop, but will get retried.

But on the other hand, for now, if it has transient errors, it is unlikely to have other errors because the current known transient errors are:
- unable to read the error file error because *-sync container is not ready yet.
- the race condition you're fixing.
They'll terminate the current parse-apply-watch loop, and retry.

tiffanny29631: Makes sense!

tiffanny29631 (approved):
/lgtm
/hold
See if Karl has other thoughts. Feel free to unhold.

karlkfi:
Any idea how we can test this?

nan-yu:
> Any idea how we can test this?

@karlkfi I added a unit test for the `run` function.

pkg/hydrate/controller.go
```diff
@@ -312,24 +316,25 @@ func SourceCommitAndDir(sourceType v1beta1.SourceType, sourceRoot cmpath.Absolut
...
-				git.SyncError(containerName, errFilePath, fmt.Sprintf("%s=%s", metadata.ReconcilerLabel, reconcilerName)))
-		} else {
-			err = errors.Wrapf(err, "unable to sync repo\n%s",
-				git.SyncError(containerName, errFilePath, fmt.Sprintf("%s=%s", metadata.ReconcilerLabel, reconcilerName)))
-		}
-		return status.SourceError.Wrap(err).Build()
+	content, err := os.ReadFile(errFilePath)
+	if err != nil && !os.IsNotExist(err) {
+		return "", "", status.TransientError(
+			fmt.Errorf("unable to sync repo\nunable to load %s: %v. Please "+
```
karlkfi, line 322: do we need the linebreak here? Would a `"` suffice?

pkg/parse/run.go
```diff
@@ -234,6 +241,12 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
 // It also updates the .status.rendering and .status.source fields.
 func read(ctx context.Context, p Parser, trigger string, state *reconcilerState, sourceState sourceState) status.MultiError {
 	hydrationStatus, sourceStatus := readFromSource(ctx, p, trigger, state, sourceState)
+	if status.HasTransientErrors(hydrationStatus.errs) {
```
karlkfi, line 244: Add a comment explaining why it's desirable and safe to fail due to transient errors. It's not immediately obvious from the code.

pkg/hydrate/controller.go
nan-yu: The `unable to sync repo` doesn't provide any value, so I removed that.

pkg/parse/run.go
nan-yu: done

pkg/parse/run_test.go
```diff
@@ -75,3 +140,188 @@ func TestSplitObjects(t *testing.T) {
...
+
+			if !status.DeepEqual(state.renderingStatus.errs, tc.expectedStateRenderingErrs) {
+				t.Errorf("expected state.renderingStatus.errs '%v', but got '%v'", tc.expectedStateRenderingErrs.Error(), state.renderingStatus.errs.Error())
+			}
+
+			rs := &v1beta1.RootSync{}
+			if err = parser.options().client.Get(context.Background(), rootsync.ObjectKey(parser.options().syncName), rs); err != nil {
+				t.Fatal(err)
+			}
+			if !reflect.DeepEqual(rs.Status.Source.Errors, tc.expectedRSSourceErrs) {
```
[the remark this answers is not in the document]
nan-yu: Good to know. Switched to use testutil.

karlkfi:
Looks like this is causing some e2e tests to fail by suppressing errors too aggressively.

I think you might need to make it only return if there's transient errors but not non-transient errors, otherwise filter out the transient errors and report the non-transient errors.

nan-yu:
> Looks like this is causing some e2e tests to fail by suppressing errors too aggressively.
> 
> I think you might need to make it only return if there's transient errors but not non-transient errors, otherwise filter out the transient errors and report the non-transient errors.

It is not suppressing errors too aggressively, but we should check the existence of the source root directory (`/repo/source`), instead of the source link directory (`/repo/source/rev`). The `/repo/source/rev` symlink is only created after *-sync successfully pulls the source configs.

PR #423, opened 2023-03-01
Change e2e test validation to use sync status output

victorpras:
e2e test for nomos status was using git server string as its validation, which can vary in other git providers. Validation is changed to use sync status instead, which value is only SYCED/PENDING/RECONCILING/STALLED

### DIFF
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -1046,7 +1046,8 @@ func TestNomosStatus(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
-	if !strings.Contains(string(out), "git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync/acme@main") {
-		nt.T.Fatalf("Expected to find Git provider string in output:\n%s\n", string(out))
+	if !strings.Contains(string(out), "SYNCED") && !strings.Contains(string(out), "PENDING") &&
+		!strings.Contains(string(out), "RECONCILING") && !strings.Contains(string(out), "STALLED") {
+		nt.T.Fatalf("Expected to find sync status in string output:\n%s\n", string(out))
 	}
 }
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -1046,8 +1046,7 @@ func TestNomosStatus(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
-	if !strings.Contains(string(out), "SYNCED") && !strings.Contains(string(out), "PENDING") &&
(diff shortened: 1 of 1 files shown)


e2e/testcases/cli_test.go
```diff
@@ -1046,7 +1046,8 @@ func TestNomosStatus(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
-	if !strings.Contains(string(out), "git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync/acme@main") {
-		nt.T.Fatalf("Expected to find Git provider string in output:\n%s\n", string(out))
+	if !strings.Contains(string(out), "SYNCED") && !strings.Contains(string(out), "PENDING") &&
```
nan-yu, line 1049: Is `!strings.Contains(string(out), "SYNCED")` correct? If it is neither of these statuses, what does the out look like?

victorpras: From my understanding, those are the possible sync status as listed on https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/cmd/nomos/status/status.go#L37, although I've not seen `STALLED` when doing manual e2e testing. Other possible output is if ConfigSync wasnt installed because rootsync isnt found, the output will look like `Failed to get the RootSync CRD: customresourcedefinitions.apiextensions.k8s.io "rootsyncs.configsync.gke.io" not found`

victorpras: Updated to check only for SYNCED

victorpras: I tried running the e2e test in my local cluster and the output looks something like this
```
  <root>:root-sync                         git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync/acme@main
  SYNCED @ 2023-03-01 20:50:41 +0000 UTC   c4453f390335a7ca0abbb7de4468e7bbd8faa487
  Managed resources:
     NAMESPACE   NAME                                                                              STATUS
                 clusterrole.rbac.authorization.k8s.io/safety-config-management-system-root-sync   Unknown
                 namespace/safety-config-management-system-root-sync                               Unknown
               ```

nan-yu: This is the previous condition `if !strings.Contains(string(out), "SYNCED") && !strings.Contains(string(out), "PENDING") && !strings.Contains(string(out), "RECONCILING") && !strings.Contains(string(out), "STALLED")`. With the above output, I couldn't see how it can pass the test. Can you change it back to the previous condition locally, and check if it passes?

nan-yu: Checked with Victor offline. The output does contain `SYNCED`, which makes it pass the condition.

nan-yu (approved):
/lgtm

PR #421, opened 2023-02-28
Enable nomos version in OSS

victorpras:
Indicates OSS installation in `nomos version` command when OSS installation is found. Moved `isOssInstallation` func to util package

PR #428, opened 2023-03-01
Enable nomos version in OSS (#421)

victorpras:
Enable nomos version in OSS

### DIFF
--- a/cmd/nomos/util/write.go
+++ b/cmd/nomos/util/write.go
@@ -31,6 +31,8 @@ const (
 	NotConfiguredMsg = "NOT CONFIGURED"
 	// UnknownMsg indicates that a field's value is unknown or unavailable.
 	UnknownMsg = "UNKNOWN"
+	// OSSMsg indicates OSS installation
+	OSSMsg = "OSS Installation"
 )
 
 // NewWriter returns a standardized writer for the CLI for writing tabular output to the console.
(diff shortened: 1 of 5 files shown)


cmd/nomos/util/write.go
```diff
@@ -31,6 +31,8 @@ const (
 	NotConfiguredMsg = "NOT CONFIGURED"
 	// UnknownMsg indicates that a field's value is unknown or unavailable.
 	UnknownMsg = "UNKNOWN"
+	// OSSMsg indicates OSS installation
+	OSSMsg = "OSS Installation"
```
sdowell, line 35: I'm not sure that `OSS Installation` is the right language here. I think something along the lines of "standalone installation` is more accurate. The important distinction IMO is whether the ACM operator is installed

sdowell: Oh is this a cherry pick? I didn't realize this was already submitted. In that case don't change this PR

victorpras: I can create a separate PR to address this

nan-yu: Good point. @victorpras Can you make a separate PR to change it to `Standalone Config Sync`?

nan-yu (approved):
/lgtm

cmd/nomos/util/write.go
victorpras: PR to change the output https://github.com/GoogleContainerTools/kpt-config-sync/pull/436

PR #431, opened 2023-03-02
Avoid unnecessary source status update

karlkfi:
Don't update the source status just because the sync status is newer. 
Do update the source status if the rendering status is newer.

### DIFF
--- a/pkg/parse/state.go
+++ b/pkg/parse/state.go
@@ -159,14 +159,38 @@ func (s *reconcilerState) resetAllButSourceState() {
 
 // needToSetSourceStatus returns true if `p.setSourceStatus` should be called.
 func (s *reconcilerState) needToSetSourceStatus(newStatus sourceStatus) bool {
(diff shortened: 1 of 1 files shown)


pkg/parse/state.go
```diff
@@ -159,14 +159,38 @@ func (s *reconcilerState) resetAllButSourceState() {
...
-		s.sourceStatus.lastUpdate.Before(&s.syncingConditionLastUpdate))
+	if status.HasTransientErrors(newStatus.errs) {
+		return false
+	}
+	// Update if not initialized
+	if s.sourceStatus.lastUpdate.IsZero() {
+		return true
+	}
+	// Update if source status was last updated before the rendering status
+	if s.sourceStatus.lastUpdate.Before(&s.renderingStatus.lastUpdate) {
```
nan-yu, line 170: I think the reason that we were comparing with the lastUpdate timestamp before is we thought the sourceStatus and renderingStatus were from the R*Sync status, but it is actually the internal cache from the new commit.
It makes sense to compare with the next phase.

nan-yu (approved):
/lgtm

PR #432, opened 2023-03-02
Make TestComposition exit early on failure

karlkfi:
Use watch to log the diff when debugging and exit early when one of the watched objects is updated after syncing.


nan-yu (approved):
/lgtm

PR #433, opened 2023-03-02
Increase e2e polling periods

karlkfi:
This should allow the retry loop to execute, allowing it to be tested. 
This should also allow the remediator more time to run between applies.


nan-yu (approved):
/lgtm

karlkfi:
/retest

`failed to init node with kubeadm` flake

PR #434, opened 2023-03-02
Print YAML instead of JSON on e2e validation error

karlkfi:
JSON takes less vertical space, but it significantly harder to read/scan quickly. Debug only runs on dev machines so this wont affect CI.


karlkfi:
/retest

Looks like flakey cluster creation problem

nan-yu (approved):
/lgtm

karlkfi:
/retest

`stream error: stream ID 2009; INTERNAL_ERROR; received from peer` go download flake

PR #435, opened 2023-03-02
Add more logs for debugging

karlkfi:
- Rename "unmanage" in the remediator (internal enum & log msgs)  to "abandon" to be more consistent with the language used by the applier. These are both the same thing. It's when the object is removed from the inventory and its management annotations are removed (`client.lifecycle.config.k8s.io/deletion: detach`, `cli-utils.sigs.k8s.io/on-remove: keep`).
- Add logs before status updates to indicate when/why


pkg/diff/diff.go
```diff
@@ -53,8 +53,8 @@ const (
 	// Error indicates the resource's management annotation in the API server is invalid.
 	Error = Operation("error")
 
-	// Unmanage indicates the resource's management annotation should be removed from the API Server.
-	Unmanage = Operation("unmanage")
+	// Abandon indicates the resource's management annotation should be removed from the API Server.
```
nan-yu, line 56: Are we going to distinguish between abandon and unmanage?

karlkfi: Is there a difference?
Abandoning is removing it from the inventory and removing the management annotations.

nan-yu: If no difference, I'm curious why we need to rename it.

karlkfi: Oh, I was just renaming it in this internal enum used by the remediator to be more consistent with the language used by the applier.

nan-yu: Gotcha.

nan-yu (approved):
/lgtm

PR #436, opened 2023-03-02
Change nomos version output in standalone OSS to match image version

victorpras:
nomos version output in standalone OSS installation will now display the matching version with the image that is used for installation

### DIFF
--- a/cmd/nomos/version/version.go
+++ b/cmd/nomos/version/version.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"k8s.io/client-go/kubernetes"
@@ -141,13 +142,24 @@ func filterConfigs(contexts []string, all map[string]*rest.Config) map[string]*r
 }
 
 func getImageVersion(deployment *v1.Deployment) (string, error) {
-	reconcilerImage := strings.Split(deployment.Spec.Template.Spec.Containers[0].Image, ":")
-	if len(reconcilerImage) <= 1 {
-		return "", fmt.Errorf("Failed to get valid image version from: %s", reconcilerImage)
+	var container corev1.Container
+	for _, c := range deployment.Spec.Template.Spec.Containers {
+		if c.Name == util.ReconcilerManagerName {
+			container = c
+			break
+		}
 	}
-	return reconcilerImage[1], nil
+
+	reconcilerManagerImage := strings.Split(container.Image, ":")
+	if len(reconcilerManagerImage) <= 1 {
+		return "", fmt.Errorf("Failed to get valid image version from: %s", reconcilerManagerImage)
+	}
+	return reconcilerManagerImage[1], nil
 }
 
+// lookupVersionAndMode will check for installation type (oss or operator)
+// and will return ACM version, isMultiRepo state, installation type, and error if needed
+// OSS installation will check version from image used, while operator will check from ConfigManagement
 func lookupVersionAndMode(ctx context.Context, cfg *rest.Config) (string, *bool, string, error) {
 	cmClient, err := util.NewConfigManagementClient(cfg)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/nomos/util/write.go
```diff
@@ -31,8 +31,8 @@ const (
 	NotConfiguredMsg = "NOT CONFIGURED"
 	// UnknownMsg indicates that a field's value is unknown or unavailable.
 	UnknownMsg = "UNKNOWN"
-	// OSSMsg indicates OSS installation
-	OSSMsg = "OSS Installation"
+	// StandaloneMsg indicates standalone OSS installation
+	StandaloneMsg = "Standalone Installation"
```
nan-yu, line 35: Here will be the new output:
```
nomos version --contexts=$(kubectl config current-context)
CURRENT   CLUSTER_CONTEXT_NAME                                    COMPONENT           VERSION
                                                                  <nomos CLI>         UNKNOWN
*         gke_nanyu-gke-dev_us-central1-c_cs-standalone-regular   config-management   Standalone Installation
```

The COMPONENT column shows `config-management`, and the VERSION column shows `Standalone Installation`.

We'll need to change the COMPONENT column to `config-sync`, which distinguish the standalone Config Sync from the ACM version of Config Sync. For the VERSION column, it would be good to extract the image tag from the reconciler-manager container. WDYT? @sdowell

sdowell: That's a good suggestion, SGTM

victorpras: > it would be good to extract the image tag from the reconciler-manager container 

since nomos version would normally query from all clusters and project in kubeconfig, will it able to access the image tag from different clusters and projects?

nan-yu: As long as the clusters are accessible, it should be able to get the reconciler-manager Deployment and extract the image tag.

victorpras: I made a prototype of this in my local, some example output: 
``` 
CURRENT   CLUSTER_CONTEXT_NAME                                       COMPONENT           VERSION
                                                                     <nomos CLI>         v1.14.3-rc.2-5-g5e27395-dirty
          gke_victortjhia-gke-dev_us-central1-c_operator-cluster-1   config-management   v1.14.1-rc.3
          gke_victortjhia-gke-dev_us-central1-c_oss-cluster-1        config-sync         latest
*         gke_victortjhia-gke-dev_us-central1-c_oss-cluster-v114     config-sync         v1.14.3-rc.2-5-g5e27395
          gke_victortjhia-gke-dev_us-east4-c_nat-test-cluster                            <error: Get "https://172.16.0.2/api?timeout=3s": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)>

```

(the one showing "latest" as version was because I was following an out of date installation instruction for config sync)

nan-yu: The output looks good to me.

cmd/nomos/version/version.go
```diff
@@ -136,44 +140,61 @@ func filterConfigs(contexts []string, all map[string]*rest.Config) map[string]*r
 	return cfgs
 }
 
-func lookupVersionAndMode(ctx context.Context, cfg *rest.Config) (string, *bool, error) {
+func getImageVersion(deployment *v1.Deployment) (string, error) {
+	reconcilerImage := strings.Split(deployment.Spec.Template.Spec.Containers[0].Image, ":")
```
nan-yu, line 144: Suggest checking if the container's name is the `reconciler-manager`. The reconciler-manager has two containers, using the container index is not reliable.

nan-yu, line 144: FYI, it is the reconciler-manager image, not the reconciler.

cmd/nomos/version/version.go
```diff
@@ -136,44 +140,61 @@ func filterConfigs(contexts []string, all map[string]*rest.Config) map[string]*r
...
+	if len(reconcilerImage) <= 1 {
+		return "", fmt.Errorf("Failed to get valid image version from: %s", reconcilerImage)
+	}
+	return reconcilerImage[1], nil
+}
+
+func lookupVersionAndMode(ctx context.Context, cfg *rest.Config) (string, *bool, string, error) {
```
nan-yu, line 151: Now the function returns multiple values. I would suggest adding a comment to explain what are those values.

cmd/nomos/version/version.go
victorpras: updated to search for container named reconciler-manager

victorpras: update var name for clarity

cmd/nomos/version/version.go
victorpras: added explanation in function description

nan-yu (approved):
/lgtm

PR #437, opened 2023-03-02
Fix remediator missing events

karlkfi:
- Fix a bug where drift changes that occurred between apply and watch update would be tolerated until there was another change. Now the drift will be detected as soon as the watch starts.
- Use ListAndWatch, instead of just Watch, to track changes to managed objects for remediation.

### DIFF
--- a/pkg/remediator/watch/filteredwatcher.go
+++ b/pkg/remediator/watch/filteredwatcher.go
@@ -84,7 +84,7 @@ const errorLoggingInterval = time.Second
 // - or managed by the same reconciler.
 type filteredWatcher struct {
 	gvk        string
-	startWatch startWatchFunc
+	startWatch WatchFunc
 	resources  *declared.Resources
 	queue      *queue.ObjectQueue
 	scope      declared.Scope
@@ -105,8 +105,8 @@ type filteredWatcher struct {
 // filteredWatcher implements the Runnable interface.
 var _ Runnable = &filteredWatcher{}
 
-// NewFiltered returns a new filtered watch initialized with the given options.
-func NewFiltered(_ context.Context, cfg watcherConfig) Runnable {
+// NewFiltered returns a new filtered watcher initialized with the given options.
+func NewFiltered(cfg watcherConfig) Runnable {
 	return &filteredWatcher{
 		gvk:                     cfg.gvk.String(),
 		startWatch:              cfg.startWatch,
@@ -268,7 +268,7 @@ func (w *filteredWatcher) Run(ctx context.Context) status.Error {
 		// 1. false, error -> We were unable to start the watch, so exit Run().
 		// 2. false, nil   -> We have been stopped via Stop(), so exit Run().
 		// 3. true,  nil   -> We have not been stopped and we started a new watch.
-		started, err := w.start(resourceVersion)
+		started, err := w.start(ctx, resourceVersion)
 		if err != nil {
 			return err
 		}
@@ -318,7 +318,7 @@ func (w *filteredWatcher) Run(ctx context.Context) status.Error {
 // threadsafe manner and returns true if the new base watch was created. Returns
 // false if the filteredWatcher is already stopped and returns error if the base
 // watch could not be started.
-func (w *filteredWatcher) start(resourceVersion string) (bool, status.Error) {
+func (w *filteredWatcher) start(ctx context.Context, resourceVersion string) (bool, status.Error) {
 	w.mux.Lock()
 	defer w.mux.Unlock()
 
@@ -337,7 +337,7 @@ func (w *filteredWatcher) start(resourceVersion string) (bool, status.Error) {
 		Watch:               true,
 	}
 
-	base, err := w.startWatch(options)
+	base, err := w.startWatch(ctx, options)
 	if err != nil {
 		return false, status.APIServerErrorf(err, "failed to start watch for %s", w.gvk)
 	}
--- a/pkg/remediator/watch/listwatch.go
+++ b/pkg/remediator/watch/listwatch.go
@@ -1,0 +1,337 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package watch
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"kpt.dev/configsync/pkg/core"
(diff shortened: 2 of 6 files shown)


pkg/remediator/watch/filteredwatcher.go
```diff
@@ -268,7 +268,7 @@ func (w *filteredWatcher) Run(ctx context.Context) status.Error {
 		// 1. false, error -> We were unable to start the watch, so exit Run().
 		// 2. false, nil   -> We have been stopped via Stop(), so exit Run().
 		// 3. true,  nil   -> We have not been stopped and we started a new watch.
-		started, err := w.start(resourceVersion)
+		started, err := w.start(ctx, resourceVersion)
```
karlkfi, line 271: I just love how before this change, the ctx was just being dropped on the floor. Sure, it was being given to the hander, but the handler can't terminate the watch and only gets run when there's an event. So before this, the watch would just keep running until it timed out, or more likely, the app terminated, leaving the connection open...

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -105,8 +105,8 @@ type filteredWatcher struct {
 // filteredWatcher implements the Runnable interface.
 var _ Runnable = &filteredWatcher{}
 
-// NewFiltered returns a new filtered watch initialized with the given options.
-func NewFiltered(_ context.Context, cfg watcherConfig) Runnable {
+// NewFiltered returns a new filtered watcher initialized with the given options.
+func NewFiltered(cfg watcherConfig) Runnable {
```
karlkfi, line 109: Why did this take a context before in the first place? Why would a constructor take a context at all?

pkg/remediator/watch/listwatch.go
```diff
@@ -0,0 +1,299 @@
...
+		stopCh:  make(chan struct{}),
+		eventCh: make(chan watch.Event),
+	}
+
+	// Spawn background thread to handle watching and pre-processing events.
+	go func() {
+		defer close(wrapper.eventCh)
+
+		// De-dupe Added events to Modified events by caching the IDs of pre-existing objects.
+		ids := make(map[core.ID]struct{}, len(cObjList.Items))
```
sdowell, line 224: Is this the primary motivation for list and watch? Namely differentiating pre-existing objects as `modified` rather than `added` for the watch?

pkg/remediator/watch/watcher.go
```diff
@@ -17,60 +17,48 @@ package watch
...
-				return dynamicClient.Resource(mapping.Resource).Watch(ctx, options)
-			}
-		} else {
-			cfg.startWatch = func(options metav1.ListOptions) (watch.Interface, error) {
-				return dynamicClient.Resource(mapping.Resource).Namespace(string(cfg.scope)).Watch(ctx, options)
+// watcherFromListerWatcher returns a new watcherFactory that uses the specified
+// ListerWatcherFactory.
+func watcherFromListerWatcher(factory ListerWatcherFactory) watcherFactory {
+	factoryPtr := factory
+	return func(ctx context.Context, cfg watcherConfig) (Runnable, status.Error) {
```
sdowell, line 50: It doesn't look like the context input of this function is used. The context from the nested `startWatch` function seems to be used instead. Should the `watcherFactory` signature be changed to not include a context input arg?

sdowell (commented):
Would it be possible to separate the context change from the list and watch change? I think that might make this easier to review

karlkfi:
> Would it be possible to separate the context change from the list and watch change?

ListAndWatch takes the context and can be cancelled with it. That's one of the primary reasons I wrote my own ListAndWatch, instead of use the client-go one. The other reason is to use Unstructured objects, instead of typed objects that need a local up-to-date Scheme for all objects.

It doesn't make sense to use ListAndWatch and not pass the context to it.

sdowell:
> > Would it be possible to separate the context change from the list and watch change?
> 
> ListAndWatch takes the context and can be cancelled with it. That's one of the primary reasons I wrote my own ListAndWatch, instead of use the client-go one. The other reason is to use Unstructured objects, instead of typed objects that need a local up-to-date Scheme for all objects.
> 
> It doesn't make sense to use ListAndWatch and not pass the context to it.

I was thinking the context change could go in first since it is relatively simple. Not a big deal

pkg/remediator/watch/listwatch.go
karlkfi: The motivation for doing a List and a Watch, instead of just a Watch, is to pull the current state so we don't effectively ignore changes before we started watching until the next event after we started watching.

This ListAndWatch is a convenience method that returns the same watch.Interface that Watch does, but it does the List first and translates the output into Added events. Then, to avoid duplicate Added events, we have to cache which ones we've seen already and convert new Added events into Modified events. And we also have to remove IDs from the cache if they're Deleted, so that we don't translate any subsequent Added event when it's re-created.

karlkfi: The same ListAndWatch pattern can be seen in the Informer: https://github.com/kubernetes/client-go/blob/release-1.26/tools/cache/reflector.go#L256

The difference is that this one is standalone, uses Unstructured objects (dynamic.Client), cancels with context instead of channel, doesn't handle retries/resync,  and handles errors better (doesn't just log and retry).

If I wanted to go further, I'd write a DynamicInformer, but the FilteredWatcher code already handles retries. We could also later write something like the Until watch tools, but we don't need them yet.

karlkfi: Actually, because this returns a watch.Interface, you can already use it with UntilWithoutRetry, if you want: https://github.com/kubernetes/client-go/blob/release-1.26/tools/watch/until.go#L56

pkg/remediator/watch/listwatch.go
```diff
@@ -0,0 +1,299 @@
...
+			return nil, fmt.Errorf("lister failed to get REST mapping for %s", gvk.String())
+		}
+		switch mapping.Scope.Name() {
+		case meta.RESTScopeNameRoot:
+			if namespace != "" {
+				return nil, fmt.Errorf("cannot list cluster-scoped resource %q in namespace %q", mapping.Resource, namespace)
+			}
+			return dynamicClient.Resource(mapping.Resource).List(ctx, options)
+		case meta.RESTScopeNameNamespace:
+			if namespace == "" {
```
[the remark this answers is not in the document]
karlkfi: sure. done

pkg/remediator/watch/watcher.go
karlkfi: good catch. removed.

pkg/remediator/watch/listwatch.go
```diff
@@ -0,0 +1,338 @@
...
+func ListAndWatch(ctx context.Context, lw ListerWatcher, opts metav1.ListOptions) (watch.Interface, error) {
+	if opts.ResourceVersion != "" {
+		// Just watch!
+		return lw.Watch(ctx, opts)
+	}
+
+	opts.AllowWatchBookmarks = false // not allowed for List
+	opts.Watch = false               // not allowed for List
+
+	// TODO: Use a short timeout for List and longer for Watch
```
sdowell, line 201: What's stopping this from happening now?

sdowell (commented):
/lgtm

pkg/remediator/watch/listwatch.go
karlkfi: Wasn't sure how to configure it cleanly.
I could hard code it, but that seemed wrong.
Also seems wrong to pass two timeouts into ListAndWatch.
I could do in the ListerWatcherFactory, but then I have to pass it through watcherConfig and duplicate the res.Config, because it only has one timeout setting, and I don't want to pass two rest.Config through a bunch of different layers when they're mostly identical.
Ideally rest.Config would be modified to support it, but that's an upstream client-go abstraction...

karlkfi: I think I can add a wrapper that returns different configs for Get, List, and Watch, but I'll need to do it in another PR, since it will touch everything...

sdowell: Ack, that seems reasonable

karlkfi: I spent hours on this and it's a giant headache.

There's a half dozen places where I'm not sure whether it should get the longer timeout or the shorter timeout, because the same client is used for CRUD, List, and Watch...

Today, only the remediator uses the longer watch timeout. Everything else uses the shorter timout, including the applier, which does its own watches, and the discovery clients, which poll long lists of CRDs...

karlkfi:
Pushed a small cleanup to NewFilteredListWatchFromClient to reduce the code duplication between list and watch, which both use the same dynamic.ResourceInterface.

PR #441, opened 2023-03-03
test: add notification server test component

sdowell:
This is an in cluster test component that enables simple testing of the notification controller using http/webhooks. It hosts a few http endpoints which record requests from the notification controller and can be queried from the test framework using a port forward.

Creating this PR before the notification controller to provide smaller/incremental code reviews and so that the test scaffolding is available before the notification controller PR is created

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -219,6 +219,7 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	if err := installWebhook(nt); err != nil {
 		nt.T.Fatal(err)
 	}
+
 	setupTestCase(nt, opts)
 	return nt
 }
@@ -365,6 +366,22 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 
 	if nt.GitProvider.Type() == e2e.Local {
 		nt.gitRepoPort = portForwardGitServer(nt, allRepos...)
+	}
+
+	// install notification server per-testcase only if the flag is provided
+	if opts.InstallNotificationServer {
+		nt.NotificationServer = &NotificationServer{}
+		nt.T.Cleanup(func() {
+			if err := nt.NotificationServer.uninstall(nt); err != nil {
+				nt.T.Errorf("failed to uninstall notification server: %v", err)
+			}
+		})
+		if err := nt.NotificationServer.install(nt); err != nil {
+			nt.T.Fatalf("failed to install notification server: %v", err)
+		}
+		if err := nt.NotificationServer.portForward(nt); err != nil {
+			nt.T.Fatalf("failed to port-forward notification server: %v", err)
+		}
 	}
 
 	for name := range opts.RootRepos {
--- a/test/docker/http-notification-webhook/main.go
+++ b/test/docker/http-notification-webhook/main.go
@@ -1,0 +1,102 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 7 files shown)


e2e/nomostest/new.go
```diff
@@ -337,6 +344,19 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
...
+		t.Fatalf("failed to install notification server: %v", err)
+	}
+	if err := nt.NotificationServer.portForward(nt); err != nil {
+		t.Fatalf("failed to port-forward notification server: %v", err)
+	}
+	t.Cleanup(func() {
+		if err := nt.NotificationServer.uninstall(nt); err != nil {
+			t.Errorf("failed to uninstall notification server: %v", err)
+		}
+	})
```
nan-yu, line 358: Should we install the test-notification-server for all tests?
Can we just install for the notification test specifically?

test/docker/http-notification-webhook/main.go
```diff
@@ -0,0 +1,102 @@
...
+	"kpt.dev/configsync/e2e/nomostest"
+)
+
+// mux is used to synchronize operations across the various endpoints
+var mux sync.Mutex
+
+// store is a simple in memory storage of notification records.
+var store nomostest.NotificationRecords
+
+func doGet(w http.ResponseWriter, r *http.Request) {
```
nan-yu, line 35: suggest renaming to `doGetAll` and `doDeleteAll`. Depending on the validation, you might need another method like `doGet` with an id.

e2e/nomostest/new.go
sdowell: Yeah I was thinking it might make sense to have an option for `nomostest.New` such as `WithInstallNotificationServer` or something similar

test/docker/http-notification-webhook/main.go
sdowell: I agree with the general premise, however I'm not sure that we have
- a use case for querying individual notifications
- a unique identifier for querying an individual notification

We could use something like an auto-incrementing integer or UUID as an ID, however I don't see that as being very usable in this use case. WDYT?

e2e/nomostest/new.go
sdowell: Done

test/docker/http-notification-webhook/main.go
nan-yu: Yes, that makes sense. We can change it later when needed.

e2e/nomostest/new.go
```diff
@@ -337,6 +346,21 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
...
+		}
+		if err := nt.NotificationServer.portForward(nt); err != nil {
+			t.Fatalf("failed to port-forward notification server: %v", err)
+		}
+		t.Cleanup(func() {
+			if err := nt.NotificationServer.uninstall(nt); err != nil {
+				t.Errorf("failed to uninstall notification server: %v", err)
+			}
+		})
+	}
```
nan-yu, line 362: Should we move the installation in `setupTestCase`?

The `InstallNotificationServer` option works when `--share--test-env` is `false`, but for shared GKE environment, it will install the notification server for all test cases. The port-forwarding will remain open for all tests, which might has the same issue as the metrics validation on autopilot clusters.

If we move it into `setupTestCase`, it will only install for the particular test with the `InstallNotificationServer` option.

If we have the code snippet in `FreshTestEnv`, you'll need

sdowell: Ah you're right the flag is enabled per-testcase which means the installation should be scoped to a testcase. Let me take another pass.

sdowell: updated to install on a per-testcase basis

nan-yu (approved):
/lgtm

PR #444, opened 2023-03-03
Change nomos version output in standalone OSS to match image version …

victorpras:
…(#436)

* Component output udpate and use image version

* Update reconciler manager container check


nan-yu (approved):
/lgtm

PR #445, opened 2023-03-03
Increase the reconcileTimeout on the Autopilot cluster

nan-yu:
The reconcileTimeout is set to 1 minute for all clusters, but it takes more time for a Deployment to be reconciled on an autopilot cluster. This commit increases the timeout to 2 minutes on autopilot clusters.

### DIFF
--- a/e2e/testcases/apiservice_test.go
+++ b/e2e/testcases/apiservice_test.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"testing"
+	"time"
 
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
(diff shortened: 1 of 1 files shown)


e2e/testcases/apiservice_test.go
```diff
@@ -31,7 +32,8 @@ const (
 )
 
 func TestCreateAPIServiceAndEndpointInTheSameCommit(t *testing.T) {
-	nt := nomostest.New(t, nomostesting.Reconciliation1, ntopts.Unstructured, ntopts.RequireGKE(t))
+	nt := nomostest.New(t, nomostesting.Reconciliation1, ntopts.Unstructured,
```
karlkfi, line 35: add comments as to why we're lengthening it, please.

nan-yu: DONE
Also added a link to the prow job in the commit.

karlkfi (approved):
/lgtm

PR #447, opened 2023-03-03
Fix abandonment

karlkfi:
- Rename disableObject to abandonObject, to be consistent with terminology used in the applier.
- Fix supervisor.abandonObject to not adopt all object fields. Use merge-patch, instead of server-side apply. SSA only works if you have the source of truth for all the fields you previously applied. The way it was being used assumed that all fields were owned by configsync.
- Fix supervisor to correctly abandon objects that have their delete or prune skipped with an AnnotationPreventedDeletionError. Previously, the configsync annotations weren't being removed and then the remediator detected it as drift and removed them.
- Use the client.Client in abandonObject. There's no need to use the dynamic.Client, because the client.Client can still populate an Unstructured object, and now that we're using merge-patch, we don't care about any other populated optional fields.
- Remove the now unused dynamic.Client from the ClientSet

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -384,14 +384,15 @@ func (a *supervisor) handleDeleteSkippedEvent(ctx context.Context, eventType eve
 	// Disable protected namespaces that were removed from the desired object set.
 	if isNamespace(obj) && differ.SpecialNamespaces[obj.GetName()] {
 		// the `client.lifecycle.config.k8s.io/deletion: detach` annotation is not a part of the Config Sync metadata, and will not be removed here.
-		err := a.disableObject(ctx, obj)
+		err := a.abandonObject(ctx, obj)
 		handleMetrics(ctx, "unmanage", err, id.WithVersion(""))
 		if err != nil {
-			errorMsg := "failed to remove the Config Sync metadata from %v (which is a special namespace): %v"
-			klog.Errorf(errorMsg, id, err)
-			return applierErrorBuilder.Wrap(fmt.Errorf(errorMsg, id, err)).Build()
+			err = fmt.Errorf("failed to remove the Config Sync metadata from %v (protected namespace): %v",
+				id, err)
+			klog.Error(err)
+			return applierErrorBuilder.Wrap(err).Build()
 		}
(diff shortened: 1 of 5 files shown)


pkg/applier/applier.go
```diff
@@ -797,30 +812,43 @@ func (a *supervisor) removeFromInventory(rg *live.InventoryResourceGroup, objs [
...
+	klog.Infof("Abandoning object: %s", core.IDOf(obj))
 	if metadata.HasConfigSyncMetadata(uObj) {
-		updated := metadata.RemoveConfigSyncMetadata(uObj)
+		// Use minimal before & after objects to simplify DeepCopy and building
+		// the merge patch.
+		fromObj := &unstructured.Unstructured{}
+		fromObj.SetGroupVersionKind(uObj.GroupVersionKind())
+		fromObj.SetNamespace(uObj.GetNamespace())
+		fromObj.SetName(uObj.GetName())
+		fromObj.SetAnnotations(uObj.GetAnnotations())
```
nan-yu, line 840: Also need to setLabels

pkg/applier/applier_test.go
```diff
@@ -226,46 +248,80 @@ func TestApply(t *testing.T) {
...
+			expectedGVKs: map[schema.GroupVersionKind]struct{}{
+				kinds.Deployment(): {},
+				testGVK:            {},
+			},
+			expectedServerObjs: []client.Object{
+				func() client.Object {
+					obj := abandonObj.DeepCopy()
+					obj.SetAnnotations(map[string]string{
+						common.LifecycleDeleteAnnotation: common.PreventDeletion,
+						// all configsync annotations removed
```
nan-yu, line 278: all configsync annotations **_and labels_** removed

pkg/applier/applier.go
karlkfi: fixed

pkg/applier/applier_test.go
karlkfi: fixed & added labels to test

nan-yu (approved):
/lgtm

karlkfi:
/retest

flaky timeout

PR #450, opened 2023-03-04
Remove GKE requirement in API service test

tiffanny29631:
The test can now run and pass on Kind clusters. Removing this constraint.


tiffanny29631:
@karlkfi The test on kind wouldn't pass back then and reporting GCE related errors.

tiffanny29631:
Misclick!

PR #451, opened 2023-03-06
Fix race condition (#415)

tiffanny29631:
In a rare condition hydration-controller and parser may proceed with file hydration or listing when git-sync see and updates the commit hash in status, which causes resource mismatch and accidental deletion.

This change blocks the risk by double checking if commit has changed during hydration or listing.

go/cs-race-condition


karlkfi (approved):
/lgtm

Note: cherrypick for v1.14.3

PR #453, opened 2023-03-07
Fix OCI tests to read latest commit from git

karlkfi:
Previously the OCI tests were using the latestSyncedCommit from the RootSync/RepoSync as the expected commit in the other status fields, which is circular logic.
This change uses git ls-remote to get the latest commit from the remote repo, without needing to clone it.
This unblocks an optimization for WaitForSync to avoid retrying the Sha1Fn.


sdowell (approved):
Great catch, thanks for fixing
/lgtm

PR #454, opened 2023-03-07
Make GetDeploymentPod return an error

karlkfi:
- Return an error, instead of failing the current test, to allow use in test setup and teardown.
- Use GetDeploymentPod to lookup the git-server pod
- Use GetDeploymentPod to lookup the otel-collector pod
- Add checks to ensure the deployment is updated and ready


nan-yu (approved):
/lgtm

PR #456, opened 2023-03-07
Fix test setup metric validation 

karlkfi:
- Always validate a standard set of metrics after test setup.
- Record the expected "declared resources" populated by test setup.
  This is used when validating standard metrics or as a base-line
  when calculating declared resources from a test.
- Clean up setupDelegatedControl and setupCentralizedControl to use
  RootRepos/NonRootRepos created by setupTestCase, rather than using
  the opts they came from.
- Configure reconcile timeout consistently.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"go.uber.org/multierr"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -32,7 +33,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e"
-	testmetrics "kpt.dev/configsync/e2e/nomostest/metrics"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	"kpt.dev/configsync/e2e/nomostest/testing"
@@ -446,17 +446,23 @@ func noOOMKilledContainer(nt *NT) Predicate {
 	}
 }
 
-func setupRootSync(nt *NT, rsName string) {
+func setupRootSync(nt *NT, rsName string, reconcileTimeout *time.Duration) {
 	// create RootSync to initialize the root reconciler.
 	rs := RootSyncObjectV1Beta1FromRootRepo(nt, rsName)
+	if reconcileTimeout != nil {
+		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(*reconcileTimeout)
+	}
 	if err := nt.Apply(rs); err != nil {
 		nt.T.Fatal(err)
 	}
 }
 
-func setupRepoSync(nt *NT, nn types.NamespacedName) {
+func setupRepoSync(nt *NT, nn types.NamespacedName, reconcileTimeout *time.Duration) {
 	// create RepoSync to initialize the Namespace reconciler.
 	rs := RepoSyncObjectV1Beta1FromNonRootRepo(nt, nn)
+	if reconcileTimeout != nil {
+		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(*reconcileTimeout)
+	}
 	if err := nt.Create(rs); err != nil {
 		nt.T.Fatal(err)
 	}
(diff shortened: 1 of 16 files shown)


e2e/nomostest/config_sync.go
```diff
@@ -979,6 +973,24 @@ func setupCentralizedControl(nt *NT, opts *ntopts.New) {
 	}
 }
 
+func validateRootSyncsExist(nt *NT, opts *ntopts.New) error {
+	var err error
+	for rsName := range opts.RootRepos {
+		err = multierr.Append(err,
+			nt.Validate(rsName, configsync.ControllerNamespace, &v1beta1.RootSync{}))
+	}
+	return nil
```
sdowell, line 982: return `err`

ditto for function below

karlkfi: fixed

karlkfi:
Significant changes. Needs another review.

e2e/nomostest/nt.go
```diff
@@ -425,25 +437,16 @@ func (nt *NT) MustMergePatch(obj client.Object, patch string, opts ...client.Pat
...
-		numObjects += numRepoSyncs               // 1 for each RepoSync
-		numObjects += numRepoSyncs               // 1 for each RepoSync RoleBinding
-		if isPSPCluster() {
-			numObjects += numRepoSyncs // 1 for each RepoSync ClusterRoleBinding
-		}
-	}
-	return numObjects
+// DefaultRootSyncObjectCount returns the number of declared resource objects in
+// the specified RootSync, according to records from test setup.
+func (nt *NT) DefaultRootSyncObjectCount(syncName string) int {
```
sdowell, line 442: I think something like `InitialR*SyncObjectCount` might be a bit more clear than `DefaultR*SyncObjectCount`

e2e/nomostest/config_sync.go
```diff
@@ -873,118 +855,121 @@ func RepoSyncObjectV1Beta1FromOtherRootRepo(nt *NT, nn types.NamespacedName, rep
...
-		if err := nt.WatchForAllSyncs(); err != nil {
-			nt.T.Fatal(err)
-		}
-		return
-	}
-
-	cr := nt.RepoSyncClusterRole()
-	nt.RootRepos[configsync.RootSyncName].Add("acme/cluster/cr.yaml", cr)
+	if len(nt.NonRootRepos) > 0 {
+		cr := nt.RepoSyncClusterRole()
```
sdowell, line 879: Why the change to only create the `RepoSyncClusterRole` if the ntopts define a RepoSync? There are some tests which create RepoSyncs within the test rather than via the test setup. This seems like it could be a gotcha

e2e/nomostest/config_sync.go
```diff
@@ -873,118 +855,121 @@ func RepoSyncObjectV1Beta1FromOtherRootRepo(nt *NT, nn types.NamespacedName, rep
...
-		// Wait for the RootSyncs and exit early
-		if err := nt.WatchForAllSyncs(); err != nil {
-			nt.T.Fatal(err)
-		}
-		return
-	}
-
-	cr := nt.RepoSyncClusterRole()
-	nt.RootRepos[configsync.RootSyncName].Add("acme/cluster/cr.yaml", cr)
+	if len(nt.NonRootRepos) > 0 {
```
sdowell, line 878: refactoring this `if` statement makes this diff unnecessarily large imo. also makes the `if` block quite long which makes it harder to follow

e2e/nomostest/nt.go
karlkfi: I had a brain wave so I'm refactoring these to `nt.expectedObjects` and `ExpectedRootSyncObjectCount` and adding Add & Remove methods.

e2e/nomostest/config_sync.go
karlkfi: Previously `if len(nt.NonRootRepos) > 0 {` was returning before this block. So the behavior hasn't changed.

e2e/nomostest/config_sync.go
karlkfi: That's why i had removed the if statement in a previous PR, but then that broke metric validation, which was being done after the If block.

e2e/nomostest/config_sync.go
sdowell: Ah I see, my mistake

e2e/nomostest/config_sync.go
karlkfi: hypothetically, but it's possible they may diverge in the future. This is just some bits extracted from https://github.com/GoogleContainerTools/kpt-config-sync/pull/412, which changes all the metrics validation impls. I just wanted them to be consistent first.

e2e/nomostest/nt.go
sdowell: ack, let me know when it's ready for review

e2e/nomostest/config_sync.go
karlkfi: done

PR #457, opened 2023-03-07
Fix e2e to avoid double setup

karlkfi:
- Run sharedTestEnv or freshTestEnv, not both.
- Only run test setup once per test
- Only run WatchForAllSyncs once per test setup
- Rename test setup funcs to make them private


sdowell (approved):
Nice change!
/lgtm

PR #462, opened 2023-03-09
test: fix ls-remote for CSR repos

sdowell:
Tests in CI run the ls-remote step against CSR repos, but these repos require authentication to access. This change adds a step to first clone such repos using gcloud, which can authenticate using workload identity and set up the git credentials.

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -30,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/testing/fake"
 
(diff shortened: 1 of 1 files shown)


e2e/nomostest/nt.go
```diff
@@ -1006,6 +1017,31 @@ func gitCommitFromSpec(nt *NT, gitSpec *v1beta1.Git) (string, error) {
...
+func cloneCloudSourceRepo(nt *NT, repo string) (string, error) {
+	tmpDir, err := os.MkdirTemp("", "sourcerepos")
+	if err != nil {
+		return "", err
+	}
+	nt.T.Cleanup(func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			nt.T.Errorf("failed to clean up tmpDir %s: %v", tmpDir, err)
+		}
+	})
```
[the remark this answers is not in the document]
sdowell: Oh good suggestion, I forgot about that dir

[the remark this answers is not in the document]
sdowell: Done

e2e/nomostest/nt.go
```diff
@@ -1006,6 +1017,26 @@ func gitCommitFromSpec(nt *NT, gitSpec *v1beta1.Git) (string, error) {
 	return columns[0], nil
 }
 
+// cloneCloudSourceRepo clones the provided Cloud Source Repository to a local
+// temp directory and returns the directory path.
+// This special logic is needed to handle authentication to the repo in CI.
+func cloneCloudSourceRepo(nt *NT, repo string) (string, error) {
+	repoName := path.Base(repo)
+	cloneDir := path.Join(nt.TmpDir, "csr-repos", repoName)
+	// return if directory was already created
```
nan-yu, line 1026: You're ahead of me. I was going to left a comment for the existence check and it just shows up after refresh. 😄

nan-yu (approved):
/lgtm

e2e/nomostest/nt.go
karlkfi: Suggest namespacing the directory by project too, not just repo name. Otherwise if you run locally there’s a chance it’s configured for a different project.

sdowell: `nt.TmpDir` is scoped to a single test case, and each test case is scoped to a single project (set by the env var). So I don't think this should be an issue, but it wouldn't hurt

karlkfi: 👍

PR #467, opened 2023-03-13
use same ignore list for add/lint license headers

sdowell:
The linting script was using a different ignore list than the add script, requiring manual fixing after adding the license headers using the script. This ensures both targets use the same ignore definitions.


nan-yu (approved):
/lgtm

PR #469, opened 2023-03-15
Nomos status rsync filter

victorpras:
Customers have been requesting a feature to filter the result of `nomos status` based on their selected RootSync object. To accommodate this, we added `name` flag that will filter the output of `nomos status` based on the provided root or repo sync name. This will work in conjunction with `namespace` flag that will filter based on namespace, and have been used by customer to filter RepoSync.

In this PR:
- add --name flag for sync name filtering
- update --namespace flag to work when the value is config-management-system
- update nomos status usage in OSS cluster to work with the new name and namespace flag
- e2e test for use cases and different combination of name and namepsace flag

### DIFF
--- a/cmd/nomos/status/client.go
+++ b/cmd/nomos/status/client.go
@@ -164,9 +164,9 @@ func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace st
 	}
 
 	if namespace == configsync.ControllerNamespace {
-		c.rootRepoClusterStatus(ctx, cs)
+		cs.Error = c.rootRepoClusterStatus(ctx, cs)
 	} else if namespace != "" {
-		c.namespaceRepoClusterStatus(ctx, cs, namespace)
+		cs.Error = c.namespaceRepoClusterStatus(ctx, cs, namespace)
 	} else if isOss || (cs.isMulti != nil && *cs.isMulti) {
 		c.multiRepoClusterStatus(ctx, cs)
 	} else {
@@ -249,16 +249,26 @@ func (c *ClusterClient) syncingConditionSupported(ctx context.Context) bool {
 // the multi repos on the ClusterClient's cluster.
 func (c *ClusterClient) multiRepoClusterStatus(ctx context.Context, cs *ClusterState) {
 	// Get the status of all RootSyncs
-	c.rootRepoClusterStatus(ctx, cs)
+	rootErr := c.rootRepoClusterStatus(ctx, cs)
 
 	// Get the status of all RepoSyncs
-	c.namespaceRepoClusterStatus(ctx, cs, "")
+	repoErr := c.namespaceRepoClusterStatus(ctx, cs, "")
+	if len(rootErr) > 0 {
+		cs.Error = fmt.Sprintf("Root repo error: %s", rootErr)
+	}
+	if len(repoErr) > 0 {
+		if len(cs.Error) > 0 {
+			cs.Error += ", "
+		}
+		cs.Error += fmt.Sprintf("Namespace repo error: %s", repoErr)
+	}
 }
 
 // rootRepoClusterStatus populates the given ClusterState with the sync status of
 // config-management-system namespace
-func (c *ClusterClient) rootRepoClusterStatus(ctx context.Context, cs *ClusterState) {
+func (c *ClusterClient) rootRepoClusterStatus(ctx context.Context, cs *ClusterState) (errorMsg string) {
 	var errs []string
+	var rootErr string
 	syncingConditionSupported := c.syncingConditionSupported(ctx)
 
 	// Get the status of all RootSyncs
@@ -293,17 +303,20 @@ func (c *ClusterClient) rootRepoClusterStatus(ctx context.Context, cs *ClusterSt
 
 	if len(errs) > 0 {
 		cs.status = util.ErrorMsg
-		cs.Error += strings.Join(errs, ", ")
+		rootErr = strings.Join(errs, ", ")
 	} else if len(cs.repos) == 0 {
 		cs.status = util.UnknownMsg
-		cs.Error = "No RootSync or RepoSync resources found"
+		rootErr = "No RootSync resources found"
 	}
+
+	return rootErr
 }
 
 // namespaceRepoClusterStatus populates the given ClusterState with the sync status of
 // the specified namespace repo on the ClusterClient's cluster.
-func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *ClusterState, ns string) {
+func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *ClusterState, ns string) (errorMsg string) {
 	var errs []string
+	var repoErr string
 	syncingConditionSupported := c.syncingConditionSupported(ctx)
 
 	var rgs []*unstructured.Unstructured
@@ -337,11 +350,13 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
 
 	if len(errs) > 0 {
 		cs.status = util.ErrorMsg
-		cs.Error += strings.Join(errs, ", ")
+		repoErr = strings.Join(errs, ", ")
 	} else if len(cs.repos) == 0 {
 		cs.status = util.UnknownMsg
-		cs.Error = "No RepoSync resources found"
+		repoErr = "No RepoSync resources found"
 	}
+
+	return repoErr
 }
 
 // IsInstalled returns true if the ClusterClient is connected to a cluster where
--- a/cmd/nomos/status/client.go
+++ b/cmd/nomos/status/client.go
@@ -146,29 +146,28 @@ func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace st
 		return cs
 	}
 
-	if isOss {
-		c.multiRepoClusterStatus(ctx, cs)
-		return cs
+	if !isOss {
+		cs.isMulti, err = c.ConfigManagement.IsMultiRepo(ctx)
+
+		if !c.IsInstalled(ctx, cs) {
+			return cs
+		}
+		if !c.IsConfigured(ctx, cs) {
+			return cs
+		}
+
+		if err != nil {
+			cs.status = util.ErrorMsg
+			cs.Error = err.Error()
+			return cs
+		}
 	}
 
-	cs.isMulti, err = c.ConfigManagement.IsMultiRepo(ctx)
-
-	if !c.IsInstalled(ctx, cs) {
-		return cs
-	}
-	if !c.IsConfigured(ctx, cs) {
-		return cs
-	}
-
-	if err != nil {
-		cs.status = util.ErrorMsg
-		cs.Error = err.Error()
-		return cs
-	}
-
-	if namespace != "" {
+	if namespace == configsync.ControllerNamespace {
+		c.rootRepoClusterStatus(ctx, cs)
+	} else if namespace != "" {
 		c.namespaceRepoClusterStatus(ctx, cs, namespace)
-	} else if cs.isMulti != nil && *cs.isMulti {
+	} else if isOss || (cs.isMulti != nil && *cs.isMulti) {
 		c.multiRepoClusterStatus(ctx, cs)
 	} else {
 		c.monoRepoClusterStatus(ctx, cs)
@@ -249,6 +248,16 @@ func (c *ClusterClient) syncingConditionSupported(ctx context.Context) bool {
 // multiRepoClusterStatus populates the given ClusterState with the sync status of
 // the multi repos on the ClusterClient's cluster.
 func (c *ClusterClient) multiRepoClusterStatus(ctx context.Context, cs *ClusterState) {
+	// Get the status of all RootSyncs
+	c.rootRepoClusterStatus(ctx, cs)
+
+	// Get the status of all RepoSyncs
+	c.namespaceRepoClusterStatus(ctx, cs, "")
+}
+
+// rootRepoClusterStatus populates the given ClusterState with the sync status of
+// config-management-system namespace
+func (c *ClusterClient) rootRepoClusterStatus(ctx context.Context, cs *ClusterState) {
 	var errs []string
 	syncingConditionSupported := c.syncingConditionSupported(ctx)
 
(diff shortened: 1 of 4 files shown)


cmd/nomos/status/cluster_state.go
```diff
@@ -55,7 +55,9 @@ func (c *ClusterState) printRows(writer io.Writer) {
 	}
 	for _, repo := range c.repos {
 		fmt.Fprintf(writer, "%s%s\n", util.Indent, util.Separator)
-		repo.printRows(writer)
+		if len(rsyncname) <= 0 || (len(rsyncname) > 1 && repo.syncName == rsyncname) {
```
sdowell, line 58: `if rsyncname == "" || rsyncname == repo.syncName`

e2e/testcases/cli_test.go
```diff
@@ -1068,3 +1068,22 @@ func TestNomosVersion(t *testing.T) {
...
+
+	// get status
+	cmd := nt.Command("nomos", "status", "--rsyncname", "repo-sync")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		nt.T.Log(string(out))
+		nt.T.Fatal(err)
+	}
+
+	if strings.Contains(string(out), "root-sync") {
```
sdowell, line 1086: This assertion isn't very strong. It should validate more about the output, e.g. that other syncs are not present in the output.

Perhaps we could mask the dynamic parts of the output (I assume commit SHA) so that we could do a more explicit assertion?

cmd/nomos/status/status.go
```diff
@@ -53,6 +54,7 @@ func init() {
 	Cmd.Flags().DurationVar(&pollingInterval, "poll", 0*time.Second, "Polling interval (leave unset to run once)")
 	Cmd.Flags().StringVar(&namespace, "namespace", "", "Namespace repo to get status for (multi-repo only, leave unset to get all repos)")
 	Cmd.Flags().BoolVar(&resourceStatus, "resources", true, "show resource level status for Namespace repo (multi-repo only)")
+	Cmd.Flags().StringVar(&rsyncname, "rsyncname", "", "rsyncname to filter rsync)")
```
sdowell, line 57: I don't think `rsyncname` is the right argument name here. We use this terminology informally to refer to RootSync|RepoSync, but I don't expect users to be familiar with this.

This also doesn't seem to give much control over filtering RootSync vs RepoSync. e.g. what if there are multiple RSyncs in the cluster that have the same `Name`? Perhaps as a user I would like to filter for RepoSyncs, or RootSyncs, or by a particular namespace, etc.

This also only allows for exact name matching. I imagine a user might want to support some form of pattern matching for more flexibility.

WDYT?

sdowell (commented):
Left some comments. Also please clean up the commit history and provide more context in the commit message

cmd/nomos/status/status.go
victorpras: yeah it might be better to use a different argument name, what would you suggest? Maybe something like `--resourceName`?

From what I understand, most feature request from the customer were about filtering for the RootSync that they are looking for. They typically find `--namespace` to be sufficient for RepoSync. In this PR I made the filter to work for both Root and Repo sync, but they can be separated, maybe with different flags for root and repo. What would you suggest to use as filter if not using the `Name`?

sdowell: I'd suggest an argument name such as `--name`

I think providing a `Name` filter makes sense, but an exact name matching filter on its own doesn't seem sufficient. At a minimum I think it should also support filtering by RootSync/RepoSync for this use case, e.g.
- an additional `sync-kind` type field that selects RootSync or RepoSync
- perhaps special handling for `--namespace=config-management-system` to list RootSyncs, although I'm not sure if this behavior might be confusing

nan-yu: Agree with using `--name` as the argument name, but should it be a plural?

I think we should filter the RootSyncs and RepoSyncs with a combination  of `--names` and `--namespaces` together.

When only `--namespaces` is specified, it should list all RepoSyncs/RootSyncs in the namespaces.

When only `--names` is specified, it should list all RepoSyncs and RootSyncs with the name in all namespaces.

When both `--namespaces` and `--names` are specified, it should only list the matching ones.

Given that RepoSync can't be in the `config-managmeent-system` namespace, while RootSync can only be in the namespace, I feel like `--sync-kind` seems unnecessary.

sdowell: My only hesitation is that `--namespace` currently seems to implicitly mean "list RepoSyncs in the provided namespace". I agree it would make more sense if this allowed listing `RootSync` as well (assuming `config-management-system` is provided), this will just require a bit larger of a change to the behavior.

Other than that caveat, +1 to @nan-yu

nan-yu: I take it back on the comments around `plural`. I thought the current `--namespace` is a plural, which supports listing RepoSyncs in multiple namespaces.

I think is is okay to use singular to filter with a particular namespace and/or name.

nan-yu: One benefit of adding `--sync-kind` is that it can support listing all RepoSyncs in all namespaces, which can't be achieved with the current arguments.

But my concern is that if `--sync-kind` is not provided, do we infer it from `--namespace`, or do we force it to be specified when `--namespace` or `--name` is provided?

sdowell: Yeah the different combinations would add quite a bit of complexity. Might be easier to discuss in a doc

victorpras: i have created a onepager doc to continue the discussion, holding this PR for now

cmd/nomos/status/cluster_state.go
victorpras: thanks, i have updated to follow your sample

cmd/nomos/status/status.go
victorpras: the use case has been updated to follow the discussion here and in the design docs

e2e/testcases/cli_test.go
victorpras: i have updated the assertion to validate namespace/resource that we expect to see and should not see in the result. Also added more test for different use case

cmd/nomos/status/client.go
```diff
@@ -168,6 +168,9 @@ func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace st
 
 	if namespace != "" {
 		c.namespaceRepoClusterStatus(ctx, cs, namespace)
+		if namespace == "config-management-system" {
+			c.multiRepoClusterStatus(ctx, cs)
```
nan-yu, line 172: The function list both RootSyncs and RepoSyncs, but you only need RootSyncs when namespace is cms.

cmd/nomos/status/client.go
```diff
@@ -168,6 +168,9 @@ func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace st
 
 	if namespace != "" {
 		c.namespaceRepoClusterStatus(ctx, cs, namespace)
```
nan-yu, line 170: `c.namespaceRepoClusterStatus` lists all RepoSyncs in the namespace no matter if the namespace is c-m-s or not.
Do you need it when namespace is c-m-s?

cmd/nomos/status/client.go
nan-yu, line 172: `c.multiRepoClusterStatus` lists all RootSyncs and RepoSyncs, but you only need RootSyncs here.

cmd/nomos/status/cluster_state.go
```diff
@@ -55,7 +55,9 @@ func (c *ClusterState) printRows(writer io.Writer) {
 	}
 	for _, repo := range c.repos {
 		fmt.Fprintf(writer, "%s%s\n", util.Indent, util.Separator)
-		repo.printRows(writer)
+		if name == "" || strings.Contains(repo.syncName, name) {
```
nan-yu, line 58: The --name flag should be an exact match. Can you check the entire string, rather than using `contains`?

e2e/testcases/cli_test.go
```diff
@@ -1068,3 +1068,46 @@ func TestNomosVersion(t *testing.T) {
 		nt.T.Fatalf("Expected to find config-sync component in output:\n%s\n", string(out))
 	}
 }
+
+func TestNomosStatusNameFilter(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI)
+
+	// get status
+	cmd := nt.Command("nomos", "status", "--name", "root")
```
nan-yu, line 1076: The RootSync name is `root-sync`, not `root`.

e2e/testcases/cli_test.go
```diff
@@ -1068,3 +1068,46 @@ func TestNomosVersion(t *testing.T) {
...
+	if strings.Contains(string(out), "bookinfo:repo-sync") {
+		nt.T.Fatalf("Expected to not find repo-sync component in output:\n%s\n", string(out))
+	}
+
+	if !strings.Contains(string(out), "<root>:root-sync") {
+		nt.T.Fatalf("Expected to find root-sync component in output:\n%s\n", string(out))
+	}
+}
+
+func TestNomosStatusNameNamespaceFilter(t *testing.T) {
```
nan-yu, line 1092: Using a separate test case requires resetting the test env when --share-test-env is true.
I would suggest combining them into a single test case.

Another suggestion is to test the combination of the --name and --namespace flag.

For example, in the test, you can create multiple RootSyncs and RepoSyncs with different namespaces.
Then your can verify:
- '--namespace=c-m-s' only should list all RootSyncs
- '--namespace=***' only should list all RepoSyncs in the specific namespace
- '--name=*** only' should list both RootSyncs and RepoSyncs with the same name (exact match)
- '--name=*** and --namespace=***' should list RootSyncs **OR** RepoSyncs that match with the flags.

nan-yu (changes requested):
Request changes

cmd/nomos/status/client.go
victorpras: the function has been split to do root only for c-m-s namespace

cmd/nomos/status/client.go
victorpras: the logic has been update to only show root for c-m-s ns

cmd/nomos/status/client.go
victorpras: as stated in previous comment, this has been split and updated to show only root

cmd/nomos/status/cluster_state.go
victorpras: the logic has been reverted to find an exact string matching

e2e/testcases/cli_test.go
victorpras: this was written with Strings.contains in mind, fixed now

e2e/testcases/cli_test.go
victorpras: the testcase has been merged to 1 big testcase that setup a cluster with multiple root and reposync and perform validation for each of the 4 listed scenario

e2e/testcases/cli_test.go
```diff
@@ -1068,3 +1068,91 @@ func TestNomosVersion(t *testing.T) {
...
 	}
 }
+
+func TestNomosStatusNameFilter(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI)
+
+	nt.MustKubectl("apply", "-f", "../testdata/filter/bookinfo.yaml")
+	nt.MustKubectl("apply", "-f", "../testdata/filter/reposync-bookinfo.yaml")
+	nt.MustKubectl("apply", "-f", "../testdata/filter/rootsync-crontab.yaml")
+	nt.MustKubectl("apply", "-f", "../testdata/filter/reposync-crontab.yaml")
```
nan-yu, line 1078: These will left the resources around after the test finishes, which causes cleanup issues in the shared test environment.

Please use the test framework to set up multiple R*Syncs.

nan-yu (changes requested):

cmd/nomos/status/client.go
```diff
@@ -147,7 +147,13 @@ func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace st
...
 
 	if isOss {
-		c.multiRepoClusterStatus(ctx, cs)
+		if namespace == configsync.ControllerNamespace {
+			c.rootRepoClusterStatus(ctx, cs)
+		} else if namespace != "" {
+			c.namespaceRepoClusterStatus(ctx, cs, namespace)
+		} else {
+			c.multiRepoClusterStatus(ctx, cs)
+		}
```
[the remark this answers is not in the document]
victorpras: reordered the sequential conditional to remove code duplication

cmd/nomos/status/client.go
```diff
@@ -253,58 +261,76 @@ func (c *ClusterClient) multiRepoClusterStatus(ctx context.Context, cs *ClusterS
 	syncingConditionSupported := c.syncingConditionSupported(ctx)
 
 	// Get the status of all RootSyncs
-	var rootRGs []*unstructured.Unstructured
-	rootSyncs, rootSyncNsAndNames, err := c.rootSyncs(ctx)
+	c.rootRepoClusterStatus(ctx, cs)
+
+	// Get the status of all RepoSyncs
```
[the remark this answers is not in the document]
victorpras: removed the identical code and have it call the func namespaceRepoClusterStatus

e2e/testcases/cli_test.go
victorpras: test setup has been changed to use the builtin from framework. All config files removed too

cmd/nomos/status/client.go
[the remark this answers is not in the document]
nan-yu: The original `multiRepoClusterStatus` function collects all errors from all RootSyncs and RepoSyncs. The current code calls the two functions, the second one will override the errors of the first one.

You need to make sure all errors are included in the result.

cmd/nomos/status/client.go
```diff
@@ -293,7 +293,7 @@ func (c *ClusterClient) rootRepoClusterStatus(ctx context.Context, cs *ClusterSt
 
 	if len(errs) > 0 {
 		cs.status = util.ErrorMsg
-		cs.Error = strings.Join(errs, ", ")
+		cs.Error += strings.Join(errs, ", ")
 	} else if len(cs.repos) == 0 {
 		cs.status = util.UnknownMsg
 		cs.Error = "No RootSync or RepoSync resources found"
```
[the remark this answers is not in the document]
victorpras: corrected error message

cmd/nomos/status/client.go
```diff
@@ -293,7 +293,7 @@ func (c *ClusterClient) rootRepoClusterStatus(ctx context.Context, cs *ClusterSt
 
 	if len(errs) > 0 {
 		cs.status = util.ErrorMsg
-		cs.Error = strings.Join(errs, ", ")
+		cs.Error += strings.Join(errs, ", ")
```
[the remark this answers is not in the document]
victorpras: Modified the 2 functions to return error message. Improved error message formatting. Examples: 

```
gke_victortjhia-gke-dev_us-central1-c_oss-cluster-2
  --------------------
  ERROR   Root repo error: resourcegroups.kpt.dev "root-sync-2" not found in namespace "config-management-system", resourcegroups.kpt.dev "root-sync-3" not found in namespace "config-management-system" 

gke_victortjhia-gke-dev_us-central1-c_operator-cluster-1
  --------------------
  ERROR   Root repo error: No RootSync resources found, Namespace repo error: resourcegroups.kpt.dev "repo-sync" not found in namespace "bookinfo"   
```

cmd/nomos/status/client.go
```diff
@@ -337,7 +337,7 @@ func (c *ClusterClient) namespaceRepoClusterStatus(ctx context.Context, cs *Clus
...
 	} else if len(cs.repos) == 0 {
 		cs.status = util.UnknownMsg
 		cs.Error = "No RepoSync resources found"
```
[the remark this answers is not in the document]
victorpras: error message formatting has been improved as per comment in https://github.com/GoogleContainerTools/kpt-config-sync/pull/469#discussion_r1152464149

nan-yu (approved):
/lgtm

PR #470, opened 2023-03-16
feat: manage notification controller

sdowell:
This change adds logic to manage the notification controller via the reconciler-manager. The reconciler-manager will create the notification controller as a pod in the reconciler deployment if the notification API is enabled in the RSync.

### DIFF
--- a/e2e/testcases/notification_test.go
+++ b/e2e/testcases/notification_test.go
@@ -1,0 +1,187 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"encoding/base64"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
+	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+)
+
+func TestNotification(t *testing.T) {
+	rootSyncNN := nomostest.RootSyncNN(configsync.RootSyncName)
+	repoSyncNN := nomostest.RepoSyncNN(backendNamespace, configsync.RepoSyncName)
+	rootReconcilerName := core.RootReconcilerName(rootSyncNN.Name)
+	backendReconcilerName := core.NsReconcilerName(repoSyncNN.Namespace, repoSyncNN.Name)
+	nt := nomostest.New(t, nomostesting.SyncSource,
+		ntopts.InstallNotificationServer,
+		ntopts.NamespaceRepo(repoSyncNN.Namespace, repoSyncNN.Name),
+	)
+	var err error
+	credentialMap := map[string]string{
+		rootSyncNN.Namespace: "pass123",
+		repoSyncNN.Namespace: "pass456",
+	}
+	for ns, pass := range credentialMap {
+		_, err = nomostest.NotificationSecret(nt, ns,
+			nomostest.WithNotificationUsername("user"),
+			nomostest.WithNotificationPassword(pass),
+		)
+		if err != nil {
+			nt.T.Fatal(err)
+		}
+		_, err = nomostest.NotificationConfigMap(nt, ns,
+			nomostest.WithOnSyncSyncedTrigger,
+			nomostest.WithSyncSyncedTemplate,
+			nomostest.WithLocalWebhookService,
+		)
+		if err != nil {
+			nt.T.Fatal(err)
+		}
+	}
+
+	repoSyncBackend := nomostest.RepoSyncObjectV1Beta1FromNonRootRepo(nt, repoSyncNN)
+	rootSync := nomostest.RootSyncObjectV1Beta1FromRootRepo(nt, rootSyncNN.Name)
+	if err := nt.Get(rootSyncNN.Name, rootSyncNN.Namespace, rootSync); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nomostest.SubscribeRootSyncNotification(rootSync, "on-sync-synced", "local")
+	if err := nt.Update(rootSync); err != nil {
+		nt.T.Fatal(err)
+	}
+	tg := taskgroup.New()
+	tg.Go(func() error {
+		return nomostest.WatchObject(nt, kinds.Deployment(), rootReconcilerName, rootSyncNN.Namespace,
+			[]nomostest.Predicate{nomostest.DeploymentHasContainer(reconcilermanager.Notification)})
+	})
+	tg.Go(func() error {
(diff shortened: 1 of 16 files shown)


pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -245,8 +246,50 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+				log.Error(updateErr, "Object status update failed",
+					logFieldObject, rsRef.String(),
+					logFieldKind, r.syncKind)
+			}
+			// Use the upsert error for metric tagging.
+			metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+			return controllerruntime.Result{}, errors.Wrap(err, "Secret validation failed")
+		}
+	}
+	log.Info("reached populateContainerEnvs")
```
nan-yu, line 290: Delete the debugging log

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -292,8 +293,51 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+			log.Error(updateErr, "Object status update failed",
+				logFieldObject, rsRef.String(),
+				logFieldKind, r.syncKind)
+		}
+		// Use the upsert error for metric tagging.
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+		return controllerruntime.Result{}, errors.Wrap(err, "Check notification subscription failed")
+	}
+
+	if notificationEnabled && rs.Spec.NotificationConfig != nil &&
```
nan-yu, line 313: It only validates if the Secret exists. We should also validate if the ConfigMap exists.

We also need to validate when it has the subscribe annotation, but NotificationConfig is nil. It should return an error.

We can probably do the validation inside the `NotificationEnabled` function.

pkg/webhook/fields.go
```diff
@@ -91,6 +92,21 @@ func ConfigSyncMetadata(set *fieldpath.Set) *fieldpath.Set {
 	return csSet
 }
 
+// OnlyNotifiedAnnotation returns if the filePath set only includes the notified annotation.
```
nan-yu, line 95: This is not needed anymore. It was used to prevent the fight between the admission webhook and the notification controller. After Karl made the change in the admission webhook (https://github.com/GoogleContainerTools/kpt-config-sync/pull/242/files#diff-9e60d0c27478b517b75a5e3e5dbd109c91c99abe0afc21b256ee8f9827d7502cR100) and our discussion, we decided to not control it at this level.

pkg/webhook/fields_test.go
nan-yu, line 1: The changes in this file is also not needed.

pkg/webhook/validate.go
nan-yu, line 1: Not needed as well.

pkg/reconcilermanager/controllers/reposync_controller_test.go
nan-yu, line 1: We can add some validation tests at the unit test level. For example, the subscribe annotation exists, but no NotificationConfig, or missing ConfigMapRef or SecretRef.

pkg/reconcilermanager/controllers/reposync_controller.go
sdowell: I moved all of the validation into the `NotificationEnabled` function and added more checks, PTAL

pkg/webhook/fields.go
sdowell: done

pkg/reconcilermanager/controllers/reposync_controller_test.go
sdowell: Added some unit tests in this package with more extensive unit testing on the `NotificationEnabled` function

pkg/util/notification.go
```diff
@@ -38,4 +47,66 @@ const (
...
+		if apierrors.IsNotFound(err) {
+			return false, fmt.Errorf("notification ConfigMap %s not found in the %s namespace", cmObjectKey.Name, cmObjectKey.Namespace)
+		}
+		return false, err
+	}
+	// subscription via ConfigMap
+	_, found := cm.Data[MultiSubscriptionsField]
+	if !hasSubscriptionAnnotation && !found {
+		return false, fmt.Errorf("notificationConfig defined without any subscriptions")
+	}
```
nan-yu, line 97: This should be allowed.
Suppose that a user originally enabled notification, but then the user just unsubscribed the annotation with the notification config left behind. It seems to be a valid case.

pkg/util/notification.go
```diff
@@ -38,4 +47,66 @@ const (
...
+		}
+	}
+	if hasSubscriptionAnnotation && notificationConfig == nil {
+		return false, fmt.Errorf("subscription annotation was provided without a notificationConfig")
+	}
+	if notificationConfig == nil {
+		return false, nil
+	}
+	// ConfigMapRef must be defined
+	configMapName := v1beta1.GetConfigMapName(notificationConfig.ConfigMapRef)
```
nan-yu, line 77: If there is no subscription annotation, should this be allowed?

e2e/testcases/notification_test.go
```diff
@@ -0,0 +1,173 @@
...
+	nomostest.SubscribeRootSyncNotification(rootSync, "on-sync-synced", "local")
+	if err := nt.Update(rootSync); err != nil {
+		nt.T.Fatal(err)
+	}
+	err = nomostest.WatchObject(nt, kinds.Deployment(), rootReconcilerName, rootSyncNN.Namespace,
+		[]nomostest.Predicate{nomostest.DeploymentHasContainer(reconcilermanager.Notification)})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	records, err := waitForNotifications(nt, 1)
```
nan-yu, line 81: It may be also worthwhile to validate the `notified` annotation.

pkg/util/notification.go
sdowell: My thinking was that `notificationConfig` should be empty (`nil`) if the user is not enabling the notification feature. And if the user _is_ enabling the notification feature, the `configMapRef` _must_ be set.

Are you suggesting that we should return `false, nil` if `notificationConfig` is not `nil` but `configMapRef` is unset/empty?

pkg/util/notification.go
sdowell: That seems reasonable - I'd say it's valid but notifications are essentially disabled in that case. I decided to return an error here because I figured it may be confusing for users if they missed a step to fully configure notifications but there is no error.

Do you think we should return true or false?

e2e/testcases/notification_test.go
sdowell: Sure, I can add validation for that as well

pkg/util/notification.go
nan-yu: My thought is `if !hasSubscriptionNotification && notificationConfig != nil`, we can return `false, nil` no matter if `configMapRef` is set/empty or not.

Basically, if the subscription is not enabled, we don't need to validate the notificationConfig content.
One example is that a user originally enabled the subscription with configMapRef, then the user unsubscribed, and the ConfigMap was deleted from the cluster. We shouldn't return a validation error for such case, because the ConfigMap is not needed anymore.

sdowell: The only issue I see with that is the ConfigMap-based default subscriptions (set via the `subscriptions` field). A valid/enabled notification configuration could have the ConfigMap set with default subscriptions but no annotations

nan-yu: If the user would like to use the ConfigMap-based subscriptions, he/she should configure the configMapRef. If it is not configured, we can just treat it like disabled.

pkg/util/notification.go
nan-yu: I feel like we should return false in this case. No subscription annotation, notificationConfig.configMapRef exists, but no subscription == subscription is disabled.

Otherwise, if a user wants to unsubscribe notification, he/she has to clean up the notificationConfig.

pkg/util/notification.go
sdowell: done

pkg/util/notification.go
sdowell: done

e2e/testcases/notification_test.go
sdowell: done

nan-yu (approved):
/lgtm

PR #472, opened 2023-03-16
test: only update ns if it was changed

sdowell:
This is an optimization to simply return if the namespace had no annotations or labels removed. We are also seeing an issue on autopilot where the update is being denied by a webhook, so this should hopefully fix that.

### DIFF
--- a/e2e/nomostest/clean.go
+++ b/e2e/nomostest/clean.go
@@ -281,6 +281,10 @@ func deleteConfigSyncAndTestAnnotationsAndLabels(nt *NT, ns *corev1.Namespace) e
 			labels[k] = v
 		}
 	}
+	// return without updating if no annotations/labels were removed
+	if len(annotations) == len(ns.Annotations) && len(labels) == len(ns.Labels) {
+		return nil
+	}
 
 	ns.Annotations = annotations
 	ns.Labels = labels


e2e/nomostest/clean.go
```diff
@@ -271,16 +271,24 @@ func isTestLabel(key string) bool {
 func deleteConfigSyncAndTestAnnotationsAndLabels(nt *NT, ns *corev1.Namespace) error {
 	var annotations = make(map[string]string)
 	var labels = make(map[string]string)
+	changed := false
```
karlkfi, line 274: nit: This might be cleared as `found`, since it's looking for CS labels and annotations. We don't actually know if it was changed and we're not really checking for other changes.

sdowell: Agreed that seems more clear

e2e/nomostest/clean.go
```diff
@@ -271,16 +271,24 @@ func isTestLabel(key string) bool {
...
 		}
 	}
 	for k, v := range ns.Labels {
 		if !isConfigSyncLabel(k, v) && !isTestLabel(k) {
 			labels[k] = v
+		} else {
+			found = true
 		}
 	}
+	if !found {
```
haiyanmeng, line 289: Why not check the length of `annotations` and `labels`?

karlkfi: that might be more clear to read, yeah.

sdowell: updated

karlkfi (approved):
/lgtm

haiyanmeng (approved):

PR #474, opened 2023-03-16
do not error test if port forward finished

sdowell:
This scenario occurs during test cleanup - if killing the port-forward process fails because the process is already finished, we can treat this as a non-fatal error.

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -870,7 +870,7 @@ func (nt *NT) ForwardToFreePort(ns, pod, port string) (int, error) {
 		cleanup = true
 		nt.T.Log("stopping port-forward %s/%s:%s process", ns, pod, port)
 		err := cmd.Process.Kill()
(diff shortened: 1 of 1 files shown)


e2e/nomostest/nt.go
```diff
@@ -870,7 +870,7 @@ func (nt *NT) ForwardToFreePort(ns, pod, port string) (int, error) {
 		cleanup = true
 		nt.T.Log("stopping port-forward %s/%s:%s process", ns, pod, port)
 		err := cmd.Process.Kill()
-		if err != nil {
+		if err != nil && !errors.Is(err, os.ErrProcessDone) {
```
karlkfi, line 873: I have code to change this to use Context cancel in https://github.com/GoogleContainerTools/kpt-config-sync/pull/412, but this works for now. Thanks!

karlkfi (approved):
/lgtm

PR #475, opened 2023-03-17
add apiservice teardown to e2e clean

sdowell:
There have been instances of the Cleanup failing where the backing Deployments are deleted but the APIServices remain. This results in API discovery errors that prevent namespaces from being deleted. This change ensures the test-generated APIServices are deleted in Cleanup.


nan-yu (approved):
/lgtm

nan-yu (approved):
/lgtm

e2e/nomostest/clean.go
```diff
@@ -606,3 +610,12 @@ func disableRootSyncDeletionPropagation(nt *NT) error {
 	}
 	return nil
 }
+
+func deleteTestAPIServices(nt *NT) error {
+	out, err := nt.Kubectl("delete", "apiservices", "-l", "testdata=true")
```
karlkfi, line 615: Would have preferred this use the existing deleteObjects that uses the SDK, instead of shelling out to kubectl.

PR #478, opened 2023-03-17
test: keep port forwards open when pod changes

sdowell:
This updates the port forwarding logic to spawn a thread which ensures the port forward stays open if the deployment pod changes. The git server also needs to have its git repositories re-initialized because it uses emptyDir storage.

### DIFF
--- a/e2e/nomostest/git.go
+++ b/e2e/nomostest/git.go
@@ -23,10 +23,12 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e/nomostest/retry"
 	"kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
 	"kpt.dev/configsync/e2e/nomostest/testlogger"
@@ -39,9 +41,6 @@ import (
 )
 
 const (
-	// remoteOrigin is static as every git repository has exactly one remote.
-	remoteOrigin = "origin"
-
 	// MainBranch is static as behavior when switching branches is never under
 	// test.
 	MainBranch = "main"
@@ -140,7 +139,11 @@ func NewRepository(nt *NT, repoType RepoType, nn types.NamespacedName, sourceFor
 		nt.T.Fatal(err)
 	}
 	g.RemoteRepoName = repoName
-	g.RemoteURL = nt.GitProvider.RemoteURL(nt.gitRepoPort, repoName)
+	port, err := nt.gitRepoPortForwarder.LocalPort()
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	g.RemoteURL = nt.GitProvider.RemoteURL(port, repoName)
 
 	g.init(nt.gitPrivateKeyPath)
 	g.initialCommit(sourceFormat)
@@ -154,8 +157,12 @@ func (g *Repository) ReInit(nt *NT, sourceFormat filesystem.SourceFormat) {
 
 	// Update test environment
 	g.T = nt.T
+	port, err := nt.gitRepoPortForwarder.LocalPort()
+	if err != nil {
+		nt.T.Fatal(err)
+	}
 	// Update URL to use latest port-forward port
-	g.RemoteURL = nt.GitProvider.RemoteURL(nt.gitRepoPort, g.RemoteRepoName)
+	g.RemoteURL = nt.GitProvider.RemoteURL(port, g.RemoteRepoName)
 	// Reset repo contents
 	g.init(nt.gitPrivateKeyPath)
 	g.initialCommit(sourceFormat)
@@ -241,8 +248,6 @@ func (g *Repository) init(privateKey string) {
 	// 2) Use the private key file we generated.
 	g.Git("config", "core.sshCommand",
 		fmt.Sprintf("ssh -q -o StrictHostKeyChecking=no -i %s", privateKey))
-	// Point the origin remote
-	g.Git("remote", "add", remoteOrigin, g.RemoteURL)
 }
 
 // Add writes a YAML or JSON representation of obj to `path` in the git
@@ -501,7 +506,35 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) {
 	g.Git("commit", "-m", msg)
 
 	g.Logger.Infof("[repo %s] committing %q (%s)", path.Base(g.Root), msg, g.Hash())
-	g.Git("push", "-f", "-u", remoteOrigin, branch)
+	g.Push(branch, "-f")
+}
+
+// Push pushes the provided refspec to the git server.
+// Performs a retry using RemoteURL, which may change if the port forwarding restarts.
+func (g *Repository) Push(refspec string, flags ...string) {
+	var out []byte
+	took, err := retry.Retry(1*time.Minute, func() error {
+		args := []string{"push"}
+		args = append(args, flags...)
+		args = append(args, g.RemoteURL, refspec)
+		cmd := g.gitCmd(args...)
+		var err error
+		out, err = cmd.CombinedOutput()
+		return err
+	})
+	if err != nil {
+		g.T.Log(string(out))
+		g.T.Fatalf("took %v, err %v", took, err)
+	}
+}
+
+func (g *Repository) pushAllToRemote(remote string) {
+	cmd := g.gitCmd("push", remote, "--all")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		g.T.Log(string(out))
+		g.T.Errorf("failed to push to remote %s: %v", remote, err)
+	}
 }
 
 // CreateBranch creates and checkouts a new branch at once.
@@ -525,8 +558,8 @@ func (g *Repository) RenameBranch(current, new string) {
 	g.T.Helper()
 
 	g.Git("branch", "-m", current, new)
-	g.Git("push", remoteOrigin, "-u", new)
-	g.Git("push", remoteOrigin, "--delete", current)
+	g.Push(new)
+	g.Push(current, "--delete")
 }
 
 // Hash returns the current hash of the git repository.
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -182,12 +182,6 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		GitProvider:             sharedNt.GitProvider,
 		RemoteRepositories:      sharedNt.RemoteRepositories,
 		WebhookDisabled:         sharedNt.WebhookDisabled,
-		// Reset git-repo port-forward for each test
-		gitRepoPort:    0,
-		gitRepoPodName: "",
-		// Reset otel-collector port-forward for each test
-		otelCollectorPort:    0,
-		otelCollectorPodName: "",
 	}
 
 	if opts.SkipConfigSyncInstall {
@@ -375,7 +369,7 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 
 	if nt.GitProvider.Type() == e2e.Local {
 		InitGitRepos(nt, allRepos...)
-		nt.PortForwardGitServer()
+		nt.portForwardGitServer()
 	}
 
 	for name := range opts.RootRepos {
@@ -414,7 +408,8 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 		nt.T.Fatalf("waiting for ConfigSync Deployments to become available: %v", err)
 	}
 
-	nt.PortForwardOtelCollector()
+	nt.portForwardOtelCollector()
+	nt.portForwardPrometheus()
 
 	nt.Control = opts.Control
 	switch opts.Control {
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
 	"strings"
@@ -29,7 +28,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
-	"kpt.dev/configsync/e2e/nomostest/retry"
+	"kpt.dev/configsync/e2e/nomostest/portforwarder"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
 	"kpt.dev/configsync/e2e/nomostest/testlogger"
@@ -148,11 +147,11 @@ type NT struct {
 	// caCertPath is the path to the CA cert used for communicating with the Git server.
 	caCertPath string
 
-	// gitRepoPort is the local port that forwards to the git repo deployment.
-	gitRepoPort int
+	// gitRepoPortForwarder is the local port forwarding for the git server deployment.
+	gitRepoPortForwarder *portforwarder.PortForwarder
 
-	// gitRepoPodName is the pod name of the git repo.
-	gitRepoPodName string
+	// prometheusPortForwarder is the local port forwarding for the prometheus deployment.
+	prometheusPortForwarder *portforwarder.PortForwarder
 
 	// kubeconfigPath is the path to the kubeconfig file for the kind cluster
 	kubeconfigPath string
@@ -160,11 +159,8 @@ type NT struct {
 	// Scheme is the Scheme for the test suite that maps from structs to GVKs.
 	Scheme *runtime.Scheme
 
-	// otelCollectorPort is the local port that forwards to the otel-collector.
-	otelCollectorPort int
-
-	// otelCollectorPodName is the pod name of the otel-collector.
-	otelCollectorPodName string
+	// otelCollectorPort is the local port forwarding for the otel-collector.
+	otelCollectorPortForwarder *portforwarder.PortForwarder
 
 	// GitProvider is the provider that hosts the Git repositories.
 	GitProvider gitproviders.GitProvider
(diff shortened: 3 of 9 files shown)


e2e/nomostest/new.go
```diff
@@ -181,11 +181,9 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		// Reset cached metrics for each test
 		ReconcilerMetrics: make(testmetrics.ConfigSyncMetrics),
 		// Reset git-repo port-forward for each test
-		gitRepoPort:    0,
-		gitRepoPodName: "",
```
karlkfi, line 185: Why remove these?

Yes, the empty string is default, but I wanted to make it clear these are being reset on each test.

sdowell: The pod name is scoped to a function closure now, so we don't need to keep track of each of these in the higher level NT struct. It's still reset on each test

e2e/nomostest/nt.go
```diff
@@ -793,59 +788,69 @@ func (nt *NT) MustDeleteGatekeeperTestData(file, name string) {
...
-			port, err := nt.ForwardToFreePort(ocmetrics.MonitoringNamespace, pod.Name, testmetrics.MetricsPort)
-			if err != nil {
-				return err
-			}
-			nt.otelCollectorPort = port
-			nt.otelCollectorPodName = pod.Name
+// portForwardOtelCollector forwards the otel-collector pod.
+func (nt *NT) portForwardOtelCollector() {
+	if nt.otelCollectorPort == 0 {
+		var mux sync.Mutex
```
karlkfi, line 794: Not sure i like having the lock outside of the struct that owns the protected field.
Won't this replace the lock if the value's set to zero?

I suggest moving this all into a PortForwarder object and having one instance each for otel-collector and git-server.

e2e/nomostest/nt.go
```diff
@@ -793,59 +788,69 @@ func (nt *NT) MustDeleteGatekeeperTestData(file, name string) {
...
+			// allRepos specifies the slice all repos for port forwarding.
+			var allRepos []types.NamespacedName
+			for repo := range nt.RootRepos {
+				allRepos = append(allRepos, RootSyncNN(repo))
+			}
+			for repo := range nt.NonRootRepos {
+				allRepos = append(allRepos, repo)
+			}
+			// re-init all repos
+			InitGitRepos(nt, allRepos...)
```
karlkfi, line 833: I don't understand why all this logic is in the setGitRepoPort function...?

sdowell: > The git server also needs to have its git repositories re-initialized because it uses emptyDir storage.

We could alternatively use a PersistentVolume but this seemed like a simpler solution

e2e/nomostest/nt.go
sdowell: This shouldn't happen because the value should never get set back to zero, but I suppose a struct pattern might be a better fit. I'll take a pass at refactoring this

sdowell: Updated to use a `portForward` struct. PTAL

e2e/nomostest/new.go
```diff
@@ -400,7 +394,7 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 		nt.T.Fatalf("waiting for ConfigSync Deployments to become available: %v", err)
 	}
 
-	nt.PortForwardOtelCollector()
+	nt.portForwardOtelCollector()
```
karlkfi, line 397: What if test setup just makes a new empty PortForwarder for the otel-collector and git-server when it creates the test NT. Then when you call nt.GitServerPortForwarder.GetPort() it lazily starts forwarding?

e2e/nomostest/nt.go
karlkfi: Do they need to be re-initialized ever time port-forwarding restarts? Or just every time it's a new pod? Or are those always the same?

e2e/nomostest/new.go
karlkfi: Usually the way to handle that problem is to have a private method that only the struct calls and then the public methods owns the lock.

e2e/nomostest/nt.go
sdowell: yeah this only gets reached when the pod restarts. There's also a check above to see if the pod name changed

e2e/nomostest/new.go
sdowell: updated to lazily start forwarding

e2e/nomostest/prometheus_metrics.go
```diff
@@ -46,22 +46,26 @@ type MetricsPredicate func(context.Context, prometheusv1.API) error
...
-	client, err := prometheusapi.NewClient(prometheusapi.Config{
-		Address: fmt.Sprintf("http://%s", address),
-	})
-	if err != nil {
-		return err
-	}
-	v1api := prometheusv1.NewAPI(client)
+	defer cancel()
+	var address string
+	var client prometheusapi.Client
```
karlkfi, line 51: the client doesn't need to be exposed outside of the retry, just the v1api that wraps i.

e2e/nomostest/port_forward.go
```diff
@@ -0,0 +1,198 @@
...
+	"os/exec"
+	"strings"
+	"sync"
+	"time"
+
+	"kpt.dev/configsync/e2e"
+	"kpt.dev/configsync/pkg/kinds"
+)
+
+type portForward struct {
```
karlkfi, line 31: Can we move this to a sub-package and rename it PortForwarder or DeploymentProxy? "port forward" is more of a verb than a noun.

e2e/nomostest/nt.go
```diff
@@ -752,123 +749,61 @@ func (nt *NT) MustDeleteGatekeeperTestData(file, name string) {
...
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		default:
-			s := stdout.String()
-			if !strings.Contains(s, "\n") {
-				return fmt.Errorf(
-					"nothing written to stdout for kubectl port-forward to pod %s/%s, stdout=%s",
-					ns, pod, s)
+	if nt.gitRepoPortForward == nil {
```
karlkfi, line 765: Don't like all these unlocked nil checks on the NT.
Can you just construct the PortForwarder/DeploymentProxy in test setup and then the places that need to use it can call nt.GitRepoProxy.LocalAddress() or nt.GitRepoProxy.LocalPort()?

e2e/nomostest/prometheus_metrics.go
sdowell: Done

e2e/nomostest/port_forward.go
karlkfi: If it's just GetDeploymentPod, you can move that to a subdir too. Make it return an error, then wrap it with one that fails in NT.

e2e/nomostest/nt.go
sdowell: This is only called in the test setup. The nil check is just to be defensive in case someone added code which calls it a second time but I can remove it if you prefer

e2e/nomostest/port_forward.go
karlkfi: Oh, it uses the nt.GetObject things too huh... maybe those also should be in a testutil package... but then they also use the debug log... sigh

sdowell: Yeah, this would require a fairly extensive refactor. We also would need to move the associated `WatchObject` functions to a subpackage.

e2e/nomostest/nt.go
karlkfi: If that's the case, make it return if not-nil (fast fail) and log a warning with why that's not necessary. Or maybe even Fatal, if we don't want it to happen.

e2e/nomostest/port_forward.go
karlkfi: Is this feasible now?

sdowell: The change has been rebased and moved to a subpackage

e2e/nomostest/nt.go
sdowell: done

e2e/nomostest/git.go
```diff
@@ -501,7 +506,35 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) {
...
+		out, err = cmd.CombinedOutput()
+		return err
+	})
+	if err != nil {
+		g.T.Log(string(out))
+		g.T.Fatalf("took %v, err %v", took, err)
+	}
+}
+
+func (g *Repository) pushAllToRemote(remote string) {
```
karlkfi, line 531: Feels like this should be public. But maybe we need to move the Repository itself into a subpackage in a subsequent PR.

e2e/nomostest/nt.go
```diff
@@ -148,23 +147,20 @@ type NT struct {
...
 
-	// gitRepoPort is the local port that forwards to the git repo deployment.
-	gitRepoPort int
+	// gitRepoPortForward is the local port forwarding for the git server deployment.
+	gitRepoPortForward *portforwarder.PortForwarder
 
-	// gitRepoPodName is the pod name of the git repo.
-	gitRepoPodName string
+	// prometheusPortForward is the local port forwarding for the prometheus deployment.
+	prometheusPortForward *portforwarder.PortForwarder
```
karlkfi, line 154: nit: these could be named `prometheusProxy` and `gitRepoProxy`. Ro at least add -er to match the structs.

e2e/nomostest/nt.go
```diff
@@ -568,123 +564,83 @@ func (nt *NT) MustDeleteGatekeeperTestData(file, name string) {
...
-			nt.otelCollectorPort = port
-			nt.otelCollectorPodName = pod.Name
-		}
-		return nil
-	})
-	if err != nil {
-		nt.T.Fatal(err)
-	}
-	nt.T.Logf("took %v to wait for otel-collector port-forward", took)
+func (nt *NT) newPortForward(ns, deployment, port string, opts ...portforwarder.PortForwardOpt) *portforwarder.PortForwarder {
```
karlkfi, line 567: nit: -er

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+// WithPortForwardSetFn registers a callback that will be invoked whenever the
+// deployment's underlying pod changes and the port-forward is recreated.
+func WithPortForwardSetFn(setFn func(int, string)) PortForwardOpt {
+	return func(pf *PortForwarder) {
+		pf.setFn = setFn
+	}
+}
+
+// NewPortForward creates a new instance of a PortForwarder
+func NewPortForward(
```
karlkfi, line 81: nit: -er

e2e/nomostest/nt.go
```diff
@@ -568,123 +564,83 @@ func (nt *NT) MustDeleteGatekeeperTestData(file, name string) {
...
-		}
-		nt.T.Cleanup(cancel) // stop port-forward
-		nt.gitRepoPort = port
-		nt.gitRepoPodName = podName
+// portForwardOtelCollector forwards the otel-collector deployment to a port.
+func (nt *NT) portForwardOtelCollector() {
+	if nt.otelCollectorPortForward != nil {
+		nt.T.Fatal("otel collector port forward already initialized")
 	}
+	nt.otelCollectorPortForward = nt.newPortForward(
```
karlkfi, line 587: What if we just construct these in the test setup func? Are they called from elsewhere? Does it need to be its own function?

karlkfi: nvm. i see portForwardGitServer is more complex and it's nice to be consistent.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+	// pod is the name of the current Pod that is being forwarded to
+	pod string
+}
+
+// PortForwardOpt is an optional parameter for PortForwarder
+type PortForwardOpt func(pf *PortForwarder)
+
+// WithPortForwardSetFn registers a callback that will be invoked whenever the
+// deployment's underlying pod changes and the port-forward is recreated.
+func WithPortForwardSetFn(setFn func(int, string)) PortForwardOpt {
```
karlkfi, line 74: "PortForward" is redundant with the package now. This can just be WithSetFn. But fwiw, I usually prefer "func" over "fn" for these. And "setFunc" isn't very self-descriptive... if it's called when the forwarding is set up, maybe "setupFunc" or "initFunc"?

e2e/nomostest/nt.go
sdowell: done

e2e/nomostest/nt.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/new.go
karlkfi: thanks!

e2e/nomostest/git.go
sdowell: Yeah if this goes into its own package then we'll need to make it public. At this moment in time though this is only intended to be used within the `nomostest` package, so I think the method being private demonstrates the intent.

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: renamed

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	for _, opt := range opts {
+		opt(pf)
+	}
+	return pf
+}
+
+// GetPort returns the current localhost port which forwards to the pod.
+// lazily starts the port forward when called the first time.
+func (pf *PortForwarder) GetPort() (int, error) {
```
karlkfi, line 109: nit... i have feelings about getters... feel free to take action or make a follow up PR or just tell me you dont care.

I think the go authors recommend avoiding "get" as a prefix for getters funcs. I don't know if I always agree with that tho. I tend to go with a get- prefix if it's a dumb function that just returns a value and avoid the get- prefix when it does some work to compute the result. This is a bit of a hybrid, due to the caching, but regardless, I think `LocalPort()` is a slightly more informative name. It also helps distinguish from the remote port being proxied to. Another option would be to have an `Address()` with the `localhost:` prefix but without a protocol. That would work even better if the name of the struct was `DeploymentPodProxy` or something, because it would be the "address of the proxy", which makes a little more sense than "address of the port forwarder". A port forwarder "forwards" as the verb, or "starts forwarding" but a proxy would have "start", "stop", and "address" as verbs (or just "address" if it automatically starts and stops).

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	pf.logger.Infof("took %v to wait for port-forward to pod %s/%s (localhost:%d)", took, pf.ns, pod, localPort)
+	pf.setPort(localPort, pod, async)
+	return cmd, nil
+}
+
+// keepPortForwardOpen creates a port forward to a deployment's pod and keeps it open,
+// even if the pod is recreated.
+// Once ctx is cancelled, the port forward will be closed and will no longer be recreated
+func (pf *PortForwarder) keepPortForwardOpen() error {
```
karlkfi, line 187: "Start"?

e2e/nomostest/prometheus_metrics.go
```diff
@@ -47,23 +47,29 @@ type MetricsPredicate func(context.Context, prometheusv1.API) error
...
 	nt.T.Log("[METRICS] validating prometheus metrics...")
 	for i, predicate := range predicates {
 		duration, err := retry.Retry(nt.DefaultMetricsTimeout, func() error {
-			err := predicate(ctx, v1api)
+			port, err := nt.prometheusPortForwarder.GetPort()
+			if err != nil {
+				return err
+			}
+			newAddress := fmt.Sprintf("localhost:%d", port)
+			if newAddress != address {
```
karlkfi, line 62: I have a feeling this is an unnecessary optimization that only really saves on a little memory churn from re-creating the client, but it's ok to leave it in. Maybe leave a comment to explain the added complexity tho.

e2e/testcases/root_sync_test.go
```diff
@@ -341,7 +341,7 @@ func TestForceRevert(t *testing.T) {
 	}
 
 	nt.RootRepos[configsync.RootSyncName].Git("reset", "--hard", "HEAD^")
-	nt.RootRepos[configsync.RootSyncName].Git("push", "-f", "origin", "main")
+	nt.RootRepos[configsync.RootSyncName].Push("main", "-f")
```
karlkfi, line 344: `MainBranch` const

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: TLDR: "proxy" sounds like a service you start and stop. "port forwarder" sounds like a utility you have to actuate with "forward". So if we rename the struct to something like "proxy" then the verbs can be more service-like: start, stop, address.

e2e/testcases/root_sync_test.go
sdowell: done

e2e/nomostest/prometheus_metrics.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: renamed to `LocalPort` and `start`

There are some existing interfaces which expect the port rather than an address (e.g. `GitProvider`), so I think at this point it makes sense to keep the port abstraction. I don't feel too strongly about the name of the struct/package - I think something like `DeploymentPodProxy` makes sense too

e2e/nomostest/git.go
```diff
@@ -501,7 +506,35 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) {
...
+	cmd := g.gitCmd("push", remote, "--all")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		g.T.Log(out)
```
karlkfi, line 535: gotta cast the out bytes to string, otherwise it'll verbosely print the byte array, which isn't human readable.

e2e/nomostest/git.go
```diff
@@ -501,7 +506,35 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) {
...
+		return err
+	})
+	if err != nil {
+		g.T.Log(string(out))
+		g.T.Fatalf("took %v, err %v", took, err)
+	}
+}
+
+func (g *Repository) pushAllToRemote(remote string) {
+	cmd := g.gitCmd("push", remote, "--all")
```
karlkfi, line 532: The distinction between Git and gitCmd are confusing, but predate this PR. In the future, we probably want to rename Git to like MustGit or something, since it fails and doesn't return the output. Then gitCmd could be named Git and return the output.

Anyway, it's fine for now, just a note.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+					pf.logger.Info(err)
+				}
+			}
+			select {
+			case <-pf.ctx.Done():
+				return
+			default:
+			}
+			pf.logger.Info("stopping port-forward process")
+			if err := cmd.Process.Kill(); err != nil && errors.Is(err, os.ErrProcessDone) {
```
karlkfi, line 211: Do we need to wait for the process to exit after being killed?
Or does the fact that it uses a random port every time mean that it doesn't matter if it's shut down when we start it up again?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+	go func() {
+		for {
+			if err := pf.watcher.WatchForNotFound(kinds.Pod(), pod.Name, pf.ns, testwatcher.WatchTimeout(1*time.Hour), testwatcher.WatchContext(pf.ctx)); err != nil {
+				if *e2e.Debug {
+					pf.logger.Info(err)
+				}
+			}
+			select {
+			case <-pf.ctx.Done():
+				return
```
karlkfi, line 207: It's mildy confusing that this doesn't kill the process. Is this because the process was started with the same context and will kill itself?
If so, can you add a comment here about why we don't need to kill it?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+			default:
+			}
+			pf.logger.Info("stopping port-forward process")
+			if err := cmd.Process.Kill(); err != nil && errors.Is(err, os.ErrProcessDone) {
+				pf.logger.Info("error: killing port-forward process: %v", err)
```
karlkfi, line 212: So i guess with the logger we can't kill the test, but it's still weird to have it log "errors". Should they be "warning"? Are we waiting for the next command that uses the port to fail to fail the tests? If so, can you leave a comment explaining that this doesn't need to be terminal because the next attempt to use the LocalPort will error?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+	cmd, portForwardErr := pf.portForwardToPod(pod.Name, false)
+	if portForwardErr != nil {
+		return portForwardErr
+	}
+	// Restart the port forward if the process exits prematurely (e.g. due to pod eviction)
+	go func() {
+		for {
+			if err := pf.watcher.WatchForNotFound(kinds.Pod(), pod.Name, pf.ns, testwatcher.WatchTimeout(1*time.Hour), testwatcher.WatchContext(pf.ctx)); err != nil {
```
karlkfi, line 200: nit: This might be a little cleaner if the for loop calls out to an inner method which returns errors with the `errors.Wrap(err, "more info")`. Then the code here in the loop can handle logging the error and return.

karlkfi (changes requested):
Just a couple minor things. Thanks

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: I don't think so. I also don't think it's strictly necessary to kill the process here, since as you said, it uses a random port and will eventually get cleaned up when the context is cancelled. The main reason I left this here was just to not leave unnecessary subprocesses running in the interim

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: Yeah since we pass the same context to the process, it should exit when the context is done. I can leave a comment

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: I originally had this calling `Fatal`, but refactored this when switching from `NT` to `Logger`. I decided to leave these as logs since we don't necessarily need to fail the test if the test is done using the port forward. But you are correct that the test would fail if it tried to use the port forward again after this failed. I can leave a comment

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: refactored

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: left a comment

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: left a comment

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: left a comment

e2e/nomostest/git.go
sdowell: done

karlkfi (approved):
/lgtm

PR #479, opened 2023-03-20
test: only update ns if it was changed (#472)

sdowell:
This is an optimization to simply return if the namespace had no annotations or labels removed. We are also seeing an issue on autopilot where the update is being denied by a webhook, so this should hopefully fix that.


karlkfi (approved):
/lgtm

PR #465, opened 2023-03-10
bump Go version to 1.19

sdowell:
Dependencies for the notification-engine require Go version 1.19

- upgrade buildenv image to support Go 1.19
- upgrade Go linter to compatible version
- allow default exclusions for linter
- fix linter issues

PR #481, opened 2023-03-21
Update golangci-lint to v1.52.0

karlkfi:
- Ignore "should have a package comment" for now (many)
- Ignore "error strings should not be capitalized", because Config Sync uses errors for user-readable status messages.


sdowell:
perhaps we could consider cherry picking https://github.com/GoogleContainerTools/kpt-config-sync/pull/465 instead? This would make it simpler to rebase the feature branch

karlkfi:
@sdowell https://github.com/GoogleContainerTools/kpt-config-sync/pull/465 looks too big. linter changes should probably be separate from version bumps.

When you update to 1.19 tho, you will need to update the staticcheck linter to 1.1.9. I'm only updating it to 1.17 here.

karlkfi:
/unhold

I fixed all the unused parameter failures.

sdowell:
Agreed it would have been an easier cherry pick if I had separated the linter change. That being said, is there an urgent reason for this PR? #465 is already submitted, so we could get the linter bump when merging the feature branch

karlkfi:
Linting is failing on https://github.com/GoogleContainerTools/kpt-config-sync/pull/412

Not super sure why, but I realized our lint-go.sh script was using master to download the install script, instead of the versioned tag. So I pinned it, but it was still failing. It also tends to hang occasionally. So I upgraded it, which caused test failures. So I fixed them

PR #480, opened 2023-03-21
Modify the metrics for CS smokedetector setup

tiffanny29631:
This change modifies metrics to align with Monarch definition

- includes rg_reconcile_duration_seconds in Monarch pipeline for future SLI setup

- clean up kcc_resource_count_total exclusion since the metric has been removed from Resource Group Controller

PR #483, opened 2023-03-21
Add type & commit metric tags

karlkfi:
- Added `type` tag to the following metrics:
  - APICallDuration
  - ApplyOperations
  - RemediateDuration
- Added `commit` tag to the following metrics:
  - DeclaredResources
  - LastApply
  - ResourceConflicts
- Filtered the `type` and `commit` tags from all metrics when exporting to Monarch

### Reasoning
- **Type** is important for validating metrics occurred because of a specific type (aka Kind) of object. Type allows verifying that the expected number of API calls were made for the expected resource types. Without this granularity, it's really hard to debug why the count is off when it doesn't match expectations.
- **Commit** is important for validating metrics occurred because of a specific source change. Without commit, it's difficult to validate that a metric was recorded after or before any given commit. Testing expiration is especially hard without commit, because many of these metrics aggregate by count, instead of by last value. So the count keeps increasing for every new occurrence and doesn't decrease until the metric timestamps becomes older than the cache boundary.

### DIFF
--- a/pkg/remediator/reconcile/worker_test.go
+++ b/pkg/remediator/reconcile/worker_test.go
@@ -22,9 +22,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/stretchr/testify/require"
-	"go.opencensus.io/stats/view"
-	"go.opencensus.io/tag"
+	"github.com/google/uuid"
 	v1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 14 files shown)


tiffanny29631:
The Monarch pipeline would start to complain about undefined fields for `status` and `commit` for the modified metrics in this PR. (Yeah they are all included in the pipeline)

The `commit` tag is easier to exclude since it has never been used in a metric that is defined in Monarch: you can specify it in the attribute/kubernetes processor. I added the exclusion in this https://github.com/GoogleContainerTools/kpt-config-sync/pull/480 but seems like it's also necessary here. 

The `status` tag is more complicated, some metrics are already defined in Monarch with it, like reconciler_error, so excluding it entirely from the pipeline may cause lose of data. For that, since `status` comes with fixed values, we could add them into Monarch definitions first, let it roll out and then proceed with this change.

tiffanny29631:
> The Monarch pipeline would start to complain about undefined fields for `status` and `commit` for the modified metrics in this PR. (Yeah they are all included in the pipeline)
> 
> The `commit` tag is easier to exclude since it has never been used in a metric that is defined in Monarch: you can specify it in the attribute/kubernetes processor. I added the exclusion in this #480 but seems like it's also necessary here.
> 
> The `status` tag is more complicated, some metrics are already defined in Monarch with it, like reconciler_error, so excluding it entirely from the pipeline may cause lose of data. For that, since `status` comes with fixed values, we could add them into Monarch definitions first, let it roll out and then proceed with this change.

To correct myself - ignore the comment regarding `status`. The `type` tag can have similar process with `commit`.

tiffanny29631 (commented):
/lgtm

tiffanny29631:
See if Haiyan or Morten has more thoughts.

pkg/remediator/reconcile/worker_test.go
```diff
@@ -85,7 +79,7 @@ func TestWorker_Run_Remediates(t *testing.T) {
 		}
 	}
 
-	d := makeDeclared(t, declaredObjs...)
+	d := makeDeclared(t, "commit-unused", declaredObjs...)
```
haiyanmeng, line 82: You may want to define a constant for "commit-unused", considering it is being used multiple times here.

karlkfi: Surprisingly, the linter didn't complain. Must not have hit the threshold for extracting constants.

karlkfi: Used a UUID generator instead of a constant. Now at least we can see if it ends up being accidentally tested.

PR #484, opened 2023-03-23
Fix the cleanup for fleet memberships

nan-yu:
In the original cleanup code, it checks the membership object on the cluster. Sometimes, the membership object is deleted successfully, while the membership resource still exists in the GCP project.

This commit updates the cleanup function to check the membership resource using gcloud.

### DIFF
--- a/e2e/testcases/csr_auth_test.go
+++ b/e2e/testcases/csr_auth_test.go
@@ -162,64 +162,59 @@ func TestFleetWIDifferentProject(t *testing.T) {
 	})
 }
 
-// getMembershipIdentityProvider fetches the workload_identity_pool if the membership exists.
-func getMembershipIdentityProvider(nt *nomostest.NT) (bool, string, error) {
-	bytes, err := nt.Kubectl("get", "membership", "membership")
+// getMembership gets the membership resource in the GCP project.
+// It returns a boolean that indicates whether the membership exists, and an error
+func getMembership(fleetMembership, gcpProject string) (bool, error) {
+	bytes, err := exec.Command("gcloud", "container", "fleet", "memberships", "describe", fleetMembership, "--project", gcpProject).CombinedOutput()
 	out := string(bytes)
 	if err != nil {
-		if strings.Contains(out, `the server doesn't have a resource type "membership"`) || strings.Contains(out, "NotFound") {
-			return false, "", nil
+		// ERROR: (gcloud.container.fleet.memberships.describe) Membership xxx not found in the fleet.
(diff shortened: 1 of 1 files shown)


e2e/testcases/csr_auth_test.go
```diff
@@ -162,64 +162,62 @@ func TestFleetWIDifferentProject(t *testing.T) {
...
+// getMembership gets the membership resource in the GCP project.
+// It returns a boolean that indicates whether the membership exists, and an error
+func getMembership(fleetMembership, gcpProject string) (bool, error) {
+	bytes, err := exec.Command("gcloud", "container", "fleet", "memberships", "describe", fleetMembership, "--project", gcpProject).CombinedOutput()
 	out := string(bytes)
 	if err != nil {
-		if strings.Contains(out, `the server doesn't have a resource type "membership"`) || strings.Contains(out, "NotFound") {
-			return false, "", nil
+		// ERROR: (gcloud.container.fleet.memberships.describe) Membership xxx not found in the fleet.
+		if strings.Contains(out, "not found") {
```
karlkfi, line 172: Seems like this could catch other similar errors. Maybe check for soemthing more exact, like `Membership %s not found in the fleet.` with Sprintf?

e2e/testcases/csr_auth_test.go
```diff
@@ -162,64 +162,62 @@ func TestFleetWIDifferentProject(t *testing.T) {
...
-	membershipExists, wiPool, err := getMembershipIdentityProvider(nt)
+	exists, err := getMembership(fleetMembership, gcpProject)
 	if err != nil {
 		nt.T.Error(err)
-		nt.T.Skip("Skip the test because unable to check if membership exists")
-	} else if membershipExists {
-		fleetProject := gcpProject
-		if len(wiPool) > 0 {
-			fleetProject = strings.TrimSuffix(wiPool, ".svc.id.goog")
+		nt.T.Skip("Skip the test because unable to check get membership and its workload identity pool")
```
karlkfi, line 187: I suspect this should just Fatal, not Skip.

e2e/testcases/csr_auth_test.go
nan-yu: Sure. My concern is that the message is from Hub controller and might change.

I guess it is okay to update the string if it changes in the future.

DONE

e2e/testcases/csr_auth_test.go
nan-yu: DONE

karlkfi (approved):
/lgtm

PR #490, opened 2023-03-24
Add a mutex to make the fight detector threadsafe

nan-yu:


pkg/syncer/reconcile/fight_detector.go
```diff
@@ -57,12 +58,15 @@ func FightWarning(frequency float64, resource client.Object) status.ResourceErro
 // Instantiate with newFightDetector().
 //
 // Performance characteristics:
-//  1. Current implementation is NOT threadsafe.
+//  1. Current implementation is threadsafe.
 //  2. Current implementation has unbounded memory usage on the order of the
 //     number of objects the Syncer updates through its lifetime.
 //  3. Updating an already-tracked resource requires no memory allocations and
 //     take approximately 30ns, ignoring logging time.
 type fightDetector struct {
```
sdowell, line 66: I think perhaps this should be defined in it's own package with public methods to safeguard against future changes directly accessing the protected variables. Currently it looks like this is all used within the same package, so the private fields aren't really protected semantically.

nan-yu: That'a s great point. Done.

pkg/syncer/reconcile/fight/logger.go
```diff
@@ -21,28 +21,28 @@ import (
...
+// Logger is used to log errors about fights from Detector at most
 // once every 60 seconds. It has similar performance characteristics as
-// fightDetector.
+// Detector.
 //
-// Instantiate with newFightLogger().
-type fightLogger struct {
-	// lastLogged is when fightLogger last logged about a given API resource.
+// Instantiate with NewLogger().
+type Logger struct {
```
sdowell, line 29: Should this be private? I assume it's only intended to be used within this package

nan-yu: Well, it is instantiated in `pkg/syncer/reconcile/apply.go`. Maybe I should just move the instantiation into the detector.

nan-yu: Done

PR #491, opened 2023-03-24
Moves the fight error to the status package

nan-yu:
This will be a user-facing error and exposed in the R*Sync status field.


karlkfi (approved):
/lgtm

PR #492, opened 2023-03-24
Replace gknn with core.ID

nan-yu:


pkg/syncer/reconcile/fight/detector.go
```diff
@@ -59,15 +59,15 @@ type Detector struct {
 
 	// fights is a record of how much the Syncer is fighting over any given
 	// API resource.
-	fights map[gknn]*fight
+	fights map[core.ID]*fight
```
sdowell, line 62: is the motivation that `core.ID` is the more canonical identifier? or is there some other reason?

nan-yu: Yes, the `gknn` struct is basically the same as `core.ID`. We should try to unify them.

karlkfi: FWIW, GKNN is used by the remediator object queue. So you'll have to translate back and forth or refactor that too.

nan-yu: The remediator object queue uses `GVKNN`, which has an additional version information than core.ID. I think the translation is still needed.

PR #493, opened 2023-03-24
Surface the fight errors in the R*Sync status field

nan-yu:
Previously, if Config Sync constantly fights with other controllers on certain resources, Config Sync only logs the fight as a warning in the reconciler log. The fight is not exposed to the user in the R*Sync status.

This commit adds the fight errors to the `status.sync` field, which is more visible to the user so they can take actions to address the fight.

### DIFF
--- a/e2e/testcases/remediator_test.go
+++ b/e2e/testcases/remediator_test.go
@@ -1,0 +1,83 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/metrics"
+	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/pkg/api/configmanagement"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/status"
+	"kpt.dev/configsync/pkg/testing/fake"
+)
+
+func TestSurfaceFightError(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.DriftControl)
+
+	nt.T.Logf("Stop the admission webhook to generate the fights")
+	nomostest.StopWebhook(nt)
+
+	ns := fake.NamespaceObject("test-ns", core.Annotation("foo", "bar"))
+	rb := roleBinding("test-rb", ns.Name, map[string]string{"foo": "bar"})
+	nt.RootRepos[configsync.RootSyncName].Add(
+		fmt.Sprintf("acme/namespaces/%s/ns.yaml", ns.Name), ns)
+	nt.RootRepos[configsync.RootSyncName].Add(
+		fmt.Sprintf("acme/namespaces/%s/rb.yaml", ns.Name), rb)
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Add Namespace and RoleBinding")
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// Make the # of updates exceed the fightThreshold defined in pkg/syncer/reconcile/fight_detector.go
+	go func() {
+		for i := 0; i <= 5; i++ {
+			nt.MustMergePatch(ns, `{"metadata": {"annotations": {"foo": "baz"}}}`)
+			nt.MustMergePatch(rb, `{"metadata": {"annotations": {"foo": "baz"}}}`)
+			time.Sleep(time.Second)
+		}
+	}()
+
+	nt.T.Log("The RootSync reports a fight error")
+	nt.WaitForRootSyncSyncError(configsync.RootSyncName, status.FightErrorCode,
+		"This may indicate Config Sync is fighting with another controller over the object.")
+
+	rootReconcilerPod, err := nt.KubeClient.GetDeploymentPod(
+		nomostest.DefaultRootReconcilerName, configmanagement.ControllerNamespace,
+		nt.DefaultWaitTimeout)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	commitHash := nt.RootRepos[configsync.RootSyncName].Hash()
+
+	err = nomostest.ValidateMetrics(nt,
+		nomostest.ReconcilerErrorMetrics(nt, rootReconcilerPod.Name, commitHash, metrics.ErrorSummary{
+			Fights: 5,
+		}))
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("The fight error should be auto-resolved if no more fights")
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+}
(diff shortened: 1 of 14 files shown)


pkg/remediator/reconcile/worker.go
```diff
@@ -35,15 +35,22 @@ import (
 // Worker pulls objects from a work queue and passes them to its reconciler for
 // remediation.
 type Worker struct {
-	objectQueue queue.Interface
-	reconciler  reconcilerInterface
+	objectQueue          queue.Interface
+	reconciler           reconcilerInterface
+	addFightErrorFunc    func(id core.ID, e status.Error)
```
karlkfi, line 40: Not a big fan of injecting functions like this. It exports business logic and makes it hard to find/debug.

Is there any way to avoid it by wrapping the Worker or injecting a FightHandler "strategy pattern" struct?

pkg/parse/updater.go
```diff
@@ -80,6 +81,16 @@ func (u *updater) conflictErrors() status.MultiError {
 	return errs
 }
 
+// fightErrors converts []Error into []MultiErrors.
+// This method is safe to call while Update is running.
+func (u *updater) fightErrors() status.MultiError {
```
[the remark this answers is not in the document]
nan-yu: Not necessary. The caller (`func (u *updater) Errors() status.MultiError`) has a lock on `errs`.

pkg/remediator/remediator.go
```diff
@@ -92,16 +101,18 @@ var _ Interface = &Remediator{}
 // It is safe for decls to be modified after they have been passed into the
 // Remediator.
 func New(scope declared.Scope, syncName string, cfg *rest.Config, applier syncerreconcile.Applier, decls *declared.Resources, numWorkers int) (*Remediator, error) {
+	remediator := &Remediator{
+		fightErrs: map[core.ID]status.Error{},
```
karlkfi, line 105: This kind of required initialization should be done with a New constructor, or modify the methods so they lazy init the map.

pkg/syncer/reconcile/fight/detector.go
```diff
@@ -73,15 +73,17 @@ func NewDetector() Detector {
 }
 
 // DetectFight detects whether the resource is needing updates too frequently.
-// If so, it increments the resource_fights metric and logs to klog.Warning.
-func (d *Detector) DetectFight(ctx ctx.Context, time time.Time, obj *unstructured.Unstructured, operation string) bool {
-	if fight := d.markUpdated(time, obj); fight != nil {
+// If so, it increments the resource_fights metric and logs to klog.Error.
+func (d *Detector) DetectFight(ctx ctx.Context, time time.Time,
+	obj *unstructured.Unstructured, operation string) (bool, status.ResourceError) {
+	fightErr := d.markUpdated(time, obj)
```
karlkfi, line 79: This is a little weird. The `markUpdated` method should return "mark errors" or "update errors". Calling them "fight errors" leaks implementation details of the function not implied by the name.

And if you change this to just `err` then it's confusing why you're passing all errors to the `logFight` function.

Maybe this function should instead do an error type check for FightError and only call logFight if it's confirmed to be a FightError?

Otherwise, maybe there's a way to combine these functions? Or were they made separate on purpose to scope the lock to not include RecordResourceFight &  logFight?

pkg/remediator/reconcile/worker.go
nan-yu: Changed to use a FightHandler to update the fight errors.

pkg/remediator/remediator.go
nan-yu: DONE. Refactored to use a New constructor for the FightHandler initialization.

pkg/syncer/reconcile/fight/detector.go
nan-yu: We don't need a type check for FightError because this method only detects the fight error, but not other types.

I think the confusion might be caused by the function name. I refactored it a little bit. Hopefully it is clearer than before.

pkg/syncer/reconcile/fight/detector_test.go
```diff
@@ -261,7 +261,7 @@ func TestResourceFightsMetricValidation(t *testing.T) {
 			u := fake.UnstructuredObject(tc.gvk)
 
 			for _, op := range tc.operations {
-				fd.DetectFight(context.Background(), time.Now(), u, op)
+				_, _ = fd.DetectFight(context.Background(), time.Now(), u, op)
```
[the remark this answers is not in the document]
nan-yu: Done. Added assertions.

nan-yu (commented):
Addressed all comments.

e2e/testcases/remediator_test.go
```diff
@@ -0,0 +1,83 @@
...
+	nt.RootRepos[configsync.RootSyncName].Add(
+		fmt.Sprintf("acme/namespaces/%s/rb.yaml", ns.Name), rb)
+	nt.RootRepos[configsync.RootSyncName].CommitAndPush("Add Namespace and RoleBinding")
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// Make the # of updates exceed the fightThreshold defined in pkg/syncer/reconcile/fight_detector.go
+	go func() {
+		for i := 0; i <= 5; i++ {
```
karlkfi, line 51: Is this going to be flakey if CS misses any of these events?

pkg/remediator/reconcile/reconciler.go
```diff
@@ -85,11 +90,21 @@ func (r *reconciler) Remediate(ctx context.Context, id core.ID, obj client.Objec
...
 	metrics.RecordRemediateDuration(ctx, metrics.StatusTagKey(err), id.WithVersion(""), start)
 
-	// Record conflict, if there was one
-	if err != nil && err.Code() == syncerclient.ResourceConflictCode {
-		metrics.RecordResourceConflict(ctx, id.WithVersion(""), commit)
+	if err != nil {
+		switch err.Code() {
+		case syncerclient.ResourceConflictCode:
+			// Record conflict, if there was one
+			metrics.RecordResourceConflict(ctx, id.WithVersion(""), commit)
```
karlkfi, line 97: Wanna do everyone a favor later and convert all the metrics record functions to just use a kind string, instead of the whole GVK, since we only record the kind? It's so silly we keep having to pass an empty version to build a GVK.

pkg/remediator/reconcile/reconciler.go
```diff
@@ -85,11 +90,21 @@ func (r *reconciler) Remediate(ctx context.Context, id core.ID, obj client.Objec
...
+		case status.FightErrorCode:
+			operation := objDiff.Operation(r.scope, r.syncName)
+			metrics.RecordResourceFight(ctx, string(operation), id.WithVersion(""))
+			r.fightHandler.AddFightError(id, err)
```
karlkfi, line 101: Now that we have a FightHandler... would it make sense to also add a ConflictHandler? It's kinda weird we add one here but not the other.

e2e/testcases/remediator_test.go
nan-yu: You mean if the patch fails? If it happens, I think the test will fail with `nt.T.Fatal(err)`.

pkg/remediator/reconcile/reconciler.go
nan-yu: I have a local PR ready for the ConflictHandler. Will push out for review once this PR is merged.

pkg/remediator/reconcile/reconciler.go
nan-yu: Sure. Can I make the change after this PR?

pkg/remediator/remediator.go
```diff
@@ -82,6 +86,8 @@ type Interface interface {
 	ManagementConflict() bool
 	// ConflictErrors returns the errors the remediator encounters.
 	ConflictErrors() []status.ManagementConflictError
+	// FightErrors returns the fight errors (KNV2005) the remediator encounters.
+	FightErrors() map[core.ID]status.Error
```
karlkfi, line 90: Should we maybe move this method to be on the FightHandler?
Then we could expose the FightHandler as a public field and call it from there. It might make it cleaner to move the locks out of the remediator. 

We could also do the same with a ConflictHandler.

karlkfi: Oh, I see you already did it this way, nice!

Never mind then.

e2e/testcases/remediator_test.go
karlkfi: No, I mean if the patch succeeds, but the remediator watch misses the event. I think this would only happen if the reconciler is rescheduled, but that can happen on autopilot clusters when the node pool is changed.

I don't know a good way to defend against it tho, so maybe it doesn't matter.

karlkfi (approved):
/lgtm

Looks great with the new FightHandler! Thanks for the cleanup.

PR #494, opened 2023-03-24
Move nomostest.Retry to retry package

karlkfi:


e2e/nomostest/predicates.go
```diff
@@ -54,7 +55,8 @@ var ErrObjectNotFound = errors.New("object not found")
 // WrongTypeErr reports that the passed type was not equivalent to the wanted
 // type.
 func WrongTypeErr(got, want interface{}) error {
-	return fmt.Errorf("%w: got %T, want %T", ErrWrongType, got, want)
```
sdowell, line 57: Do we still have a need for `ErrWrongType`? It looks like at this point it's basically just used as a string const

karlkfi: Probably not... Next time ;)

PR #496, opened 2023-03-27
test: refactor watch/predicates into subpackages

sdowell:
This is intended to decouple the watch and predicate code from the nomostest package. This provides better separation and enables building abstractions on top of watch/predicates without needing to couple them with the nomostest package.


e2e/nomostest/nt.go
```diff
@@ -386,6 +390,7 @@ func DefaultRepoSha1Fn(nt *NT, nn types.NamespacedName) (string, error) {
 func (nt *NT) RenewClient() {
 	nt.T.Helper()
 	nt.KubeClient = newTestClient(nt)
+	nt.Watcher = testwatcher.NewWatcher(nt.Context, nt.Logger, nt.KubeClient, nt.Config, nt.Scheme, nt.DefaultWaitTimeout)
```
karlkfi, line 393: nit: testwatcher.New should return a TestWatcher

karlkfi: I guess testwatcher.NewWatcher returning a Watcher is ok... It's not watching the test, i guess.

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -430,7 +432,7 @@ func DeploymentMissingEnvVar(containerName, key string) Predicate {
 // IsManagedBy checks that the object is managed by configsync, has the expected
 // resource manager, and has a valid resource-id.
 // Use diff.IsManager if you just need a boolean without errors.
-func IsManagedBy(nt *NT, scope declared.Scope, syncName string) Predicate {
+func IsManagedBy(scheme *runtime.Scheme, scope declared.Scope, syncName string) Predicate {
```
karlkfi, line 435: nit: in other package funcs that take the scheme, the scheme is the last arg, like kinds.Lookup. But I don't know if it really matters. It's kind of like a Context, which is always the first arg...

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -817,16 +819,15 @@ func MissingDeletionTimestamp(o client.Object) error {
 }
 
 // ObjectNotFoundPredicate returns an error unless the object is nil (not found).
-func ObjectNotFoundPredicate(nt *NT) Predicate {
-	ntPtr := nt
+func ObjectNotFoundPredicate(scheme *runtime.Scheme) Predicate {
```
karlkfi, line 822: These schemes all make me think this should be a PredicateFactory, but that seems potentially tedious...

e2e/nomostest/testwatcher/watch.go
```diff
@@ -52,37 +56,67 @@ func WatchTimeout(timeout time.Duration) WatchOption {
 	}
 }
 
+// Watcher is used for performing various Watch operations
```
karlkfi, line 59: generic desc. maybe add something about k8s objects

e2e/nomostest/testutils/validate.go
```diff
@@ -0,0 +1,42 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/util/log"
+)
+
+// ValidateError returns true if the specified errors contain an error
+// with the specified error code and (partial) message.
+func ValidateError(errs []v1beta1.ConfigSyncError, code, message string) error {
```
karlkfi, line 27: Are there other Validate functions that might want extraction or are they all tightly coupled to the NT?

e2e/nomostest/testwatcher/watch.go
```diff
@@ -52,37 +56,67 @@ func WatchTimeout(timeout time.Duration) WatchOption {
...
+	ctx                context.Context
+	defaultWaitTimeout time.Duration
+	logger             *testlogger.TestLogger
+	restConfig         *rest.Config
+	kubeClient         *testkubeclient.KubeClient
+	scheme             *runtime.Scheme
+}
+
+// NewWatcher constructs a new Watcher
+func NewWatcher(
```
karlkfi, line 70: When ALL of the fields need to be populated, I usually just skip the constructor and make them all public. That's what I did for the other test package structs. I'm not sure there's really any value in hiding them as private fields. Yes, it makes them not mutable by other packages, but it doesn't really hide any complexity. And it makes a constructor with too many args that you have to then document in the constructor, instead of just on the fields themselves.

karlkfi: Also, in the testing utility context, we don't really need to try to protect the user from shooting themselves in the foot as much as we might in a public facing interface that we need to maintain reverse compatibility on.

karlkfi:
Looks pretty good. Just a few nits & questions.

e2e/nomostest/testwatcher/watch.go
sdowell: In my view the constructor w/ private fields adds a few things:
- guarantee of immutability from other packages (as you mentioned)
- guarantee of required fields being provided. When looking at the constructor you know that all of the necessary fields have been set for a given instantiation. Conversely, a new person looking at the code can quickly understand what fields they need to provide to create a new instance. If it's just a struct with a bunch of fields, it requires more investigation for a new person to figure out how to instantiate the struct and what fields are required (which may be all or a subset of fields).

I acknowledge that we don't need to be as strict with test utilities compared to a public interface, however I think that mentality is what leads to the situation with `nomostest` where it's an amalgamation of organically grown interfaces. The looseness of the APIs makes it take longer to understand how everything works (i.e. more cyclomatic complexity), rather than easier to use - IMO.

e2e/nomostest/testutils/validate.go
sdowell: There might be but I don't think it needs to be in scope for this refactor

e2e/nomostest/testpredicates/predicates.go
sdowell: This sounds tedious to change but if it matters I can switch the argument ordering. I'm not familiar with that convention so let me know how important this is

e2e/nomostest/new.go
```diff
@@ -275,6 +276,12 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	// Speed up the delay between rendering attempts to speed up testing (default: 5s)
 	nt.HydrationPollingPeriod = 1 * time.Second
 
+	if nt.IsGKEAutopilot {
+		nt.DefaultWaitTimeout = 10 * time.Minute
```
nan-yu, line 280: nt.IsGKEAutopilot is set after `nt.detectGKEAutopilot`, it has to be after.

sdowell: Great catch, thank you! Fixed

e2e/nomostest/validate.go
tiffanny29631, line 1: Can this be merged under testutils pkg?

sdowell: No, because the majority of the methods in here are coupled with `NT`. We could potentially refactor this but I'd like to limit the scope of this PR

karlkfi: I don't love the name "testutils" as a package. Too generic. But you can change it in another PR. It's always hard to name when there's only one function in a package and you're not sure what else might go there, but the "utils" name is generally good to avoid because it becomes a kitchen sink.

karlkfi (approved):
/lgtm

Good enough for now. We can always iterate more on it in subsequent PRs. Thanks.

PR #497, opened 2023-03-28
Clean up kubevirt-related apiservices

haiyanmeng:


e2e/nomostest/clean.go
```diff
@@ -601,8 +601,14 @@ func disableRootSyncDeletionPropagation(nt *NT) error {
...
-func deleteTestAPIServices(nt *NT) error {
-	out, err := nt.Shell.Kubectl("delete", "apiservices", "-l", "testdata=true")
+func deleteAPIServices(nt *NT) error {
+	out, err := nt.Shell.Kubectl("delete", "apiservices", "-l", "testdata=true", "--ignore-not-found")
+	if err != nil {
+		nt.T.Logf("%v", out)
+		nt.T.Errorf("cleaning up apiservices: %v", err)
+	}
+
+	out, err = nt.Shell.Kubectl("delete", "apiservices", "v1.subresources.kubevirt.io", "v1alpha3.subresources.kubevirt.io", "--ignore-not-found")
```
karlkfi, line 611: Do these not have the testdata label?
Is there some way we could add it when they're applied?

haiyanmeng: I believe those are installed by the kubevirt-operator, and we don't have control over the logic.

e2e/nomostest/clean.go
```diff
@@ -601,8 +601,14 @@ func disableRootSyncDeletionPropagation(nt *NT) error {
 	return nil
 }
 
-func deleteTestAPIServices(nt *NT) error {
-	out, err := nt.Shell.Kubectl("delete", "apiservices", "-l", "testdata=true")
+func deleteAPIServices(nt *NT) error {
+	out, err := nt.Shell.Kubectl("delete", "apiservices", "-l", "testdata=true", "--ignore-not-found")
```
karlkfi, line 605: Would be nice if this was using the delete and watch functions in the reset code. We would get better errors and it already handles both NotFound and the test label filter.

e2e/nomostest/clean.go
karlkfi: Please add a comment explaining that. We probably also want to loop over a list of known APIServices that are created without the test label, and then wait for them to be notfound before continuing.

haiyanmeng: https://github.com/GoogleContainerTools/kpt-config-sync/pull/503 fixed the issue. Close this PR.

PR #498, opened 2023-03-28
feat: Add reconciler-manager finalizer

karlkfi:
- The reconciler-manager now adds a finalizer onto all RootSyncs &
  RepoSyncs ("configsync.gke.io/reconciler-manager") which blocks
  removal of these objects until garbage collection is complete.
  This ensures that garbage collection has completed when the object
  is reconciled (Not Found), making it safer to use RootSyncs &
  RepoSyncs as dependencies.
- Watch for object deletion before completing garbage collection.
  This ensures the objects are Not Found before allowing RSync
  removal.
- Add watch.DeleteAndWait & watch.UntilDeletedWithSync to help work
  around limitations of the controller-runtime client.Delete.

b/236892981

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/582
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/583
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/587
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/650

### DIFF
--- a/cmd/reconciler-manager/main.go
+++ b/cmd/reconciler-manager/main.go
@@ -33,6 +33,7 @@ import (
 	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -76,6 +77,18 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	// The Client built by ctrl.NewManager uses caching by default, and doesn't
+	// support the Watch method. So build another with shared config.
+	// This one can be used for watching and bypassing the cache, as needed.
+	// Use with discretion.
+	watcher, err := client.NewWithWatch(mgr.GetConfig(), client.Options{
+		Scheme: mgr.GetScheme(),
+		Mapper: mgr.GetRESTMapper(),
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to create watching client")
+		os.Exit(1)
+	}
 	dynamicClient, err := dynamic.NewForConfig(mgr.GetConfig())
 	if err != nil {
 		setupLog.Error(err, "failed to build dynamic client")
@@ -83,7 +96,8 @@ func main() {
 	}
 	watchFleetMembership := fleetMembershipCRDExists(dynamicClient, mgr.GetRESTMapper())
 
-	repoSync := controllers.NewRepoSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod, mgr.GetClient(), dynamicClient,
+	repoSync := controllers.NewRepoSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod,
+		mgr.GetClient(), watcher, dynamicClient,
 		ctrl.Log.WithName("controllers").WithName(configsync.RepoSyncKind),
 		mgr.GetScheme())
 	if err := repoSync.SetupWithManager(mgr, watchFleetMembership); err != nil {
@@ -91,7 +105,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	rootSync := controllers.NewRootSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod, mgr.GetClient(), dynamicClient,
+	rootSync := controllers.NewRootSyncReconciler(*clusterName, *reconcilerPollingPeriod, *hydrationPollingPeriod,
+		mgr.GetClient(), watcher, dynamicClient,
 		ctrl.Log.WithName("controllers").WithName(configsync.RootSyncKind),
 		mgr.GetScheme())
 	if err := rootSync.SetupWithManager(mgr, watchFleetMembership); err != nil {
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -18,22 +18,30 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
+	"kpt.dev/configsync/e2e"
 	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
 	"kpt.dev/configsync/pkg/testing/fake"
 	"kpt.dev/configsync/pkg/util"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -54,6 +62,149 @@ const (
 	expectedFirstContainerMemoryLimit1 = "100Mi"
 	expectedFirstContainerCPULimit1    = "180m"
 )
+
+// TestReconcilerManagerTeardown validates that when a RootSync or RepoSync is
+// deleted, the reconciler-manager finalizer handles deletion of the reconciler
+// and its dependencies managed by the reconciler-manager.
+func TestReconcilerManagerTeardown(t *testing.T) {
+	testNamespace := "teardown"
+	nt := nomostest.New(t, nomostesting.ACMController,
+		ntopts.WithDelegatedControl, ntopts.Unstructured,
+		ntopts.NamespaceRepo(testNamespace, configsync.RepoSyncName))
+
+	t.Log("Validate the reconciler-manager deployment")
+	reconcilerManager := &appsv1.Deployment{}
+	setNN(reconcilerManager, client.ObjectKey{Name: reconcilermanager.ManagerName, Namespace: v1.NSConfigManagementSystem})
+	err := nt.Validate(reconcilerManager.Name, reconcilerManager.Namespace, reconcilerManager)
+	require.NoError(t, err)
+
+	t.Log("Validate the RootSync")
+	rootSync := &v1beta1.RootSync{}
+	setNN(rootSync, client.ObjectKey{Name: configsync.RootSyncName, Namespace: v1.NSConfigManagementSystem})
+	err = nt.Watcher.WatchObject(kinds.RootSyncV1Beta1(), rootSync.Name, rootSync.Namespace, []testpredicates.Predicate{
+		testpredicates.StatusEquals(nt.Scheme, kstatus.CurrentStatus),
+		testpredicates.HasFinalizer(metadata.ReconcilerManagerFinalizer),
+	})
+	require.NoError(t, err)
+
+	t.Log("Validate the RootSync reconciler and its dependencies")
+	var rootSyncFriends []client.Object
+
+	rootSyncReconciler := &appsv1.Deployment{}
+	setNN(rootSyncReconciler, core.RootReconcilerObjectKey(rootSync.Name))
+	rootSyncFriends = append(rootSyncFriends, rootSyncReconciler)
+
+	// Note: reconciler-manager no longer applies ConfigMaps to configure the
+	// reconciler. So we don't need to validate their deletion. The deletion
+	// only happens when upgrading from a very old unsupported version.
+
+	rootSyncCRB := &rbacv1.ClusterRoleBinding{}
+	setNN(rootSyncCRB, client.ObjectKey{Name: controllers.RootSyncPermissionsName()})
+	rootSyncFriends = append(rootSyncFriends, rootSyncCRB)
+
+	rootSyncSA := &corev1.ServiceAccount{}
+	setNN(rootSyncSA, client.ObjectKeyFromObject(rootSyncReconciler))
+	rootSyncFriends = append(rootSyncFriends, rootSyncSA)
+
+	for _, obj := range rootSyncFriends {
+		err := nt.Validate(obj.GetName(), obj.GetNamespace(), obj)
+		require.NoError(t, err)
+	}
+
+	t.Log("Validate the RepoSync")
+	repoSync := &v1beta1.RepoSync{}
+	setNN(repoSync, client.ObjectKey{Name: configsync.RepoSyncName, Namespace: testNamespace})
+	err = nt.Watcher.WatchObject(kinds.RepoSyncV1Beta1(), repoSync.Name, repoSync.Namespace, []testpredicates.Predicate{
+		testpredicates.StatusEquals(nt.Scheme, kstatus.CurrentStatus),
+		testpredicates.HasFinalizer(metadata.ReconcilerManagerFinalizer),
+	})
+	require.NoError(t, err)
+
+	t.Log("Validate the RepoSync reconciler and its dependencies")
+	var repoSyncFriends []client.Object
+
+	repoSyncReconciler := &appsv1.Deployment{}
+	setNN(repoSyncReconciler, core.NsReconcilerObjectKey(repoSync.Namespace, repoSync.Name))
+	repoSyncFriends = append(repoSyncFriends, repoSyncReconciler)
+
+	// Note: reconciler-manager no longer applies ConfigMaps to configure the
+	// reconciler. So we don't need to validate their deletion. The deletion
+	// only happens when upgrading from a very old unsupported version.
+
+	repoSyncRB := &rbacv1.RoleBinding{}
+	setNN(repoSyncRB, client.ObjectKey{
+		Name:      controllers.RepoSyncPermissionsName(),
+		Namespace: testNamespace,
+	})
+	repoSyncFriends = append(repoSyncFriends, repoSyncRB)
+
+	repoSyncSA := &corev1.ServiceAccount{}
+	setNN(repoSyncSA, client.ObjectKeyFromObject(repoSyncReconciler))
+	repoSyncFriends = append(repoSyncFriends, repoSyncSA)
+
+	// See nomostest.CreateNamespaceSecret for creation of user secrets.
+	// This is a managed secret with a derivative name.
+	repoSyncAuthSecret := &corev1.Secret{}
+	setNN(repoSyncAuthSecret, client.ObjectKey{
+		Name:      controllers.ReconcilerResourceName(repoSyncReconciler.Name, nomostest.NamespaceAuthSecretName),
+		Namespace: repoSyncReconciler.Namespace,
+	})
+	repoSyncFriends = append(repoSyncFriends, repoSyncAuthSecret)
+
+	// See nomostest.CreateNamespaceSecret for creation of user secrets.
+	// This is a managed secret with a derivative name.
+	// For local kind clusters, the CA Certs are provided to authenticate the git server.
+	if nt.GitProvider.Type() == e2e.Local {
+		repoSyncCACertSecret := &corev1.Secret{}
+		setNN(repoSyncCACertSecret, client.ObjectKey{
+			Name:      controllers.ReconcilerResourceName(repoSyncReconciler.Name, nomostest.NamespaceAuthSecretName),
+			Namespace: repoSyncReconciler.Namespace,
+		})
+		repoSyncFriends = append(repoSyncFriends, repoSyncCACertSecret)
+	}
+
+	for _, obj := range repoSyncFriends {
+		err := nt.Validate(obj.GetName(), obj.GetNamespace(), obj)
+		require.NoError(t, err)
+	}
+
+	t.Log("Delete the RootSync and wait for it to be not found")
+	err = nomostest.DeleteObjectsAndWait(nt, rootSync)
+	require.NoError(t, err)
+
+	t.Log("Validate the RootSync reconciler and its dependencies were deleted")
+	for _, obj := range rootSyncFriends {
+		gvk, err := kinds.Lookup(obj, nt.Scheme)
+		require.NoError(t, err)
+		rObj, err := kinds.NewObjectForGVK(gvk, nt.Scheme)
+		require.NoError(t, err)
+		cObj, err := kinds.ObjectAsClientObject(rObj)
+		require.NoError(t, err)
+		err = nt.ValidateNotFound(obj.GetName(), obj.GetNamespace(), cObj)
+		require.NoError(t, err)
+	}
+
+	t.Log("Delete the RepoSync and wait for it to be not found")
+	err = nomostest.DeleteObjectsAndWait(nt, repoSync)
+	require.NoError(t, err)
+
+	t.Log("Validate the RepoSync reconciler and its dependencies were deleted")
+	for _, obj := range repoSyncFriends {
+		gvk, err := kinds.Lookup(obj, nt.Scheme)
+		require.NoError(t, err)
+		rObj, err := kinds.NewObjectForGVK(gvk, nt.Scheme)
+		require.NoError(t, err)
+		cObj, err := kinds.ObjectAsClientObject(rObj)
+		require.NoError(t, err)
+		err = nt.ValidateNotFound(obj.GetName(), obj.GetNamespace(), cObj)
+		require.NoError(t, err)
+	}
+}
+
+func setNN(obj client.Object, nn types.NamespacedName) {
+	obj.SetName(nn.Name)
+	obj.SetNamespace(nn.Namespace)
+}
 
 func TestManagingReconciler(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.ACMController)
--- a/pkg/reconcilermanager/controllers/garbage_collector.go
+++ b/pkg/reconcilermanager/controllers/garbage_collector.go
@@ -17,19 +17,34 @@ package controllers
 import (
 	"context"
 	"strings"
+	"time"
 
+	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/reconcilermanager"
+	"kpt.dev/configsync/pkg/util/watch"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// deleteWaitTimeout is the time to block watching for an object to be deleted.
+//
+// Shorter means the reconciler-manager will be more responsive.
+// Longer means the reconciler-manager will be more efficient.
+//
+// Since reconcile timeout doesn't cause a Stalled condition, it's ok to timeout
+// and wait for Reconcile to be triggered by another event.
+const deleteWaitTimeout = 30 * time.Second
+
+// cleanup deletes the object and waits for it to be fully deleted.
 func (r *reconcilerBase) cleanup(ctx context.Context, obj client.Object) error {
 	gvk, err := kinds.Lookup(obj, r.scheme)
 	if err != nil {
@@ -41,20 +56,31 @@ func (r *reconcilerBase) cleanup(ctx context.Context, obj client.Object) error {
 	}
 	// Convert Object to Unstructured to avoid checking UID, ResourceVersion, or
 	// modifying input object.
-	u := &unstructured.Unstructured{}
-	u.SetName(id.Name)
-	u.SetNamespace(id.Namespace)
-	u.SetGroupVersionKind(gvk)
+	uObj := &unstructured.Unstructured{}
+	uObj.SetName(id.Name)
+	uObj.SetNamespace(id.Namespace)
+	uObj.SetGroupVersionKind(gvk)
 	r.logger(ctx).Info("Deleting managed object",
 		logFieldObjectRef, id.ObjectKey.String(),
 		logFieldObjectKind, id.Kind)
-	if err := r.client.Delete(ctx, u); err != nil {
-		if apierrors.IsNotFound(err) {
-			r.logger(ctx).Info("Managed object already deleted",
-				logFieldObjectRef, id.ObjectKey.String(),
-				logFieldObjectKind, id.Kind)
-			return nil
+	err = watch.DeleteAndWait(ctx, r.watcher, uObj, deleteWaitTimeout)
+	if err != nil {
+		// ErrWaitTimeout is returned when the context is Done, but doesn't
+		// specify cancelled or deadline exceeded.
+		if errors.Is(err, wait.ErrWaitTimeout) {
+			// If the DeleteAndWait observed any events, the object should have been updated.
+			if uObj.GetDeletionTimestamp().IsZero() {
+				// object not deleted
+				return NewObjectOperationErrorWithID(err, id, OperationDelete)
+			}
+			result, computeErr := kstatus.Compute(uObj)
+			if computeErr != nil {
+				return errors.Wrapf(computeErr, "computing %s (%s) status failed", id.Kind, id.ObjectKey)
+			}
+			// Since DeletionTimestamp is set, this will be either Terminating or Failed
+			return NewObjectReconcileErrorWithID(err, id, result.Status)
 		}
+		// If not a timeout, assume the delete failed.
 		return NewObjectOperationErrorWithID(err, id, OperationDelete)
 	}
 	r.logger(ctx).Info("Managed object delete successful",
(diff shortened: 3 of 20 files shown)


e2e/nomostest/testpredicates/predicates.go
```diff
@@ -441,7 +440,7 @@ func IsManagedBy(scheme *runtime.Scheme, scope declared.Scope, syncName string)
 		gvk := obj.GetObjectKind().GroupVersionKind()
 		if gvk.Empty() {
 			var err error
-			gvk, err = apiutil.GVKForObject(obj, scheme)
+			gvk, err = kinds.Lookup(obj, scheme)
```
karlkfi, line 443: These Lookups can be pulled out to a new PR. They're nice to have, but not required for this PR any more.

e2e/testcases/kcc_test.go
```diff
@@ -97,35 +94,13 @@ func TestKCCResourcesOnCSR(t *testing.T) {
...
-	}
-	u := o.(*unstructured.Unstructured)
-	conditions, found, err := unstructured.NestedSlice(u.Object, "status", "conditions")
-	if err != nil || !found || len(conditions) == 0 {
-		return fmt.Errorf(".status.conditions not found %v", err)
-	}
-	condition := (conditions[0]).(map[string]interface{})
-	if condition["type"] != "Ready" || condition["status"] != "True" {
-		return fmt.Errorf("resource is not ready %v", condition)
+	if err := nt.Watcher.WatchForCurrentStatus(gvk, name, namespace); err != nil {
```
karlkfi, line 97: This refactor to Watch can be moved to another PR

pkg/api/configsync/v1beta1/constants.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"kpt.dev/configsync/pkg/api/configmanagement"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+)
+
+// SyncFinalizer is a finalizer handled by ReconcilerManager to ensure
+// deletion of managed resource objects complete before the RootSync/RepoSync is
+// deleted
+const SyncFinalizer = reconcilermanager.ManagerName + "." + configmanagement.GroupName
```
karlkfi, line 25: Change annotation to be consistent with the reconciler finalizer domain

pkg/kinds/lookup.go
```diff
@@ -12,19 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//
```
karlkfi, line 15: This fix can go in the Lookup PR

pkg/reconciler/finalizer/reposync_finalizer_test.go
```diff
@@ -239,9 +237,20 @@ func TestRepoSyncFinalize(t *testing.T) {
 		{
 			name:  "rsync not found",
 			rsync: repoSync1.DeepCopy(),
-			setup: func() error {
+			setup: func(fakeClient *fake.Client) error {
```
karlkfi, line 240: The fake.Client upgrades to support Watch can be extracted to another PR

pkg/syncer/syncertest/fake/client.go
```diff
@@ -59,14 +61,25 @@ var _ client.StatusWriter = &statusWriter{}
...
+	test    *testing.T
 	scheme  *runtime.Scheme
 	codecs  serializer.CodecFactory
 	mapper  meta.RESTMapper
 	Objects map[core.ID]client.Object
+	eventCh chan watch.Event
+	Now     func() metav1.Time
+
+	lock     sync.RWMutex
+	watchers map[chan watch.Event]struct{}
```
karlkfi, line 73: The fake.Client upgrades to support Watch can be extracted to another PR

pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -17,78 +17,41 @@ package controllers
...
+// This should be long enough for foreground deletion of any one managed object,
+// including garbage collection. The longest is usually the Deployment.
+// However, even if the Deployment takes longer to delete, it's ok to timeout.
+// Timeout will just cause an error, which causes reconcile to be retried by the
+// controller-manager. Waiting is just an optimization to reduce API calls and
+// error log messages, which could be red herrings.
+const deleteWaitTimeout = 1 * time.Minute
+
+// cleanup deletes the object and waits for it to be fully deleted.
+func (r *reconcilerBase) cleanup(ctx context.Context, obj client.Object) error {
```
karlkfi, line 43: move cleanup down to where it was before to make the diff cleaner

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -95,34 +94,45 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
 	rsRef := req.NamespacedName
-	log := r.log.WithValues("reposync", rsRef.String())
 	start := time.Now()
 	var err error
 	rs := &v1beta1.RepoSync{}
 	reconcilerRef := types.NamespacedName{
 		Namespace: v1.NSConfigManagementSystem,
 		Name:      core.NsReconcilerName(rsRef.Namespace, rsRef.Name),
 	}
+	log := r.log.WithValues(
```
karlkfi, line 104: these log changes can be extracted to another PR

pkg/api/configsync/v1beta1/constants.go
karlkfi: And rename to include ReconcilerManager.
And define in the same place with the ReconcilerFinalizer

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -171,10 +176,11 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		return controllerruntime.Result{}, updateErr
 	}
 
+	// Create a tombstone, so we know to clean up managed objects during termination.
```
karlkfi, line 179: This tombstone method won't survive rescheduling of the reconciler-manager pod. We might want to skip it and rely on the more durable method of just calling deleteManagedObjects again. 

Do we might need to verify that the RSync is managed by this reconciler-manager? The RSyncs don't have owner references and now neither do the other objects. Perhaps there's an annotation to identify the reconciler-manager?

Can multiple Config Syncs even co-exist on one cluster? Or do they end up fighting over every RSync? If not, we can remove the cache all together.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -184,6 +190,90 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+
+	// Cross-namespace owner references are disallowed by Kubernetes.
+	// So we can't rely on Kubernetes' built-in garbage collection to delete
+	// managed objects in the central "config-management-system" namespace, when
+	// the RepoSync is in another namespace.
+	// To avoid leaking managed resources, we use a custom finalizer to force
+	// foreground deletion. This way, the RepoSync will not be fully deleted
+	// until all its managed resources are also deleted.
+
+	if rs.ObjectMeta.DeletionTimestamp.IsZero() {
```
karlkfi, line 203: This if/else is a lot of code. Should we pull it out to a method?

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -184,6 +190,90 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+	if err := r.upsertManagedObjects(ctx, reconcilerRef, currentRS, rs); err != nil {
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+		return controllerruntime.Result{}, err
+	}
+
+	metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(nil), start)
+	return controllerruntime.Result{}, nil
+}
+
+func (r *RepoSyncReconciler) upsertManagedObjects(ctx context.Context, reconcilerRef types.NamespacedName, currentRS, rs *v1beta1.RepoSync) error {
```
karlkfi, line 259: Extracting upsertManagedObjects could be in a seperate PR, perhaps with the log changes.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -440,13 +645,13 @@ func (r *RepoSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
 		Watches(&source.Kind{Type: &corev1.Secret{}},
 			handler.EnqueueRequestsFromMapFunc(r.mapSecretToRepoSyncs),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
-		Watches(&source.Kind{Type: &appsv1.Deployment{}},
+		Watches(&source.Kind{Type: withNamespace(&appsv1.Deployment{}, configsync.ControllerNamespace)},
```
karlkfi, line 648: Down-scoping these watches to a namespace could be its own PR

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -102,33 +103,54 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
 
 	if err = r.client.Get(ctx, rsRef, rs); err != nil {
 		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
 		if apierrors.IsNotFound(err) {
 			r.clearLastReconciled(rsRef)
-			r.log.Info("Deleting managed objects",
-				logFieldObject, rsRef.String(),
-				logFieldKind, r.syncKind)
-			return controllerruntime.Result{}, r.deleteClusterRoleBinding(ctx, reconcilerRef)
+			// TODO: Handle cleanup of RootSync deleted while reconciler-manager is down.
```
karlkfi, line 122: This TODO can be handled just by removing the sync cache. I don't think we need it. Just have the reconciler-manager run deleteManagedObjects when it's triggered. It's ok if it happens multiple times, as long as it's not happening continuously.

pkg/syncer/client/client_test.go
```diff
@@ -30,6 +31,11 @@ import (
 )
 
 func TestClient_Create(t *testing.T) {
+	scheme := runtime.NewScheme()
```
karlkfi, line 34: Remove. Already using core.Scheme.

pkg/testing/fake/cluster_config.go
```diff
@@ -47,3 +47,10 @@ func ClusterConfigObject(opts ...ClusterConfigMutator) *v1.ClusterConfig {
 
 	return result
 }
+
+// ClusterConfigResourceVersion sets the ClusterConfig's resource version to the given value.
+func ClusterConfigResourceVersion(version string) ClusterConfigMutator {
```
karlkfi, line 52: Don't see this used. Is it needed?

pkg/api/configsync/v1beta1/constants.go
karlkfi: configsync.gke.io/reconciler-manager

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -102,33 +103,54 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+					logFieldObject, rsRef.String(),
+					logFieldKind, r.syncKind)
+				// return `controllerruntime.Result{}, nil` here to make sure the request will not be requeued.
+				return controllerruntime.Result{}, nil
+			}
+			// Cleanup after already deleted RootSync.
+			// This code path is unlikely, because the custom finalizer should
+			// have already deleted the managed resources and removed the
+			// rootSyncs cache entry. But if we get here, clean up anyway.
+			if err := r.deleteManagedObjects(ctx, reconcilerRef, nil, rs, false); err != nil {
```
karlkfi, line 134: Need to add a check to see if the `configsync.gke.io/reconciler` is present. If so, log that the reconciler-manager finalizer is blocked waiting on the reconciler finalizer. 

Otherwise, the reconciler will be deleted before the user objects have been deleted.

pkg/util/watch/until.go
```diff
@@ -0,0 +1,385 @@
...
+//
+// If the object is not yet deleted, the delete call will use foreground
+// deletion, waiting until garbage collection completes too. However, if the
+// object was already deleted
+//
+// If the object is Unstructured, the GVK must be set.
+//
+// The ResourceVersion of the input object is ignored, because a new Informer is
+// created which will do a ListAndWatch, and List needs the collection
+// ResourceVersion instead of the object ResourceVersion.
```
karlkfi, line 47: This is technically incorrect. There is no such thing as an object RV. There is only the RV of the whole resource. We could in fact just do a Watch with "time-travel", if we already have the RV of the object we want to see deleted.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -545,3 +547,71 @@ func (r *reconcilerBase) addTypeInformationToObject(obj runtime.Object) error {
...
+// deletion, which allows for stricter contracts when uninstalling.
+func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Object, setupFn, teardownFn func(context.Context) error) error {
+	if syncObj.GetDeletionTimestamp().IsZero() {
+		// The object is NOT being deleted.
+		if !controllerutil.ContainsFinalizer(syncObj, metadata.ReconcilerManagerFinalizer) {
+			// The object is new and doesn't have our finalizer yet.
+			// Add our finalizer and update the object.
+			controllerutil.AddFinalizer(syncObj, metadata.ReconcilerManagerFinalizer)
+			if err := r.client.Update(ctx, syncObj); err != nil {
+				err = status.APIServerError(err, "failed to update RepoSync to add finalizer")
```
haiyanmeng, line 574: The object may be RootSync.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -545,3 +547,71 @@ func (r *reconcilerBase) addTypeInformationToObject(obj runtime.Object) error {
...
+				r.logger(ctx).Error(err, "Finalizer injection failed")
+				return err
+			}
+			r.logger(ctx).Info("Finalizer injection successful")
+		}
+
+		// Our finalizer is present, so setup managed resource objects.
+		if err := setupFn(ctx); err != nil {
+			// Failed to delete a managed object.
```
haiyanmeng, line 583: The comment seems inaccurate.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -170,7 +171,12 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		return controllerruntime.Result{}, updateErr
 	}
 
-	if err := r.setup(ctx, reconcilerRef, rsRef, currentRS, rs); err != nil {
+	err := r.setupOrTeardown(ctx, rs, func(ctx context.Context) error {
+		return r.setup(ctx, reconcilerRef, rsRef, currentRS, rs)
+	}, func(ctx context.Context) error {
+		return r.teardown(ctx, reconcilerRef, rsRef, currentRS, rs)
+	})
```
haiyanmeng, line 178: Can you define two variables for the last two parameters of `setupOrTeardown`? 

```
var setupFn = ...
var teardownFn = ...
err := r.setupOrTeardown(ctx, rs, setupFn, teardownFn)
```

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -305,6 +311,34 @@ func (r *RepoSyncReconciler) setup(ctx context.Context, reconcilerRef, rsRef typ
 	return nil
 }
 
+// teardown performs the following teardown steps:
+// - Set the Reconciling condition with the Teardown reason (in-progress)
+// - Call deleteManagedObjects and update the Stalled condition if it errors
```
haiyanmeng, line 316: Swap these two lines to match the implementation.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -305,6 +311,34 @@ func (r *RootSyncReconciler) setup(ctx context.Context, reconcilerRef, rsRef typ
 	return nil
 }
 
+// teardown performs the following teardown steps:
+// - Set the Reconciling condition with the Teardown reason (in-progress)
+// - Call deleteManagedObjects and update the Stalled condition if it errors
+// - Update the Reconciling condition with the Teardown reason (complete)
```
haiyanmeng, line 317: Swap these two lines to match the implementation.

pkg/util/watch/delete.go
```diff
@@ -0,0 +1,185 @@
+// Copyright 2022 Google LLC
```
haiyanmeng, line 1: 2022 -> 2023

pkg/util/watch/delete.go
```diff
@@ -0,0 +1,185 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// These log fields
```
haiyanmeng, line 31: Make the comment more helpful.

pkg/util/watch/delete.go
```diff
@@ -0,0 +1,185 @@
...
+// DeleteAndWait deletes the object (if not already deleted) and blocks until it
+// is fully deleted (NotFound, not just deleteTimestamp). This allows for any
+// finalizers to complete.
+//
+// Exits early if the object is already deleted or doesn't exist.
+//
+// The Object contents will be replaced with updated state for each new event
+// received while waiting for deletion.
+//
+// If you want to time out the full function, use a context.WithTimeout.
```
haiyanmeng, line 46: It seems reconcileTimeout still takes effect even if a context.WithTimeout is used. Right?

pkg/util/watch/until.go
```diff
@@ -0,0 +1,403 @@
+// Copyright 2022 Google LLC
```
haiyanmeng, line 1: 2022 -> 2023

pkg/util/watch/delete.go
karlkfi: lol. i wrote this code in 2022

pkg/util/watch/delete.go
karlkfi: Correct.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -545,3 +547,71 @@ func (r *reconcilerBase) addTypeInformationToObject(obj runtime.Object) error {
...
+		if err := teardownFn(ctx); err != nil {
+			// Failed to delete a managed object.
+			// Return an error to trigger retry.
+			return errors.Wrap(err, "teardown failed")
+		}
+
+		// Remove our finalizer and update the object.
+		controllerutil.RemoveFinalizer(syncObj, metadata.ReconcilerManagerFinalizer)
+		if err := r.client.Update(ctx, syncObj); err != nil {
+			err = status.APIServerError(err, "failed to update RepoSync to remove finalizer")
```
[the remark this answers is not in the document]
karlkfi: fixed.

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: done

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: fixed.

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: fixed.

pkg/util/watch/delete.go
karlkfi: fixed.

pkg/util/watch/delete.go
karlkfi: fixed.

pkg/util/watch/until.go
karlkfi: fixed.

pkg/reconcilermanager/controllers/reconciler_base.go
karlkfi: removed

pkg/reconcilermanager/controllers/reconciler_base.go
karlkfi: fixed.

cmd/reconciler-manager/main.go
```diff
@@ -76,22 +77,36 @@ func main() {
...
+	// support the Watch method. So build another with shared config.
+	// This one can be used for watching and bypassing the cache, as needed.
+	// Use with discretion.
+	watcher, err := client.NewWithWatch(mgr.GetConfig(), client.Options{
+		Scheme: mgr.GetScheme(),
+		Mapper: mgr.GetRESTMapper(),
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to create watching client")
+		os.Exit(1)
```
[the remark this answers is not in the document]
karlkfi: feel free to follow up with a refactor. I just followed the existing convention for this file. But I agree with you in theory.

pkg/util/watch/until.go
```diff
@@ -0,0 +1,403 @@
...
+		}
+	}
+	return &opts
+}
+
+// MergeListOptions merges two sets of ListOptions.
+// - For Namespace, Limit, and Continue, at most one may be specified, unless they're the same
+// - For FieldSelector and LabelSelector, the result requires both filters to match (AND)
+// - For Raw, at most one may be specified
+func MergeListOptions(a, b *client.ListOptions) (*client.ListOptions, error) {
```
sdowell, line 235: These util functions could use some unit tests

pkg/util/watch/until.go
```diff
@@ -0,0 +1,403 @@
...
+	case b.FieldSelector != nil:
+		c.FieldSelector = b.FieldSelector
+	}
+
+	switch {
+	case a.LabelSelector != nil && b.LabelSelector != nil:
+		bReqs, selectable := b.LabelSelector.Requirements()
+		if selectable {
+			c.LabelSelector = a.LabelSelector.Add(bReqs...)
+		} else {
```
sdowell, line 274: Is it desirable to silently drop `b.LabelSelector` in this case? Seems like that could hide issues, and we should return an error here - although I don't have full understanding of what `selectable` means

pkg/util/watch/until.go
```diff
@@ -0,0 +1,403 @@
...
+		c.Continue = a.Continue
+	case a.Continue != "":
+		c.Continue = a.Continue
+	case b.Continue != "":
+		c.Continue = b.Continue
+	}
+
+	switch {
+	case a.Raw != nil && b.Raw != nil:
+		// TODO: It should be possible to merge raw ListOptions, but we don't need it yet
```
sdowell, line 311: Curious why we don't need it - do we just not define any raw `ListOptions` in code currently?

pkg/util/watch/until.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	if mOpts.Continue != "" {
+		cOpts.Continue = mOpts.Continue
+	}
+
+	// Pass raw metav1.ListOptions, in case some options are specified that
+	// aren't fields of client.ListOptions.
+	// The fields on client.ListOptions will take precedence over the raw
+	// metav1.ListOptions fields.
+	cOpts.Raw = mOpts
```
sdowell, line 357: Seems inconsistent that we support raw `ListOptions` here but not in the above function

sdowell (commented):
Generally looks good, although I'd like to see e2e tests for the reconciler-manager finalizer. Makes it difficult to be confident the overall state of the feature is good if it's not being submitted with tests

pkg/util/watch/until.go
karlkfi: The only code using this conversion method today is the flexWatcher, implementing ListerWatcher for UntilDeletedWithSync. 

UntilDeletedWithSync calls client-go's UntilWithSync which builds an Informer which wraps a Reflector. So the Reflector is the only thing calling the List & Watch methods on the flexWatcher. It specifies ResourceVersion, TimeoutSeconds, and AllowWatchBookmarks for Watch (which adds Watch=true) and ResourceVersion, ResourceVersionMatch, Limit, and Continue for List.

The metav1.ListOptions then passes through ConvertListOptions, MergeListOptions, and UnrollListOptions in the flexWatcher List & Watch methods. 
- ConvertListOptions converts the metav1.ListOptions from the Reflector into client.ListOptions (including some Raw options that don't have client.ListOptions fields)
- MergeListOptions merges the options from the Reflector with the options from flexWatcher.listOptions()
- UnrollListOptions converts the client.ListOptions into an array of client.ListOption

So there's no need to merge the Raw options unless we start using these conversion functions elsewhere, which we might in the future, since they're kinda handy when you have a client.Client and want to make an Informer to watch things.

pkg/util/watch/until.go
karlkfi: We support ONE of the ListOptions containing Raw fields, just not both.

pkg/util/watch/until.go
sdowell: Ack

pkg/util/watch/until.go
karlkfi: `labels.internalSelector.Requirements()` basically returns self, because `internalSelector` is an alias for `[]Requirement`. And for some unknown reason it always returns true, even if the list is empty/nil. So this is probably always be selectable unless someone makes a different `labels.Requirements` impl.

Anyway, when it's selectable, the next line adds the "b" requirements to the "a" selector (which is itself a list of requirements).

pkg/util/watch/until.go
karlkfi: yes mother

sdowell: image

e2e/testcases/reconciler_manager_test.go
```diff
@@ -55,6 +63,149 @@ const (
...
+		err := nt.Validate(obj.GetName(), obj.GetNamespace(), obj)
+		require.NoError(t, err)
+	}
+
+	t.Log("Delete the RootSync and wait for it to be not found")
+	err = nomostest.DeleteObjectsAndWait(nt, rootSync)
+	require.NoError(t, err)
+
+	t.Log("Validate the RootSync reconciler and its dependencies were deleted")
+	for _, obj := range rootSyncFriends {
```
sdowell, line 176: love the naming of `r*SyncFriends`

karlkfi: rootSyncManagedObjects seemed confusing... they're managed by the reconciler-manager, not the reconciler... and rootSyncReconcilerManagerManagedObjects seemed too long...

pkg/util/watch/until.go
karlkfi: Looks like `nothingSelector.Requirements()` actually returns false. So maybe it doesn't mean what I thought it meant. If it's not selectable the merge would be the the nothingSelector.

pkg/util/watch/until.go
karlkfi: my room is clean now, mother.

pkg/util/watch/until.go
karlkfi: Fixed.

karlkfi:
e2e test added (& changed how Reconciling condition is set to be True when Stalled)

unit tests added for MergeListOptions, ConvertListOptions, UnrollListOptions (& fixed a few bugs)

e2e/testcases/reconciler_manager_test.go
```diff
@@ -55,6 +63,149 @@ const (
...
+		gvk, err := kinds.Lookup(obj, nt.Scheme)
```
sdowell, line 177: WDYT about putting the body of this in a taskgroup so that we can validate the friends in parallel?

karlkfi: It's just a bunch of Get calls which the apiserver caches and the QPS is pretty high. I don't think doing 3-5 in parallel would really be all that much of time savings unless they're timing out.

sdowell: ack, sgtm

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -324,6 +360,7 @@ func (r *RepoSyncReconciler) handleReconcileError(ctx context.Context, err error
 			logFieldObjectRef, opErr.ID.ObjectKey.String(),
 			logFieldObjectKind, opErr.ID.Kind,
 			logFieldOperation, opErr.Operation)
+		reposync.SetReconciling(rs, stage, fmt.Sprintf("%s stalled", stage))
```
sdowell, line 363: Curious why are we setting the reconciling condition in these areas now?

pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
```diff
@@ -93,29 +95,22 @@ func TestRepoSyncReconcilerDeploymentLifecycle(t *testing.T) {
...
-		if event.Type == watch.Deleted {
-			reconcilerObj = event.Object.(*appsv1.Deployment)
-			// success! deployment was deleted.
-			return nil
-		}
-		// keep watching
-		return errors.Errorf("reconciler deployment %s", event.Type)
-	})
-	require.NoError(t, err)
+	// All managed objects should have been deleted by the reconciler-manager finalizer.
```
sdowell, line 109: Should there be an assertion that the managed objects are deleted?

pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
```diff
@@ -93,29 +95,22 @@ func TestRepoSyncReconcilerDeploymentLifecycle(t *testing.T) {
...
+	// Only the user Secret should remain.
+	secretObj.SetUID("1")
+	t.Log("verifying all managed objects were deleted")
```
sdowell, line 112: Log message seems misleading

pkg/reconcilermanager/controllers/rootsync_controller_manager_test.go
```diff
@@ -105,29 +107,22 @@ func TestRootSyncReconcilerDeploymentLifecycle(t *testing.T) {
...
-			return nil
-		}
-		// keep watching
-		return errors.Errorf("reconciler deployment %s", event.Type)
-	})
-	require.NoError(t, err)
+	// All managed objects should have been deleted by the reconciler-manager finalizer.
+	// Only the user Secret should remain.
+	secretObj.SetUID("1")
+	t.Log("verifying all managed objects were deleted")
```
sdowell, line 124: ditto

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: It seemed like a good idea to convey that it was Reconciling that was Stalled and that Reconciling was still not complete because it was Stalled. Otherwise, it's not super clear what Stalled means.

AFAICT, everything that knows to look for Stalled and Reconciling (kstatus, nomos status) looks for Stalled first, so it shouldn't cause any problems. I should probably verify with the UI team tho that it won't break anything there.

pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
karlkfi: That's what Check does.

pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
karlkfi: nah, it's correct. It's checking that the supplied set exactly matches all objects in storage.

pkg/util/watch/delete.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// Use cancellable context to delay starting the timer until after the
+	// watcher has synchronized and the delete call has completed.
+	watchCtx, cancel := context.WithCancel(ctx)
+	// Stop early, if done or errored before timeout.
+	defer cancel()
+	// Delay timer start until after Sync and Delete.
+	startTimer := func() {
+		timerCtx, cancel2 := context.WithTimeout(watchCtx, reconcileTimeout)
```
sdowell, line 66: Maybe something like `timeoutCancel` instead of `cancel2`?

pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
sdowell: Ack, my bad for not looking closer at the implementation of `fakeClient.Check`

pkg/util/watch/listerwatcher_test.go
```diff
@@ -0,0 +1,895 @@
...
+			},
+			b: &client.ListOptions{
+				LabelSelector: labels.NewSelector().Add(
+					newLabelRequirement(t, "key", selection.Equals, []string{"value"}),
+				),
+			},
+			expectedOpts: &client.ListOptions{
+				LabelSelector: labels.NewSelector().Add(
+					newLabelRequirement(t, "key", selection.Equals, []string{"value"}),
+					// TODO: add LabelSelector de-duping (minor optimization)
```
sdowell, line 173: Worth creating a ticket for?

pkg/util/watch/listerwatcher_test.go
```diff
@@ -0,0 +1,895 @@
...
+				Raw: &metav1.ListOptions{
+					ResourceVersion: "abc123",
+				},
+			},
+			expectedError: testutil.EqualError(
+				errors.New("not yet implemented: merging two different raw ListOptions: " +
+					"&ListOptions{LabelSelector:,FieldSelector:,Watch:false,ResourceVersion:,TimeoutSeconds:nil,Limit:0,Continue:,AllowWatchBookmarks:false,ResourceVersionMatch:,} & " +
+					"&ListOptions{LabelSelector:,FieldSelector:,Watch:false,ResourceVersion:abc123,TimeoutSeconds:nil,Limit:0,Continue:,AllowWatchBookmarks:false,ResourceVersionMatch:,}")),
+		},
+		// AllTheThings!
```
sdowell, line 543: This pleases mother

pkg/util/watch/listerwatcher_test.go
karlkfi: meh. it doesn't effect our usage. we only need it if we expand usage of these utility functions. like the fake clients, they're only as good as they need to be, because making them perfect is expensive and unnecessary. I mostly just put it in as a warning to anyone who runs into needing it later that it was intentionally left out for expedience.

pkg/util/watch/delete.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+			// Stop waiting
+			return false, err
+		}
+		klog.V(3).Infof("Object delete successful %q=%q, %q=%q",
+			logFieldObjectRef, id.ObjectKey.String(),
+			logFieldObjectKind, id.Kind)
+
+		// Continue waiting until fully deleted
+		return false, nil
```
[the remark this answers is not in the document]
karlkfi: Returning false tells UntilDeletedWithSync to keep processing events. In this case, it hasn't received/processed the Deleted event yet. When it does, it will stop and return.

pkg/util/watch/delete.go
karlkfi: done

sdowell (approved):

PR #499, opened 2023-03-28
Refactor the conflict error functions to ConflictHandler

nan-yu:
It also unifies FightHandler to use the orderedMap.


pkg/remediator/conflict/handler.go
```diff
@@ -0,0 +1,86 @@
...
+func (h *handler) AddConflictError(e status.ManagementConflictError) {
+	h.mux.Lock()
+	defer h.mux.Unlock()
+
+	for _, existingErr := range h.conflictErrs {
+		if e.Error() == existingErr.Error() {
+			return
+		}
+	}
+	h.conflictErrs = append(h.conflictErrs, e)
```
karlkfi, line 60: Should we sort on add so they stay in a consistent order? Or are they always just ordered by order of appearance?

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -109,17 +109,16 @@ var _ Runnable = &filteredWatcher{}
...
-		removeConflictErrorFunc: cfg.removeConflictErrorFunc,
+		gvk:             cfg.gvk.String(),
+		startWatch:      cfg.startWatch,
+		resources:       cfg.resources,
+		queue:           cfg.queue,
+		scope:           cfg.scope,
+		syncName:        cfg.syncName,
+		base:            watch.NewEmptyWatch(),
+		errorTracker:    make(map[string]time.Time),
+		conflictErrMap:  make(map[queue.GVKNN]status.ManagementConflictError),
```
karlkfi, line 120: Is this map something the ConflictHandler could handle? Does it need to be locked the same way?

karlkfi: Seems redundant to store as both a map and a list, unless we care about the performance at scale. If performance and order are important, we could used an ordered map, like https://pkg.go.dev/github.com/wk8/go-ordered-map

pkg/remediator/conflict/handler.go
nan-yu: What about sorting on get?

nan-yu: Switched to use orderedmap.

pkg/remediator/watch/filteredwatcher.go
nan-yu: Switched to use ordered map.

karlkfi (approved):
/lgtm

Looks much better, thanks!

PR #500, opened 2023-03-28
Pass Kind instead of GVK to the record metrics functions

nan-yu:
The record functions only use the Kind tag. Passing GVK requires providing an empty version string to the callers.

It also sets the name tag.

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -230,12 +230,12 @@ func processApplyEvent(ctx context.Context, e event.ApplyEvent, s *stats.ApplyEv
 
 	case event.ApplySuccessful:
 		objectStatus.Actuation = actuation.ActuationSucceeded
-		handleMetrics(ctx, "update", e.Error, id.WithVersion(""))
+		handleMetrics(ctx, "update", e.Error, id.Kind)
 		return nil
 
 	case event.ApplyFailed:
 		objectStatus.Actuation = actuation.ActuationFailed
-		handleMetrics(ctx, "update", e.Error, id.WithVersion(""))
+		handleMetrics(ctx, "update", e.Error, id.Kind)
 		switch e.Error.(type) {
 		case *applyerror.UnknownTypeError:
 			unknownTypeResources[id] = struct{}{}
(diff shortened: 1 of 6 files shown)


pkg/applier/applier.go
```diff
@@ -770,7 +770,7 @@ func (a *supervisor) handleDisabledObjects(ctx context.Context, rg *live.Invento
 	for _, obj := range objs {
 		id := core.IDOf(obj)
 		err := a.abandonObject(ctx, obj)
-		handleMetrics(ctx, "unmanage", err, obj.GetObjectKind().GroupVersionKind())
+		handleMetrics(ctx, "unmanage", err, obj.GetObjectKind().GroupVersionKind().Kind)
```
karlkfi, line 773: Is this always Unstructured? If not, the GVK might not be set. Should we check?

pkg/metrics/record.go
```diff
@@ -139,31 +139,33 @@ func RecordApplyDuration(ctx context.Context, status, commit string, startTime t
...
-func RecordResourceFight(ctx context.Context, _ string, _ schema.GroupVersionKind) {
-	//tagCtx, _ := tag.New(ctx,
-	//tag.Upsert(KeyName, GetResourceLabels()),
-	//tag.Upsert(KeyOperation, operation),
-	//tag.Upsert(KeyType, gvk.Kind),
-	//)
+func RecordResourceFight(ctx context.Context, operation, kind string) {
+	reconcilerName := os.Getenv(reconcilermanager.ReconcilerNameKey)
+	tagCtx, _ := tag.New(ctx,
+		tag.Upsert(KeyName, reconcilerName),
```
karlkfi, line 145: I don't think these can/should be added if they're not registered on the metric View. Otherwise there will be errors when sending tags that aren't in the metric scheme.

Same goes for all the other tags you uncommented.

karlkfi: /cc @tiffanny29631

nan-yu: Should we add these tags to the metric views?

pkg/applier/applier.go
nan-yu: I traced the root where `objs` was initialized. It is unstructured, but with GVK populated: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/kinds/convert.go#L35. We should be good here.  We can also use `id.Kind` for simplicity.

pkg/metrics/record.go
karlkfi: I think that should be a different PR, if we do. And we need to discuss the cardinality effects, which is why they were removed.

nan-yu: Make sense. Commented back.

karlkfi (approved):

PR #501, opened 2023-03-29
Detect the new autopilot webhook used on GKE 1.26+

haiyanmeng:
Before GKE 1.26, we determine an Autopilot cluster based on the presence of the `workload-defaulter.config.common-webhooks.networking.gke.io` MutatingWebhookConfiguration.

On GKE 1.26+, an Autopilot cluster replaces
`workload-defaulter.config.common-webhooks.networking.gke.io` with `sasecret-redacter.config.common-webhooks.networking.gke.io`.

### DIFF
--- a/pkg/util/autopilot.go
+++ b/pkg/util/autopilot.go
@@ -43,8 +43,12 @@ var AutopilotManagedKinds = []schema.GroupVersionKind{
 	admissionregistrationv1.SchemeGroupVersion.WithKind("MutatingWebhookConfigurationList"),
 }
 
-// The mutating webhook to determine an Autopilot cluster.
-const autopilotWebhook = "workload-defaulter.config.common-webhooks.networking.gke.io"
+var autopilotWebhooks = []string{
+	// The mutating webhook to determine an Autopilot cluster before GKE 1.26.
+	"workload-defaulter.config.common-webhooks.networking.gke.io",
+	// The mutating webhook to determine an Autopilot cluster on GKE 1.26+.
+	"sasecret-redacter.config.common-webhooks.networking.gke.io",
+}
 
 // IsAutopilotManagedNamespace returns if the input object is a namespace managed by the Autopilot cluster.
 func IsAutopilotManagedNamespace(o client.Object) bool {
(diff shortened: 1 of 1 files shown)


pkg/util/autopilot.go
```diff
@@ -55,13 +59,20 @@ func IsAutopilotManagedNamespace(o client.Object) bool {
...
-		return true, nil
+	var autopilotWebhookNotFoundCount int
+	var err error
+	for _, autopilotWebhook := range autopilotWebhooks {
+		err = c.Get(context.Background(), client.ObjectKey{Name: autopilotWebhook}, &admissionregistrationv1.MutatingWebhookConfiguration{})
+		if err == nil {
+			return true, nil
+		}
+		if apierrors.IsNotFound(err) {
+			autopilotWebhookNotFoundCount++
```
nan-yu, line 72: nit: you don't need the `autopilotWebhookNotFoundCount` variable. You can `return false, err` `if !apierrors.IsNotFound(err)`, and `return false, nil` outside the for-loop.

haiyanmeng: If the first iteration encounters an error other than `NotFoundErr`, we should still go to the second iteration.

nan-yu: If the first iteration failed with an error other than `NotFoundErr`, that seems to indicate some k8s-apiserver error. Should we return the error and requeue the reconciliation request?

With the current approach, if the Get call for both webhooks failed, the first error will be overridden by the second one.

pkg/util/autopilot.go
```diff
@@ -55,13 +59,20 @@ func IsAutopilotManagedNamespace(o client.Object) bool {
...
-// It leverages the existence of the workload-defaulter mutating webhook configuration, which exists and only exists on the Autopilot clusters.
+// It determines an Autopilot cluster by the presence of autopilot webhooks.
 func IsGKEAutopilotCluster(c client.Client) (bool, error) {
-	err := c.Get(context.Background(), client.ObjectKey{Name: autopilotWebhook}, &admissionregistrationv1.MutatingWebhookConfiguration{})
-	if err == nil {
-		return true, nil
+	var autopilotWebhookNotFoundCount int
+	var err error
+	for _, autopilotWebhook := range autopilotWebhooks {
+		err = c.Get(context.Background(), client.ObjectKey{Name: autopilotWebhook}, &admissionregistrationv1.MutatingWebhookConfiguration{})
```
karlkfi, line 67: Might be slightly more efficient to do a LIST instead of a GET. That way it only makes one API call.

haiyanmeng: Two `Get` calls is okay. It does not worth the effort to optimize to save one API call.

pkg/util/autopilot.go
haiyanmeng: If the first iteration failed with an error other than NotFoundErr, it it still possible that the second iteration succeeds and identify the cluster as an Autopilot cluster. In this case, I don't think it is necessary to return the error in the first iteration.

If both Get calls failed with errors other than NotFoundErr, it is very likely that they have the same root cause. Therefore, it should be okay to just return the error in the second iteration.

nan-yu: SG.

nan-yu (approved):
/lgtm

PR #502, opened 2023-03-29
Detect the new autopilot webhook used on GKE 1.26+ (#501)

haiyanmeng:
Before GKE 1.26, we determine an Autopilot cluster based on the presence of the `workload-defaulter.config.common-webhooks.networking.gke.io` MutatingWebhookConfiguration.

On GKE 1.26+, an Autopilot cluster replaces
`workload-defaulter.config.common-webhooks.networking.gke.io` with `sasecret-redacter.config.common-webhooks.networking.gke.io`.


nan-yu (approved):
/lgtm

PR #504, opened 2023-03-30
Add script to report image vulnerabilities

karlkfi:
Example output:

```
$ scripts/vulnerabilities.sh
Scanning..........
IMAGE                                                                          VULNERABILITIES
gcr.io/config-management-release/admission-webhook:v1.15.0-rc.1                12
gcr.io/config-management-release/git-sync:v3.6.4-gke.1__linux_amd64            0
gcr.io/config-management-release/helm-sync:v1.15.0-rc.1                        46
gcr.io/config-management-release/hydration-controller:v1.15.0-rc.1             86
gcr.io/config-management-release/hydration-controller-with-shell:v1.15.0-rc.1  86
gcr.io/config-management-release/nomos:v1.15.0-rc.1                            12
gcr.io/config-management-release/oci-sync:v1.15.0-rc.1                         12
gcr.io/config-management-release/reconciler-manager:v1.15.0-rc.1               12
gcr.io/config-management-release/reconciler:v1.15.0-rc.1                       12
gcr.io/config-management-release/resource-group-controller:v1.0.9              0
ERROR: 278 critical, high, or medium vulnerabilities are fixable
```

### DIFF
--- a/scripts/vulnerabilities.sh
+++ b/scripts/vulnerabilities.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+#
+# Copyright 2023 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script prints a table of fixable vulnerabilities of medium severity or
+# higher, based on the latest or specified git tag.
+#
+# USAGE: $ scripts/vulnerabilities.sh [IMAGE_TAG]
+#
+# The IMAGE_TAG is optional. If not specified, it defaults to the latest tag of
+# the current local environment.
+#
+# The GCP_PROJECT environment variable is used to configure the project that
+# hosts the container images in Google Container Registry.
+# Default: config-management-release
+#
+# Requires gcloud authn/authz to read the results of vulnerability scans from
+# Google Container Registry in the specified project.
+# Required access:
+# - Storage Object Viewer on "artifacts.${GCP_PROJECT}.appspot.com"
+# - Container Registry Service Agent on Container Registry API
+#   (containerregistry.googleapis.com)
+# - Container Analysis Service Agent on Container Analysis API
+#   (containerscanning.googleapis.com)
+
+set -euo pipefail
+
+configsync_latest_image_tag="$(git describe --tags --abbrev=0)"
+configsync_image_tag="${1:-${configsync_latest_image_tag}}"
+
+GCP_PROJECT="${GCP_PROJECT:-config-management-release}"
+
+# TODO: share with Makefile to keep in sync
+configsync_images=(
+  "gcr.io/${GCP_PROJECT}/reconciler"
+  "gcr.io/${GCP_PROJECT}/reconciler-manager"
+  "gcr.io/${GCP_PROJECT}/admission-webhook"
+  "gcr.io/${GCP_PROJECT}/hydration-controller"
+  "gcr.io/${GCP_PROJECT}/hydration-controller-with-shell"
+  "gcr.io/${GCP_PROJECT}/oci-sync"
+  "gcr.io/${GCP_PROJECT}/helm-sync"
+  "gcr.io/${GCP_PROJECT}/nomos"
+)
+
+dependency_images=(
+  "gcr.io/${GCP_PROJECT}/resource-group-controller"
+  "gcr.io/${GCP_PROJECT}/git-sync"
+  "gcr.io/${GCP_PROJECT}/otelcontribcol"
+)
+
+# find the "name:tag" of images in the manifest directory by name
+find_manifest_images_by_name() {
+  image_name="$1"
+  grep -roh --color=never \
+    "${image_name}:\S*" \
+    ./manifests/ | uniq -u
+}
+
+# Build a full list of images with tags
+images=()
+
+# Add the specified tag to all configsync images
+for image_name in "${configsync_images[@]}"; do
+  images+=("${image_name}:${configsync_image_tag}")
+done
+
+# Add dependencies from the manifests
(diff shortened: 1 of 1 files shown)


scripts/vulnerabilities.sh
```diff
@@ -0,0 +1,114 @@
...
+  grep -roh --color=never \
+    "${image_name}:\S*" \
+    ./manifests/ | uniq -u
+}
+
+# Build a full list of images with tags
+images=()
+
+# Add the specified tag to all configsync images
+for image_name in "${configsync_images[@]}"; do
```
sdowell, line 69: An alternative approach I'm considering is taking in a `config-sync-manifest.yaml` and parsing the images from there. This would make the script work better across versions and help avoid the situation of someone forgetting to update this script (i.e. mismatch of actual images included in a release version versus what is declared in this script). The script could potentially accept either a local file path (e.g. `.output/staging/operator/config-sync-manifest.yaml`) or a GCS path. WDYT?

scripts/vulnerabilities.sh
```diff
@@ -0,0 +1,114 @@
...
+#   (containerregistry.googleapis.com)
+# - Container Analysis Service Agent on Container Analysis API
+#   (containerscanning.googleapis.com)
+
+set -euo pipefail
+
+configsync_latest_image_tag="$(git describe --tags --abbrev=0)"
+configsync_image_tag="${1:-${configsync_latest_image_tag}}"
+
+# TODO: share with Makefile to keep in sync
```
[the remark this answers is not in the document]
karlkfi: How would that work for the dependencies? They all need to be in the same project, but the dependencies are all hard-coded in the manifests, not dynamic like the ones from the makefile.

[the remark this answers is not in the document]
karlkfi: I guess the dependencies don't technically need to be in the same project if they both have scanning enabled, but I'd have to make the project configurable, not just the whole image prefix, since the project needs to be passed to the gcloud command.

scripts/vulnerabilities.sh
karlkfi: > An alternative approach I'm considering is taking in a config-sync-manifest.yaml and parsing the images from there.

The manifest isn't rendered. It doesn't have the image names in it. We would have to render them first to be able to pull them out of the yaml. I'm already pulling the hardcoded dependencies out of the manifest yaml.

scripts/vulnerabilities.sh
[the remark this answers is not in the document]
karlkfi: Ok, I made the project configurable with GCP_PROJECT

karlkfi:
/retest

[Build log](https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/504/kpt-config-sync-presubmit-e2e-multi-repo-test-group3/1641529785018486784) is weirdly empty, even though it ran 10m...

nan-yu (commented):
This script only works for the OSS version of Config Sync. Given that the mono-repo hasn't been deprecated yet, should we move the script to the operator repo and expand it to support mono-repo images as well?

scripts/vulnerabilities.sh
sdowell: One could either target already built manifests published in GCS, or checkout the tagged commit and run the `manifests` target to build the manifests at that SHA. This ensures there isn't a mismatch between what is included in the manifest bundle and what we scan for here

karlkfi: Perhaps. I didn't want to make this depend on the manifests being rendered locally. I suppose we could download the published manifests? Or do we only publish the ACM ones?

karlkfi:
> This script only works for the OSS version of Config Sync. Given that the mono-repo hasn't been deprecated yet, should we move the script to the operator repo and expand it to support mono-repo images as well?

That's the point. We want to make sure these are discovered and fixed before cutting a release candidate AND updating ACM AND cutting an ACM release candidate. 

Shift Left!

scripts/vulnerabilities.sh
sdowell: For the CI/CD use case I'm thinking we would pull published manifests from GCS (probably from a bucket in the CI project).

The flexibility to use a manifest on the filesystem would be a convenience for the manual/local use case, IMO. You could either download a published manifest or use manifests that you built locally

sdowell:
+1 to @karlkfi 

We should aim to shift this left, meaning we should probably have a similar script for both ACM and monorepo

scripts/vulnerabilities.sh
karlkfi: What if we commit the generated manifest into the repo? We could add a CI check to make sure it's up to date. Then we can just pull the latest from the local path AND we'll be able to download it raw from github for the ACM release process.

sdowell: I'm not sure how practical it would be. Images are typically tagged based on something after the commit is created, e.g. a commit tag or commit SHA - so it becomes a bit of a chicken and egg problem. We could use something arbitrary like a timestamp, but the image tag loses meaning/context. We also would need to choose a registry for these snapshot builds, which begs the question of when/how the images get published. Are they published before the change is submitted, or in a postsubmit job after?

I think the pattern of publishing the manifest and images together after the commit is submitted is a bit more of a natural fit

karlkfi: Fair enough. The other option is to publish the rendered manifest to GCS, like we do for ACM.

karlkfi: Can we do this later? I don't feel it's urgent to get into this PR.

sdowell: Sure, can be a future improvement

scripts/vulnerabilities.sh
```diff
@@ -0,0 +1,120 @@
...
+declare -A vuln_map
+
+echo -n "Scanning" >&2
+
+# Sum the fixable vulnerabilities with severity CRITICAL, HIGH, or MEDIUM
+for image in "${images[@]}"; do
+  echo -n "."
+  vulnerabilities=$(gcloud beta container images describe --project "${GCP_PROJECT}" --show-package-vulnerability --format json --verbosity error "${image}" |
+    jq -r '.package_vulnerability_summary.vulnerabilities')
+  # echo "Scan Results: ${vulnerabilities}"
```
sdowell, line 99: remove commented code here and below

sdowell (commented):
LGTM with one nit

scripts/vulnerabilities.sh
karlkfi: Done.

PR #507, opened 2023-03-30
Update Go to v1.19

karlkfi:
- Fix several CVEs from Go 1.17
- Update buildenv image to use golang:1.19.7-buster
- Pull out variable for golangci-lint version

This started as a cherry-pick of https://github.com/GoogleContainerTools/kpt-config-sync/pull/465, but some different choices were made previously when upgrading golangci-lint, so there weren't any more lint failures to fix. And we don't need to use the default exclusions, because we already have exclusions configured in .golangci.yaml.


sdowell:
/retest

One more time with feeling

karlkfi:
Realized the cherrypick was actually downgrading the linter compared to head, so i fixed it.

PR #511, opened 2023-04-03
Bitbucket e2e test create repository fix

victorpras:
Create repository API in bitbucket (POST https://api.bitbucket.org/2.0/repositories/) will return an error if the repository name ends in a non alphanumeric character. Given that we are currently using `uuid.NewRandom` to generate a unique repository name and perform truncation to keep the name within maximum length, the repository name can end with a hyphen '-' character. To prevent this, all the hyphens in the name and uuid will be stripped

### DIFF
--- a/e2e/nomostest/gitproviders/bitbucket.go
+++ b/e2e/nomostest/gitproviders/bitbucket.go
@@ -37,6 +37,7 @@ const (
 	PrivateSSHKey = "config-sync-ci-ssh-private-key"
 
 	repoNameMaxLength = 62
+	localNameLength   = 30
 )
 
 // BitbucketClient is the client that calls the Bitbucket REST APIs.
@@ -88,8 +89,17 @@ func (b *BitbucketClient) CreateRepository(localName string) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "failed to generate a new UUID")
 	}
-	// Make the remote repoName unique in order to run multiple tests in parallel.
-	repoName := strings.ReplaceAll(localName, "/", "-") + "-" + u.String()
+
+	// strip all hyphens from localName and uuid before appending
+	localName = strings.ReplaceAll(localName, "/", "")
+	localName = strings.ReplaceAll(localName, "-", "")
+	if len(localName) > localNameLength {
+		localName = localName[:localNameLength]
+	}
+
+	uuid := strings.ReplaceAll(u.String(), "-", "")
+
+	repoName := localName + uuid
 	if len(repoName) > repoNameMaxLength {
 		repoName = repoName[:repoNameMaxLength]
 	}


e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -94,6 +95,11 @@ func (b *BitbucketClient) CreateRepository(localName string) (string, error) {
 		repoName = repoName[:repoNameMaxLength]
 	}
 
+	// last character should be alphanumeric
+	if !regexp.MustCompile(`^[a-zA-Z0-9]*$`).MatchString(repoName[len(repoName)-1:]) {
+		repoName = repoName[:len(repoName)-1]
```
karlkfi, line 100: what repo names end in non-alphanumeric characters???

karlkfi: can we just add validation here that fails the test and then go fix the test to not have a crazy name?

karlkfi: or it it just that it gets truncated in the middle of the name that it might end in a hyphen?

victorpras: To create a unique repository name, we generate a random string using `uuid.NewRandom()`, append it to the provided r*sync name, and truncate it to keep it within maximum length. The truncated name can sometimes ends with a dash ('-') character that will cause the bitbucket create repo API to throw an error

karlkfi: Here's a better idea, taken from a similar problem we had with GKE subnet names in firewall rules:

1. Use a shorter max-length for the localName (like max-8, to allow at least 7 UUID characters), and truncate that first. This leaves enough characters to append a minimum length uuid suffix, which ensures that extra long names don't become non-unique after truncation.
2. Strip all the hyphens out of the UUID before appending it to the localName. This ensures the string will never end in a hyphen.
3. Do another max-length truncation after appending. This ensures use of the longest possible UUID, depending on localName length.

victorpras: localName always start with the namespace name, so typically RootSync resource name will look like `config-management-system-root-sync`. Meanwhile, UUID is always 36 characters, 32 if we stripped all the hyphens. The current max length for repo name is set to 62, so maybe we can allocate 30 characters for localName and 32 for UUID (all stripped of hyphens)? With that, typical repoName will look something like: `configmanagementsystemrootsync43b555fe7764f76bbb1497f2492d8ed`

victorpras: updated to follow Karl's solution in https://github.com/GoogleContainerTools/kpt-config-sync/pull/511#discussion_r1156511615

e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -88,8 +90,20 @@ func (b *BitbucketClient) CreateRepository(localName string) (string, error) {
...
+
+	// strip all hyphens from localName and uuid before appending
+	localName = strings.ReplaceAll(localName, "/", "")
+	localName = strings.ReplaceAll(localName, "-", "")
+	if len(localName) > localNameLength {
+		localName = localName[:localNameLength]
+	}
+
+	uuid := strings.ReplaceAll(u.String(), "-", "")
+	if len(uuid) > uuidLength {
```
sdowell, line 102: UUID should be a fixed length, so this if statement seems unnecessary

victorpras: you're right, uuid is always 36 characters long (32 after trimming hyphens). uuid length validation has been removed

karlkfi (approved):
/lgtm

PR #512, opened 2023-04-04
add Notification status CRD and controller logic

sdowell:
This api surfaces the status of the notification controller, which informs users on the status of:
- which notifications have been delivered
- whether any errors occurred
- whether any warnings occurred

This is intended to provide a better UX for understanding the state of the Notification configuration. The lifecycle of the Notification CR is managed by reconciler-manager, and the status is updated by the notification container in the reconciler.


pkg/notifications/notifications.go
```diff
@@ -0,0 +1,70 @@
...
+			Recipient:       d.Destination.Recipient,
+			AlreadyNotified: d.AlreadyNotified,
+		})
+	}
+	return notificationStatus
+}
+
+// IsNotificationStatusSame returns whether the two provided NotificationStatus
+// objects are equal. Ignores certain fields such as timestamps.
+func IsNotificationStatusSame(left, right v1beta1.NotificationStatus) bool {
```
nan-yu, line 56: Should it include ObservedGeneration as well?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -556,3 +556,53 @@ func (r *reconcilerBase) isLastReconciled(nn types.NamespacedName, resourceVersi
 	}
 	return resourceVersion == lastReconciled
 }
+
+func (r *reconcilerBase) upsertNotificationCR(ctx context.Context, rs client.Object) (*v1beta1.Notification, error) {
```
nan-yu, line 560: Suggest adding labels to the notification CR for querying with labelSelector
```
configsync.gke.io/sync-namespace: ...
configsync.gke.io/sync-name: ...
configsync.gke.io/sync-kind: ...
```

pkg/notifications/notifications.go
sdowell: Hmm, I'm not certain. I think the question is whether the Notification status should be updated whenever the RSync generation changes. WDYT?

pkg/reconcilermanager/controllers/reconciler_base.go
sdowell: I'm not opposed, but I'm curious about the value add. Given the Notification has the same name/namespace as the RSync (and the namespace implies the sync-kind), it seems like a user could query using name/namespace. I suppose perhaps it's an implementation detail that the name/namespace is the same, and that might not be transparent to the user.

Did you have a use case in mind?

pkg/notifications/notifications.go
nan-yu: I would think so. When RSync generation changes, the reconciler Pod will be restarted, and a new notification event would be triggered. 
Even thought the same event might already be delivered, the observeredGeneration of Notification CR should match RSync.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: The use case I can think of is to filter certain Notification CRs with the labels via CLI/UI.

After a second thought, since the CR has the same name/namespace as the RSync, we can also use the namespace/name to select the objects. For UI, it is most likely the Notification status is just a column of the Package, rather than a separate page.

I think it is okay to not add the labels for now.

pkg/notifications/notifications.go
sdowell: Updated to compare the `ObservedGeneration`

nan-yu (approved):
/lgtm

PR #515, opened 2023-04-04
Add helm & kustomize licenses into images

karlkfi:
- Move creation of LICENSES.txt into the release build image. This allows inclusion of the binary dependency licenses, which are not vendored.

PR #517, opened 2023-04-05
Validate helm & kustomize checksums

karlkfi:
Since wget validates the TLS cert of github when downloading, we can trust that the file is from the expected source without MITM modification. Validating the checksum proves that the file downloaded without corruption. This mirrors what a package manager would do.


karlkfi:
/hold

Blocked on https://github.com/GoogleContainerTools/kpt-config-sync/pull/515

haiyanmeng (approved):

PR #518, opened 2023-04-05
test: fix git portforward to support non-local provider

sdowell:
The port argument only needs to be provided for the in-cluster git provider. In other cases, such as Gitlab or Bitbucket, the gitRepoPortForwarder is nil and the port argument is unused. This updates the logic for git repository setup to support the other git provider types.


karlkfi (approved):
/lgtm

PR #520, opened 2023-04-05
Fix vulnerabilities.sh uniqueness filter

karlkfi:
The -u flag was showing only the unique values, which isn't what we wanted. Now it reduces duplicate entries to single entries as intended.

### DIFF
--- a/scripts/vulnerabilities.sh
+++ b/scripts/vulnerabilities.sh
@@ -65,7 +65,7 @@ find_manifest_images_by_name() {
   image_name="$1"
   grep -roh --color=never \
     "${image_name}:\S*" \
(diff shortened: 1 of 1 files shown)


karlkfi:
/retest

Unrelated metrics flake:
```
=== CONT  TestObjectReactsToChangeInLegacyClusterSelector
    new.go:440: ERROR: timed out waiting for metrics predicate[0]: no results from prometheus query: config_sync_last_sync_timestamp{commit="e901c7b5c5b45acfe6165d1d7d7441d6432c1ebb", component="otel-collector", k8s_pod_name="root-reconciler-5b769b4447-w2znj", status="success"}
```

PR #521, opened 2023-04-05
test: refactor git portforwarder under gitprovider

sdowell:
This refactors the git port forwarder abstraction to be underneath the git provider. The git port forwarder is only relevant with the in-cluster git server, and thus fits as an implementation detail of the LocalProvider.

### DIFF
--- a/e2e/nomostest/git.go
+++ b/e2e/nomostest/git.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/retry"
 	"kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
@@ -96,9 +97,8 @@ type Repository struct {
 	// For other git providers, it appends a UUID to Name for uniqueness.
 	RemoteRepoName string
 
-	// RemoteURL is the remote URL of the repository.
-	// It is used to set the url for the remote origin using `git remote add origin <REMOTE_URL>.
-	RemoteURL string
+	// GitProvider is the provider that hosts the Git repositories.
+	GitProvider gitproviders.GitProvider
 
 	// Scheme used for encoding and decoding objects.
 	Scheme *runtime.Scheme
@@ -130,6 +130,7 @@ func NewRepository(nt *NT, repoType RepoType, nn types.NamespacedName, sourceFor
 		SafetyClusterRolePath: fmt.Sprintf("acme/cluster/cluster-role-%s.yaml", safetyName),
 		Scheme:                nt.Scheme,
 		Logger:                nt.Logger,
+		GitProvider:           nt.GitProvider,
 	}
 
 	repoName, err := nt.GitProvider.CreateRepository(namespacedName)
@@ -139,16 +140,6 @@ func NewRepository(nt *NT, repoType RepoType, nn types.NamespacedName, sourceFor
 		nt.T.Fatal(err)
 	}
 	g.RemoteRepoName = repoName
-	var port int
-	// port argument is used for in-cluster git provider with the local port forward.
-	// the argument is ignored for other providers, e.g. Gitlab/Bitbucket
-	if nt.gitRepoPortForwarder != nil {
-		port, err = nt.gitRepoPortForwarder.LocalPort()
-		if err != nil {
-			nt.T.Fatal(err)
-		}
-	}
-	g.RemoteURL = nt.GitProvider.RemoteURL(port, repoName)
 
 	g.init(nt.gitPrivateKeyPath)
 	g.initialCommit(sourceFormat)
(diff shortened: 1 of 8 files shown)


e2e/nomostest/git.go
```diff
@@ -139,16 +140,6 @@ func NewRepository(nt *NT, repoType RepoType, nn types.NamespacedName, sourceFor
 		nt.T.Fatal(err)
 	}
 	g.RemoteRepoName = repoName
-	var port int
```
karlkfi, line 142: yay so much cleaner setup!

e2e/nomostest/gitproviders/git-provider.go
```diff
@@ -33,7 +34,7 @@ type GitProvider interface {
 	// For the testing git-server, RemoteURL uses localhost and forwarded port, while SyncURL uses the DNS.
 	// For other git providers, RemoteURL should be the same as SyncURL.
 	// name refers to the repo name in the format of <NAMESPACE>/<NAME> of RootSync|RepoSync.
-	RemoteURL(port int, name string) string
+	RemoteURL(name string) (string, error)
```
karlkfi, line 37: Prolly just want to rename this URL or Address, since it's not always remote.

e2e/nomostest/new.go
```diff
@@ -318,8 +316,7 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	if err := nt.KubeClient.Create(fake.NamespaceObject(metrics.MonitoringNamespace)); err != nil {
 		nt.T.Fatal(err)
 	}
-
-	if nt.GitProvider.Type() == e2e.Local {
+	if *e2e.GitProvider == e2e.Local {
```
karlkfi, line 319: Not for this PR, but it might be worth thinking if this logic to set up the auth for git can also be abstracted.

e2e/nomostest/nt.go
```diff
@@ -615,19 +609,18 @@ func (nt *NT) portForwardGitServer() {
...
-			remoteURL := nt.GitProvider.RemoteURL(newPort, nn.String())
-			remoteRepo.pushAllToRemote(remoteURL)
-			remoteRepo.RemoteURL = remoteURL
+		for _, remoteRepo := range nt.RemoteRepositories {
+			remoteRepo.pushAllToRemote()
 		}
 		prevPodName = podName
 	}
-	nt.gitRepoPortForwarder = nt.newPortForwarder(
+	gitRepoPortForwarder := nt.newPortForwarder(
```
karlkfi, line 617: If you're only gonna use this var on the next line, you can just return it here and save an allocation.

e2e/nomostest/nt.go
```diff
@@ -615,19 +609,18 @@ func (nt *NT) portForwardGitServer() {
 		// re-init all repos
 		InitGitRepos(nt, allRepos...)
 		// attempt to recover by re-pushing the local repo states
-		for nn, remoteRepo := range nt.RemoteRepositories {
-			remoteURL := nt.GitProvider.RemoteURL(newPort, nn.String())
-			remoteRepo.pushAllToRemote(remoteURL)
-			remoteRepo.RemoteURL = remoteURL
+		for _, remoteRepo := range nt.RemoteRepositories {
+			remoteRepo.pushAllToRemote()
```
karlkfi, line 613: Kinda surprised this isn't public yet. Is it because Repository is still in this package?

Maybe that's another PR to pull out.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -118,17 +118,19 @@ func (pf *PortForwarder) LocalPort() (int, error) {
...
-	if pf.setPortCallback != nil {
-		pf.setPortCallback(port, pod)
 	}
 	pf.cmd = cmd
 	pf.localPort = port
 	pf.pod = pod
+	if async {
+		pf.mux.Unlock()
+	}
+	if pf.setPortCallback != nil {
```
karlkfi, line 131: I dont think we need it now, but one way to make this more flexible, slightly cleaner, and reduce the number of branches is to make the callbacks an array and for loop over it. The only problem with this approach for hooks/callbacks in general is removal. If you need removal and ordering, then an OrderedMap is required.

karlkfi (changes requested):
A couple comments. Mostly looking good tho!

e2e/nomostest/gitproviders/git-provider.go
sdowell: I think `Remote` is fairly applicable here, assuming it's referring to `remote` in git terminology

e2e/nomostest/new.go
sdowell: Agreed, that instantiation could be abstracted to the provider as well

e2e/nomostest/nt.go
sdowell: Yeah it's still in this package and the method is only used within the package currently. The method will need to be public if it's moved to its own package

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: Right now we're just registering a single function, but I see your point to flexility if we need to support multiple callbacks

e2e/nomostest/nt.go
sdowell: done

karlkfi (approved):
/lgtm

e2e/nomostest/gitproviders/git-provider.go
karlkfi: I see where you're coming from, and I've approved, but I think the "remote" is from the perspective of the local repo copy. From the provider's perspective, it's actually the address of the local repo. Semantics tho. I don't think it super matters.

PR #523, opened 2023-04-06
test: init GitProvider before setupTestCase

sdowell:
There are certain code paths which expect the GitProvider to be set, such as Clean for shared test environments on GKE. Initializing the GitProvider later in setupTestCase was leading to a null pointer exception. This moves the initialization to FreshTestEnv/SharedTestEnv.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -24,7 +24,6 @@ import (
 
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e"
-	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	testmetrics "kpt.dev/configsync/e2e/nomostest/metrics"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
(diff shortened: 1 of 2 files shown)


e2e/nomostest/new.go
```diff
@@ -203,6 +202,11 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		}
 	})
 
+	// re-init the GitProvider.
+	// The setPortCallback must be registered for this NT so that it initializes
+	// the correct repositories.
+	nt.initGitProvider()
```
karlkfi, line 208: Is it deliberate that this is set after the Cleanup is registered? Construction of the provider and forwarder doesn't look like it does anything that needs cleanup.

karlkfi: In fact, it's before Cleanup in FreshTestEnv, so it should probably be before cleanup here too.

sdowell: `Cleanup` is essentially a no-op until the end of the test so I don't think it really matters, but I can change the ordering

e2e/nomostest/nt.go
```diff
@@ -622,6 +622,14 @@ func (nt *NT) portForwardGitServer() *portforwarder.PortForwarder {
 	)
 }
 
+func (nt *NT) initGitProvider() {
```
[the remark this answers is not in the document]
sdowell: It still needs access to the `nt` object for the `portForwardGitServer` method etc, so it can't be inline with the declaration

e2e/nomostest/new.go
sdowell: reordered

e2e/nomostest/nt.go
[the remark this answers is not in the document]
karlkfi: ugh. that stuff all needs more cleanup... this is fine for now.

karlkfi (approved):
/lgtm

PR #525, opened 2023-04-06
test: set GitProvider in ReInit

sdowell:
The GitProvider must be set here, otherwise the Repository may be using a GitProvider from a previous NT. Using a stale GitProvider means that the PortForwarder will be closed, since it is closed at the end of the test.


karlkfi (approved):
/lgtm

PR #526, opened 2023-04-06
test: use RemoteRepositories for setPortCallback

sdowell:
The RootRepos/NonRootRepos fields are reset for every testcase, which required for the PortForwarder to be recreated with each testcase. This switches the reset function to use RemoteRepositories, which represents all Repositories that have been created (across test cases). This change enables us to create a single GitProvider in FreshTestEnv and share it across shared test environments. This also means there is only a single PortForwarder to the git server active at a given time, which is preferred.


karlkfi (approved):
/lgtm

PR #530, opened 2023-04-07
test: restart PortForwarder if process is killed

sdowell:
Prior to this change the port forward was restarted only if the Pod was
restarted. This change makes it so that the port forward is restarted
either if the Pod is restarted or if the port-forward process is killed.

### DIFF
--- a/e2e/nomostest/gitproviders/git-provider.go
+++ b/e2e/nomostest/gitproviders/git-provider.go
@@ -16,7 +16,6 @@ package gitproviders
 
 import (
 	"kpt.dev/configsync/e2e"
-	"kpt.dev/configsync/e2e/nomostest/portforwarder"
 	"kpt.dev/configsync/e2e/nomostest/testing"
 )
 
@@ -44,29 +43,8 @@ type GitProvider interface {
 	DeleteObsoleteRepos() error
 }
 
-// GitProviderOpt is an optional parameter for instantiating a new GitProvider
-type GitProviderOpt func(opts *GitProviderOpts)
-
-// GitProviderOpts is the set of optional parameters for instantiating a new GitProvider
-type GitProviderOpts struct {
-	portForwarder *portforwarder.PortForwarder
-}
-
-// WithPortForwarder provides a PortForwarder for the GitProvider to use.
-// Required for LocalProvider in order to establish a PortForwarder to the in-cluster
-// git server.
-func WithPortForwarder(portForwarder *portforwarder.PortForwarder) GitProviderOpt {
-	return func(opts *GitProviderOpts) {
-		opts.portForwarder = portForwarder
-	}
-}
-
 // NewGitProvider creates a GitProvider for the specific provider type.
-func NewGitProvider(t testing.NTB, provider string, opts ...GitProviderOpt) GitProvider {
-	options := GitProviderOpts{}
-	for _, opt := range opts {
-		opt(&options)
-	}
+func NewGitProvider(t testing.NTB, provider string) GitProvider {
 	switch provider {
 	case e2e.Bitbucket:
 		client, err := newBitbucketClient()
@@ -81,10 +59,6 @@ func NewGitProvider(t testing.NTB, provider string, opts ...GitProviderOpt) GitP
 		}
 		return client
 	default:
-		client, err := newLocalProvider(options)
-		if err != nil {
-			t.Fatal(err)
-		}
-		return client
+		return &LocalProvider{}
 	}
 }
--- a/e2e/nomostest/gitproviders/local.go
+++ b/e2e/nomostest/gitproviders/local.go
@@ -23,17 +23,9 @@ import (
 
 // LocalProvider refers to the test git-server running on the same test cluster.
 type LocalProvider struct {
-	portForwarder *portforwarder.PortForwarder
-}
-
-func newLocalProvider(opts GitProviderOpts) (*LocalProvider, error) {
-	if opts.portForwarder == nil {
-		return nil, fmt.Errorf("must provide a PortForwarderFactory for LocalProvider")
-	}
-	provider := &LocalProvider{
-		portForwarder: opts.portForwarder,
-	}
-	return provider, nil
+	// PortForwarder is a port forwarder to the in-cluster git server.
+	// This is used to communicate from the tests to the in-cluster git server.
+	PortForwarder *portforwarder.PortForwarder
 }
 
 // Type returns the provider type.
@@ -44,11 +36,21 @@ func (l *LocalProvider) Type() string {
 // RemoteURL returns the Git URL for connecting to the test git-server.
 // name refers to the repo name in the format of <NAMESPACE>/<NAME> of RootSync|RepoSync.
 func (l *LocalProvider) RemoteURL(name string) (string, error) {
-	port, err := l.portForwarder.LocalPort()
+	if l.PortForwarder == nil {
+		return "", fmt.Errorf("PortForwarder must be set for LocalProvider.RemoteURL()")
+	}
+	port, err := l.PortForwarder.LocalPort()
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("ssh://git@localhost:%d/git-server/repos/%s", port, name), nil
+	return l.RemoteURLWithPort(port, name)
+}
+
+// RemoteURLWithPort returns the Git URL for connecting to the test git-server.
+// localPort refers to the local port the PortForwarder is listening on.
+// name refers to the repo name in the format of <NAMESPACE>/<NAME> of RootSync|RepoSync.
+func (l *LocalProvider) RemoteURLWithPort(localPort int, name string) (string, error) {
+	return fmt.Sprintf("ssh://git@localhost:%d/git-server/repos/%s", localPort, name), nil
 }
 
 // SyncURL returns a URL for Config Sync to sync from.
--- a/e2e/nomostest/gitproviders/repository.go
+++ b/e2e/nomostest/gitproviders/repository.go
@@ -118,6 +118,8 @@ type Repository struct {
 	yamlSerializer *jserializer.Serializer
 
 	defaultWaitTimeout time.Duration
+	// isEmpty indicates whether the local repo has any commits
+	isEmpty bool
 }
 
 // NewRepository creates a remote repo on the git provider.
@@ -151,6 +153,7 @@ func NewRepository(
 		PrivateKeyPath:        privateKeyPath,
 		yamlSerializer:        jserializer.NewYAMLSerializer(jserializer.DefaultMetaFactory, scheme, scheme),
 		defaultWaitTimeout:    defaultWaitTimeout,
+		isEmpty:               true,
 	}
 }
 
@@ -162,16 +165,7 @@ func (g *Repository) Create() error {
 		return errors.Wrapf(err, "creating repo: %s", g.Name)
 	}
 	g.RemoteRepoName = repoName
-	return g.Reformat(g.Format)
-}
-
-// Reformat the repo using the specified source format.
-// Use Repository.Format to keep the format the same.
-func (g *Repository) Reformat(format filesystem.SourceFormat) error {
-	if err := g.init(); err != nil {
-		return err
-	}
-	return g.initialCommit(format)
+	return nil
 }
 
 // Git executes the command from the repo root.
@@ -205,8 +199,8 @@ func (g *Repository) BulkGit(cmds ...[]string) error {
 	return nil
 }
 
-// initialCommit initializes the Nomos repo with the Repo object.
-func (g *Repository) initialCommit(sourceFormat filesystem.SourceFormat) error {
+// InitialCommit initializes the Nomos repo with the Repo object.
+func (g *Repository) InitialCommit(sourceFormat filesystem.SourceFormat) error {
 	// Add .gitkeep to retain dir when empty, otherwise configsync will error.
 	if err := g.AddEmptyDir(DefaultSyncDir); err != nil {
 		return err
@@ -238,12 +232,13 @@ func (g *Repository) initialCommit(sourceFormat filesystem.SourceFormat) error {
 	return g.CommitAndPush("initial commit")
 }
 
-// init initializes this git repository and configures it to talk to the cluster
+// Init initializes this git repository and configures it to talk to the cluster
 // under test.
-func (g *Repository) init() error {
+func (g *Repository) Init() error {
 	if err := os.RemoveAll(g.Root); err != nil {
 		return errors.Wrapf(err, "deleting root directory: %s", g.Root)
 	}
+	g.isEmpty = true
 
 	err := os.MkdirAll(g.Root, fileMode)
 	if err != nil {
@@ -554,7 +549,11 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) error {
 		return err
 	}
 	g.Logger.Infof("[repo %s] pushing commit: %s", path.Base(g.Root), commit)
-	return g.Push(branch, "-f")
+	if err := g.Push(branch, "-f"); err != nil {
+		return err
+	}
+	g.isEmpty = false
+	return nil
 }
 
 // Push pushes the provided refspec to the git server.
@@ -571,8 +570,7 @@ func (g *Repository) Push(args ...string) error {
 		if err != nil {
 			return err
 		}
-		_, err = g.Git(append([]string{"push", remoteURL}, args...)...)
-		return err
+		return g.push(remoteURL, args...)
 	})
 	if err != nil {
 		return errors.Wrapf(err, "failed to push after retrying for %v", took)
@@ -581,9 +579,21 @@ func (g *Repository) Push(args ...string) error {
 	return nil
 }
 
+func (g *Repository) push(remoteURL string, args ...string) error {
+	_, err := g.Git(append([]string{"push", remoteURL}, args...)...)
+	return err
+}
+
 // PushAllBranches push all local branches to the git server.
-func (g *Repository) PushAllBranches() error {
-	return g.Push("--all")
+// This is currently intended to only be called from the OnReadyCallback for the
+// in-cluster git server. Accepts a remoteURL to avoid calls to LocalPort, as this
+// would lead to a deadlock
+func (g *Repository) PushAllBranches(remoteURL string) error {
+	if g.isEmpty {
+		// empty repository, nothing to push
+		return nil
+	}
+	return g.push(remoteURL, "--all")
 }
 
 // CreateBranch creates and checkouts a new branch at once.
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -260,6 +260,7 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		RemoteRepositories:      make(map[types.NamespacedName]*gitproviders.Repository),
 		WebhookDisabled:         &webhookDisabled,
 		DefaultMetricsTimeout:   30 * time.Second,
+		GitProvider:             gitproviders.NewGitProvider(t, *e2e.GitProvider),
 	}
 
 	// Speed up the delay between sync attempts to speed up testing
@@ -283,9 +284,6 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	} else {
 		nt.DefaultWaitTimeout = 6 * time.Minute
 	}
-
-	// GitProvider is used by Clean, so this must be instantiated before the call to Clean
-	nt.initGitProvider()
 
 	if *e2e.TestCluster == e2e.Kind {
 		// We're using an ephemeral Kind cluster, so connect to the local Docker
@@ -358,25 +356,28 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 
 func setupTestCase(nt *NT, opts *ntopts.New) {
 	nt.T.Log("[SETUP] New test case")
-
-	// allRepos specifies the slice all repos for port forwarding.
-	var allRepos []types.NamespacedName
-	for repo := range opts.RootRepos {
-		allRepos = append(allRepos, RootSyncNN(repo))
-	}
-	for repo := range opts.NamespaceRepos {
-		allRepos = append(allRepos, repo)
-	}
-
-	if *e2e.GitProvider == e2e.Local {
-		InitGitRepos(nt, allRepos...)
-	}
-
+	// init all repositories. This ensures that:
+	// - local repo is initialized and empty
+	// - remote repo exists (except for LocalProvider, which is done in portForwardGitServer)
 	for name := range opts.RootRepos {
-		nt.RootRepos[name] = ResetRepository(nt, gitproviders.RootRepo, RootSyncNN(name), opts.SourceFormat)
+		nt.RootRepos[name] = initRepository(nt, gitproviders.RootRepo, RootSyncNN(name), opts.SourceFormat)
 	}
 	for nsr := range opts.NamespaceRepos {
-		nt.NonRootRepos[nsr] = ResetRepository(nt, gitproviders.NamespaceRepo, nsr, filesystem.SourceFormatUnstructured)
+		nt.NonRootRepos[nsr] = initRepository(nt, gitproviders.NamespaceRepo, nsr, filesystem.SourceFormatUnstructured)
+	}
+	// set up port forward if using in-cluster git server
+	if *e2e.GitProvider == e2e.Local {
+		nt.portForwardGitServer()
+	}
+	// The following prerequisites have been met, so we can now push commits
+	// - local repo initialized
+	// - remote repo exists
+	// - port forward started (only applicable to in-cluster git server)
+	for name := range opts.RootRepos {
+		initialCommit(nt, gitproviders.RootRepo, RootSyncNN(name), opts.SourceFormat)
+	}
+	for nsr := range opts.NamespaceRepos {
+		initialCommit(nt, gitproviders.NamespaceRepo, nsr, filesystem.SourceFormatUnstructured)
 	}
 
 	if opts.InitialCommit != nil {
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -582,10 +582,7 @@ func (nt *NT) MustDeleteGatekeeperTestData(file, name string) {
 }
 
 func (nt *NT) newPortForwarder(ns, deployment, port string, opts ...portforwarder.PortForwardOpt) *portforwarder.PortForwarder {
-	ctx, cancel := context.WithCancel(nt.Context)
-	nt.T.Cleanup(cancel)
 	return portforwarder.NewPortForwarder(
-		ctx,
 		nt.KubeClient,
 		nt.Watcher,
 		nt.Logger,
@@ -594,6 +591,42 @@ func (nt *NT) newPortForwarder(ns, deployment, port string, opts ...portforwarde
 		ns, deployment, port,
 		opts...,
 	)
+}
+
+func (nt *NT) startPortForwarder(ns, deployment string, pf *portforwarder.PortForwarder) {
+	ctx, cancel := context.WithCancel(nt.Context)
+	eventCh := pf.Start(ctx)
+	doneCh := make(chan struct{})
+	readyCh := make(chan struct{})
+	nt.T.Cleanup(func() {
+		nt.T.Logf("killing port-forward %s/%s", ns, deployment)
+		cancel()
+		<-doneCh
+		nt.T.Logf("finished waiting for port-forward %s/%s to exit", ns, deployment)
+	})
+	go func() {
+		defer close(doneCh)
+		// first event should be Init. signal if it's ready
+		event := <-eventCh
+		if event.Type == portforwarder.InitEvent {
+			close(readyCh)
+		}
+		for event := range eventCh {
+			if event.Type == portforwarder.ErrorEvent {
+				nt.T.Error(event.Message)
+			}
+		}
+	}()
+	timeoutCtx, timeoutCancel := context.WithTimeout(ctx, nt.DefaultWaitTimeout)
+	defer timeoutCancel()
+	now := time.Now()
+	select {
+	case <-timeoutCtx.Done():
+		since := time.Since(now)
+		nt.T.Fatalf("Waited %v for PortForwarder %s/%s to become ready: %v", since, ns, deployment, timeoutCtx.Err())
+	case <-readyCh:
+		nt.T.Logf("PortForwarder for %s/%s is ready", ns, deployment)
+	}
 }
 
 // portForwardOtelCollector forwards the otel-collector deployment to a port.
@@ -606,16 +639,21 @@ func (nt *NT) portForwardOtelCollector() {
 		ocmetrics.OtelCollectorName,
 		fmt.Sprintf(":%d", testmetrics.OtelCollectorMetricsPort),
 	)
+	nt.startPortForwarder(
+		ocmetrics.MonitoringNamespace,
+		ocmetrics.OtelCollectorName,
+		nt.otelCollectorPortForwarder,
+	)
 }
 
 // portForwardGitServer forwards the git-server deployment to a port.
-func (nt *NT) portForwardGitServer() *portforwarder.PortForwarder {
+func (nt *NT) portForwardGitServer() {
 	nt.T.Helper()
+	provider := nt.GitProvider.(*gitproviders.LocalProvider)
 	prevPodName := ""
 	resetGitRepos := func(newPort int, podName string) {
 		// pod unchanged, don't reset
-		if prevPodName == "" || prevPodName == podName {
-			prevPodName = podName
+		if prevPodName == podName {
 			return
 		}
 		// allRepos specifies the slice all repos for port forwarding.
@@ -625,28 +663,34 @@ func (nt *NT) portForwardGitServer() *portforwarder.PortForwarder {
 		}
 		// re-init all repos
 		InitGitRepos(nt, allRepos...)
-		// attempt to recover by re-pushing the local repo states
-		for _, remoteRepo := range nt.RemoteRepositories {
-			if err := remoteRepo.PushAllBranches(); err != nil {
+		for _, repo := range allRepos {
+			// construct remoteURL with provided port. Calling LocalPort would lead to deadlock
+			remoteURL, err := provider.RemoteURLWithPort(newPort, repo.String())
+			if err != nil {
+				nt.T.Fatal(err)
+			}
+			if err := nt.RemoteRepositories[repo].PushAllBranches(remoteURL); err != nil {
 				nt.T.Fatal(err)
 			}
 		}
 		prevPodName = podName
 	}
-	return nt.newPortForwarder(
+	nt.T.Cleanup(func() {
+		// clear PortForwarder after each test. at this point it has been stopped
+		// a new PortForwarder is created for each test.
+		provider.PortForwarder = nil
+	})
+	provider.PortForwarder = nt.newPortForwarder(
 		testGitNamespace,
 		testGitServer,
 		":22",
-		portforwarder.WithSetPortCallback(resetGitRepos),
+		portforwarder.WithOnReadyCallback(resetGitRepos),
 	)
-}
-
-func (nt *NT) initGitProvider() {
-	var gitProviderOpts []gitproviders.GitProviderOpt
-	if *e2e.GitProvider == e2e.Local {
-		gitProviderOpts = append(gitProviderOpts, gitproviders.WithPortForwarder(nt.portForwardGitServer()))
-	}
-	nt.GitProvider = gitproviders.NewGitProvider(nt.T, *e2e.GitProvider, gitProviderOpts...)
+	nt.startPortForwarder(
+		testGitNamespace,
+		testGitServer,
+		provider.PortForwarder,
+	)
 }
 
 // portForwardGitServer forwards the prometheus deployment to a port.
@@ -659,6 +703,11 @@ func (nt *NT) portForwardPrometheus() {
 		prometheusNamespace,
 		prometheusServerDeploymentName,
 		fmt.Sprintf(":%d", prometheusServerPort),
+	)
+	nt.startPortForwarder(
+		prometheusNamespace,
+		prometheusServerDeploymentName,
+		nt.prometheusPortForwarder,
 	)
 }
 
--- a/e2e/nomostest/portforwarder/port_forwarder.go
+++ b/e2e/nomostest/portforwarder/port_forwarder.go
@@ -31,6 +31,34 @@ import (
 	"kpt.dev/configsync/pkg/kinds"
 )
 
+var errTerminal = errors.New("terminal PortForwarder error")
+
+func isTerminalError(err error) bool {
+	return errors.Is(err, errTerminal)
+}
+
+func newTerminalError(format string, args ...interface{}) error {
+	return errors.Wrapf(errTerminal, format, args...)
+}
+
+// EventType is a type of event sent by PortForwarder
+type EventType string
+
+const (
+	// ErrorEvent indicates a fatal error occurred which caused PortForwarder to exit
+	ErrorEvent EventType = "ERROR"
+	// InitEvent indicates PortForwarder was started and is ready
+	InitEvent EventType = "INIT"
+	// DoneEvent indicates PortForwarder exited gracefully
+	DoneEvent EventType = "DONE"
+)
+
+// Event is a signal that PortForwarder emits on its event channel
+type Event struct {
+	Type    EventType
+	Message string
+}
+
 // PortForwarder creates a port forwarding from localhost to a Deployment in the
 // cluster. Assumes the Deployment has a single Pod, and keeps the port forward
 // open if the Pod is deleted.
@@ -54,9 +82,9 @@ type PortForwarder struct {
 	deployment string
 	// port is the port on the Pod to forward to
 	port string
-	// setPortCallback is a callback which will be invoked whenever the Pod changes and the
+	// onReadyCallback is a callback which will be invoked whenever the Pod changes and the
 	// port-forward is recreated.
-	setPortCallback func(int, string)
+	onReadyCallback func(int, string)
 	// mux is a mutex to synchronize getting/setting the port-forward value
 	mux sync.Mutex
 	// localPort is the local port which forwards to the Pod
@@ -65,22 +93,27 @@ type PortForwarder struct {
 	pod string
 	// cmd is the port forward subprocess
 	cmd *exec.Cmd
+	// isReady indicates whether the port forward is currently ready
+	isReady bool
+	// isStarted indicates whether the PortForwarder has been started
+	isStarted bool
 }
 
 // PortForwardOpt is an optional parameter for PortForwarder
 type PortForwardOpt func(pf *PortForwarder)
 
-// WithSetPortCallback registers a callback that will be invoked whenever the
+// WithOnReadyCallback registers a callback that will be invoked whenever the
 // deployment's underlying pod changes and the port-forward is recreated.
-func WithSetPortCallback(setFn func(int, string)) PortForwardOpt {
+// The callback must not invoke LocalPort from inside the callback, as this will
+// lead to a deadlock.
+func WithOnReadyCallback(onReadyFunc func(int, string)) PortForwardOpt {
 	return func(pf *PortForwarder) {
-		pf.setPortCallback = setFn
+		pf.onReadyCallback = onReadyFunc
 	}
 }
 
 // NewPortForwarder creates a new instance of a PortForwarder
 func NewPortForwarder(
-	ctx context.Context,
 	kubeClient *testkubeclient.KubeClient,
 	watcher *testwatcher.Watcher,
 	logger *testlogger.TestLogger,
@@ -92,7 +125,6 @@ func NewPortForwarder(
 		kubeClient:     kubeClient,
 		watcher:        watcher,
 		logger:         logger,
-		ctx:            ctx,
 		kubeConfigPath: kubeConfigPath,
 		ns:             ns,
 		deployment:     deployment,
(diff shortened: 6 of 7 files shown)


e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -254,3 +259,37 @@ func (pf *PortForwarder) start() error {
...
+func (pf *PortForwarder) signalOnFailure(fns ...func() error) chan struct{} {
+	restartChan := make(chan struct{})
+	// signalled indicates whether a signal has been sent
+	var signalled bool
+	var mux sync.Mutex
+	for _, fn := range fns {
+		fn := fn
+		go func() {
+			err := fn()
+			mux.Lock()
```
karlkfi, line 275: wont this block the 2nd function from starting until the 1st function is done?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,14 +223,29 @@ func (pf *PortForwarder) start() error {
...
+				func() error {
+					return pf.cmd.Wait()
+				},
+				func() error {
+					return pf.watcher.WatchForNotFound(kinds.Pod(), pf.pod, pf.ns,
+						testwatcher.WatchTimeout(0), testwatcher.WatchContext(watchCtx))
+				},
+			)
+			select {
+			case <-pf.ctx.Done():
```
karlkfi, line 242: why not just put this blocking on ctx.Done() as one of the parallel functions?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,14 +223,29 @@ func (pf *PortForwarder) start() error {
...
-				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-					// stop retrying, the subprocess will be killed with the context.
-					return
-				}
-				// Watch automatically retries, so this shouldn't happen unless
-				// there's a terminal error.
-				pf.logger.Infof("Watch exited, restarting port-forward for %s/%s: %v", pf.ns, pf.deployment, err)
+			// create an inner context for the watch so we can cancel threads between iterations
+			// the subprocess will be killed inside portForwardToDeployment
+			watchCtx, watchCancel := context.WithCancel(pf.ctx)
```
karlkfi, line 228: always need a `defer cancel()` immediately after creating one.

karlkfi:
I like where this is going, but I think it needs some work.

It might be better to go with https://github.com/GoogleContainerTools/kpt-config-sync/pull/532 first. It adds tracking of readiness, which is a combination of process still running and watch still running. Then you can add restart on unready afterwards.

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: No, because the assumption is that the `fn` function should be blocking. The mux stops them from entering the signalling section at the same time

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: I'd need to put this in another nested function to use `defer`, but yeah that might be cleaner

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: We would need another signal to tell it to exit instead of restart (e.g. `doneChan`). At that point we're creating another signal which is pretty much just a proxy for `ctx.Done`, so I feel that may just be making things more complicated

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +223,80 @@ func (pf *PortForwarder) start() error {
...
+		func() error {
+			return pf.cmd.Wait()
+		},
+		func() error {
+			return pf.watcher.WatchForNotFound(kinds.Pod(), pf.pod, pf.ns,
+				testwatcher.WatchTimeout(0), testwatcher.WatchContext(watchCtx))
+		},
+	)
+	select {
+	case <-pf.ctx.Done():
```
karlkfi, line 255: can this also just be added to the signalOnFailure list? 
```
func() error {
   <-watchCtx.Done()
   return watchCtx.Err()
}
```

Then we only need to wait for <-restartChan (maybe rename doneCh) here, right? And we can check if the error is DeadlineExceeded or Cancelled to decide if we want to restart or not.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +223,80 @@ func (pf *PortForwarder) start() error {
...
+			defer func() {
+				signalled = true
+				mux.Unlock()
+			}()
+			if signalled {
+				return
+			}
+			select {
+			case <-pf.ctx.Done():
+				// if the parent context closes, we no longer need to restart
```
karlkfi, line 292: Looks like the restartChan is never closed if the context is done.
There's probably a clearer way to do this.

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +223,77 @@ func (pf *PortForwarder) start() error {
...
+		// tries to use the local port.
+		return err
+	}
+	return nil
+}
+
+// signalOnFailure returns a chan which will be closed if any of the provided
+// functions return. The functions should block and only return if a failure
+// has occurred and the port-forward must be restarted
+func (pf *PortForwarder) signalOnFailure(fns ...func() error) chan error {
```
karlkfi, line 276: I spent way too long trying to figure understand how this works.
I think it mostly comes down to not using the conventional language.
But when I finally figured it out, I realized there was a better way.

The return channel what I would call a `doneCh`, indicating that the goroutines it's waiting for are done. `restartChan` is confusing, because this the `signalOnFailure` func doesn't handle restarting anything, and it can be closed even if restart is not required (e.g. context cancelled).

With that renamed, it follows that the boolean should be named `closed` or `done`. Yes, it's a signal to the caller, but it's not a signal within the context of this method, because there's no reader here, only writers. Then you'd end up with a channel and boolean with similar names, which is the smell that might have lead discovery of a simpler way, because...

When you're closing a channel, it's generally recommended to only close it on a single goroutine, not with a mutex and locking on each goroutine. This can be accomplished by having the goroutines send a message to an intermediate channel and spawning another goroutine to block waiting for a message on the intermediate channel.

The next question becomes, who should close the channels? The answer is always "the writer". And that's another tip: you should only have one writer for each channel, otherwise there's no way to cleanly close it, because you can't write to a closed channel. From there it follows that we need an intermediate channel for each goroutine, and a dedicated goroutine to close the channel we returned to the caller.

The fan-in aggregation goroutine also needs to wait until all the other goroutines are done, otherwise we could end up with memory leaks.

The nest question is "how do we block waiting for one of the channels in a list of channels?". The answer is the [Select](https://golang.org/pkg/reflect/#Select) function from the [reflect](http://golang.org/pkg/reflect/) package! https://stackoverflow.com/a/19992525 Block waiting for ONE of them to finish, then signal them all to stop (cancel()), then block waiting for ALL of them to finish, then close the `doneCh` you gave to the caller.

Do all that and you can replace the mutex/bool with channels, avoid memory leaks, and make it all easier to read.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +223,77 @@ func (pf *PortForwarder) start() error {
...
+// no longer healthy. If that occurs, it will restart the port-forward
+func (pf *PortForwarder) waitAndRestart() error {
+	// create an inner context for the watch so we can cancel threads between iterations
+	// the subprocess will be killed inside portForwardToDeployment
+	watchCtx, watchCancel := context.WithCancel(pf.ctx)
+	defer watchCancel()
+	// restart the port-forward if:
+	// - the subprocess is killed
+	// - the pod is restarted
+	restartChan := pf.signalOnFailure(
```
karlkfi, line 245: The name "signalOnFailure" doesn't make sense here. Both Wait and Watch can return nil errors, which both need to trigger restart. So you're actually waiting for any one of these functions to return. And then once one of them returns you want to cancel the others. You'll need to pass a context to each function so they can be cancelled in a self-contained way without side effects. Then you can push the `WithCancel` setup into the `signalOnFailure` method. Then it makes sense to rename the method to something like `waitForAny` or even `race`, since it's a controlled race with only one winner.

karlkfi (changes requested):

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: Oh, the more obvious smell i missed is mixing channels and locks. 

If you're using channels you can probably do it all with channels, but it might require multiple. 
If you're using mutexes, you can probably do it all with a single lock (sync.Mutex or sync.RWMutex or sync.Cond).
Usually go with channels unless there's a good reason you can't.

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: The "failure" I was intending to refer to here is the higher level failure of either:
- the subprocess exiting, i.e. "failing"
- the pod exiting, i.e. "failing"

The way we detect those failures is using blocking `Wait` statements, so I can understand the confusion. We can pass a context to `Watch` (which we do), however AFAIK we can't pass a context to `cmd.Wait()`. I think we could work around this by creating a cancellable context for the underlying `Command` and cancelling that here, which would ultimately lead to `cmd.Wait()` returning. I'll take a pass at refactoring

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: I admittedly have much more experience with multi-threaded programming in C/C++ than Go, so I'm not as familiar with conventions around channels. I'll take a stab at rewriting this with your suggestions in mind

sdowell: Channel-ification done. PTAL

sdowell:
```
    custom_resources_test.go:257: ERROR: timed out waiting for metrics predicate[0]: no results from prometheus query: config_sync_resource_conflicts_total{commit="3d3819069a854ba77eea54573c684c4f9ed7c0c9", component="otel-collector", k8s_pod_name="root-reconciler-d5c976ddf-h9b29"}
```
seems to be a metrics flake
/retest

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
-				pf.logger.Infof("Watch exited, restarting port-forward for %s/%s: %v", pf.ns, pf.deployment, err)
-			}
-			if err := pf.portForwardToDeployment(true); err != nil {
-				pf.logger.Info(err)
-				// this is a potentially fatal error since we failed to restart the port
-				// forward. we log the error and return in case the test is already done
-				// with the port forward. the test will otherwise error the next time it
-				// tries to use the local port.
+			if err := pf.waitAndRestart(); err != nil {
+				pf.logger.Infof("PortForwarder finished: %v", err)
```
karlkfi, line 218: Prolly want to log this even if there's not an error? Or does it always error? If so, you dont need the if...

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
+}
+
+// functionWithChannel takes a blocking function and invokes it asynchronously in a
+// goroutine. A channel is returned which will be written to when the blocking
+// function returns.
+func functionWithChannel(fn func() error) chan error {
+	doneCh := make(chan error)
+	go func() {
+		err := fn()
+		doneCh <- err
```
karlkfi, line 304: nit: `doneCh <- fn()`

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
+	case err := <-podChan:
+		return err
+	}
+}
+
+// functionWithChannel takes a blocking function and invokes it asynchronously in a
+// goroutine. A channel is returned which will be written to when the blocking
+// function returns.
+func functionWithChannel(fn func() error) chan error {
+	doneCh := make(chan error)
```
karlkfi, line 301: nit: "doneCh" is usually used for `chan struct{}` that just close when done. This is more of an errorCh or resultCh, since it also sends a nil when there's no error. Speaking of which... should it? Or does that need an `if err != nil` check?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
+// waitAndRestart will block until seeing an indicator that the port-forward is
+// no longer healthy. If that occurs, it will restart the port-forward
+func (pf *PortForwarder) waitAndRestart() error {
+	err := pf.waitForFailure()
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return fmt.Errorf("exiting - PortForwarder for pod %s/%s received done signal", pf.ns, pf.pod)
+	}
+	pf.logger.Infof("received signal for port-forward for %s/%s. err: %v", pf.ns, pf.pod, err)
+	if err := pf.portForwardToDeployment(true); err != nil {
+		// this is a potentially fatal error since we failed to restart the port
```
karlkfi, line 235: Should we retry if the port-forward errors?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
+// - the pod is restarted
+func (pf *PortForwarder) waitForFailure() error {
+	ctx, cancel := context.WithCancel(pf.ctx)
+	cmdChan := functionWithChannel(func() error {
+		// cmd.Wait does not accept a context, so create a wrapper that cancels with the ctx
+		cmdWaitChan := functionWithChannel(pf.cmd.Wait)
+		var chanErr error
+		select {
+		case <-ctx.Done():
+			chanErr = ctx.Err()
```
karlkfi, line 256: comment: stop waiting for the command to exit so we can kill it.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
+	return nil
+}
+
+// waitForFailure waits for any failure condition which indicates the port-forward
+// must be restarted. The port-forward must be restarted if:
+// - the subprocess is killed
+// - the pod is restarted
+func (pf *PortForwarder) waitForFailure() error {
+	ctx, cancel := context.WithCancel(pf.ctx)
+	cmdChan := functionWithChannel(func() error {
```
karlkfi, line 250: The functionWithChannel is cute, but these anon functions might be easier to read if they were methods or package functions with names.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
+		var chanErr error
+		select {
+		case <-ctx.Done():
+			chanErr = ctx.Err()
+		case err := <-cmdWaitChan:
+			chanErr = err
+		}
+		// kill the subprocess. This should force the cmd.Wait goroutine to return, if it hasn't already.
+		if err := pf.cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) {
+			// potentially bad if this fails, cmd.Wait may never return?
```
karlkfi, line 262: This should probably return the kill error. If we fail to kill a process we need to stop the tests. If we exit the test process, it won't matter that the wait was still running in a goroutine.

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: nevermind. this is fine.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
+			pf.logger.Infof("stopped port-forward process for %s/%s", pf.ns, pf.pod)
+		}
+		// wait for cmdWaitChan to close, i.e. cmd.Wait returns
+		<-cmdWaitChan
+		return chanErr
+	})
+	podChan := functionWithChannel(func() error {
+		err := pf.watcher.WatchForNotFound(kinds.Pod(), pf.pod, pf.ns,
+			testwatcher.WatchTimeout(0), testwatcher.WatchContext(ctx))
+		// We're using WatchForNotFound to detect if the Pod is deleted, but a nil
```
karlkfi, line 274: nit: this comment is mildly confusing. What we need is a function description. So maybe just move it up and say "Wait for the Pod to be deleted, then return an error to trigger port-forward restart"

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: unless we shouldn't send a nil error on the channel?

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: The question is - if this fails, do we still wait for the goroutine to exit? AFAICT this would be an unexpected error which means we may not be able to kill the subprocess, which means we may be waiting forever.

If we do wait for `cmd.Wait` to exit, this means the subprocess did exit and we probably want to return the original `chanErr` so that we can proceed as normal.

If we don't wait for `cmd.Wait` to exit, then we could return the `Kill` error but we may be leaking the goroutine

We could potentially do something like `panic` or `Fatal`. WDYT?

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: I think it's okay to send a `nil` error, the intent is just to pass through whatever error was returned by the function

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: Maybe, although I think we might have to put some thought into how we recover from this if we fail during the port-forward setup. This is an unexpected scenario that shouldn't generally happen, so I'd be inclined to leave this as-is and take a closer look if we see this pop up as a failure mode.

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: renamed to `errorCh`

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,95 @@ func (pf *PortForwarder) start() error {
...
 	}()
 	return nil
 }
+
+// waitAndRestart will block until seeing an indicator that the port-forward is
+// no longer healthy. If that occurs, it will restart the port-forward
+func (pf *PortForwarder) waitAndRestart() error {
+	err := pf.waitForFailure()
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return fmt.Errorf("exiting - PortForwarder for pod %s/%s received done signal", pf.ns, pf.pod)
```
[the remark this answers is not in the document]
sdowell: done

sdowell:
```
    new.go:440: ERROR: timed out waiting for metrics predicate[0]: no results from prometheus query: config_sync_last_sync_timestamp{commit="7ab83d567477a90a95ed777af490550444140a27", component="otel-collector", k8s_pod_name="ns-reconciler-test-ns-rs-test-7-db56b4797-66trq", status="success"}
```
/retest

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: Fair enough.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,107 @@ func (pf *PortForwarder) start() error {
...
+func (pf *PortForwarder) waitForCmdFailure(ctx context.Context) error {
+	// cmd.Wait does not accept a context, so create a wrapper that cancels with the ctx
+	cmdWaitChan := wrapWithChannel(pf.cmd.Wait)
+	var chanErr error
+	select {
+	case <-ctx.Done():
+		// stop waiting for the command to exit, so we can kill it
+		chanErr = ctx.Err()
+	case err := <-cmdWaitChan:
+		chanErr = err
```
karlkfi, line 283: I think if the cmd exits, you can just return here. You don't need to kill it or wait for it to exit any more.

sdowell: fair point, done

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,107 @@ func (pf *PortForwarder) start() error {
...
 }
+
+// waitAndRestart will block until seeing an indicator that the port-forward is
+// no longer healthy. If that occurs, it will restart the port-forward
+func (pf *PortForwarder) waitAndRestart() error {
+	err := pf.waitForFailure()
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return errors.Wrap(err, fmt.Sprintf("exiting - PortForwarder for pod %s/%s received done signal", pf.ns, pf.pod))
+	}
+	pf.logger.Infof("received signal for port-forward for %s/%s. err: %v", pf.ns, pf.pod, err)
```
karlkfi, line 233: "signal" is a bit ambiguous to people who've never seen this code, especially if the error in nil. 
How about using different log msg for err vs nil?
Since we're not waiting for the watch and command to exit, and the watch always returns an error, we can be more specific:
`port-forward exited for pod %s/%s: %v`
else
`port-forward exited without error for pod %s/%s`

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -233,24 +214,107 @@ func (pf *PortForwarder) start() error {
...
 	}()
 	return nil
 }
+
+// waitAndRestart will block until seeing an indicator that the port-forward is
+// no longer healthy. If that occurs, it will restart the port-forward
+func (pf *PortForwarder) waitAndRestart() error {
+	err := pf.waitForFailure()
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return errors.Wrap(err, fmt.Sprintf("exiting - PortForwarder for pod %s/%s received done signal", pf.ns, pf.pod))
```
karlkfi, line 231: don't need Sprintf here. use Wrapf.

e2e/nomostest/gitproviders/local.go
```diff
@@ -75,3 +68,8 @@ func (l *LocalProvider) DeleteRepositories(...string) error {
 func (l *LocalProvider) DeleteObsoleteRepos() error {
 	return nil
 }
+
+// WithPortForwarder sets the PortForwarder for the in-cluster git server
+func (l *LocalProvider) WithPortForwarder(pf *portforwarder.PortForwarder) {
```
karlkfi, line 73: With- is usually used for options and builders. This is just a setter. But we usually avoid setters in Go and just make the field public.

karlkfi: The general rule is that all public fields need to be set by the caller who constructs the struct instance, or at least before other method use.

sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -190,67 +196,155 @@ func (pf *PortForwarder) portForwardToPod(pod string, async bool) error {
...
+// Returns a channel. An error will be sent to the channel, and the channel will
+// be closed, when the PortForwarder stops.
+func (pf *PortForwarder) Start() chan error {
+	pf.mux.Lock()
+	defer pf.mux.Unlock()
+	// defend against the PortForwarder being started more than once
+	if pf.isStarted {
+		panic("PortForwarder already started")
 	}
+	pf.isStarted = true
```
karlkfi, line 235: Do we want to wait for ready wait here?

In the k8s informer, for example, there's a pattern of waiting until "synced", which in the informer means the List has completed and returned a ResourceVersion. That indicates it's safe to make changes because even if the Watch hasn't started yet, when it starts it will start at the ResourceVersion, so you won't miss any events.

Here, we could block until ready before returning from Start, and that would guarantee that the port-forwarder has come up and the callback was called at least once. It would help guard against race conditions by obviating the need to call InitGitRepos outside the callback, which is inside the PortForwarder lock.

Another option would be to expose a way to wait for ready, either with a WaitForReady func (maybe with ctx cancel to stop waiting) or a ReadyChan method that returns a readyCh which is closed when the PF is next ready (or already closed if it's already ready). I'm also open to adding a new field called `initCh` and having it closed after the first time ready == true, but for that we probably still need a `readyCh`.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -190,67 +196,155 @@ func (pf *PortForwarder) portForwardToPod(pod string, async bool) error {
...
+			} else if err != nil && isTerminalError(err) {
+				err = errors.Wrapf(err, "PortForwarder for Pod %s/%s encountered a fatal error", pf.ns, pf.pod)
 				pf.logger.Info(err)
-				// this is a potentially fatal error since we failed to restart the port
-				// forward. we log the error and return in case the test is already done
-				// with the port forward. the test will otherwise error the next time it
-				// tries to use the local port.
-				return
+				return err
+			} else if err != nil {
```
karlkfi, line 247: Can you pull this `if err != nil` out and make 2 layers of ifs? I think it would be more readable than this switch-like flow.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -110,32 +124,34 @@ func NewPortForwarder(
...
-		pf.mux.Unlock()
-	}
+	pf.isReady = true
+	pf.mux.Unlock()
 	if pf.setPortCallback != nil {
 		pf.setPortCallback(port, pod)
 	}
 }
 
+func (pf *PortForwarder) notReady() {
```
karlkfi, line 146: I think you can just re-use the above `update` func and pass a ready arg.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -190,67 +196,155 @@ func (pf *PortForwarder) portForwardToPod(pod string, async bool) error {
...
+	podChan := wrapWithChannel(func() error {
+		return pf.waitForPodFailure(ctx)
+	})
+	defer func() {
+		// before we return, cancel and wait for goroutines to exit
+		cancel()
+		// If any channel returns a TerminalError, return it instead of original error.
+		// This informs upstream logic that a fatal error has occurred, and we must exit the retry loop.
+		cmdErr, ok := <-cmdChan
+		if ok && isTerminalError(cmdErr) {
```
karlkfi, line 285: I think you lost the checks for Cancelled and DeadlineExceeded here. Or are they wrapped as terminal somewhere i missed?

karlkfi (changes requested):
Looking better!

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -31,6 +31,16 @@ import (
 	"kpt.dev/configsync/pkg/kinds"
 )
 
+var errTerminal = errors.New("terminal PortForwarder error")
```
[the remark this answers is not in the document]
sdowell: I initially was going to reuse that one, but I decided to scope this error to be private to this package so that consumers don't need to worry about this error. All of the checking for terminal errors is done within the package

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: We don't want to invoke the callback here, or necessarily need to set any of the other parameters. I think we would be bloating the `update` function if we did this

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: The check is done in the top level retry loop

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -190,67 +196,155 @@ func (pf *PortForwarder) portForwardToPod(pod string, async bool) error {
...
-	if err := pf.portForwardToDeployment(false); err != nil {
-		return err
+// Once ctx is cancelled, the port forward will be closed and will no longer be recreated.
+// Returns a channel. An error will be sent to the channel, and the channel will
+// be closed, when the PortForwarder stops.
+func (pf *PortForwarder) Start() chan error {
+	pf.mux.Lock()
+	defer pf.mux.Unlock()
+	// defend against the PortForwarder being started more than once
+	if pf.isStarted {
```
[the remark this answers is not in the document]
sdowell: We didn't have to do this, it was a decision I made to change the `PortForwarder` to be scoped to a testcase rather than the entire test suite. It was scoped to the test case originally, and the other `PortForwarder`s are also scoped to the testcase.

It can work either way, but the downside to scoping it to the entire test suite is that if the `PortForwarder` becomes unhealthy it will fail all subsequent tests

e2e/nomostest/new.go
```diff
@@ -369,6 +367,11 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 	}
 
 	if *e2e.GitProvider == e2e.Local {
+		provider := nt.GitProvider.(*gitproviders.LocalProvider)
+		nt.T.Cleanup(func() {
+			provider.PortForwarder = nil
+		})
+		provider.PortForwarder = nt.portForwardGitServer()
 		InitGitRepos(nt, allRepos...)
```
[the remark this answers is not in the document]
sdowell: I like this suggestion, I'll see what I can make of this

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -110,32 +124,34 @@ func NewPortForwarder(
...
-		err := pf.start()
-		return pf.localPort, err
+	if !pf.isReady {
+		return pf.localPort, fmt.Errorf("PortForwarder for Deployment %s/%s is not ready", pf.ns, pf.deployment)
 	}
 	return pf.localPort, nil
 }
 
-func (pf *PortForwarder) setPort(cmd *exec.Cmd, port int, pod string, async bool) {
+func (pf *PortForwarder) setPort(cmd *exec.Cmd, port int, pod string) {
```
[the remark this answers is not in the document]
sdowell: Makes sense, I'll _update_ this

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: > which is inside the PortForwarder lock.

Oops. The callback is NOT in the lock, because it calls LocalPort indirectly. But I think that might be ok, as long as we wait for callback completion, rather than just the first ready=true. I guess that means we need to synchronously re-lock after the callback and close the initCh (with sync.Once maybe, since we can't re-close it). Then the Start can block on <-initCh or return the initCh so the caller can.

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: sure

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: > Then the Start can block on <-initCh or return the initCh so the caller can.

Returning the initCh from Start probably the better option. That way Start doesn't also need to take another context to stop waiting if something happens. The caller can then decide if they want to timeout or do something else async, without needing that complexity in Start.

I guess that means Start is returning an initCh and an errorCh... that makes it seem like maybe they should be combined into an eventCh with an Event struct that can pass InitEvent and ErrorEvent and close when fully done, after being stopped/cancelled/terminated.

e2e/nomostest/portforwarder/port_forwarder.go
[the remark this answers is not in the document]
karlkfi: > if the PortForwarder becomes unhealthy it will fail all subsequent tests

This seems fine. We're making it not flakey :P

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: You can just wrap the callback in `if ready`, but :shrug:

e2e/nomostest/portforwarder/port_forwarder.go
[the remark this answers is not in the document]
sdowell: If we revert this, we also won't be able to consolidate `InitGitRepos` - because that needs to be done in `setupTestCase`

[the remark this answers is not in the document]
sdowell: Note this is kind of tangential from the `isStarted` logic that I added here. The `isStarated` logic is just to help protect developers from calling `Start` more than once, because that would break the assumption of a single goroutine running for a `PortForwarder`. We would probably still want this regardless of whether we are scoping to a `testcase` or to the whole test suite. The reason this is introduced now is because we are no longer lazily starting, and we've exposed a public `Start` method.

With the logic written as it is now, it wouldn't call `Start` more than once for a `PortForwarder` - but IMO it's good to fail fast if someone accidentally introduces a bug which causes `Start` to be called more than once

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: Oh, I see what's happening. I was confused because this defer is above the select.

If a terminal error is received here, it REPLACES the `err` returned by the select block. Otherwise, if a 2nd error was received, it seems to be ignored in favor of the first err (which could be nil, if Cmd.Wait returns nil).

I guess we do want to ignore Cancelled here, since one being completed will cancel the other. But I'm not sure if we want to ignore other non-terminal non-cancel errors... We probably at least want to log them?

sdowell: The intention here is just to block until any error happens which requires us to restart, i.e. a race where we don't care who wins. However if any terminal errors occur, we exit instead of restarting - hence we return the terminal error instead.

I think we could log other errors as well (errors other than cancel or terminal), although I feel it might just be unneeded noise since all we really care about is that we need to restart the port-forward

e2e/nomostest/new.go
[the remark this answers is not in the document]
sdowell: Consolidated `InitGitRepos` into the callback. Note that this had some knock-on effects, because:
- the way `ResetRepository` was written assumed that the port forward was ready, because it pushes an initial commit
- the callback assumes `RemoteRepositories` has been set, which was also handled by the `ResetRepository` setup after the port-forward setup

I was previously working around this by returning early in the callback for the first iteration via the `prevPodName == ""` check

To work around this with the consolidation I split up `ResetRepository` into two helper functions - one to `init` the repo and one to push the initial commit. This way we can set up the repositories before starting the port forward and waiting for it to be ready, and push the initial commits after the port forward is ready. I played with a few different approaches to solve this problem, and I preferred this approach as it minimizes the differences in how we handle the `LocalProvider` vs other git providers. Let me know what you think

e2e/nomostest/portforwarder/port_forwarder.go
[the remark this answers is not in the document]
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: Done. It now waits for a ready event from the channel

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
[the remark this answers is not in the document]
karlkfi: The only reason to expose it, I think, would be if someone wanted to use it in the callback, right?
Or are callback errors always terminal?

e2e/nomostest/gitproviders/repository.go
```diff
@@ -554,7 +549,11 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) error {
 		return err
 	}
 	g.Logger.Infof("[repo %s] pushing commit: %s", path.Base(g.Root), commit)
-	return g.Push(branch, "-f")
+	if err := g.Push(branch, "-f"); err != nil {
+		return err
+	}
+	g.isEmpty = false
```
karlkfi, line 555: `isEmpty = false` -> `initialized = true`

The description on the field is technically incorrect too. It's not whether there are any commits (because there might be some pre-exiting, when attaching to an existing remote repo). It's actually just whether `Init` has been called yet, which force-pushes on top of any optionally existing commits to "reset" the contents of the repo to the default contents.

e2e/nomostest/gitproviders/repository.go
```diff
@@ -581,9 +579,21 @@ func (g *Repository) Push(args ...string) error {
 	return nil
 }
 
+func (g *Repository) push(remoteURL string, args ...string) error {
```
karlkfi, line 582: Why pull out an inner push method?
Don't we want to retry when calling PushAllBranches?

e2e/nomostest/gitproviders/git-provider.go
```diff
@@ -44,29 +43,8 @@ type GitProvider interface {
...
-	}
-}
-
 // NewGitProvider creates a GitProvider for the specific provider type.
-func NewGitProvider(t testing.NTB, provider string, opts ...GitProviderOpt) GitProvider {
-	options := GitProviderOpts{}
-	for _, opt := range opts {
-		opt(&options)
-	}
+func NewGitProvider(t testing.NTB, provider string) GitProvider {
```
karlkfi, line 47: This isn't a constructor. It's a test specific initializer. Should be a method on NT. But I'll fix it later.

e2e/nomostest/nt.go
```diff
@@ -596,6 +593,40 @@ func (nt *NT) newPortForwarder(ns, deployment, port string, opts ...portforwarde
...
+		nt.T.Logf("killing port-forward %s/%s", ns, deployment)
+		cancel()
+		<-doneCh
+		nt.T.Logf("finished waiting for port-forward %s/%s to exit", ns, deployment)
+	})
+	go func() {
+		defer close(doneCh)
+		// first event should be Init. signal if it's ready
+		event := <-eventCh
+		if event.Type == portforwarder.InitEvent {
```
karlkfi, line 611: nit: Minor simplification: You could just do this in the for loop with a switch. It would panic on close if you received init twice.

e2e/nomostest/nt.go
```diff
@@ -596,6 +593,40 @@ func (nt *NT) newPortForwarder(ns, deployment, port string, opts ...portforwarde
...
+			if event.Type == portforwarder.ErrorEvent {
+				nt.T.Error(event.Message)
+			}
+		}
+	}()
+	timeoutCtx, timeoutCancel := context.WithTimeout(ctx, nt.DefaultWaitTimeout)
+	defer timeoutCancel()
+	select {
+	case <-timeoutCtx.Done():
+		nt.T.Fatalf("PortForwarder for %s/%s not ready: %v", ns, deployment, timeoutCtx.Err())
```
karlkfi, line 624: The error won't include the duration. So you probably want to say something like `Timed out after %v waiting for  PortForwarder for %s/%s to become ready`.

e2e/nomostest/nt.go
```diff
@@ -606,15 +637,21 @@ func (nt *NT) portForwardOtelCollector() {
...
-func (nt *NT) portForwardGitServer() *portforwarder.PortForwarder {
+func (nt *NT) portForwardGitServer() {
 	nt.T.Helper()
+	provider := nt.GitProvider.(*gitproviders.LocalProvider)
 	prevPodName := ""
 	resetGitRepos := func(newPort int, podName string) {
 		// pod unchanged, don't reset
-		if prevPodName == "" || prevPodName == podName {
+		if prevPodName == podName {
 			prevPodName = podName
```
karlkfi, line 655: nit: this line seems redundant.

e2e/nomostest/nt.go
```diff
@@ -625,28 +662,32 @@ func (nt *NT) portForwardGitServer() *portforwarder.PortForwarder {
...
+				nt.T.Fatal(err)
+			}
+			if err := nt.RemoteRepositories[repo].PushAllBranches(remoteURL); err != nil {
 				nt.T.Fatal(err)
 			}
 		}
 		prevPodName = podName
 	}
-	return nt.newPortForwarder(
+	nt.T.Cleanup(func() {
```
karlkfi, line 677: comment: // clear PortFowarder after each test, so it gets recreated in test setup.

Is this actually necessary or does test setup replace it anyway?

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -65,22 +93,27 @@ type PortForwarder struct {
...
 // PortForwardOpt is an optional parameter for PortForwarder
 type PortForwardOpt func(pf *PortForwarder)
 
-// WithSetPortCallback registers a callback that will be invoked whenever the
+// WithOnReadyCallback registers a callback that will be invoked whenever the
 // deployment's underlying pod changes and the port-forward is recreated.
-func WithSetPortCallback(setFn func(int, string)) PortForwardOpt {
+// The callback must not invoke LocalPort from inside the callback, as this will
+// lead to a deadlock.
+func WithOnReadyCallback(setFn func(int, string)) PortForwardOpt {
```
karlkfi, line 109: setFn -> onReady or onReadyFn or onReadyFunc or just fn

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -31,6 +31,34 @@ import (
...
+// EventType is a type of event sent by PortForwarder
+type EventType string
+
+const (
+	// ErrorEvent indicates a fatal error occurred which caused PortForwarder to exit
+	ErrorEvent EventType = "ERROR"
+	// InitEvent indicates PortForwarder was started and is ready
+	InitEvent EventType = "INIT"
+	// FinishedEvent indicates PortForwarder exited gracefully
+	FinishedEvent EventType = "FINISHED"
```
karlkfi, line 53: Finished -> Done.

"Done" is conventional go lingo for finished/complete, explicitly including teardown. The pre-teardown convention is "Stopped". e.g. `Stop()` or closing `stopCh` flips the `stopping/stopped` bool, and `doneCh` is closed when teardown is complete.

I don't know if we need a Stopped event here, but we can always add once later if we do.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -190,67 +214,165 @@ func (pf *PortForwarder) portForwardToPod(pod string, async bool) error {
...
+		// stop waiting for the command to exit, so we can kill it
+		chanErr = ctx.Err()
+	case err := <-cmdWaitChan:
+		return err
+	}
+	// kill the subprocess. This should force the cmd.Wait goroutine to return, if it hasn't already.
+	if err := pf.cmd.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) {
+		return newTerminalError("failed to kill port-forward process: %v", err)
+	} else if err == nil {
+		pf.logger.Infof("stopped port-forward process for %s/%s", pf.ns, pf.pod)
```
karlkfi, line 348: Move this up above the Kill call and log it as `killing the port-forward process for %s/%s`.
This could be debug, but I think for now it's ok as Info, since we'll need to debug in CI if it fails.

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: I think we should log these, but it doesn't need to block this PR. We're gonna need all the info we can get debugging this if it fails. We can always remove noisy logs later.

e2e/nomostest/new.go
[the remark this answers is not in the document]
karlkfi: seems alright. 

I feel like there's more room to clean up by pushing down into the GitProvider abstraction, but we can do more cleanup later.

e2e/nomostest/gitproviders/repository.go
sdowell: This is so we can pass the `remote` with `PushAllBranches` in order to use the `port` value passed into the callback instead of calling the `LocalPort` method. This enables us to wrap the callback in the lock without causing a deadlock.

In this case the `RemoteURL` is static, so we don't retry for the `RemoteURL`. I could refactor this to retry for failures in `git push`, but I opted for just pushing without a retry

e2e/nomostest/portforwarder/port_forwarder.go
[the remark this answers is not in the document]
sdowell: The callback does not return an error. Errors within the callback call `Fatal`

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/portforwarder/port_forwarder.go
sdowell: done

e2e/nomostest/nt.go
sdowell: It's not strictly necessary but I thought it would be good form. At this point in the `Cleanup` stack the `PortForwarder` will have been stopped, so it can no longer be used. Added a comment

e2e/nomostest/nt.go
sdowell: removed

e2e/nomostest/nt.go
sdowell: Is there a way to get that from the context? Or are you suggesting I use `nt.DefaultWaitTimeout`?

e2e/nomostest/nt.go
```diff
@@ -594,6 +593,17 @@ func (nt *NT) newPortForwarder(ns, deployment, port string, opts ...portforwarde
 		ns, deployment, port,
 		opts...,
 	)
+	errorCh := pf.Start()
+	nt.T.Cleanup(func() {
```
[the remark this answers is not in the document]
sdowell: https://www.youtube.com/watch?v=aoyKxSYCzRA

e2e/nomostest/portforwarder/port_forwarder.go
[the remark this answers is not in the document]
karlkfi: We should probably make them return errors in a follow up PR, and define the expected behavior when the callback errors..

e2e/nomostest/nt.go
sdowell: done. evaluated using `time`

sdowell:
```
new.go:441: ERROR: timed out waiting for metrics predicate[0]: no results from prometheus query: config_sync_last_sync_timestamp{commit="f6235da2861dca72223cbd34294d0b1ea3da2d07", component="otel-collector", k8s_pod_name="root-reconciler-5fb6c947b5-v7t5s", status="success"}
```
metrics flake
/retest

karlkfi (approved):
/lgtm

Thanks for the hard work and many iterations!

PR #531, opened 2023-04-08
Fix e2e Watch to distinguish cancel vs timeout

karlkfi:


karlkfi:
/retest

Found a race condition in the Scheme, which isn't thread-safe. But it's not related to this change.

e2e/nomostest/testwatcher/watch.go
```diff
@@ -271,34 +271,36 @@ func (w *Watcher) WatchObject(gvk schema.GroupVersionKind, name, namespace strin
...
-					cause:  multierr.Combine(evalErrs...),
-					prefix: errPrefix,
-				}
+			// Until returns ErrWaitTimeout for any ctx.Done
+			// https://github.com/kubernetes/apimachinery/blob/v0.26.3/pkg/util/wait/wait.go#L594
+			// https://github.com/kubernetes/client-go/blob/v0.26.3/tools/watch/until.go#L89
+			// So use the context error instead, if done.
+			// Use errors.Is to detect DeadlineExceeded or Cancelled.
+			select {
+			case <-ctx.Done():
```
sdowell, line 280: Should we also check if `spec.timeout == 0` before switching this to the context err (i.e. the watch was configured without a timeout)?

karlkfi: Until returns ErrWaitTimeout even if the context was cancelled. So using `ctx.Err()` lets us distiguish between Cancelled and DeadlineExceeded. If there's no timeout configured, it can still be cancelled.

sdowell (approved):

PR #532, opened 2023-04-08
Add readiness tracking to PortForwarder

karlkfi:
- Refactor PortForwarder:
  - Add a ready field, which is toggled off when either the
    process exits or the pod is not found. Then only toggle it on
    after the port-forwarding is ready.
  - Use a subprocess struct to track the current port-forward
    command. This makes it easier to update atomically.
  - Replace setPort with updateSubprocess, and have it handle
    state changes, not just field changes.
  - Add updateSubprocessAsync to wrap updateSubprocess with the
    mutex, when necessary.
  - Wait for subprocess exit inside portForwardToPod, to handle
    early termination.
  - Wait for subprocess exit in a goroutine launched by
    updateSubprocess, after it becomes ready, to handle updating
    the readiness of the PortForwarder.
  - Refactor portForwardToPod and portForwardToDeployment to
    avoid needing to access fields protected by the mutex. This
    allows LocalPort and updateSubprocessAsync to manage the locks,
    without the subfunctions needing to know if they're being
    called async or not.
- Refactor Repository:
  - Change Push to accept only a vararg, making the refspec
    optional.
  - Rename pushAllToRemote to PushAllBranches and delegate to Push.
    This takes advantage of the retry loop in Push to avoid errors.
  - Increase the Push timeout to 5 minutes (from 1m) when using the
    LocalProvider, to allow enough time for the local git-server to
    be rescheduled on GKE Autopilot, when node spin-up is required.

### DIFF
--- a/e2e/nomostest/portforwarder/port_forwarder.go
+++ b/e2e/nomostest/portforwarder/port_forwarder.go
@@ -54,27 +54,35 @@ type PortForwarder struct {
 	deployment string
 	// port is the port on the Pod to forward to
 	port string
-	// setPortCallback is a callback which will be invoked whenever the Pod changes and the
-	// port-forward is recreated.
-	setPortCallback func(int, string)
-	// mux is a mutex to synchronize getting/setting the port-forward value
+	// onReadyFunc is a callback which will be invoked whenever the
+	// PortForwarder becomes ready.
+	onReadyFunc func(int, string)
+
+	// mux protects the subprocess and ready flag
 	mux sync.Mutex
-	// localPort is the local port which forwards to the Pod
-	localPort int
-	// pod is the name of the current Pod that is being forwarded to
-	pod string
+	// subprocess is a reference to the currently running port-forward command
+	subprocess *subprocess
+	// ready is true if the deployment is ready and the port-forward is running.
+	ready bool
+}
+
+type subprocess struct {
 	// cmd is the port forward subprocess
 	cmd *exec.Cmd
+	// pod is the name of the Pod being forwarded to
+	pod string
+	// localPort is the local port which forwards to the Pod
+	localPort int
 }
 
 // PortForwardOpt is an optional parameter for PortForwarder
 type PortForwardOpt func(pf *PortForwarder)
 
-// WithSetPortCallback registers a callback that will be invoked whenever the
-// deployment's underlying pod changes and the port-forward is recreated.
-func WithSetPortCallback(setFn func(int, string)) PortForwardOpt {
+// WithOnReady registers a callback that will be invoked whenever the
+// PortForwarder becomes ready.
+func WithOnReady(onReadyFunc func(int, string)) PortForwardOpt {
 	return func(pf *PortForwarder) {
-		pf.setPortCallback = setFn
+		pf.onReadyFunc = onReadyFunc
 	}
 }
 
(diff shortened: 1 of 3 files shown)


e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -110,32 +118,72 @@ func NewPortForwarder(
 func (pf *PortForwarder) LocalPort() (int, error) {
 	pf.mux.Lock()
 	defer pf.mux.Unlock()
-	if pf.localPort == 0 {
+	if pf.subprocess == nil {
 		err := pf.start()
-		return pf.localPort, err
+		return pf.subprocess.localPort, err
```
sdowell, line 123: looks like this could cause a null pointer exception if `start` fails to set the `subprocess`

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -110,32 +118,72 @@ func NewPortForwarder(
...
+// updateSubprocess handles changes to the subprocess
+func (pf *PortForwarder) updateSubprocess(proc *subprocess, ready bool) {
+	switch {
+	case ready:
+		// Subprocess become ready
+		pf.logger.Infof("updating port-forward localhost:%d -> %s:%d (ready: %v)", proc.localPort, proc.pod, pf.port, ready)
+		pf.subprocess = proc
+		pf.ready = true
+		if pf.onReadyFunc != nil {
+			pf.onReadyFunc(proc.localPort, proc.pod)
```
sdowell, line 148: This will cause a deadlock if we enter `updateSubprocess` with the lock, and the `onReadyFunc` calls `LocalPort` and tries to re-obtain the lock (which it currently does for the git PortForwarder).

I confirmed after testing this locally that this leads to a deadlock.

e2e/nomostest/portforwarder/port_forwarder.go
```diff
@@ -183,74 +235,92 @@ func (pf *PortForwarder) portForwardToPod(pod string, async bool) error {
...
 	})
 	if err != nil {
-		return err
+		return nil, err
 	}
+	// We can't stop cmd.Process.Wait(), but we can at least consume the error
+	// when it's done, to let its goroutine return.
+	// The caller can start their own wait, if they want.
+	go func() {
+		<-exitCh
```
sdowell, line 249: It looks like this error just gets swallowed, what is the purpose of this and the goroutine on line 206 which captures the error?

karlkfi: `portForwardToPod` only cares about premature exit now. It watches until the retry loop is done, and then doesn't care about any exit signal after that, but there's no way to stop the `cmd.Process.Wait()` call. So it spins up another goroutine to consume the error or nil from the channel, to allow both goroutines to return.

After `portForwardToPod` returns and `portForwardToDeployment` returns, `start` calls `updateSubprocess` which calls `waitAndUpdateSubprocess` in a goroutine. `waitAndUpdateSubprocess` calls `proc.cmd.Wait()` and blocks until the process has exited.

This complexity allows `portForwardToPod` to be ignorant of what happens to the process after it returns. It doesn't need to know how to update the subprocess on the PortForwarder, and it doesn't need to know how to run async or lock. It just returns and lets the caller figure that out.

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: Good catch. How about we launch the callback in a goroutine?

We need waitAndUpdateSubprocess to run in the background anyway, while the callback is running, in case the port-forward stops while the callback is running. 

It occurs to me, that if the callback fails, we want to keep going anyway, because the next callback might work. But for that to work, we also need to give the callback its own lock, to prevent multiple running simultaneously.

sdowell: My first iteration was to avoid calling `LocalPort` from inside the callback, and the second iteration was to invoke the callback outside of the lock. I think either of those approaches work without too much added complexity

sdowell: I think running the callback async would work but adds some unnecessary complexity

karlkfi: It is more complex to run in a goroutine, but it should allow us to refactor the callback to allow it to fail and retry with a new port-forward. Unfortunately, that requires refactoring Repository first to return errors, instead of failing the test.

e2e/nomostest/portforwarder/port_forwarder.go
karlkfi: Indeed. Added error check.

karlkfi:
/retest

unrelated metrics error:
```
=== CONT  TestDeleteRepoSync_Centralized_AndRepoSyncV1Alpha1
    namespace_repo_test.go:278: ERROR: timed out waiting for metrics predicate[0]: no results from prometheus query: config_sync_last_sync_timestamp{commit="1d874966f306d59a75cdf3ba26abf8e263d88e7b", component="otel-collector", k8s_pod_name="ns-reconciler-bookstore-67cc4cccd4-fq99l", status="success"}
E0410 20:28:02.421245   20685 retrywatcher.go:130] "Watch failed" err="context canceled"
```

karlkfi:
Obsolete. Replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/530

PR #537, opened 2023-04-11
Improved caCert validation

victorpras:
The reconciler-manager should validate that the Secret referred to by caCertSecretRef is well-formed. If the key named cert is not set in the secret, then the reconciler-manager should report an error to the user.

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -707,6 +707,9 @@ func (r *RepoSyncReconciler) validateGitSpec(ctx context.Context, rs *v1beta1.Re
 	if err := validate.GitSpec(rs.Spec.Git, rs); err != nil {
 		return err
 	}
+	if err := r.validateCACertSecret(ctx, rs.Namespace, v1beta1.GetSecretName(rs.Spec.Git.CACertSecretRef)); err != nil {
+		return err
+	}
 	return r.validateNamespaceSecret(ctx, rs, reconcilerName)
 }
 
@@ -736,7 +739,10 @@ func (r *RepoSyncReconciler) validateNamespaceSecret(ctx context.Context, repoSy
 		repoSync.Namespace,
 		r.client)
 	if err != nil {
-		return err
+		if apierrors.IsNotFound(err) {
+			return errors.Errorf("Secret %s not found: create one to allow client authentication", namespaceSecretName)
+		}
+		return errors.Wrapf(err, "Secret %s get failed", namespaceSecretName)
 	}
 	return validateSecretData(authType, secret)
 }
--- a/pkg/reconcilermanager/controllers/reconciler_base.go
+++ b/pkg/reconcilermanager/controllers/reconciler_base.go
@@ -556,3 +556,24 @@ func (r *reconcilerBase) isLastReconciled(nn types.NamespacedName, resourceVersi
 	}
 	return resourceVersion == lastReconciled
 }
+
+// validateCACertSecret verify that caCertSecretRef is well formed with a key named "cert"
+func (r *reconcilerBase) validateCACertSecret(ctx context.Context, namespace, caCertSecretRefName string) error {
+	if useCACert(caCertSecretRefName) {
+		secret, err := validateSecretExist(ctx,
+			caCertSecretRefName,
+			namespace,
+			r.client)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return errors.Errorf("Secret %s not found, create one to allow client connections with CA certificate", caCertSecretRefName)
+			}
+			return errors.Wrapf(err, "Secret %s get failed", caCertSecretRefName)
+		}
+		if _, ok := secret.Data[CACertSecretKey]; !ok {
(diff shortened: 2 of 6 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -748,6 +751,19 @@ func (r *RepoSyncReconciler) validateNamespaceName(namespaceName string) error {
 	return nil
 }
 
+func (r *RepoSyncReconciler) validateCACertSecret(ctx context.Context, repoSync *v1beta1.RepoSync) error {
```
sdowell, line 754: Need unit tests for both reposync_controller and rootsync_controller. This should include test cases with valid and invalid CACertSecret configurations

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -748,6 +751,19 @@ func (r *RepoSyncReconciler) validateNamespaceName(namespaceName string) error {
...
+	caCertSecretRefName := v1beta1.GetSecretName(repoSync.Spec.Git.CACertSecretRef)
+
+	if useCACert(caCertSecretRefName) {
+		caCertSecretRef := client.ObjectKey{Namespace: repoSync.Namespace, Name: caCertSecretRefName}
+		keys := GetSecretKeys(ctx, r.client, caCertSecretRef)
```
sdowell, line 759: It looks like `GetSecretKeys` swallows an error and returns `nil` if the `Secret` does not exist. We need to handle the error where the `Secret` does not exist and return a separate error message.

Using `validateSecretExist` seems like the right approach, and using the `Secret` returned by that function to validate that the key is present. Note if we do that we should probably refactor the `NotFound` error message to support both use cases, since [currently it is specific to authentication](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/validate_secret.go#L39).

Note we will also want some additional unit tests for this unhappy path

victorpras: refactored `validateSecretExist` to return generic error and error handling is moved to function caller. Added more unit test wrt NotFound error

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -736,7 +739,10 @@ func (r *RepoSyncReconciler) validateNamespaceSecret(ctx context.Context, repoSy
 		repoSync.Namespace,
 		r.client)
 	if err != nil {
-		return err
+		if apierrors.IsNotFound(err) {
+			return errors.Errorf("secret %s not found: create one to allow client authentication", namespaceSecretName)
```
sdowell, line 743: nit: use capital `Secret` when referring to k8s Secrets. Ack that you're refactoring this but might as well fix with this change

ditto for the other parts of the diff

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -748,6 +754,27 @@ func (r *RepoSyncReconciler) validateNamespaceName(namespaceName string) error {
...
+	caCertSecretRefName := v1beta1.GetSecretName(repoSync.Spec.Git.CACertSecretRef)
+
+	if useCACert(caCertSecretRefName) {
+		secret, err := validateSecretExist(ctx,
+			caCertSecretRefName,
+			repoSync.Namespace,
+			r.client)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return errors.Errorf("Secret %s not found", caCertSecretRefName)
```
sdowell, line 767: nit: use a more descriptive error message, consistent with the other Secret error. e.g.
`Secret %s not found, create one to allow client connections with CA certificate`

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -748,6 +754,27 @@ func (r *RepoSyncReconciler) validateNamespaceName(namespaceName string) error {
 	return nil
 }
 
+func (r *RepoSyncReconciler) validateCACertSecret(ctx context.Context, repoSync *v1beta1.RepoSync) error {
```
sdowell, line 757: If we change the signature to accept either `Git` or `SecretReference` (and `namespace`) rather than a `RootSync`/`RepoSync`, we can move this declaration to `reconcilerBase`. This would be preferable because then we only need to define this logic in one place (rather than duplicated for RootSync/RepoSync)

sdowell (commented):
Close to LGTM, a few more comments

pkg/reconcilermanager/controllers/reposync_controller.go
victorpras: updated error msg to use capital letters

pkg/reconcilermanager/controllers/reposync_controller.go
victorpras: updated error message to be more descriptive

pkg/reconcilermanager/controllers/reposync_controller.go
victorpras: refactored function to reconciler_base. Updated to take secretRef string and namespace param

sdowell (approved):

PR #538, opened 2023-04-12
Move Repository into the gitproviders package

karlkfi:
- Refactor most methods to return an error, instead of failing the current test.
- Refactored most callers to use a new NT.Must method that handles checking arguments for errors. Any non-nil errors cause immediate test failure. This is similar to require.NoError, except it uses NT's Test field and allows an number of arguments, including non-errors, which are ignored.
- This cleanup paves the way to further integrate Repository and GitProvider, to remove provider-specific code from NT.


e2e/nomostest/gitproviders/repository.go
```diff
@@ -51,193 +52,223 @@ const (
...
+	repoName, err := g.GitProvider.CreateRepository(g.Name)
 	if err != nil {
-		nt.T.Fatal(err)
+		return errors.Wrapf(err, "creating repo: %s", g.Name)
 	}
 	g.RemoteRepoName = repoName
 
-	g.init(nt.gitPrivateKeyPath)
-	g.initialCommit(sourceFormat)
+	// TODO: Move this to the caller or GitProvider
```
sdowell, line 161: Maybe we could have an NT wrapper method that calls this and sets the entry in `RemoteRepositories`

EDIT: looks like you addressed this, so I think this comment can be removed

e2e/nomostest/gitproviders/repository.go
```diff
@@ -51,193 +52,223 @@ const (
...
-//
-// Fails immediately if any git command fails.
-func (g *Repository) Git(command ...string) {
-	g.T.Helper()
-
-	cmd := g.gitCmd(command...)
-	out, err := cmd.CombinedOutput()
+	out, err := exec.Command(args[0], args[1:]...).CombinedOutput()
 	if err != nil {
+		if !g.Logger.IsDebugEnabled() {
```
sdowell, line 188: If debug is not enabled - is this right? I'd think we'd want to log the extra information if debug is enabled, or just `Logger.Debugf`

e2e/nomostest/gitproviders/repository.go
```diff
@@ -248,82 +279,80 @@ func (g *Repository) init(privateKey string) {
...
 		// If you're seeing this error, use "AddFile" instead to test ignoring
 		// files with extensions we ignore.
-		err = fmt.Errorf("invalid extension to write object to, %q, use .AddFile() instead", ext)
-	}
-	if err != nil {
-		g.T.Fatal(err)
+		return fmt.Errorf("invalid extension to write object to, %q, use .AddFile() instead", ext)
 	}
-
-	g.AddFile(path, bytes)
```
sdowell, line 286: How come this shared call was moved up to each of the cases? This seemed fine with the outer scoped `bytes` declaration

e2e/nomostest/gitproviders/repository.go
```diff
@@ -332,46 +361,72 @@ func (g *Repository) Get(path string) client.Object {
...
 	err = runtime.DefaultUnstructuredConverter.FromUnstructured(uObj.UnstructuredContent(), obj)
 	if err != nil {
-		g.T.Fatal(err)
+		return nil, errors.Wrap(err, "converting typed object to unstructured")
+	}
+	return obj, nil
+}
+
+// MustGet calls Get and fails the test on error, logging the result.
+func (g *Repository) MustGet(t testing.NTB, path string) client.Object {
```
sdowell, line 376: If we have this `MustGet` abstraction here, it seems like we could either keep the default behavior of the other methods or define a `Must` alternative (e.g. `MustCommitAndPush`). WDYT?

Is the reason for this method that `nt.Must` swallows the return value?

e2e/nomostest/gitproviders/repository.go
```diff
@@ -382,219 +437,220 @@ func (g *Repository) GetAll(dirPath string, recursive bool) []client.Object {
...
 			return err
 		}
-		args = append(args, remoteURL, refspec)
-		cmd := g.gitCmd(args...)
-		out, err = cmd.CombinedOutput()
+		_, err = g.Git(append([]string{"push", remoteURL}, args...)...)
 		return err
 	})
 	if err != nil {
-		g.T.Log(string(out))
```
sdowell, line 519: Was it intentional to remove the stdout/stderr logging?

sdowell (commented):
I can confirm that `Must` no longer seems like a real word after staring at it for too long. It's kind of like `Musk` or `Mist`... but less real...

e2e/nomostest/gitproviders/repository.go
karlkfi: Yeah, I moved it into ResetRepository, which is the sole caller of NewRepository now.
I'll remove the TODO.

e2e/nomostest/gitproviders/repository.go
karlkfi: The command is printed to the debug log BEFORE the call. So if debug is NOT enabled, we need to print the command AFTER the call errors.

This is how it works elsewhere for command failure. Alternatively we could just always print the command to the normal log before the call.

sdowell: Ah I missed that, sounds good then

e2e/nomostest/gitproviders/repository.go
karlkfi: The `Git` method handles logging on error now, so the callers don't need to.

e2e/nomostest/gitproviders/repository.go
karlkfi: Just a matter of taste. When I replaced the Fatal calls with returns I realized that some of the switch cases were returning inside the switch and some weren't. So I move all the returns into the switch.

Yes, it means this AddFile call is duplicated in 2 cases, but it has some benefits:
1. The compiler now validates that all possible cases return
2. It's more obvious that the default case cannot ever lead to the AddFile code branch.
3. It becomes more clear that AsUnstructuredSanitized is not returning just an error, otherwise we wouldn't need the `var err error` to down-cast it.
4. We don't need the `var bytes []byte` or `var u *unstructured.Unstructured` outside the switch.

I could change it back, if you prefer, but I think this is slightly better.

sdowell (approved):

PR #539, opened 2023-04-12
Add a new reconciler into reconciler-manager

haiyanmeng:
Currently, to export CS metrics to Cloud Monitoring and Cloud Monarch on a GKE cluster with Workload Identity enabled, the customers need to: 
1) bind the `default` service account in the    `config-management-monitoring` namespace to a GSA with the metric    writer role;
2) set the `iam.gke.io/gcp-service-account` annotation of the k8s        service account to the email address of the GSA;
3) restart the `otel-collector` Deployment.

This PR automates the third step by adding a new reconciler into reconciler-manager. The new reconciler watches the default service account under the config-management-monitoring namespace, and adds/updates the `iam.gke.io/gcp-service-account` annotation in the `spec.template.annotation` field of the otel-collector Deployment. This triggers the Deployment to restart in the event of an annotation upate.

The automation of the first two steps will be explored seperately.

### DIFF
--- a/cmd/reconciler-manager/main.go
+++ b/cmd/reconciler-manager/main.go
@@ -107,6 +107,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	otelSA := controllers.NewOtelSAReconciler(*clusterName, mgr.GetClient(),
+		ctrl.Log.WithName("controllers").WithName(controllers.OtelSALoggerName),
+		mgr.GetScheme())
+	if err := otelSA.SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "OtelSA")
+		os.Exit(1)
+	}
+
 	// Register the OpenCensus views
 	if err := metrics.RegisterReconcilerManagerMetricsViews(); err != nil {
 		setupLog.Error(err, "failed to register OpenCensus views")
--- a/pkg/reconcilermanager/controllers/otel_sa_controller.go
+++ b/pkg/reconcilermanager/controllers/otel_sa_controller.go
@@ -1,0 +1,116 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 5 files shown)


pkg/reconcilermanager/controllers/otel_sa_controller.go
```diff
@@ -0,0 +1,112 @@
...
+		client:      client,
+		log:         log,
+		scheme:      scheme,
+	}
+}
+
+// Reconcile reconciles the default service account under the config-management-monitoring namespace and updates the Deployment annotation.
+// This triggers the `otel-collector` Deployment to restart in the event of an annotation update.
+func (r *OtelSAReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := r.log.WithValues("otelSA", req.NamespacedName.String())
```
tiffanny29631, line 63: Same here, use a var?

pkg/reconcilermanager/controllers/otel_sa_controller.go
```diff
@@ -0,0 +1,112 @@
...
+	clusterName string
+	client      client.Client
+	log         logr.Logger
+	scheme      *runtime.Scheme
+}
+
+// NewOtelSAReconciler returns a new OtelSAReconciler.
+func NewOtelSAReconciler(clusterName string, client client.Client, log logr.Logger, scheme *runtime.Scheme) *OtelSAReconciler {
+	if clusterName == "" {
+		clusterName = "unknown_cluster"
```
tiffanny29631, line 50: Should we error out if no cluster is specified?

cmd/reconciler-manager/main.go
```diff
@@ -107,6 +107,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	otelSA := controllers.NewOtelSAReconciler(*clusterName, mgr.GetClient(),
+		ctrl.Log.WithName("controllers").WithName("OtelSA"),
```
tiffanny29631, line 111: Extract this into a var?

tiffanny29631 (commented):
The name of the PR can be shorter, just add the remaining to the description.

cmd/reconciler-manager/main.go
haiyanmeng: What is your concern here?

haiyanmeng: Are you suggesting making a var for `OtelSA`?

pkg/reconcilermanager/controllers/otel_sa_controller.go
haiyanmeng: Are you suggesting making a var for `otelSA`?

pkg/reconcilermanager/controllers/otel_sa_controller.go
tiffanny29631: Interesting that it's not used but still specified, could be a pattern copied from the r*sync reconcilers. We might remove it as it's not used in here, unless we want to include cluster name in the log

haiyanmeng: Let us keep it to be consistent with [kpt-config-sync/pkg/reconcilermanager/controllers/otel_controller.go](https://github.com/GoogleContainerTools/kpt-config-sync/blob/688b7a1d7c57e3f8b5167b34d318125ec7c782f5/pkg/reconcilermanager/controllers/otel_controller.go#L55).

PR #541, opened 2023-04-12
test: increase timeout for git push retry

sdowell:
The git server can take a long time to restart when running on autopilot. This extends the retry timeout when using the in-cluster git server.


e2e/nomostest/gitproviders/repository.go
```diff
@@ -555,7 +556,13 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) error {
 // Push pushes the provided refspec to the git server.
 // Performs a retry using RemoteURL, which may change if the port forwarding restarts.
 func (g *Repository) Push(args ...string) error {
-	took, err := retry.Retry(1*time.Minute, func() error {
+	timeout := 1 * time.Minute
+	if g.GitProvider.Type() == e2e.Local {
```
karlkfi, line 560: Not for this PR, but we should really move these provider type constants into the gitproviders package. It doesn't make sense to be importing the e2e package for this.

e2e/nomostest/gitproviders/repository.go
```diff
@@ -555,7 +556,13 @@ func (g *Repository) CommitAndPushBranch(msg, branch string) error {
...
+		// Wait long enough for the local git-server pod to be rescheduled
+		// on GKE Autopilot, which may involve spinning up a new node.
+		timeout = 5 * time.Minute
```
karlkfi, line 563: It occurs to me that this could use DefaultWaitTimeout, but it would need to be routed in through the constructor. That way it would be longer for autopilot and shorter otherwise.

sdowell: done

e2e/nomostest/gitproviders/repository.go
```diff
@@ -115,6 +116,8 @@ type Repository struct {
 	Logger *testlogger.TestLogger
 
 	yamlSerializer *jserializer.Serializer
+
+	defaultWaitTimeout time.Duration
```
karlkfi, line 120: nit: since it can't be overridden here, it's not really "Default". It's more like "retryTimeout" or "pushRetryTimeout"

karlkfi: otoh, "defaultWaitTimeout" makes it sorta obvious it's linked to that e2e setting.

karlkfi (approved):
/lgtm

PR #543, opened 2023-04-18
feat: filter Args from error json in RSync status

sdowell:
The Args field sometimes contains values which change between reconciles, such as the failCount in the *-sync containers. Having errors which flap between iterations makes it unfeasible to treat the error message as a unique key for notification delivery, as the user will get repeat notifications for errors with trivial differences. This change adds logic to filter out the Args field of the error payload before passing it through to the RSync status.

### DIFF
--- a/pkg/hydrate/controller.go
+++ b/pkg/hydrate/controller.go
@@ -15,6 +15,7 @@
 package hydrate
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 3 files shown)


sdowell:
This change is a prerequisite for https://github.com/GoogleContainerTools/kpt-config-sync/pull/542

nan-yu (approved):
/lgtm

PR #544, opened 2023-04-19
test: use RootRepos/NonRootRepos in init

sdowell:
RemoteRepositories contains Repositories for all testcases, including previous tests which have finished. The local repo is stored in a temporary directory which is deleted at the end of the test, which caused this to fail when there was a leftover Repository in RemoteRepositories which was not defined in the current test. This switches the logic to only reset repositories used by the test.


karlkfi (approved):
/lgtm

PR #545, opened 2023-04-19
design docs folder seeding

mikebz:
Looks like one of the docs by @karlkfi is a design doc and we can move that as an example of where public design docs can go.

In addition the change to remove the internal shortcut http://cloud from an external doc.


karlkfi:
It’s not a design doc. Design docs are written before implementation to propose a feature. This is a usage doc, describing the feature to users and maintainers.

PR #549, opened 2023-04-19
docs: add user guide for notification feature

sdowell:


docs/user-guides/notifications/catalog/triggers.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+## on-reconciler-stalled
+
+The `RootSync`/`RepoSync` object is stalled. Triggers whenever the `Stalled`
+condition is true.
+
+```yaml
+  trigger.on-reconciler-stalled: |
+    - when: any(sync.status.conditions, {.type == 'Stalled' && .status == 'True')
```
sdowell, line 40: I'm wondering if we should try to provide a `oncePer` condition for all of these which don't already have one

docs/user-guides/notifications/catalog/templates.md
```diff
@@ -0,0 +1,101 @@
+# Triggers Catalog
```
nan-yu, line 1: Templates Catalog

docs/user-guides/notifications/catalog/templates.md
```diff
@@ -0,0 +1,101 @@
...
+
+## reconciler-created
+
+```yaml
+  template.reconciler-created: |
+    message: |
+      {{.sync.kind}} {{.sync.metadata.namespace}}/{{.sync.metadata.name}} has been created.
+    email:
+      subject: |
+        A new {{.sync.kind}} has been created.
```
nan-yu, line 18: The subject can be the same as message. Ditto for other templates.

docs/user-guides/notifications/catalog/templates.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+## reconciler-stalled
+
+```yaml
+  template.reconciler-stalled: |
+    message: |
+      {{.sync.kind}} {{.sync.metadata.namespace}}/{{.sync.metadata.name}} is Stalled.
+      
+      {{range $index, $e := call .utils.ConfigSyncErrors}}{{js $e.ErrorMessage}}\n\n{{end}}
```
nan-yu, line 39: The stalled error should be included in the Stalled condition, instead of ConfigSyncErrors.

docs/user-guides/notifications/catalog/templates.md
```diff
@@ -0,0 +1,101 @@
...
+      subject: |
+        A {{.sync.kind}} is Stalled.
+```
+
+## reconciler-reconciling
+
+```yaml
+  template.reconciler-reconciling: |
+    message: |
+      {{.sync.kind}} {{.sync.metadata.namespace}}/{{.sync.metadata.name}} is reconciling.
```
nan-yu, line 50: This is misleading. It is the reconciler Deployment is reconciling, rather than the R*Sync is reconciling the managed resources.

Suggest updating to
```
The reconciler Deployment for {{.sync.kind}} {{.sync.metadata.namespace}}/{{.sync.metadata.name}} is reconciling.
```

docs/user-guides/notifications/catalog/templates.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+## sync-synced
+
+This example shows how to configure Github notifications in addition to email.
+
+```yaml
+  template.sync-synced: |
+    message: |
+      {{.sync.kind}} {{.sync.metadata.name}} is synced to {{.sync.status.lastSyncedCommit}}!
```
nan-yu, line 63: Suggest adding the namespace for RepoSyncs.

docs/user-guides/notifications/catalog/templates.md
```diff
@@ -0,0 +1,101 @@
...
+    email:
+      subject: |
+        {{.sync.kind}} {{.sync.metadata.name}} is synced.
+    github:
+      repoURLPath: "{{.sync.spec.git.repo}}"
+      revisionPath: "{{.sync.status.lastSyncedCommit}}"
+      status:
+        state: success
+        label: "continuous-delivery/{{.sync.metadata.name}}"
```
nan-yu, line 72: Add namespace

docs/user-guides/notifications/catalog/templates.md
```diff
@@ -0,0 +1,101 @@
...
+      {{range $index, $e := call .utils.ConfigSyncErrors}}{{js $e.ErrorMessage}}\n\n{{end}}
+    email:
+      subject: |
+        A {{.sync.kind}} failed to sync.
+```
+
+## sync-pending
+
+```yaml
+  template.sync-failed: |
```
nan-yu, line 92: sync-pending

docs/user-guides/notifications/catalog/triggers.md
```diff
@@ -0,0 +1,90 @@
...
+## on-reconciler-created
+
+The `RootSync`/`RepoSync` object is created. Triggers once for each newly created
+`RootSync`/`RepoSync` object.
+
+```yaml
+  trigger.on-reconciler-created: |
+    - when: True
+      oncePer: sync.metadata.name
+      send: [on-reconciler-created]
```
nan-yu, line 18: replace `on-reconciler-created` with `reconciler-created`. We should use the template name here.

Same for other triggers.

docs/user-guides/notifications/catalog/templates.md
sdowell: done

docs/user-guides/notifications/catalog/triggers.md
nan-yu: Missed this comment. Currently, only the following triggers don't have a `oncePer` condition.
- `on-reconciler-stalled`: we would like to have a stalled notification for every R*Sync whose reconciler Deployment becomes STALLED. For the same reconciler, it might become STALLED multiple times, and we would like to capture each STALLED case.
- `on-reconciler-reconciling`: same as above
- `on-sync-pending`:  probably yes, and set `oncePer` for `utils.LastCommit()`, which matches `on-sync-synced`.

docs/user-guides/notifications/catalog/templates.md
nan-yu: https://github.com/argoproj/argo-cd/blob/master/notifications_catalog/templates/app-created.yaml shows an easy way to reference `message` in the subject. Does it work here?

docs/user-guides/notifications/catalog/templates.md
nan-yu: I don't think it is necessary add `conditions` to `ConfigSyncErrors`.
For STALLED and RECONCILING conditions, the error is from its condition, not `ConfigSyncErrors`.
For SYNCING condition, the error is from `ConfigSyncErrors`, not the condition.

docs/user-guides/notifications/catalog/templates.md
sdowell: Yeah looks like that should work, done

docs/user-guides/notifications/catalog/triggers.md
sdowell: updated `on-sync-pending`

nan-yu (approved):
/lgtm

PR #565, opened 2023-05-02
test: replace metrics timeout with wait timeout

sdowell:
The metrics validation now gets the dynamic port forward in the retry loop. If the port forward has to restart during the loop, this can take longer than the default metrics timeout (esp on autopilot). This is because under this scenario we may need to wait for the Pod to be ready before we can get a valid port forward. This change replaces the metrics timeout with the standard wait timeout to account for longer delays.

Context:
- https://testgrid.k8s.io/googleoss-kpt-config-sync-release-multi-repo-8#autopilot-rapid-latest
- https://oss.gprow.dev/view/gcs/oss-prow-build-kpt-config-sync/logs/release-multi-repo-8-autopilot-rapid-latest/1653362701939773440
- https://oss.gprow.dev/view/gcs/oss-prow-build-kpt-config-sync/logs/release-multi-repo-8-autopilot-rapid-latest/1653331999286366208


karlkfi (approved):
/lgtm

PR #567, opened 2023-05-02
[CHERRYPICK] test: replace metrics timeout with wait timeout (#565)

sdowell:
The metrics validation now gets the dynamic port forward in the retry loop. If the port forward has to restart during the loop, this can take longer than the default metrics timeout (esp on autopilot). This is because under this scenario we may need to wait for the Pod to be ready before we can get a valid port forward. This change replaces the metrics timeout with the standard wait timeout to account for longer delays.


karlkfi (approved):
/lgtm

PR #569, opened 2023-05-03
Increase the test prometheus memory request

nan-yu:
Recent test failures in release-multi-repo-8-autopilot-rapid-latest were caused by OOMKilled failures in the prometheus container:
```
containerStatuses:
  - containerID: containerd://eda177e63e2c28e473c95a17287015cae17af4503ad8ad04f99a138fa14dfd6f
    image: docker.io/prom/prometheus:v2.37.6
    imageID: docker.io/prom/prometheus@sha256:92ceb93400dd4c887c76685d258bd75b9dcfe3419b71932821e9dcc70288d851
    lastState:
      terminated:
        containerID: containerd://eda177e63e2c28e473c95a17287015cae17af4503ad8ad04f99a138fa14dfd6f
        exitCode: 137
        finishedAt: "2023-05-03T18:05:49Z"
        reason: OOMKilled
        startedAt: "2023-05-03T18:05:16Z"
    name: prometheus
    ready: false
    restartCount: 7
    started: false
    state:
      waiting:
        message: back-off 5m0s restarting failed container=prometheus pod=prometheus-deployment-76f4b58fb4-k7mgj_prometheus(779e3acb-82fc-4430-9946-714aa471ba16)
        reason: CrashLoopBackOff
```
This commit increases the memory request from 500Mi to 1Gi.

### DIFF
--- a/e2e/testdata/prometheus/prometheus.yaml
+++ b/e2e/testdata/prometheus/prometheus.yaml
@@ -174,7 +174,7 @@ spec:
           resources:
             requests:
               cpu: 500m
-              memory: 500M
+              memory: 1Gi
             limits:
               cpu: 1
               memory: 1Gi


e2e/testdata/prometheus/prometheus.yaml
```diff
@@ -174,7 +174,7 @@ spec:
           resources:
             requests:
               cpu: 500m
-              memory: 500M
+              memory: 1Gi
             limits:
               cpu: 1
               memory: 1Gi
```
sdowell, line 180: Do you think we should bump the limit or leave as-is?

karlkfi: If it's showing OOMKilled in the container status, it's probably because of the limit, not the request. Request might get it killed by the kernel oomkiller, but that usually shows up as a different failure.

karlkfi (changes requested):
The limit needs bumping. Request only affects scheduling, not oomkill.

e2e/testdata/prometheus/prometheus.yaml
nan-yu: I think the current limit (1Gi) should be okay.

The failure only happens on the `autopilot-rapid-latest` test cluster. For Autopilot clusters, the limit is reset to be the same as request (adjusted), which is `512Mi`.

nan-yu:
> The limit needs bumping. Request only affects scheduling, not oomkill.

Left a comment around limit. Autopilot sets limit to be the same as request, so bumping limit won't help on Autopilot clusters.
For standard clusters, the current limit is good enough.

sdowell:
/lgtm
/approve

@karlkfi you'll need to remove your `changes requested` state. Tide doesn't play nice with that Github state https://github.com/kubernetes/test-infra/issues/27750

karlkfi (approved):
/lgtm

karlkfi:
FWIW, you can remove my change requested by hitting the "re-request review" button top right.

PR #573, opened 2023-05-05
helm api: allow user to specify a version range

natasha41575:
This PR adds support for allowing users to specify a range in `spec.helm.version` in the RSync object and we pull down the latest one from the range.

- The version range syntax is natively supported by Helm CLI. See the syntax here: [https://github.com/Masterminds/semver#hyphen-range-comparisons](https://www.google.com/url?q=https://github.com/Masterminds/semver%23hyphen-range-comparisons) 
- This PR has some changes to ensure that the RSync object is reporting the actual version that was pulled down, instead of just displaying whatever was specified in `spec.helm.version`
- If someone pushes a new helm chart, helm-sync will pick up the new chart without requiring changes to the RSync object.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -190,46 +190,36 @@ func TestHelmDefaultNamespace(t *testing.T) {
 // TestHelmLatestVersion verifies the Config Sync behavior for helm charts when helm.spec.version is not specified. The helm-sync
 // binary should pull down the latest available version in this case. It also tests that if a new helm chart gets pushed, the
 // chart version gets automatically updated by Config Sync.
-// This test will work only with following pre-requisites:
-// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for accessing images in Artifact Registry.
-// A JSON key file is generated for this service account and stored in Secret Manager
+// The test will run on a GKE cluster only with following pre-requisites
+//
+// 1. Workload Identity is enabled.
+// 2. The Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for access image in Artifact Registry.
+// 3. An IAM policy binding is created between the Google service account and the Kubernetes service accounts with the `roles/iam.workloadIdentityUser` role.
+//
+//	gcloud iam service-accounts add-iam-policy-binding --project=${GCP_PROJECT} \
+//	   --role roles/iam.workloadIdentityUser \
+//	   --member "serviceAccount:${GCP_PROJECT}.svc.id.goog[config-management-system/root-reconciler]" \
+//	   e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com
+//
+// 4. The following environment variables are set: GCP_PROJECT, GCP_CLUSTER, GCP_REGION|GCP_ZONE.
 func TestHelmLatestVersion(t *testing.T) {
 	nt := nomostest.New(t,
-		nomostesting.SyncSource,
+		nomostesting.WorkloadIdentity,
 		ntopts.Unstructured,
 		ntopts.RequireGKE(t),
 	)
 
 	rs := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
-	nt.T.Log("Fetch password from Secret Manager")
-	key, err := gitproviders.FetchCloudSecret("config-sync-ci-ar-key")
-	if err != nil {
-		nt.T.Fatal(err)
-	}
-
-	nt.T.Log("Create secret for authentication")
-	_, err = nt.Shell.Kubectl("create", "secret", "generic", "foo", fmt.Sprintf("--namespace=%s", v1.NSConfigManagementSystem), "--from-literal=username=_json_key", fmt.Sprintf("--from-literal=password=%s", key))
-	if err != nil {
-		nt.T.Fatalf("failed to create secret, err: %v", err)
-	}
-	nt.T.Cleanup(func() {
-		nt.MustKubectl("delete", "secret", "foo", "-n", v1.NSConfigManagementSystem, "--ignore-not-found")
-	})
-
 	remoteHelmChart, err := helm.PushHelmChart(nt, privateSimpleHelmChart, privateSimpleHelmChartVersion)
 	if err != nil {
 		nt.T.Fatalf("failed to push helm chart: %v", err)
 	}
 
 	nt.T.Log("Update RootSync to sync from a private Artifact Registry")
-	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "git": null, "helm": {"repo": "%s", "chart": "%s", "auth": "token", "version": "", "namespace": "", "deployNamespace": "simple", "secretRef": {"name" : "foo"}}}}`,
-		v1beta1.HelmSource, helm.PrivateARHelmRegistry, remoteHelmChart.ChartName))
-	err = nt.WatchForAllSyncs(nomostest.WithRootSha1Func(helmChartVersion(remoteHelmChart.ChartName+":"+privateSimpleHelmChartVersion)),
-		nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: remoteHelmChart.ChartName}))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
-	if err := nt.Validate("deploy-default", "simple", &appsv1.Deployment{}, testpredicates.HasLabel("version", privateSimpleHelmChartVersion)); err != nil {
+	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "helm": {"chart": "%s", "repo": "%s", "version": "", "auth": "gcpserviceaccount", "gcpServiceAccountEmail": "%s", "deployNamespace": "simple"}, "git": null}}`,
+		v1beta1.HelmSource, remoteHelmChart.ChartName, helm.PrivateARHelmRegistry, gsaARReaderEmail))
+	if err = nt.Watcher.WatchObject(kinds.Deployment(), "deploy-default", "simple",
+		[]testpredicates.Predicate{testpredicates.HasLabel("version", privateSimpleHelmChartVersion)}); err != nil {
 		nt.T.Error(err)
 	}
 
@@ -241,12 +231,8 @@ func TestHelmLatestVersion(t *testing.T) {
 	if err := remoteHelmChart.Push(); err != nil {
 		nt.T.Fatal("failed to push helm chart update: %v", err)
 	}
-	err = nt.WatchForAllSyncs(nomostest.WithRootSha1Func(helmChartVersion(remoteHelmChart.ChartName+":"+newVersion)),
-		nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: remoteHelmChart.ChartName}))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
-	if err := nt.Validate("deploy-default", "simple", &appsv1.Deployment{}, testpredicates.HasLabel("version", newVersion)); err != nil {
+	if err = nt.Watcher.WatchObject(kinds.Deployment(), "deploy-default", "simple",
+		[]testpredicates.Predicate{testpredicates.HasLabel("version", newVersion)}); err != nil {
 		nt.T.Error(err)
 	}
 }
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -165,16 +165,16 @@ func TestHelmDefaultNamespace(t *testing.T) {
 		nt.MustKubectl("delete", "secret", "foo", "-n", v1.NSConfigManagementSystem, "--ignore-not-found")
 	})
 
-	chartName, err := helm.PushHelmChart(nt, privateSimpleHelmChart, privateSimpleHelmChartVersion)
+	remoteHelmChart, err := helm.PushHelmChart(nt, privateSimpleHelmChart, privateSimpleHelmChartVersion)
 	if err != nil {
 		nt.T.Fatalf("failed to push helm chart: %v", err)
 	}
 
 	nt.T.Log("Update RootSync to sync from a private Artifact Registry")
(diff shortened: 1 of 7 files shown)


e2e/testdata/helm-charts/simple-2.5.3/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A simple test helm chart.
+name: simple
+version: 2.5.3
```
natasha41575, line 5: I don't have permission to push this chart to the `kpt-config-sync-ci-release` project, so I will need someone to do that for me

pkg/reconcilermanager/controllers/util.go
```diff
@@ -93,11 +93,7 @@ func reconcilerEnvs(clusterName, syncName, reconcilerName string, reconcilerScop
 	case v1beta1.HelmSource:
 		syncRepo = helmConfig.Repo
 		syncDir = helmConfig.Chart
-		if helmConfig.Version != "" {
-			syncRevision = helmConfig.Version
-		} else {
-			syncRevision = "latest"
```
natasha41575, line 99: If the version is left out entirely, helm CLI will automatically pull the latest one, so there is no need to set a default here. It was discussed offline that supporting polling from a latest tag with helm could be inefficient/not scalable.

e2e/testdata/root-sync-helm-chart-version-range-cr.yaml
```diff
@@ -0,0 +1,53 @@
...
+metadata:
+  name: root-sync
+  namespace: config-management-system
+spec:
+  sourceFormat: unstructured
+  sourceType: helm
+  helm:
+    repo: https://charts.bitnami.com/bitnami
+    chart: wordpress
+    version: ^15.0.0
```
natasha41575, line 26: This syntax is natively supported by helm CLI.

See [https://github.com/Masterminds/semver#hyphen-range-comparisons](https://www.google.com/url?q=https://github.com/Masterminds/semver%23hyphen-range-comparisons) for more details on the syntax

pkg/parse/root.go
```diff
@@ -234,10 +235,21 @@ func setSourceStatusFields(source *v1beta1.SourceStatus, p Parser, newStatus sou
...
+		// but that could potentially be provided to use as a range of
+		// versions from which we pick the latest. We should display the
+		// version that was actually pulled down if we can.
+		// source.Commit is expected to be of the format `chart:version`,
+		// so we parse it to grab the version.
+		version := p.options().SourceRev
+		split := strings.Split(source.Commit, ":")
+		if len(split) == 2 {
+			version = split[1]
+		}
```
janetkuo, line 248: Would you make this a function?

pkg/reconcilermanager/controllers/util.go
janetkuo: We could also consider keeping it, so that if helm CLI decides to break this default behavior later, we wouldn't be affected.

natasha41575: Thank you for your comment, as it led me to further investigate this code. I learned that this does not actually affect what gets passed to helm CLI as it populates the `reconciler` container, not the `helm-sync` container. It seems that the primary purpose of this default value is to display `version: latest` in the RSync status when the version is empty (rather than displaying `version: ""`). Since another part of my PR overwrites the RSync status to display the version that was actually pulled down, I think in most cases this bit of code is inconsequential. That said, if something were to go wrong with the version parsing, it would make more sense to display "latest" than "", so I have reverted this change.

pkg/parse/root.go
natasha41575: Good call, done.

e2e/testdata/root-sync-helm-chart-version-range-cr.yaml
janetkuo: Will a new version trigger an update automatically? Is there a way test that? 

FYI - Because the update is triggered without changing R*Sync spec, this is going to make `status.observedGeneration` less useful. Tagging @sdowell given that we discussed this earlier.

janetkuo:
I added a comment https://github.com/GoogleContainerTools/kpt-config-sync/pull/573#discussion_r1190451574 but it's not published as a separate review in the bottom of the page.

e2e/testdata/root-sync-helm-chart-version-range-cr.yaml
natasha41575: > Will a new version trigger an update automatically? Is there a way test that?

I tried it manually and yes, it does an automatic update. 

I'll see if I can add a test. The test would presumably involve pushing a new chart version to the test registry and then deleting it after, which I think is possible, but I will have to look more into it.

natasha41575: Actually, do you know if the test env has helm CLI on it? I think that would be needed for adding a test

sdowell: Assuming you are referring to the e2e test env, [yes it has helm CLI](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/build/test-e2e-go/gke/Dockerfile#L41). Looks like the e2e test runner GSA currently does not have permission to delete images, but that could be added. Why does it need to delete images? For test cleanup?

This would be another use case where using an in-cluster/ephemeral/hermetic registry would be preferred. That shouldn't be a blocker for you here, just wanted to flag it

natasha41575: > Why does it need to delete images? For test cleanup?

Yes for test cleanup; this test would need to push a new tag and clean up the tag afterwards. Since we can add the permissions, seems like there are no blockers so I will try to add the test.

I agree hermetic registry would be a better solution in the long term.

natasha41575: Pushed the new test.

/hold (until the e2e test env GSA to have the right permissions)

natasha41575:
/hold 

(see https://github.com/GoogleContainerTools/kpt-config-sync/pull/573#discussion_r1195752922)

e2e/testdata/root-sync-helm-chart-version-range-cr.yaml
janetkuo: @natasha41575 Thanks for adding the test. Would it trigger an upgrade (to the latest version) when a version range is specified, too?

janetkuo (approved):
/lgtm

natasha41575:
Thanks @janetkuo for the review! I had to make a change to how we push the new helm chart due to the GSA permissions, so will need another LGTM. Also had another question https://github.com/GoogleContainerTools/kpt-config-sync/pull/573#discussion_r1196953251 that we should consider before merging this with the new test

e2e/testcases/helm_sync_test.go
```diff
@@ -178,6 +185,100 @@ func TestHelmDefaultNamespace(t *testing.T) {
...
+		nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: privateSimpleHelmChart}))
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	if err := nt.Validate("deploy-2.5.3", "default", &appsv1.Deployment{}); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Push a new chart version to the private Artifact Registry")
+	token, err := exec.Command("gcloud", "auth", "print-access-token").CombinedOutput()
```
sdowell, line 231: Does `gcloud auth configure-docker us-docker.pkg.dev` work? From the [helm](https://cloud.google.com/artifact-registry/docs/helm/authentication#methods)/[gcloud](https://cloud.google.com/artifact-registry/docs/docker/authentication#gcloud-helper) docs it sounds like helm should work with the credential helper.

I think it would be preferred to avoid printing/storing an access token to minimize the risk of leaking a token. Using a credential helper is the most secure option AFAICT

e2e/testdata/root-sync-helm-chart-version-range-cr.yaml
sdowell: Yes, multiple instances of this test may be running concurrently (e.g. for the different GKE channels) - and that's something we need to account for

natasha41575: @janetkuo What would you like to do here? Should we remove the part of the test I added for automatic updates and merge the PR with just the other tests for now, or should we block on rewriting these tests to be more hermetic/not use an external AR?

janetkuo: I'd prefer we make the tests more hermetic if possible. It will likely block us from adding more tests in the future.

e2e/testcases/helm_sync_test.go
natasha41575: I tried `gcloud auth configure-docker us-docker.pkg.dev` and it didn't work. If you have other suggestions or know how to get that working I'm happy to try it out, but so far the PAT is the most secure option that I've been able to get to work (and FWIW they are short-lived).

e2e/testdata/root-sync-helm-chart-version-range-cr.yaml
natasha41575: This will be unblocked by https://github.com/GoogleContainerTools/kpt-config-sync/pull/636. When that is merged I will rebase this PR

e2e/testcases/helm_sync_test.go
```diff
@@ -187,6 +187,90 @@ func TestHelmDefaultNamespace(t *testing.T) {
...
+
+	rs := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
+	nt.T.Log("Fetch password from Secret Manager")
+	key, err := gitproviders.FetchCloudSecret("config-sync-ci-ar-key")
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Create secret for authentication")
+	_, err = nt.Shell.Kubectl("create", "secret", "generic", "foo", fmt.Sprintf("--namespace=%s", v1.NSConfigManagementSystem), "--from-literal=username=_json_key", fmt.Sprintf("--from-literal=password=%s", key))
```
sdowell, line 211: Could this test use workload identity instead (with `e2e-test-ar-reader`)?

Doesn't seem like we're specifically testing token auth here, and I think it would be preferred to use WI as its easier to manage. The token auth with the exported key is more of an edge case IMO and is harder to manage. I'd suggest we stick to just using the exported token for tests which are specifically verifying token auth.

e2e/testcases/helm_sync_test.go
```diff
@@ -187,6 +187,90 @@ func TestHelmDefaultNamespace(t *testing.T) {
...
+
+	remoteHelmChart, err := helm.PushHelmChart(nt, privateSimpleHelmChart, privateSimpleHelmChartVersion)
+	if err != nil {
+		nt.T.Fatalf("failed to push helm chart: %v", err)
+	}
+
+	nt.T.Log("Update RootSync to sync from a private Artifact Registry")
+	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "git": null, "helm": {"repo": "%s", "chart": "%s", "auth": "token", "version": "", "namespace": "", "deployNamespace": "simple", "secretRef": {"name" : "foo"}}}}`,
+		v1beta1.HelmSource, helm.PrivateARHelmRegistry, remoteHelmChart.ChartName))
+	err = nt.WatchForAllSyncs(nomostest.WithRootSha1Func(helmChartVersion(remoteHelmChart.ChartName+":"+privateSimpleHelmChartVersion)),
```
sdowell, line 227: The assertions with `WatchForAllSyncs`/`Validate` here and below could be simplified to the following (with the updated version in the latter assertion):
```
nt.Watcher.WatchObject(
  kinds.Deployment(), "deploy-default", "simple",
  []testpredicates.Predicate{testpredicates.HasLabel("version", privateSimpleHelmChartVersion)}
)
```
This will wait for the deployment to have the label in the predicate

e2e/testcases/helm_sync_test.go
natasha41575: Yup makes sense, updated.

e2e/testcases/helm_sync_test.go
natasha41575: Thanks for the pointer! The test looks a lot simpler now.

e2e/testcases/helm_sync_test.go
sdowell: it looks like the test is still using token auth

natasha41575: Are you possibly looking at the first commit? I changed auth type in the latest one https://github.com/GoogleContainerTools/kpt-config-sync/pull/573/commits/96a6f44b35f2ab008d1775473d66647d21232d11

e2e/testcases/helm_sync_test.go
```diff
@@ -187,6 +187,76 @@ func TestHelmDefaultNamespace(t *testing.T) {
...
+	rootSyncFilePath := "../testdata/root-sync-helm-chart-version-range-cr.yaml"
+	nt.T.Logf("Apply the RootSync object defined in %s", rootSyncFilePath)
+	nt.MustKubectl("apply", "-f", rootSyncFilePath)
+	err := nt.WatchForAllSyncs(nomostest.WithRootSha1Func(helmChartVersion("wordpress:15.4.1")),
+		nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: "wordpress"}))
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	if err := nt.Validate("my-wordpress", "wordpress", &appsv1.Deployment{}); err != nil {
```
sdowell, line 255: Does this test not need to check a version label on the deployment?

e2e/testcases/helm_sync_test.go
sdowell: Ah you're right looks like I was looking at a stale commit, looks good now!

e2e/testcases/helm_sync_test.go
sdowell: No problem! Agreed looks good

sdowell (commented):
Final comments, mostly LGTM

e2e/testcases/helm_sync_test.go
natasha41575: This is a public helm chart made by bitnami, so no, the deployment template does not have a version label.

The other test is a chart one we wrote ourselves, and I added a version label to it myself in this PR for the automatic update test, so that I could check that the new version did indeed get resynced to the cluster. 

This test however is just testing that we grab the latest version from a version range, so checking that the correct version was pulled down is sufficient here, which we do by checking the commit in L249.

sdowell: Ack, thank you for clarifying

sdowell (approved):

PR #579, opened 2023-05-08
Empty Prometheus cache before each test case

nan-yu:
b/281576844
b/281508753
b/281432500

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -24,11 +24,14 @@ import (
 	"sync"
 	"time"
 
+	prometheusapi "github.com/prometheus/client_golang/api"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/e2e/nomostest/portforwarder"
+	"kpt.dev/configsync/e2e/nomostest/retry"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
 	"kpt.dev/configsync/e2e/nomostest/testlogger"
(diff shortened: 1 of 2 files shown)


e2e/nomostest/nt.go
```diff
@@ -711,6 +713,26 @@ func (nt *NT) portForwardPrometheus() {
...
+}
+
+// emptyPrometheusCache empty the cache in Prometheus.
+func (nt *NT) emptyPrometheusCache() error {
+	nt.T.Helper()
+	ctx, cancel := context.WithCancel(nt.Context)
+	defer cancel()
+
+	c, err := prometheusapi.NewClient(prometheusapi.Config{
+		Address: fmt.Sprintf("http://localhost:%d", prometheusServerPort),
```
sdowell, line 729: This is using the port forward on localhost, so you'll need to get the port from `nt.prometheusPortForwarder.LocalPort()` (and you'll probably want to wrap this in a retry loop)

nan-yu: Good catch. Done.

nan-yu:
/hold

Need to verify if we can pass an empty `match` parameter to delete all

e2e/nomostest/nt.go
```diff
@@ -711,6 +714,42 @@ func (nt *NT) portForwardPrometheus() {
...
+		c, err := prometheusapi.NewClient(prometheusapi.Config{
+			Address: fmt.Sprintf("http://localhost:%d", port),
+		})
+		if err != nil {
+			return err
+		}
+		v1api := prometheusv1.NewAPI(c)
+		// DeleteSeries only marks the time series for deletion
+		if err = v1api.DeleteSeries(ctx, nil, time.Time{}, time.Now()); err != nil {
+			return err
```
karlkfi, line 743: Add extra error message context for both of these calls to the api. The errors aren't very descriptive. ex: `server_error: server error: 500`

e2e/nomostest/nt.go
```diff
@@ -711,6 +714,42 @@ func (nt *NT) portForwardPrometheus() {
...
+		}
+		c, err := prometheusapi.NewClient(prometheusapi.Config{
+			Address: fmt.Sprintf("http://localhost:%d", port),
+		})
+		if err != nil {
+			return err
+		}
+		v1api := prometheusv1.NewAPI(c)
+		// DeleteSeries only marks the time series for deletion
+		if err = v1api.DeleteSeries(ctx, nil, time.Time{}, time.Now()); err != nil {
```
karlkfi, line 742: I believe `delete_series` is part of the admin API that needs to be enabled with the `--web.enable-admin-api` flag on the container command. 
- https://prometheus.io/docs/operating/security/#prometheus
- https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testdata/prometheus/prometheus.yaml#L157

nan-yu: Yes, I have to add `--web.enable-admin-api` in the deployment. Done locally. However, we have to provide a valid matcher. 
I am trying to use cluster name as the matcher.

e2e/nomostest/nt.go
nan-yu: done

e2e/nomostest/nt.go
```diff
@@ -711,6 +714,45 @@ func (nt *NT) portForwardPrometheus() {
...
+			return err
+		}
+		v1api := prometheusv1.NewAPI(c)
+		// DeleteSeries only marks the time series for deletion
+		if err = v1api.DeleteSeries(ctx, []string{`{__name__=~".+"}`}, time.Time{}, time.Now()); err != nil {
+			return fmt.Errorf("failed to delete all data from Prometheus cache: %v", err)
+		}
+		// CleanTombstones remove them from the disk to clear the space instantly
+		if err = v1api.CleanTombstones(ctx); err != nil {
+			return fmt.Errorf("failed to remove tombstones from the dish: %v", err)
```
karlkfi, line 747: s/dish/disk

sdowell (approved):

sdowell:
/hold
Putting a hold to address Karl's comment

sdowell:
re-lgtm
/lgtm

Feel free to unhold when you're ready to submit

karlkfi (approved):
/lgtm

Thanks!

PR #581, opened 2023-05-09
Bump the test prometheus memory request to 1.5Gi

nan-yu:
We're still seeing OOMKilled issues for the Prometheus Pod running on Autopilot rapid-latest channel after clearing the cache.

This commit further increases the memory request to 1.5GiB.


karlkfi (approved):
/lgtm

WCGW?
(We know the nodes/clusters have enough spare memory, right?)

PR #583, opened 2023-05-09
Use watch after deleting objects in tests

karlkfi:


e2e/testcases/oci_sync_test.go
```diff
@@ -325,6 +325,18 @@ func TestSwitchFromGitToOci(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
+	implictNs := &corev1.Namespace{}
+	implictNs.Name = namespace
+
+	t.Cleanup(func() {
+		// Stop the Config Sync webhook and delete the implicit namespace manually.
+		// The default cleanup won't do it after we change to use invalid specs.
```
sdowell, line 333: Maybe instead, the cleanup block could fix the spec to be valid?

karlkfi: It occurs to me that the updated comment is wrong. The finalizer will delete anything in the RG, even if the current spec is invalid. It just won't delete objects with one of the deletion prevention annotations. And the implicit namespace is given the `client.lifecycle.config.k8s.io/deletion: detatch` annotation. So that's actually why we need to delete it in cleanup.

I think the remediator might actually un-delete it tho...

sdowell (approved):

PR #584, opened 2023-05-09
add commit log to release tag script

sdowell:
this is intended to provide more context to the releaser when they are running the script


nan-yu (approved):
/lgtm

PR #587, opened 2023-05-09
Fake client enhancements

karlkfi:
- Add Watch impl
- Add Finalizer & deletionTimestamp impl
- Ignore Limit requests, instead of erroring. This allows informers to work. They just get all the objects in the first request, instead of paginating. We can add pagination support later, if needed.
- Treat both DeletePropagationForeground and DeletePropagationBackground requests as Foreground, instead of erroing. We can add background support later, if needed.

### DIFF
--- a/pkg/reconciler/finalizer/reposync_finalizer_test.go
+++ b/pkg/reconciler/finalizer/reposync_finalizer_test.go
@@ -63,12 +63,10 @@ func TestRepoSyncFinalize(t *testing.T) {
 		cmpopts.IgnoreFields(metav1.Time{}, "Time"),
 	)
 
-	var fakeClient *fake.Client
-
 	testCases := []struct {
 		name                       string
 		rsync                      client.Object
-		setup                      func() error
+		setup                      func(*fake.Client) error
 		destroyErrs                status.MultiError
 		expectedRsyncBeforeDestroy client.Object
 		expectedError              error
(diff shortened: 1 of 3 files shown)


pkg/syncer/syncertest/fake/client.go
```diff
@@ -388,8 +501,30 @@ func (c *Client) Delete(_ context.Context, obj client.Object, opts ...client.Del
...
+		pp := metav1.DeletePropagationBackground
+		options.PropagationPolicy = &pp
+	}
+
+	switch *options.PropagationPolicy {
+	case metav1.DeletePropagationForeground:
+		c.deleteManagedObjects(id)
+		delete(c.Objects, id)
+	case metav1.DeletePropagationBackground:
+		// TODO: implement thread-safe background deletion propagation.
```
sdowell, line 515: How important is proper emulation of this behavior for the quality of our unit tests (in your opinion)?

pkg/reconciler/finalizer/reposync_finalizer_test.go
```diff
@@ -239,9 +237,17 @@ func TestRepoSyncFinalize(t *testing.T) {
...
 			name:  "rsync not found",
 			rsync: repoSync1.DeepCopy(),
-			setup: func() error {
+			setup: func(fakeClient *fake.Client) error {
+				// remove the finalizer that blocks deletion
+				ctx := context.Background()
+				rs := &v1beta1.RepoSync{}
+				rs.Name = repoSync1.Name
+				rs.Namespace = repoSync1.Namespace
+				if err := updateToRemoveFinalizers(ctx, fakeClient, rs); err != nil {
```
sdowell, line 246: Why do we need to manually remove the finalizers? Will they not get resolved on their own when using the fake client?

karlkfi:
/retest 

flake:
```
=== CONT  TestCRDDeleteBeforeRemoveCustomResourceV1
    custom_resources_test.go:257: ERROR: timed out waiting for metrics predicate[0]: no results from prometheus query: config_sync_resource_conflicts_total{commit="22be3d37e14c584e77c869f32cf6607a46299332", component="otel-collector", k8s_pod_name="root-reconciler-6945567586-tqdlx"}
```

pkg/syncer/syncertest/fake/client.go
karlkfi: Not very. None of the unit tests depend on that behavior yet. Background is just the default. And since the garbage collector can run at any time, there's no reason to assume, even if you deliberately selected background, that the objects are or are not yet deleted.

The foreground behavior is more important. When you use foreground you assume the managed objects are deleted before the client returns a successful response.

pkg/reconciler/finalizer/reposync_finalizer_test.go
karlkfi: These are Finalizer object unit tests, not integration tests. The ones here where I'm deleting the finalizer and then deleting the RSync aren't running the full finalizer.Controller, so they don't respond to deletion events. They're only testing a specific method on the Finalizer object. In this case it's `Finalizer.Finalize`, which would normally be executed by the finalizer.Controller.

sdowell (approved):

PR #592, opened 2023-05-10
Rewrite sync controller garbage collection

karlkfi:
- Convert RootSync controller to manage cleanup the same way the
  RepoSync controller does: delete everything after the RSync is
  Not Found, instead of using the K8s Garbage collector.
  This makes the behavior and code more consistent, in prep for the
  upcoming reconciler-manager finalizer.
- Combine cleanup methods in the reconcilerBase class.
- Extract upsertManagedObjects and deleteManagedObjects methods
- Log & skip managed object upserts when an RSync is being deleted.

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/594
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/597
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/598
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/637
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/640

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -63,6 +63,7 @@ import (
 type RepoSyncReconciler struct {
 	reconcilerBase
 
+	// lock ensures that the Reconcile method only runs one at a time.
 	lock sync.Mutex
 }
 
@@ -105,16 +106,21 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 	rs := &v1beta1.RepoSync{}
 
 	if err := r.client.Get(ctx, rsRef, rs); err != nil {
-		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
 		if apierrors.IsNotFound(err) {
-			// Namespace controller resources are cleaned up when reposync no longer present.
-			//
-			// Note: Update cleanup resources in cleanupNSControllerResources(...) when
-			// resources created by namespace controller changes.
-			cleanupErr := r.cleanupNSControllerResources(ctx, rsRef, reconcilerRef)
-			// if cleanupErr != nil, the request will be requeued.
-			return controllerruntime.Result{}, cleanupErr
+			// Cleanup after already deleted RepoSync.
+			if err := r.deleteManagedObjects(ctx, reconcilerRef, rsRef); err != nil {
+				r.logger(ctx).Error(err, "Failed to delete managed objects")
+				// Failed to delete a managed object.
+				// Return an error to trigger retry.
+				metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+				// requeue for retry
+				return controllerruntime.Result{}, errors.Wrap(err, "failed to delete managed objects")
+			}
+			// cleanup successful
+			metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(nil), start)
+			return controllerruntime.Result{}, nil
 		}
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
 		return controllerruntime.Result{}, status.APIServerError(err, "failed to get RepoSync")
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -169,6 +166,29 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+	if err := r.upsertManagedObjects(ctx, reconcilerRef, currentRS, rs); err != nil {
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+		return controllerruntime.Result{}, err
+	}
+
+	metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(nil), start)
+	return controllerruntime.Result{}, nil
+}
+
+func (r *RepoSyncReconciler) upsertManagedObjects(ctx context.Context, reconcilerRef types.NamespacedName, currentRS, rs *v1beta1.RepoSync) error {
```
sdowell, line 178: nice decomposition, I think it would be good to decompose this function further in the future

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -169,6 +166,29 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+	rsRef := client.ObjectKeyFromObject(rs)
+	r.logger(ctx).V(3).Info("Reconciling managed objects")
+
+	// If Reconciling condition does not exist, set it to "Initialization".
+	// Avoid updating the timestamps on every retry.
+	cond := reposync.GetCondition(rs.Status.Conditions, v1beta1.RepoSyncReconciling)
+	if cond == nil {
+		reposync.SetReconciling(rs, "Initialization", "Setup in progress")
```
sdowell, line 186: Looks like this is new - I'm curious what is the goal?

Looks like this `Initialization` status will exist for a new RSync until the reconciler Deployment is created?

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -1656,14 +1656,6 @@ func TestRepoSyncSwitchAuthTypes(t *testing.T) {
 		t.Fatalf("unexpected reconciliation error, got error: %q, want error: nil", err)
 	}
 
-	wantRepoSyncs := map[types.NamespacedName]struct{}{
-		{Namespace: rs.Namespace, Name: rs.Name}: {},
-	}
-	// compare repoSyncs.
-	if diff := cmp.Diff(wantRepoSyncs, testReconciler.repoSyncs, cmpopts.EquateEmpty()); diff != "" {
```
sdowell, line 1663: Why are these assertions removed?

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -116,11 +115,21 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 	rs := &v1beta1.RootSync{}
 
 	if err := r.client.Get(ctx, rsRef, rs); err != nil {
-		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
 		if apierrors.IsNotFound(err) {
-			r.logger(ctx).Info("Deleting managed objects")
-			return controllerruntime.Result{}, r.deleteClusterRoleBinding(ctx, reconcilerRef)
```
sdowell, line 122: Hmm, so this was only deleting the ClusterRoleBinding before?

karlkfi: The ClusterRoleBinding was the only one that wasn’t being deleted by the k8s GC.

karlkfi: Because it’s not in the same namespace, which is a limitation of ownership.

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: Ah yeah. I added this as the opposite of Finalization. It means the controller has validated the spec but the deployment hasn’t been created yet. 

I’ll rip it out and put it in a follow up with unit tests for it.

pkg/reconcilermanager/controllers/reposync_controller_test.go
karlkfi: The RepoSyncs cache was removed. So it doesn’t need to be validated. 

The cache was preventing garbage collection in scenarios where the reconciler was restarted/rescheduled.

pkg/reconcilermanager/controllers/reposync_controller.go
sdowell: Thank you, that will help with isolating/testing the change

karlkfi: Pulled out. Will make a follow-up with it once this is merged.

sdowell (approved):

PR #593, opened 2023-05-11
Template and e2e test for reconciling notification

victorpras:
Template and e2e test for reconciling notification


e2e/testcases/notification_test.go
```diff
@@ -702,6 +702,70 @@ func TestNotificationOnSyncStalled(t *testing.T) {
 	}
 }
 
+func TestNotificationOnSyncReconciling(t *testing.T) {
+	rsName := "root-sync-5"
```
sdowell, line 706: Just curious, what is the reason for the incrementing number in the name?

victorpras: I think most of these test works very similarly with each other. Giving a unique name for the test data won't affect how the test works, but might reduce the human error factor when working with these.

e2e/testcases/notification_test.go
```diff
@@ -702,6 +702,70 @@ func TestNotificationOnSyncStalled(t *testing.T) {
...
+	records, err := waitForNotifications(nt, 1)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// validate notification
+	require.Equal(nt.T, nomostest.NotificationRecords{
+		Records: []nomostest.NotificationRecord{
+			{
+				Message: "{\n  \"content\": {\n    \"raw\": \"RootSync root-sync-5 is reconciling\"\n  }\n}",
```
sdowell, line 755: This feels a bit too generic, such that it doesn't provide much confidence that the `Reconciling` condition is the cause of the trigger. Perhaps we could include some more information in the template, e.g. the `reason`/`message` from the condition. WDYT?

victorpras: the message has been udpated to include message and reason from the reconciling status.

sdowell (approved):

PR #596, opened 2023-05-11
Update k8s.io & controller-runtime

karlkfi:
- Update k8s.io/* libs to v0.25.9
- Update sigs.k8s.io/controller-runtime to v0.13.1 (latest that uses k8s.io v0.25 and not v0.26)
- Update fake client to error if GetOptions are specified. (Not currently used by our unit tests.)
- Update test-e2e-go to use Go 1.19 (required to build now)


karlkfi:
/retest

Flakey kpt.dev/github pull?

`ERROR: failed to solve: process "/bin/sh -c go install kpt.dev/configsync/cmd/nomos" did not complete successfully: exit code: 2`

karlkfi:
Looks like the nomos install error wasn't a flake. Go 1.19 is now required to build the nomos binary (and probably the rest of the config-sync binaries too). So I updated the e2e image to Go 1.9.

build/test-e2e-go/gke/Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM gcr.io/google.com/cloudsdktool/cloud-sdk:425.0.0-slim as gcloud-install
 RUN apt-get install -y kubectl google-cloud-sdk-gke-gcloud-auth-plugin
 
 # Build e2e image
-FROM golang:1.17.7-alpine as kpt-config-sync-e2e
```
sdowell, line 20: Could we pull out this Go bump as a separate change, e.g. in case we need to revert one of these changes?

karlkfi: The rest of the code was already updated to 1.19, it's just the e2e image here. It's required to build with the newer libraries. So if we want to revert the Go bump it'll require reverting these lib updates too.

karlkfi: But I can pull out the Go image change, if you think it's meaningful.

sdowell: I 'd suggest submitting the Go bump first, but I won't give you a hard time about it

PR #597, opened 2023-05-12
Reduce watch scope for reconciler-manager

karlkfi:
- The backing informer will still watch at cluster scope, but the controller-manager will pre-filter the events and only trigger reconcile for events within the specified namespace. This should reduce the number of reconcile attempts, and thus API calls made by the reconciler-manager.
- Add watches for all objects managed by the reconciler-manager for the RootSync. This should allow the reconciler-manager to revert any drift on the objects it manages, at least on fields it specifies.

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/617
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/618
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/622
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/623
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/624
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/625
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/628
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/629
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/633
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/634


pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -352,9 +352,18 @@ func (r *RootSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
 			MaxConcurrentReconciles: 1,
 		}).
 		For(&v1beta1.RootSync{}).
-		Owns(&appsv1.Deployment{}).
-		Watches(&source.Kind{Type: &corev1.Secret{}},
+		// Custom Watch to trigger Reconcile for objects created by RootSync controller.
```
sdowell, line 355: > Add watches for all objects managed by the reconciler-manager for the RootSync. This should allow the reconciler-manager to revert any drift on the objects it manages, at least on fields it specifies.

Could you add some unit and/or e2e tests for this functionality?

karlkfi: I was afraid you were going to ask this. There's no unit tests that actually spin up these controllers with the controller manager. So I'm going to have to scaffold that up first and solve any problems when trying to run it with the fake client.

pkg/api/configsync/v1alpha1/conversion.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	conversion "k8s.io/apimachinery/pkg/conversion"
+	v1beta1 "kpt.dev/configsync/pkg/api/configsync/v1beta1"
+)
+
+// Convert_v1beta1_HelmRootSync_To_v1alpha1_HelmRootSync converts HelmRootSync
+// from v1beta1 to v1alpha1.
+//
+// This conversion is manual, because it is lossy.
+// The `DeployNamespace` field is in v1beta1, but not v1alpha1
```
sdowell, line 26: I wonder if this was intentional, I expect it may not have been. Maybe we should add the field to v1alpha1

karlkfi: Seems fine to add new fields to new versions. The reason the automation makes you stub it out is to confirm that it's lossy. And in this case, the newer version is the hub/storage version, so nothing is lost when being stored, only if you request the older version.

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: Added with https://github.com/GoogleContainerTools/kpt-config-sync/pull/634

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -356,9 +356,18 @@ func (r *RootSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
...
+		Watches(&source.Kind{Type: withNamespace(&corev1.Secret{}, configsync.ControllerNamespace)},
 			handler.EnqueueRequestsFromMapFunc(r.mapSecretToRootSyncs),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
```
sdowell, line 362: Would you mind adding unit tests for drift correction on each of these resource types?

pkg/api/configsync/v1alpha1/conversion.go
karlkfi: You’re Killin Me Small Sandlot Porter

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: Done. 

Only RepoSync manages secrets tho. 
RootSync just watches the user managed secrets.

karlkfi:
/retest

Looks like the collector failed to start due to port in use. Mighta been after a restart. I have a feeling it doesn’t always shut down cleanly.

sdowell (approved):

PR #598, opened 2023-05-12
Report Stalled condition if reconciler Terminating

karlkfi:
- Error if reconciler deployment status is invalid/unknown
- RecordReconcileDuration in a few extra failure cases

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -306,6 +306,8 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 			if updateErr != nil {
 				r.logger(ctx).Error(updateErr, "Sync status update failed")
 			}
+			// Use the deployment get error for metric tagging.
+			metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
 			return controllerruntime.Result{}, err
 		}
 	}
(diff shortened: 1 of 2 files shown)


karlkfi:
/retest

Interesting TestRootSyncReconcilerDeploymentDriftProtection unit test failure. Passes locally. Doesn't seem related to this change. Might be a new flakey from the previous PR. :(

```
spec.template.spec.serviceAccountName changes expected [root-reconciler-my-root-sync seanboswell root-reconciler-my-root-sync], but found [root-reconciler-my-root-sync]
```

Looks like the watch didn't even see the manual change to `seanboswell`, nor the revert. The watch starts before the update happens...

I have a feeling this is due to the RSync ResourceVersion cache that exits early if the RSync hasn't changed. In this case, only the reconciler Deployment changed, not the RSync. So we might need to remove that "optimization" to pick up other changes.

karlkfi:
Hmm. failed again. concerning.

I'll make another PR to remove the RV optimization.

karlkfi:
Fix for unit test flake: https://github.com/GoogleContainerTools/kpt-config-sync/pull/637

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -306,6 +306,8 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 			if updateErr != nil {
 				r.logger(ctx).Error(updateErr, "Sync status update failed")
 			}
+			// Use the deployment get error for metric tagging.
+			metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
```
sdowell, line 310: Should we just call this in a `defer`? It seems like we call this before every single return - or is the behavior more nuanced than that?

karlkfi: I have a partial for that in https://github.com/GoogleContainerTools/kpt-config-sync/pull/592 which pulls out  upsertManagedObjects and deleteManagedObjects methods.

The gotcha with a defer is changing err to be at the reconcile scope and making sure all the child scopes populate the error in the parent scope, which can be hard to enforce/discover.

sdowell (approved):

PR #599, opened 2023-05-15
test: fix test cleanup for TestSwitchFromGitToOci

sdowell:
The cleanup of this test had a few issues:
- Not adding the deletion propagation annotation to the RepoSync
- Not properly resetting the spec when switching between field managers

This change sets the deletion propagation annotation and ensures that the fields get properly reset when switching field managers.

### DIFF
--- a/e2e/testcases/oci_sync_test.go
+++ b/e2e/testcases/oci_sync_test.go
@@ -16,16 +16,13 @@ package e2e
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
 	"testing"
 
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
@@ -35,6 +32,7 @@ import (
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/declared"
+	"kpt.dev/configsync/pkg/importer/filesystem"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/testing/fake"
@@ -302,10 +300,9 @@ func TestSwitchFromGitToOci(t *testing.T) {
 		// bookinfo repo contains ServiceAccount
 		ntopts.RepoSyncPermissions(policy.RBACAdmin(), policy.CoreAdmin()),
 	)
+	var err error
 	namespace := "bookinfo"
 	managerScope := string(declared.RootReconciler)
-	// file path to the local RepoSync YAML file which syncs from a public Git repo that contains a service account object.
-	rsGitYAMLFile := "../testdata/reconciler-manager/reposync-sample.yaml"
 	// file path to the RepoSync config in the root repository.
 	repoSyncPath := "acme/reposync-bookinfo.yaml"
 	rsNN := types.NamespacedName{
@@ -317,9 +314,11 @@ func TestSwitchFromGitToOci(t *testing.T) {
 
 	rsCR := nt.RepoSyncClusterRole()
 	rsRB := nomostest.RepoSyncRoleBinding(rsNN)
-	repoSyncGit, err := parseRepoSyncFromFile(nt, rsGitYAMLFile)
-	if err != nil {
-		nt.T.Fatal(err)
+	repoSyncGit := nomostest.RepoSyncObjectV1Beta1(rsNN, "", filesystem.SourceFormatUnstructured)
+	repoSyncGit.Spec.Git = &v1beta1.Git{
+		Repo:   "https://github.com/config-sync-examples/namespace-repo-bookinfo",
+		Branch: "main",
+		Auth:   configsync.AuthNone,
 	}
 	// Ensure the RoleBinding & ClusterRole are deleted after the RepoSync
 	if err := nomostest.SetDependencies(repoSyncGit, rsRB, rsCR); err != nil {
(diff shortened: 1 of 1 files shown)


e2e/testcases/oci_sync_test.go
```diff
@@ -397,6 +398,8 @@ func TestSwitchFromGitToOci(t *testing.T) {
 		Image: imageURL,
 		Auth:  configsync.AuthNone,
 	}
+	// Enable deletion propagation for RepoSync
+	nomostest.EnableDeletionPropagation(repoSyncOCI)
```
karlkfi, line 402: RepoSyncObjectV1Beta1 should do this for you

e2e/testcases/oci_sync_test.go
```diff
@@ -475,6 +478,14 @@ func TestSwitchFromGitToOci(t *testing.T) {
 	if err := nt.ValidateNotFound("bookinfo-sa", namespace, &corev1.ServiceAccount{}); err != nil {
 		nt.T.Error(err)
 	}
+	nt.T.Cleanup(func() {
```
karlkfi, line 481: Do we really want to reset to OCI and then reset again to Git? 
The git cleanup should be sufficient, right?

e2e/testcases/oci_sync_test.go
```diff
@@ -302,10 +299,9 @@ func TestSwitchFromGitToOci(t *testing.T) {
 		// bookinfo repo contains ServiceAccount
 		ntopts.RepoSyncPermissions(policy.RBACAdmin(), policy.CoreAdmin()),
 	)
+	var err error
 	namespace := "bookinfo"
 	managerScope := string(declared.RootReconciler)
-	// file path to the local RepoSync YAML file which syncs from a public Git repo that contains a service account object.
-	rsGitYAMLFile := "../testdata/reconciler-manager/reposync-sample.yaml"
```
karlkfi, line 308: Can we delet ethe test file or are we using it elsewhere?

e2e/testcases/oci_sync_test.go
```diff
@@ -317,10 +313,15 @@ func TestSwitchFromGitToOci(t *testing.T) {
...
-	}
+	repoSyncGit := fake.RepoSyncObjectV1Beta1(namespace, configsync.RepoSyncName)
+	repoSyncGit.Spec.SourceType = string(v1beta1.GitSource)
+	repoSyncGit.Spec.Git = &v1beta1.Git{
+		Repo:   "https://github.com/config-sync-examples/namespace-repo-bookinfo",
+		Branch: "main",
+		Auth:   configsync.AuthNone,
+	}
+	// Enable deletion propagation for RepoSync
+	nomostest.EnableDeletionPropagation(repoSyncGit)
```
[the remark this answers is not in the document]
sdowell: I used `nomostest.RepoSyncObjectV1Beta1` initially, but we have to overwrite the git spec anyways so I figured we might as well just build it from scratch using `fake.RepoSyncObjectV1Beta1`. I can switch to that if you prefer

e2e/testcases/oci_sync_test.go
sdowell: This fixes one of the issues with the field manager swapping, where the `oci` spec is leftover if we go directly to `git` - because the `oci` spec was applied from the test suite field manager.

e.g.
```yaml
  spec:
    git:
      auth: none
      branch: main
      period: 0s
      repo: https://github.com/config-sync-examples/namespace-repo-bookinfo
    oci:
      auth: none
      image: us-docker.pkg.dev/stolos-dev/config-sync-test-public/namespace-repo-bookinfo
    sourceType: oci
```

This was causing an issue because the `repoSyncGit` wasn't specifying a `sourceType`, although that issue should be fixed now with the updated `repoSyncGit`. I don't think we strictly _need_ this block but I think it's good form to clean things up in the reverse order which they were applied in.

Another alternative is that this block could delete the suite-created `RepoSync`.

e2e/testcases/oci_sync_test.go
sdowell: It's referenced in a few legacy scripts, but that's the only instance I can find

https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/scripts/multi-repo.sh#L56
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/scripts/debug-reconciler.sh#L107

AFAIK these scripts don't have a use case any more, although I'm not certain

e2e/testcases/oci_sync_test.go
sdowell: Not the `fake.RepoSyncObjectV1Beta1` that is currently used. We would need to switch to `nomostest.RepoSyncObjectV1Beta1` and reconfigure the spec from `git` -> `oci`. I don't mind this refactor, although seems like a bit of a wash with the extra setup required

e2e/testcases/oci_sync_test.go
karlkfi: Check the prow test configs. If they're not there, I say nuke um. 

Maybe check who made them in the old repo and ask if they care.

e2e/testcases/oci_sync_test.go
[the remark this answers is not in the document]
karlkfi: Everywhere is getting it implicitly from the nomostest version. I didn't really want to leak it into all the tests.

e2e/testcases/oci_sync_test.go
karlkfi: Maybe we need a nomostest.RepoSyncObjectV1Beta1WithOCI ?

I'd prefer the tests get the annotation implicitly, without needing to know about it. But this is obviously a bit of an edge case where the cleanup needs to know how it works tho.

e2e/testcases/oci_sync_test.go
karlkfi: I guess this is fine. It's just slower because it has to wait for sync twice.

e2e/testcases/oci_sync_test.go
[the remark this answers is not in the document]
sdowell: Refactored, PTAL

e2e/testcases/oci_sync_test.go
sdowell: Refactored, PTAL

sdowell:
presubmit flake
/retest

karlkfi (approved):
/lgtm

Thanks for tracking this down!

PR #601, opened 2023-05-16
Set Prometheus retention size to limit the memory usage

nan-yu:
The Prometheus Pod consumes more than 1.5GiB memory on Autopilot 1.26 clusters. We tried to clear the cache before each test case (PR#579), but WAL replays the deleted data series, which caused the Pod to crash again.

This commit limits the retention size to 1GB, so that old data can be removed to avoid exceeding the limit.


karlkfi (approved):
/lgtm

sdowell (approved):

PR #603, opened 2023-05-16
ci: add postsubmit targets

sdowell:
These targets will be used by a new postsubmit prowjob to publish artifacts.


nan-yu (approved):
/lgtm

PR #604, opened 2023-05-16
Drop the k8s client api latency metric to fix the OOMKilled issue

nan-yu:
Autopilot 1.26 sends more than 220k data series for the `cilium_k8s_client_api_latency_time_seconds` metric, which caused OOMkilled issue.

This commit drops the metric to fix the CI failures.

b/283001264

### DIFF
--- a/e2e/testdata/prometheus/prometheus.yaml
+++ b/e2e/testdata/prometheus/prometheus.yaml
@@ -77,6 +77,10 @@ data:
       - job_name: 'kubernetes-pods'
         kubernetes_sd_configs:
         - role: pod
+        metric_relabel_configs:
+        - source_labels: [__name__]
+          regex: 'cilium_k8s_client_api_latency_time_seconds.*'
+          action: drop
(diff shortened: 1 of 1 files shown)


sdowell (approved):

janetkuo (approved):
/lgtm

nan-yu:
Per [Prometheus doc](https://prometheus.io/docs/prometheus/2.37/configuration/configuration/#regex:~:text=One%20use%20for%20this%20is%20to%20exclude%20time%20series%20that%20are%20too%20expensive%20to%20ingest.), we should use `metric_relabel_configs` to drop expensive metrics. However, the drop regex doesn't support [negative lookahead](https://github.com/google/re2/wiki/Syntax#:~:text=(%3F!re)-,before%20text%20not%20matching%20re%20(NOT%20SUPPORTED),-(%3F%3C%3Dre)). Therefore, we just drop the most expensive metric for now.

PR #607, opened 2023-05-17
update vulnerability scan script for CI usage

sdowell:
- Use manifests for getting image tags. This makes it straightforward to pull down the manifest and get all images included with that manifest.
- Fix assumptions about GCP project/registries. It should be assumed that all images are hosted on either GCR/GAR with container scanning enabled. These may also be hosted in different projects/registries, e.g. git-sync, gcenode-askpass-sidecar, etc.

### DIFF
--- a/scripts/pull-postsubmit-retry.sh
+++ b/scripts/pull-postsubmit-retry.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The postsubmit job which publishes artifacts takes some time to complete.
+# This script retries pulling the postsubmit artifacts for a reasonable amount
+# of time and then errors if not found.
+
+set -euo pipefail
+
+# Wait up to 20 minutes for postsubmit artifacts
+n=0
+num_intervals=80
+interval=15
+SECONDS=0
+until [[ "$n" -ge $num_intervals ]]; do
+   make pull-gcs-postsubmit && exit 0
+   echo "++++ Failed to pull postsubmit artifacts. Waiting ${interval} seconds to retry."
+   n=$((n+1))
+   sleep "${interval}"
+done
+
+echo "++++ Postsubmit artifacts not found after retrying for ${SECONDS} seconds"
+exit 1
--- a/scripts/vulnerabilities.sh
+++ b/scripts/vulnerabilities.sh
@@ -16,30 +16,19 @@
 # This script prints a table of fixable vulnerabilities of medium severity or
 # higher, based on the latest or specified git tag.
 #
-# USAGE: $ scripts/vulnerabilities.sh [IMAGE_TAG]
+# USAGE: $ scripts/vulnerabilities.sh
 #
-# The IMAGE_TAG is optional. If not specified, it defaults to the latest tag of
-# the current local environment.
-#
-# The GCP_PROJECT environment variable is used to configure the project that
-# hosts the container images in Google Container Registry.
-# Default: config-management-release
+# The script will look for the rendered manifests at .output/staging to parse
+# image tags.
 #
 # Requires gcloud authn/authz to read the results of vulnerability scans from
-# Google Container Registry in the specified project.
+# GCR/GAR in the specified project.
 # Required access:
-# - Storage Object Viewer on "artifacts.${GCP_PROJECT}.appspot.com"
-# - Container Registry Service Agent on Container Registry API
-#   (containerregistry.googleapis.com)
-# - Container Analysis Service Agent on Container Analysis API
-#   (containerscanning.googleapis.com)
+# - Storage Object Viewer on "artifacts.${GCP_PROJECT}.appspot.com" (GCR - if applicable)
+# - Artifact Registry Reader (GAR - if applicable)
+# - Container Analysis Occurrences Viewer
 
 set -euo pipefail
-
-configsync_latest_image_tag="$(git describe --tags --abbrev=0)"
-configsync_image_tag="${1:-${configsync_latest_image_tag}}"
-
-GCP_PROJECT="${GCP_PROJECT:-config-management-release}"
 
 # find the tag for the gcenode-askpass-sidecar image by reading it from a Go constant.
 # This sidecar is injected at runtime. So it's not in any local manifest files.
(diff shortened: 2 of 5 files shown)


scripts/vulnerabilities.sh
```diff
@@ -50,39 +42,34 @@ find_askpass_image_tag() {
...
-  "gcr.io/${GCP_PROJECT}/hydration-controller"
-  "gcr.io/${GCP_PROJECT}/hydration-controller-with-shell"
-  "gcr.io/${GCP_PROJECT}/oci-sync"
-  "gcr.io/${GCP_PROJECT}/helm-sync"
-  "gcr.io/${GCP_PROJECT}/nomos"
-)
-
 # Build a full list of images with tags
 images=(
-  "gcr.io/${GCP_PROJECT}/gcenode-askpass-sidecar:${GCENODE_ASKPASS_TAG}"
```
karlkfi, line 67: Why hardcode the project? Don't we want to support deploying to other projects?

scripts/vulnerabilities.sh
```diff
@@ -16,31 +16,23 @@
 # This script prints a table of fixable vulnerabilities of medium severity or
 # higher, based on the latest or specified git tag.
 #
-# USAGE: $ scripts/vulnerabilities.sh [IMAGE_TAG]
```
karlkfi, line 19: The optional image tag was added so you could scan a version that wasn't checked out.
Now you can't do that...

scripts/vulnerabilities.sh
```diff
@@ -50,39 +42,34 @@ find_askpass_image_tag() {
...
-
 # Build a full list of images with tags
 images=(
-  "gcr.io/${GCP_PROJECT}/gcenode-askpass-sidecar:${GCENODE_ASKPASS_TAG}"
+  "gcr.io/config-management-release/gcenode-askpass-sidecar:${GCENODE_ASKPASS_TAG}"
 )
 
 # find the "name:tag" of images in the manifests directory.
 find_manifest_images() {
-  grep -rohE --color=never "image:\s+\S+:\S+" manifests/**/*.yaml |
```
karlkfi, line 72: This requires rendering the manifests beforehand now?

scripts/vulnerabilities.sh
```diff
@@ -91,17 +78,12 @@ echo -n "Scanning" >&2
 # Sum the fixable vulnerabilities with severity CRITICAL, HIGH, or MEDIUM
 for image in "${images[@]}"; do
   echo -n "."
-  if [[ "${image}" == "gcr.io/${GCP_PROJECT}/"* ]]; then
```
karlkfi, line 94: This check was filtering out images that aren't re-hosted in our project. They have to be in our project for image scanning to work.

scripts/pull-postsubmit-retry.sh
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The postsubmit job which publishes artifacts takes some time to complete.
+# This script retries pulling the postsubmit artifacts for a reasonable amount
+# of time and then errors if not found.
+
+set -euo pipefail
+
+# Wait up to 20 minutes for postsubmit artifacts
```
karlkfi, line 22: Blocking a CI job for 20m waiting for another CI to finish seems... inefficient.
They can't be merged together?

If we do decide to do it this way, I would recommend adding some kind of progress reporting, like an echo for each attempt, so we can see in the log how many times it retried. And/or maybe a time print out to see how long we waited.

scripts/vulnerabilities.sh
sdowell: This is released from a different project and should be expected to be in this registry. The actual name of this registry isn't from `GCP_PROJECT` but rather whatever value is hardcoded [here](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/gcenode_askpass_sidecar.go#L47).

This is just a hack to get around the fact that we declare this image in code rather than the manifests. The real solution would be to manage the image in manifests, IMO.

scripts/vulnerabilities.sh
sdowell: The input is the rendered manifest now, so the recommended path would be to pull down the intended manifest from GCS

scripts/vulnerabilities.sh
sdowell: Yes, either built locally or pulled from GCS

scripts/vulnerabilities.sh
sdowell: They don't need to be in a single project for this to work. The assumptions here are that:
- all images should be hosted either in GCR/GAR
- all image registry should have container scanning enabled

I've tested this locally with the current postsubmit artifacts and it works as expected

scripts/pull-postsubmit-retry.sh
sdowell: This should actually be faster than merging the jobs together, as the postsubmit starts immediately after the commit is submitted. The periodic job could start at the same time as the postsubmit, but most likely would start some time after. So the build of the artifacts (postsubmit) should always have a head start on when the periodic job runs. I'm planning to investigate a similar approach for the other periodic jobs.

Sure, I can improve the output.

scripts/vulnerabilities.sh
sdowell: Also as a side note it seems like before this change you did actually need to check out the version to get a valid scan result, since it was using the unrendered manifests from the current checkout (for git-sync et al)

scripts/pull-postsubmit-retry.sh
```diff
@@ -0,0 +1,32 @@
...
+n=0
+until [ "$n" -ge 80 ]
```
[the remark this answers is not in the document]
sdowell: done

scripts/pull-postsubmit-retry.sh
sdowell: Updated the output

scripts/vulnerabilities.sh
```diff
@@ -50,39 +39,34 @@ find_askpass_image_tag() {
...
 # Build a full list of images with tags
 images=(
-  "gcr.io/${GCP_PROJECT}/gcenode-askpass-sidecar:${GCENODE_ASKPASS_TAG}"
+  "gcr.io/config-management-release/gcenode-askpass-sidecar:${GCENODE_ASKPASS_TAG}"
 )
 
 # find the "name:tag" of images in the manifests directory.
 find_manifest_images() {
-  grep -rohE --color=never "image:\s+\S+:\S+" manifests/**/*.yaml |
+  grep -rohE --color=never "image:\s+\S+:\S+" .output/staging/**/*.yaml |
```
karlkfi, line 49: Can you add a check that validates this dir and/or yaml files exist? It should prompt you to render or download the manifests with the make targets to call.

It should probably also print something up front to remind you that it's extracting images from local files at this location, in case you forget and run it on stale manifests without re-rendering.

scripts/vulnerabilities.sh
karlkfi: Does `gcloud beta container images describe` work for both GCR and GAR? I thought they had different gcloud commands for the vuln scanner features.

sdowell: Seems to work for both GCR and GAR from my testing. AFAICT they use the same container analysis api

scripts/vulnerabilities.sh
sdowell: Added a check and some extra logging

scripts/pull-postsubmit-retry.sh
```diff
@@ -0,0 +1,35 @@
...
+interval=15
+SECONDS=0
+until [[ "$n" -ge $num_intervals ]]; do
+   make pull-gcs-postsubmit && exit 0
+   echo "++++ Failed to pull postsubmit artifacts. Waiting ${interval} seconds to retry."
+   n=$((n+1))
+   sleep "${interval}"
+done
+
+echo "++++ Postsubmit artifacts not found after retrying for ${SECONDS} seconds"
```
karlkfi, line 34: What’s updating SECONDS? does this always print 0?

sdowell: It's a special bash variable https://tldp.org/LDP/abs/html/internalvariables.html

karlkfi (approved):
/lgtm

are you planning to change these in monorepo and nomos-operator too?

sdowell:
> are you planning to change these in monorepo and nomos-operator too?

@karlkfi yeah I expect I'll be porting these changes to other repos as well to assist with setting up automation

PR #609, opened 2023-05-17
Add comment for why dropping k8s client api metric

nan-yu:


karlkfi (approved):
/lgtm

Thanks.

FWIW, it sounds like skipping it in discovery might further reduce memory by scraping less.

nan-yu:
> FWIW, it sounds like skipping it in discovery might further reduce memory by scraping less.

Could be, but per https://prometheus.io/docs/prometheus/2.37/configuration/configuration/#regex:~:text=One%20use%20for%20this%20is%20to%20exclude%20time%20series%20that%20are%20too%20expensive%20to%20ingest., using `metric_relabel_configs` seems to be the recommended way to exclude time series.

karlkfi:
> > FWIW, it sounds like skipping it in discovery might further reduce memory by scraping less.
> 
> Could be, but per https://prometheus.io/docs/prometheus/2.37/configuration/configuration/#regex:~:text=One%20use%20for%20this%20is%20to%20exclude%20time%20series%20that%20are%20too%20expensive%20to%20ingest., using `metric_relabel_configs` seems to be the recommended way to exclude time series.

I think you’re missing the point. This isn’t a general purpose Prometheus for monitoring everything on the cluster. It’s exclusively needed for testing. We don’t need it to monitor anything we’re not testing. We don’t need it to monitor cilium at all. We could change how discovery is configured and make it so it’s only monitoring the components we care about. That would reduce memory/cpu use, make it more predictable, and more it more similar in different environments (like GKE vs autopilot)

PR #610, opened 2023-05-17
Template and e2e for pending notification

victorpras:
Template and e2e for pending notification


e2e/testcases/notification_test.go
```diff
@@ -766,6 +766,70 @@ func TestNotificationOnSyncReconciling(t *testing.T) {
...
+	records, err := waitForNotifications(nt, 1)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// validate notification
+	require.Equal(nt.T, nomostest.NotificationRecords{
+		Records: []nomostest.NotificationRecord{
+			{
+				Message: "{\n  \"content\": {\n    \"raw\": \"RootSync root-sync-5 is currently pending.\"\n  }\n}",
```
sdowell, line 819: Same comment here, I'd like to get some more context from the notification template so that we can have a stronger assertion

victorpras: Message and reason from pending status are don't have value, so I added status and type

sdowell: AFAICT that would be unexpected, the `Syncing` condition should have additional values:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/root.go#L381

Can you provide the full condition object that you're observing?

victorpras: looking at the full condition object: 
```
[
map[
lastTransitionTime:2023-05-19T00:07:13Z 
lastUpdateTime:2023-05-19T00:07:13Z 
status:False 
type:Reconciling] 
map[
commit:42ffae9062a3044450556a3bd1a781449ee14f37 
errorSummary:map[] 
lastTransitionTime:2023-05-19T00:07:17Z 
lastUpdateTime:2023-05-19T00:07:17Z 
message:Rendering skipped 
reason:Rendering 
status:True 
type:Syncing]
]
```

it looks like status True and type Syncing actually came from rendering, I'm not really sure if this is the validation condition that we want?

sdowell: I think that's fine. @nan-yu can you confirm this is intended behavior?

In this case I'd suggest including `commit`/`message`/`reason` in the template, so that we can have a stronger assertion

nan-yu: Confirm this is intended behavior.

The syncing condition is updated after the rendering step is done ([code](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/root.go#L285)). `status` is `True` because it will continue with the next step.

victorpras: updated the validation to include commit, message, and reason

sdowell (approved):

PR #611, opened 2023-05-17
Log transient errors as warnings instead of errors

nan-yu:
Transient errors are often red-herrings and will be auto-resolved later. This commit logs them as warnings instead of errors.


pkg/parse/state.go
```diff
@@ -107,7 +107,11 @@ func (s *reconcilerState) reset() {
 // invalidate logs the errors, clears the state tracking information.
 // invalidate does not clean up the `s.cache`.
 func (s *reconcilerState) invalidate(errs status.MultiError) {
-	klog.Errorf("Invalidating reconciler checkpoint: %v", status.FormatSingleLine(errs))
+	if status.HasTransientErrors(errs) {
```
karlkfi, line 110: Will there ever be some transient and some non-transient?

nan-yu: It is possible.

The `invalidate` function is called after `setSourceStatus`: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L164.

If `setSourceStatus` has errors, it will be appended to the transient error.

PR #612, opened 2023-05-18
Nomos status filter separator fix

victorpras:
When user is using `nomos status` command with `--name` filter, the separator "--------------------" is displayed for every RootSync even when the RootSync isn't being printed. This fix will display the separator only when the RootSync is printed in the result.

### DIFF
--- a/cmd/nomos/status/cluster_state.go
+++ b/cmd/nomos/status/cluster_state.go
@@ -54,8 +54,8 @@ func (c *ClusterState) printRows(writer io.Writer) {
 		fmt.Fprintf(writer, "%s%s\t%s\n", util.Indent, c.status, c.Error)
 	}
 	for _, repo := range c.repos {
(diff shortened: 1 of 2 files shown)


cmd/nomos/status/cluster_state.go
```diff
@@ -54,8 +54,8 @@ func (c *ClusterState) printRows(writer io.Writer) {
 		fmt.Fprintf(writer, "%s%s\t%s\n", util.Indent, c.status, c.Error)
 	}
 	for _, repo := range c.repos {
-		fmt.Fprintf(writer, "%s%s\n", util.Indent, util.Separator)
 		if name == "" || name == repo.syncName {
+			fmt.Fprintf(writer, "%s%s\n", util.Indent, util.Separator)
```
nan-yu, line 58: Can you add unit tests for the printRows output?

victorpras: unit test with name filter using printRows has been added

PR #620, opened 2023-05-25
fix: move prerequisites to pull-gcs recipe

sdowell:
The clean/output_dir targets are required by the underlying pull-gcs target, which pull-gcs-postsubmit uses as a building block.


karlkfi (approved):
/lgtm

PR #622, opened 2023-05-26
Add conversion-gen for configsync.gke.io/v1alpha1

karlkfi:
- This is required to support v1alpha1 <-> v1beta1 conversion.
- The conversion from v1beta1 -> v1alpha1 is lossy, because v1alpha1 does not support specifying spec.helm.deployNamespace
- This conversion unblocks automatic conversion to the default version by the fake.Client.


scripts/generate-clientset.sh
```diff
@@ -120,7 +123,7 @@ echo "deepcopy"
 "${GOBASE}/bin/deepcopy-gen" \
   ${LOGGING_FLAGS} \
   --input-dirs="${informer_inputs}" \
-  --output-file-base="types.generated" \
+  --output-file-base zz_generated.deepcopy \
```
sdowell, line 126: I was expecting this would result in the renaming of a file (`types.generated` -> `zz_generated.deepcopy`), but I don't see any such files. What is this change doing?

karlkfi: hmm. you'e right. I was also expecting that but forgot cause it was like a week ago.
I wonder if it didn't regen any because it found them in the package...

karlkfi: Ah. It's because they're already named that...

https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/api/configsync/v1alpha1/zz_generated.deepcopy.go

Looks like they've been named that since import.

sdowell: Ah I see, looks like we aren't generating these and checking for a diff in our presubmit checks

karlkfi: Been named that since 2020, in fact...

sdowell (approved):

PR #627, opened 2023-05-30
move images to test infra registry

sdowell:
This consolidates test infra related images to the test infra registry. This registry hosts images related to general building (e.g. buildenv) as well as e2e test images used for public registry testing. The e2e test images used for private registry testing (e.g. auth testing) are still managed on a per-environment basis so that credential management can be delegated.


e2e/nomostest/testing/env.go
```diff
@@ -16,6 +16,13 @@ package testing
 
 import "os"
 
+const (
+	// TestInfraContainerRegistry is the Config Sync test-infra registry hosted on GCR
+	TestInfraContainerRegistry = "gcr.io/kpt-config-sync-ci-artifacts"
```
nan-yu, line 21: Since GCR will be deprecated soon, we can remove test with GCR hosts, but that would be a separate PR.

nan-yu (approved):
/lgtm

nan-yu (approved):
/lgtm

PR #636, opened 2023-05-31
Rewrite the helm-sync tests to push and cleanup their own charts

natasha41575:
Each helm-sync test will use the project id + cluster name + random string as a chart name and use that chart for its tests.


e2e/testcases/helm_sync_test.go
```diff
@@ -380,3 +409,75 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
...
+		}
+	}
+	if _, err := nt.Shell.Command("sed", "-i", "-e", fmt.Sprintf("s/name: %s/name: %s/g", helmchart, chartName), filepath.Join(tmpDir, "Chart.yaml")).CombinedOutput(); err != nil {
+		nt.T.Fatalf("failed to rename helm chart: %v", err)
+	}
+	if _, err := nt.Shell.Helm("package", tmpDir, "--destination", tmpDir); err != nil {
+		nt.T.Fatalf("failed to package helm chart: %v", err)
+	}
+
+	authCmd := nt.Shell.Command("gcloud", "auth", "print-access-token")
```
natasha41575, line 444: Continuation from discussion in https://github.com/GoogleContainerTools/kpt-config-sync/pull/573#discussion_r1210809193, I slightly modified this so that it pipes the token directly into `helm registry login`. Happy to try other options; of the things that I've tried that seems to work, this seems the most secure so far.

e2e/testcases/helm_sync_test.go
```diff
@@ -380,3 +409,75 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
...
+// removes helm charts created during e2e testing
+func cleanHelmImages(nt *nomostest.NT, chartName string) {
+	if _, err := nt.Shell.Command("gcloud", "artifacts", "docker", "images", "delete", fmt.Sprintf("us-docker.pkg.dev/%s/config-sync-test-ar-helm/%s", nomostesting.GCPProjectIDFromEnv, chartName), "--delete-tags").CombinedOutput(); err != nil {
+		nt.T.Errorf("failed to cleanup helm chart image from registry: %v", err)
+	}
+}
+
+// generates a 5 character long random string to reduce chance of name conflict if multiple tests are running in the
+// same cluster
+func generateRandomString() string {
```
sdowell, line 475: I think something like `BUILD_ID` (prow env var) might be a better fit as a unique string that still has some meaning. Alternatively a timestamp might also provide context at least around when it was created

e2e/testcases/helm_sync_test.go
```diff
@@ -380,3 +409,75 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
 		return chartVersion, nil
 	}
 }
+
+// pushes a new helm chart for use during an e2e test
+func pushHelmChart(nt *nomostest.NT, t *testing.T, helmchart, version string) string {
+	t.Helper()
+
+	nt.T.Log("Push helm chart to the artifact registry")
+	tmpDir := t.TempDir()
```
sdowell, line 418: Use `nt.TmpDir`

e2e/testcases/helm_sync_test.go
```diff
@@ -380,3 +409,75 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
 		return chartVersion, nil
 	}
 }
+
+// pushes a new helm chart for use during an e2e test
+func pushHelmChart(nt *nomostest.NT, t *testing.T, helmchart, version string) string {
```
sdowell, line 414: This looks like it should be a utility in `nomostest`. Also should return an error so that the caller can decide what to do with the error

e2e/testcases/helm_sync_test.go
```diff
@@ -380,3 +409,75 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
...
+		chartName = chartName[len(chartName)-50:]
+		chartName = strings.Trim(chartName, "-")
+	}
+
+	paths, err := filepath.Glob(fmt.Sprintf("../testdata/helm-charts/%s/*", helmchart))
+	if err != nil {
+		nt.T.Fatalf("failed to get chart filepaths: %v", err)
+	}
+	for _, path := range paths {
+		if _, err := nt.Shell.Command("cp", "-r", path, tmpDir).CombinedOutput(); err != nil {
```
sdowell, line 433: I think for these filesystem manipulations (e.g. `cp` and `sed`) it would be preferred to use a pure Go alternative rather than shelling out to another binary

e2e/testcases/helm_sync_test.go
```diff
@@ -380,3 +409,75 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
...
+			nt.T.Fatalf("failed to copy helm chart: %v", err)
+		}
+	}
+	if _, err := nt.Shell.Command("sed", "-i", "-e", fmt.Sprintf("s/name: %s/name: %s/g", helmchart, chartName), filepath.Join(tmpDir, "Chart.yaml")).CombinedOutput(); err != nil {
+		nt.T.Fatalf("failed to rename helm chart: %v", err)
+	}
+	if _, err := nt.Shell.Helm("package", tmpDir, "--destination", tmpDir); err != nil {
```
sdowell, line 440: For each of these `helm` subcommands I think it would be nice to have a layer of abstraction (i.e. a function/method). I think that would help with reuse/readability

e2e/testcases/helm_sync_test.go
```diff
@@ -321,17 +339,21 @@ func TestHelmGCENode(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.Unstructured,
 		ntopts.RequireGKE(t), ntopts.GCENodeTest)
 
+	chartName := pushHelmChart(nt, t, privateCoreDNSHelmChart, privateCoreDNSHelmChartVersion)
+	defer func() {
```
sdowell, line 343: I'd suggest implementing this with a `nt.T.Cleanup` that's registered inside the `pushHelmChart` function, ideally registered before the chart is pushed so that the cleanup is guaranteed to run even if there is an error

e2e/testcases/helm_sync_test.go
```diff
@@ -380,3 +409,75 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
...
+
+	chartName := fmt.Sprintf("%s-%s-%s-%s", helmchart, nomostesting.GCPProjectIDFromEnv, nomostesting.GCPClusterFromEnv, generateRandomString())
+	if len(chartName) > 50 {
+		// the chartName + releaseName is used as the metadata.name of resources in the coredns helm chart, so we must trim this down
+		// to keep it under the k8s length limit
+		chartName = chartName[len(chartName)-50:]
+		chartName = strings.Trim(chartName, "-")
+	}
+
+	paths, err := filepath.Glob(fmt.Sprintf("../testdata/helm-charts/%s/*", helmchart))
```
sdowell, line 428: This fits with the existing layout of statically provisioned helm charts, but I'd like to see this follow a similar pattern as the [Git abstraction](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/gitproviders/repository.go#L561).

i.e. the test adds files to the local repo/workspace and then pushes the artifact to the remote sync source (git server in the case of git, helm registry in the case of helm).

I'd be okay with going this route for now and building on this in the future, but I'd like to at least set up these interfaces/abstractions so that it will be easy to make that change

e2e/testcases/helm_sync_test.go

e2e/testcases/helm_sync_test.go

e2e/testcases/helm_sync_test.go
natasha41575: Moved it out to another file with an abstraction layer!

e2e/testcases/helm_sync_test.go
natasha41575: Fixed

e2e/testcases/helm_sync_test.go
natasha41575: Using BUILD_ID sounds good to me. Do you know how many characters long it is and if we set a unique BUILD_ID for each job? (Want to make sure that if we use BUILD_ID, that the truncation that we do later still makes sense.)

e2e/testcases/helm_sync_test.go
natasha41575: Done.

e2e/testcases/helm_sync_test.go
natasha41575: Yeah that makes sense. I added a new abstraction layer under `nomostest` for this, please feel free to leave feedback on it

e2e/testcases/helm_sync_test.go
sdowell: `BUILD_ID` is unique and provided by prow: https://docs.prow.k8s.io/docs/jobs/#job-environment-variables

e.g. for your presubmit job you can see the BUILD_ID is `1664393314247380992`: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/636/kpt-config-sync-presubmit-e2e-multi-repo-test-group1/1664393314247380992

If we were to use this, we might need to double check that the env var makes it to the test process (e.g. through the [layers of test scaffolding](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/Makefile.e2e.ci#L18)).

I was thinking that using something like BUILD_ID or timestamp might provide a bit more context (e.g. find the job that created the registry). Not a huge deal though if this makes it too complicated

natasha41575: I like your idea of using the prow BUILD_ID, but if we are not sure that it gets propagated to the test process then I think it is better to use a timestamp - I agree that the extra context is beneficial over a random string. I'll make that update, thanks!

e2e/nomostest/helm/helm.go
```diff
@@ -0,0 +1,180 @@
...
+		chartName = chartName[len(chartName)-50:]
+		chartName = strings.Trim(chartName, "-")
+	}
+	return chartName
+}
+
+// removes helm charts created during e2e testing
+func cleanHelmImages(nt *nomostest.NT, chartName string) {
+	if _, err := nt.Shell.Command("gcloud", "artifacts", "docker", "images", "delete", fmt.Sprintf("us-docker.pkg.dev/%s/config-sync-test-ar-helm/%s", nomostesting.GCPProjectIDFromEnv, chartName), "--delete-tags").CombinedOutput(); err != nil {
+		nt.T.Errorf("failed to cleanup helm chart image from registry: %v", err)
```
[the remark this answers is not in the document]
natasha41575: Done, thanks

e2e/testcases/helm_sync_test.go
natasha41575: Updated to use a timestamp

e2e/nomostest/helm/helm.go
```diff
@@ -0,0 +1,177 @@
...
+
+// RemoteHelmChart represents a remote OCI-based helm chart
+type RemoteHelmChart struct {
+	// Shell is a helper utility to execute shell commands in a test.
+	Shell *testshell.TestShell
+
+	// Host is the host URL, e.g. https://us-docker.pkg.dev
+	Host string
+
+	// Registry is the registry URL, e.g. oci://us-docker.pkg.dev/stolos-dev/config-sync-test-ar-helm
```
sdowell, line 45: `stolos-dev` is legacy at this point, suggest using a different example/placeholder

e2e/nomostest/helm/helm.go
```diff
@@ -0,0 +1,177 @@
...
+	if err := remoteHelmChart.Push(); err != nil {
+		return "", err
+	}
+
+	return chartName, nil
+}
+
+// creates a chart name from current project id, cluster name, and timestamp
+func generateChartName(helmchart string) string {
+	chartName := fmt.Sprintf("%s-%s-%s-%s", helmchart, nomostesting.GCPProjectIDFromEnv, nomostesting.GCPClusterFromEnv, timestampAsString())
```
sdowell, line 143: Probably don't need the GCP project in the name, given the artifact registry is already scoped to the project. Might help with truncation if this is omitted

e2e/nomostest/helm/helm.go
natasha41575: Done!

e2e/nomostest/helm/helm.go
natasha41575: > Might help with truncation if this is omitted 

Thanks! That does help a lot. Done.

sdowell (approved):

PR #638, opened 2023-06-01
Nomos status filter separator fix (#612)

victorpras:
When user is using `nomos status` command with `--name` filter, the separator "--------------------" is displayed for every RootSync even when the RootSync isn't being printed. This fix will display the separator only when the RootSync is printed in the result.


tiffanny29631:
Could you populate the commit message a bit to explain what this PR does? Or maybe copy from the original PR

PR #641, opened 2023-06-02
test: various enhancement to e2e UX

sdowell:
- only call os.Exit in top level TestMain. This ensures proper cleanup
- add usage flag to e2e. This enables easier discovery of arguments
- Move GCP args to flags. The env var remains as a default but this allows easier discovery and usage of flags
- Add parallel flag to e2e. The test.parallel flag defaults to GOMAXPROCS which is undesirable for expensive e2e tests where a cluster is created per thread. Override the test.parallel flag with our parallel flag.


e2e/testcases/main_test.go
```diff
@@ -19,33 +19,56 @@ import (
...
+// parameter of `go test`. The `go test` defaults to GOMAXPROCS, which is sane
+// for unit tests but is an undesirable default for e2e tests which create
+// a cluster per thread. Default to 1 instead.
+func setParallelFlag() error {
+	err := flag.Set("test.parallel", strconv.Itoa(e2e.NumParallel()))
+	return err
+}
+
 func TestMain(m *testing.M) {
+	os.Exit(main(m))
```
karlkfi, line 40: Might be cleaner to have main return an error and handle printing it and exiting 1 or 0.

karlkfi: oh, i see m.Run returns a code... bleh.
I suppose you could make an error type that has an error code field and type check it.

sdowell: I would generally agree, however the `m.Run()` method returns an exit code so I think returning an exit code is cleaner in this context

sdowell:
Not sure why it's timing out, tried reproducing locally and didn't see this. going to retry
/retest

karlkfi (approved):
/lgtm

PR #643, opened 2023-06-05
Remove the kubevirt e2e test

haiyanmeng:
The kubevirt e2e test is very expensive to maintain: it is very difficult to clean up completely and correctly
(https://kubevirt.io/user-guide/operations/updating_and_deletion/#deleting-kubevirt).

https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/gatekeeper_test.go tests having a `ConstraintTemplate` CR and a `K8sAllowedRepos` CR in the SOT. The test does not run a gatekeeper controller, but simulates the Gatekeeper's controller behavior by applying the `k8sallowedrepos.constraints.gatekeeper.sh` CRD with kubectl: https://github.com/GoogleContainerTools/kpt-config-sync/blob/f7c498ae97013c726cf50ae302d8ff60b2566c46/e2e/testcases/gatekeeper_test.go#L86-L94. This is good enough for testing purpose.


e2e/testcases/apply_scoped_resource_test.go
```diff
@@ -1,138 +0,0 @@
...
-	"kpt.dev/configsync/e2e/nomostest"
-	"kpt.dev/configsync/e2e/nomostest/kubevirt/v1"
-	"kpt.dev/configsync/e2e/nomostest/metrics"
-	"kpt.dev/configsync/e2e/nomostest/ntopts"
-	"kpt.dev/configsync/e2e/nomostest/retry"
-	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
-	"kpt.dev/configsync/pkg/api/configsync"
-)
-
-func TestApplyScopedResources(t *testing.T) {
```
sdowell, line 31: It seems like we should be capturing the spirit of this test with a different (simpler) resource type rather than removing this test completely

e2e/testcases/apply_scoped_resource_test.go
```diff
@@ -1,138 +0,0 @@
...
-	nt := nomostest.New(t, nomostesting.Reconciliation1, ntopts.Unstructured, ntopts.SkipAutopilotCluster)
-
-	nt.Must(nt.RootRepos[configsync.RootSyncName].Copy("../../examples/kubevirt-compiled/.", "acme"))
```
sdowell, line 34: looks like this test data can be removed

haiyanmeng: The test data is still used in https://github.com/GoogleContainerTools/kpt-config-sync/blob/be033580abc5e9f91ed4164065a87c558ab46cd6/e2e/testcases/cli_test.go#L536.

sdowell: If we refactor this test to use a different resource rather than remove it, sounds like we could also refactor that test to use the same new resource. At that point we could remove the kubevirt resources

e2e/testcases/apply_scoped_resource_test.go
haiyanmeng: @karlkfi , @sdowell , https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/gatekeeper_test.go tests having a `ConstraintTemplate` CR and a `K8sAllowedRepos` CR in the SOT. However, the test does not run a gatekeeper controller, and simulate the Gatekeeper's controller behavior by applying the `k8sallowedrepos.constraints.gatekeeper.sh` CRD with kubectl: https://github.com/GoogleContainerTools/kpt-config-sync/blob/f7c498ae97013c726cf50ae302d8ff60b2566c46/e2e/testcases/gatekeeper_test.go#L86-L94.

haiyanmeng: Thinking more about this, simulating the Gatekeeper's controller behavior by applying the k8sallowedrepos.constraints.gatekeeper.sh CRD with kubectl is enough for testing purpose.

e2e/nomostest/clean.go
```diff
@@ -103,10 +101,6 @@ func Clean(nt *NT) error {
 	if err := uninstallConfigSync(nt); err != nil {
 		return err
 	}
-	// Delete the Kubevirt resources
-	if err := deleteKubevirt(nt); err != nil {
```
karlkfi, line 107: Still need to keep this to clean up clusters that errored during a previous run with the kubevirt applied.

karlkfi: Add a TODO to delete the cleanup in the future, once all the test clusters are clean.

haiyanmeng: Good point. Done.

karlkfi (approved):
/lgtm

PR #645, opened 2023-06-06
test: support parallel e2e tests on GKE

sdowell:
This change adds support for running the e2e tests in parallel on GKE. This uses a similar approach as the parallel kind clusters, where the clusters are created at the beginning of the tests suite and reused between tests. The clusters are torn down by cleanup at the end of the test suite.


e2e/nomostest/ntopts/gke.go
```diff
@@ -27,8 +29,18 @@ import (
...
-		getGKECredentials(t, opt.KubeconfigPath)
+		clusterName := *e2e.GCPCluster
+		if *e2e.CreateClusters {
+			opt.IsEphemeralCluster = true
+			clusterName = opt.Name
+			t.Cleanup(func() {
+				deleteGKECluster(t, clusterName)
+			})
+			createGKECluster(t, clusterName)
+			// TODO: Investigate what we need to wait for before cluster is healthy
```
karlkfi, line 40: Clusters often take much longer than a minute to come up.
The Gcloud SDK (and CLI) have a thing called Operations. The creation commands should return an operation number and then you can wait until the operation is complete. The GKE create operation usually includes apiserver and controller readiness.

karlkfi: Use `--async` to return immediately and print the operation number. Then you can run:
https://serverfault.com/a/962032

The operational wait can also be used for deletion.

karlkfi: Note: If you don't use async, the create command can timeout and auto-delete the cluster.

e2e/nomostest/ntopts/gke.go
```diff
@@ -27,8 +29,18 @@ import (
 func GKECluster(t testing.NTB) Opt {
```
karlkfi, line 29: Suggest extracting a GKECluster struct with CRUD methods to a sub-package.

e2e/nomostest/ntopts/gke.go
sdowell: Looks like the issue here, as we discussed offline, is an issue with the default machine type being insufficient for some of the components we deploy in the tests. After setting the default node pool to be the same shape as what we use in the other e2e tests, I no longer observed this issue.

Note that the link you provided is specifically for provisioning node pools rather than the whole GKE cluster. I'm open to using `--async` and monitoring the operation manually, but I think it will increase the code complexity. I'm inclined to leave this as-is and revisit using `--async` if we have an observable use-case/need (e.g. if we see issues after running this in CI)

e2e/nomostest/ntopts/gke.go
sdowell: I moved GKE/Kind out to a sub-package and have some ideas to define a common interface that can be implemented by both GKE and Kind clusters. This is looking like a bigger change though so I'd like to implement that as a followup PR

karlkfi (approved):
/lgtm

PR #650, opened 2023-06-07
Rewrite reconciler-manager error handling

karlkfi:
- Extract setup method to wrap upsertManagedObjects and handle updating the sync object status conditions.
- Add ObjectOperationError & ObjectReconcileError to pass more context about errors between abstraction layers.
- Update upsert and delete methods to return these new errors.
- Add CreateOrUpdate func that returns ObjectOperationErrors where applicable.
- Avoid updating RoleBinding and ClusterRoleBinding when noting changed.

### DIFF
--- a/pkg/reconcilermanager/controllers/errors.go
+++ b/pkg/reconcilermanager/controllers/errors.go
@@ -1,0 +1,175 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controllers
+
+import (
+	"fmt"
+
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// OperationCreate is the create operation
+	OperationCreate = Operation("create")
+	// OperationUpdate is the update operation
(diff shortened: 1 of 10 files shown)


pkg/reconcilermanager/controllers/errors.go
```diff
@@ -0,0 +1,173 @@
...
+func NewObjectOperationError(err error, obj client.Object, op Operation) *ObjectOperationError {
+	id := core.IDOf(obj)
+	if id.GroupKind.Empty() {
+		gvk, lookupErr := kinds.Lookup(obj, core.Scheme)
+		if lookupErr != nil {
+			// Invalid or unregistered resource type - err probably already includes the same message
+			// Fake the kind with the struct type
+			id.Kind = fmt.Sprintf("%T", obj)
+		}
+		id.GroupKind = gvk.GroupKind()
```
sdowell, line 74: Is `gvk` valid if the `lookupErr` is non-nil? I expect perhaps we want this to be in an `else`

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -1594,13 +1612,17 @@ func TestRepoSyncUpdateOverrideAPIServerTimeout(t *testing.T) {
...
 		containerEnvMutator(repoContainerEnv),
+		setResourceVersion("2"),
 	)
 	updatedRepoDeployment.ResourceVersion = "2"
 	wantDeployments[core.IDOf(repoDeployment)] = updatedRepoDeployment
 
 	if err := validateDeployments(wantDeployments, fakeDynamicClient); err != nil {
 		t.Errorf("Deployment validation failed. err: %v", err)
 	}
+	if t.Failed() {
```
sdowell, line 1623: nit: can replace `Errorf` with `Fatalf` to fail now

ditto elsewhere

karlkfi: This whole file is a little inconsistent about that. Sometimes it uses Fatal and sometimes it uses Error and fails later.

I think the goal was to show multiple errors when there was one, perhaps to help debug after a Reconcile. So I tried to make sure there was a Fail if Failed before every Log printing success. Some have multiple tests after the Reconcile and some one have one. So I left the check the same for all of them.

pkg/reconcilermanager/controllers/errors.go
karlkfi: fixed

sdowell (approved):

PR #651, opened 2023-06-07
test: refactor e2e clusters to interface

sdowell:
This provides a common interface for use by both GKE and kind clusters. This is intended to provide a clearer abstraction and consistent behavior between the cluster types.


e2e/nomostest/client.go
```diff
@@ -99,21 +99,69 @@ func newScheme(t testing.NTB) *runtime.Scheme {
...
+		cluster = &clusters.KindCluster{
+			T:                 t,
+			Name:              clusterName,
+			KubeConfigPath:    opts.KubeconfigPath,
+			TmpDir:            opts.TmpDir,
+			KubernetesVersion: *e2e.KubernetesVersion,
+		}
 	case e2e.GKE:
-		clusters.GKECluster(t)(opts)
+		if !*e2e.CreateClusters {
```
karlkfi, line 133: could this be moved out of the gke part? could this work with existing kind clusters, or other pre-existing clusters?

e2e/nomostest/clusters/gke.go
```diff
@@ -21,26 +21,34 @@ import (
...
+	T testing.NTB
+	// Name is the name of the cluster
+	Name string
+	// KubeConfigPath is the path to save the kube config
+	KubeConfigPath string
+}
+
+// Create the GKE cluster
+func (c *GKECluster) Create() error {
+	return createGKECluster(c.T, c.Name)
```
karlkfi, line 41: can these functions just be inlined?

e2e/nomostest/client.go
sdowell: Yeah we could generalize this to support providing the name of a pre-existing kind cluster.

We would need to modify the UX of the command line args a bit, as we currently use the `--gcp-cluster`/`GCP_CLUSTER` arg for passing in the name of a pre-existing GKE cluster. I figure we would probably want a single "cluster name" arg.

That being said, I'm not sure we really have a use case for providing a pre-provisioned kind cluster as it's relatively inexpensive to create a new cluster. However I do like the idea of pushing for more consistent behavior between the cluster types, as the small differences in behavior tend to become a pain point (e.g. things pass in kind and then explode when trying to run on gke)

e2e/nomostest/clusters/gke.go
sdowell: They could - I opted to call out to the nested functions because I figured it might help with readability. But I can move the implementation inline

karlkfi (approved):
/lgtm

PR #653, opened 2023-06-08
test: use WI for helm-sync tests

sdowell:
We have a targeted test for verifying token auth with the helm-sync container. For other tests, it is preferred to use workload identity auth as it is simpler to manage.


karlkfi (approved):
/lgtm

It'd be nice if these worked on Kind, which doesn't support WI, but I guess it's a fine optimization for now, since these only run on GKE.

sdowell:
> It'd be nice if these worked on Kind, which doesn't support WI, but I guess it's a fine optimization for now, since these only run on GKE.

Agreed. Ideally for these tests which aren't specifically testing features of GCP/AR we use an in-cluster/hermetic registry (analogous to the in-cluster git server) so that we can run the test identically on arbitrary k8s clusters without external dependencies.

For now these tests already require GKE to run, and the exported SA key is a bit of a pain to manage. I'd like to isolate our usage of the exported SA key to just targeted tests

PR #654, opened 2023-06-08
fix: install/use binaries from local build dir

sdowell:
The install-helm target was installing helm into a nested subdirectory, causing it to not be discoverable in the PATH. This refactors the target to reuse the buildenv-dirs target, which correctly creates the directory.

Also set the PATH in the makefile so that we use the correct binaries, prioritizing the local build dir, then the go bins, then the parent PATH.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,9 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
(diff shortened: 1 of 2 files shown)


nan-yu (approved):
/lgtm

PR #655, opened 2023-06-09
ci: add make targets for parallel gke e2e

sdowell:
Adds a series of make targets to support running the newly added parallel gke e2e tests in CI.

Add Dockerfile for gke-e2e image. This is based on and intended to replace the test-e2e-go/gke image. The latter image is rebuilt on every CI test execution and invoked using docker-in-docker inside the prowjob. The new prowjob is intended to not require docker-in-docker, and will use this new image which is built once and used many times. This is intended to simplify and reduce the runtime of the prowjob. Versioned dependencies such as kustomize and helm are still installed at runtime.

This change adds some make targets which are similar to existing targets - this is intentional. New prowjobs will be defined which use the new targets, and the old targets will be cleaned up once the existing prowjobs are retired.

### DIFF
--- a/Makefile.oss.prow
+++ b/Makefile.oss.prow
@@ -13,28 +13,46 @@
 # limitations under the License.
 
 # special prow related things
+# Images are tagged with an arbitrary semver and git SHA. The git SHA ensures
+# unique tags and makes the images traceable back to a commit. The semver should
+# be revved upon any major changes to the image.
 
-# This is the image that our prow nomos-presubmit-e2e job runs so that it can
-# use KIND instead of setting up a GKE cluster per test run.
+# This is the image that our prow kind/e2e presubmit job runs so that it can
+# use KIND instead of setting up a GKE cluster per test run. Based on kubekins-e2e
+# for docker-in-docker support to build docker images and run kind.
 
 # Note: nothing builds this, if you update the version, you will need to rebuild manually.
-# For the version tag, concatenate the kubekins version and KIND version
-E2E_TEST_IMAGE_OSS_PROW_KUBEKINS_REGISTRY := k8s-staging-test-infra
-E2E_TEST_IMAGE_OSS_PROW_KUBEKINS := v20230309-9a6b1b3121-1.23
-E2E_TEST_IMAGE_OSS_PROW_KIND := v0.14.0
-E2E_TEST_IMAGE_OSS_PROW_TAG := kubekins-e2e-$(E2E_TEST_IMAGE_OSS_PROW_KUBEKINS)-kind-$(E2E_TEST_IMAGE_OSS_PROW_KIND)
-E2E_TEST_IMAGE_OSS_PROW := gcr.io/oss-prow-build-kpt-config-sync/e2e-prow:$(E2E_TEST_IMAGE_OSS_PROW_TAG)
-image-oss-prow: build/prow/kubekins-kind/Dockerfile
-	@echo "+++ Building the image for the oss-prow presubmit run"
+# TODO: setup postsubmit job to build this image
+KUBEKINS_E2E_TAG := v1.0.0-$(shell git rev-parse --short HEAD)
+KUBEKINS_E2E_IMAGE := $(TEST_INFRA_REGISTRY)/kubekins-e2e:$(KUBEKINS_E2E_TAG)
+.PHONY: build-kubekins-e2e
+build-kubekins-e2e: build/prow/kubekins-e2e/Dockerfile
+	@echo "+++ Building $(KUBEKINS_E2E_IMAGE)"
 	docker buildx build \
-		-t $(E2E_TEST_IMAGE_OSS_PROW) \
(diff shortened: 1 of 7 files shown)


sdowell:
looks like a flaky unit test.
```
--- FAIL: TestRun (0.74s)
    --- FAIL: TestRun/successful_read (0.14s)
        run_test.go:294: 
            	Error Trace:	/go/src/kpt.dev/configsync/pkg/parse/asserter.go:51
            	            				/go/src/kpt.dev/configsync/pkg/parse/asserter.go:59
            	            				/go/src/kpt.dev/configsync/pkg/parse/run_test.go:294
            	Error:      	
            	            	Expected
            	            	    <string>: Syncing
            	            	to deeply equal
            	            	    <string>: Sync Completed
            	            	Diff (- Expected, + Actual):
            	            	      string(
            	            	    - 	"Sync Completed",
            	            	    + 	"Syncing",
            	            	      )
            	            	    
            	Test:       	TestRun/successful_read
            	Messages:   	[successful read] unexpected syncing message return
```
/retest

build/prow/gke-e2e/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Copyright 2022 Google LLC
```
haiyanmeng, line 1: 2022 -> 2023

sdowell: fixed

Makefile.oss.prow
```diff
@@ -13,28 +13,46 @@
...
-# For the version tag, concatenate the kubekins version and KIND version
-E2E_TEST_IMAGE_OSS_PROW_KUBEKINS_REGISTRY := k8s-staging-test-infra
-E2E_TEST_IMAGE_OSS_PROW_KUBEKINS := v20230309-9a6b1b3121-1.23
-E2E_TEST_IMAGE_OSS_PROW_KIND := v0.14.0
-E2E_TEST_IMAGE_OSS_PROW_TAG := kubekins-e2e-$(E2E_TEST_IMAGE_OSS_PROW_KUBEKINS)-kind-$(E2E_TEST_IMAGE_OSS_PROW_KIND)
-E2E_TEST_IMAGE_OSS_PROW := gcr.io/oss-prow-build-kpt-config-sync/e2e-prow:$(E2E_TEST_IMAGE_OSS_PROW_TAG)
-image-oss-prow: build/prow/kubekins-kind/Dockerfile
-	@echo "+++ Building the image for the oss-prow presubmit run"
+# TODO: setup postsubmit job to build this image
+KUBEKINS_E2E_TAG := v1.0.0-$(shell git rev-parse --short HEAD)
```
karlkfi, line 26: wont this mean you have to build it for every commit? otherwise the tag wont exist, even if nothing changed

build/prow/kubekins-e2e/Dockerfile
```diff
@@ -22,11 +22,11 @@
...
 
-FROM gcr.io/${KUBEKINS_REGISTRY}/kubekins-e2e:${KUBEKINS}
-
-ARG KIND_RELEASE="v0.14.0"
-RUN wget -q -O /bin/kind https://kind.sigs.k8s.io/dl/${KIND_RELEASE}/kind-linux-amd64
+# TODO: using a pre-installed kind binary makes it very painful to upgrade the
+# kind version. Update the e2e tests to either:
+# - install kind binary at runtime
+# - consolidate kind usage to Go dependency rather than mixing usage with binary
+RUN wget -q -O /bin/kind https://kind.sigs.k8s.io/dl/v0.14.0/kind-linux-amd64
```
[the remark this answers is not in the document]
sdowell: It was hard coded in the Makefile so that it could be set as both the image tag and passed in as a build arg. Now that it's not being used in the image tag, I just moved the hard coding into the Dockerfile. As the TODO mentions I think we should eventually update this so that we aren't installing the kind binary into the prow image. It makes upgrading to versions pretty much impossible without breaking CI

Makefile.oss.prow
karlkfi: I'm mildly concerned people will just re-build this without incrementing the semver. But i guess it's not a big deal

build/prow/kubekins-e2e/Dockerfile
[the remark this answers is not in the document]
karlkfi: I'm fine with not hard coding the values in the Makefile. I kinda liked the args here because it called out that they were versions that needed to be updated periodically, but I guess it's not a big deal.

karlkfi (approved):
/lgtm

Makefile.oss.prow
sdowell: That's fair. I'm not as opinionated about the semver, but I'd like to have the git SHA in the tag so that images can be traced back to the commit which built them

build/prow/kubekins-e2e/Dockerfile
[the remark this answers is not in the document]
sdowell: I can put the arg back in the makefile, just figured this would be simpler if the var is only used for passing into the docker build

PR #657, opened 2023-06-09
test: enable workload identity for new gke clusters

sdowell:
The cluster creation was missing the workload identity option, causing tests which require workload identity to fail.

### DIFF
--- a/e2e/nomostest/clusters/gke.go
+++ b/e2e/nomostest/clusters/gke.go
@@ -85,7 +85,10 @@ func createGKECluster(t testing.NTB, name string) error {
 	} else {
 		args = append(args, "create")
 	}
-	args = append(args, name, "--project", *e2e.GCPProject)
+	args = append(args, name,
+		"--project", *e2e.GCPProject,
+		"--workload-pool", fmt.Sprintf("%s.svc.id.goog", *e2e.GCPProject),
+	)
 	if *e2e.GCPZone != "" {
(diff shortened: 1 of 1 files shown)


e2e/nomostest/clusters/gke.go
```diff
@@ -85,7 +85,10 @@ func createGKECluster(t testing.NTB, name string) error {
 	} else {
 		args = append(args, "create")
 	}
-	args = append(args, name, "--project", *e2e.GCPProject)
+	args = append(args, name,
```
karlkfi, line 88: do you want these to have fleet membership too?

sdowell: We don't configure that with cluster creation afaict. It looks like we register the fleet memberships in code here:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/csr_auth_test.go#L394

e2e/nomostest/clusters/gke.go
```diff
@@ -85,7 +85,10 @@ func createGKECluster(t testing.NTB, name string) error {
...
+		"--project", *e2e.GCPProject,
```
karlkfi, line 89: should we always specify the project?

karlkfi: oh we are...

karlkfi (approved):
/lgtm

PR #658, opened 2023-06-09
ci: fix workload identity for reposync

sdowell:
There is no workload identity binding for ns reconcilers, which was causing this test to fail. This change adds a workload identity binding for repo-sync and updates the test to create the correctly named RepoSync.

Also removes the AR/CSR reader permissions from the default compute SA, as this was causing false positives when testing with a cluster that did not have workload identity enabled.


tiffanny29631 (commented):
/lgtm
/approve

PR #659, opened 2023-06-09
test: fix usage of flag pointers in tests

sdowell:
The flag pointers are updated after module initialization, causing these string to be evaluated with the default value instead of the parsed value. This works when the desired value is the default or and environment variable, but not when the desired value is the command line flag. Update the variables to functions so that the pointer value is evaluated after the flag has been parsed.


karlkfi (approved):
/lgtm

PR #663, opened 2023-06-10
ci: use us.gcr.io for private GCR image

sdowell:
The tests assume the GCR host is us.gcr.io rather than gcr.io. This error was introduced with a recent change.


karlkfi:
Why do they assume us? Can we fix that?

sdowell:
It's hardcoded here:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/oci_sync_test.go#L62

We could pass in an argument to make it configurable, similar to the make targets which publish them

karlkfi:
Any idea why it was hardcoded that way? commit message share any insight?

sdowell:
> Any idea why it was hardcoded that way? commit message share any insight?

@karlkfi Looks like this change comes from [this PR](https://github.com/GoogleContainerTools/kpt-config-sync/pull/121). I assume the reason the location is set is because GCR lets you set visibility (public/private) based on location. In this case `gcr.io` was public and `us.gcr.io` was private in the test project.

This has since changed because the public registries are hosted in a dedicated test-infra project now, so it can be assumed that the registries in the test environment are private. I've updated this change to have the tests use gcr.io rather than updated the make target to publish to us.gcr.io.

PR #665, opened 2023-06-12
tweak memory request for some helm sync tests

natasha41575:
follow up experimentation https://github.com/GoogleContainerTools/kpt-config-sync/pull/662 - trying to find the minimum resource request we can make and still have these tests pass in CI


haiyanmeng (approved):

sdowell:
Is this indicative of a performance regression? I'm wondering if this may be a breaking change for some users

natasha41575:
@sdowell You are correct that there is a huge performance hit that we should fix. I will try to address it.

PR #667, opened 2023-06-12
ci: observe NUM_CLUSTERS env var for e2e

sdowell:


e2e/flags.go
```diff
@@ -90,7 +90,7 @@ var TestFeatures = flag.String("test-features", "",
 // NumClusters is the number of clusters to run tests on. Each cluster only has
 // a single test running on it at a given time. The number of clusters equals the
 // number of test threads which can run concurrently.
-var NumClusters = flag.Int("num-clusters", 1,
+var NumClusters = flag.Int("num-clusters", util.EnvInt("NUM_CLUSTERS", 1),
```
karlkfi, line 93: do we want some kind of prefix for e2e env vars, like `E2E_`

karlkfi: E2E_FLEET_SIZE would also be ok

sdowell: yes dad

sdowell: consider the E2E prefix slapped

karlkfi (approved):
/lgtm

PR #668, opened 2023-06-13
ci: copy nomos binary into path

sdowell:
This change adds a copy-cli target which copies the nomos binary from the os/arch subdirectory into the top level bin directory. This enables it to be in the PATH for subsequent commands. Call copy-cli after pulling postsubmit artifacts so that the binary can be used. This enables the periodic CI jobs to use the binary that is pulled from the postsubmit artifacts.


tiffanny29631 (approved):
/lgtm

PR #669, opened 2023-06-13
test: add bookstore namespace to sync order tests

sdowell:
These namespaces were being created implicitly, and thus not being deleted by the end of the test. This change adds the namespace to the git repo so that it will be created explicitly and deleted at the end of the test.


karlkfi:
Should the reset code delete implicit namespaces?
Deletion propagation won't but we could add extra code to do so.

sdowell:
Yeah might be reasonable to delete implicit namespace in the reset code, although I think it's still good form for the tests to clean up after themselves where possible

sdowell:
agreed

karlkfi (approved):
/lgtm

I think we need to avoid implicit namespaces unless we're explicitly testing them and cleaning them. The consequences are too complex to deal with, especially if multiple reposyncs own the same implicit namespace.

PR #670, opened 2023-06-13
test: log bugreport output in e2e test

sdowell:
This is intended to provide more context on the call to nomos bugreport, particularly when the test fails.


e2e/testcases/cli_test.go
```diff
@@ -850,8 +850,8 @@ func TestCLIBugreportNomosRunningCorrectly(t *testing.T) {
 	cmd := nt.Shell.Command("nomos", "bugreport")
 	cmd.Dir = nt.TmpDir
 	out, err := cmd.CombinedOutput()
+	nt.T.Log(string(out))
```
karlkfi, line 853: does this normally print output when it doesn't error?

sdowell: for context the motivation is trying to debug this failure: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/standard-regular/1668615208798523392

karlkfi: if you're going to print the output, i suggest also printing the command that was executed first.

karlkfi: you might want to make it like the other comannd execs and only print when debug is enabled or there's an error

sdowell: The error comes later in the test - the command succeeds but the output ends up being something other than what is expected. I suppose I could add a cleanup block that prints the output only if the test failed.

I figured the output isn't too verbose and is really the core of the test (there isn't much else going on other than the call to `nomos bugreport`), so it wouldn't hurt to always log

karlkfi: then just add a log for the command so we know what the output is from.

sdowell: added

karlkfi (approved):
/lgtm

PR #673, opened 2023-06-13
test: stricter namespace cleanup

sdowell:
The existence of managed namespaces can cause issues in subsequent tests when they are not cleaned up. Make the reset code fail upon the existence of managed namespaces so that we can quickly discover tests which do not clean up after themselves.

Include implicit namespaces in the namespaces list for the reset code. This ensures implicit namespaces are either deleted if they do not have a manager or error if they do have a manager.


karlkfi (approved):
/lgtm

PR #675, opened 2023-06-13
ci: hack to enable nomos bugreport in pod

sdowell:
There is a known issue around nomos bugreport that breaks workload identity based authentication when running in a Pod. This change adds the known workaround to enable running the command in a Pod. This should be removed once the underlying issue is fixed.


karlkfi (approved):
/lgtm

PR #677, opened 2023-06-14
Make log level configurable in r*sync API

victorpras:
Currently, the only way to configure the log verbosity level is to patch the reconciler Deployment YAML in the ConfigMap. We want to add a new field in the R*Sync API to make reconciler log verbosity level configurable. 

These PR will remove the default verbosity args from the manifest and add it programmatically during reconciling, with the default value if override is not assigned. Unit test is updated to include default args with verbosity and validate args during deployment validation


sdowell:
update commit message with context and remove `[WIP]`

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -452,6 +453,20 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
 				}
 				c.Resources.Limits[corev1.ResourceMemory] = override.MemoryLimit
 			}
+			if override.LogLevel >= 0 {
+				// replace/append args with the new log level
+				index := -1
+				for i, arg := range c.Args {
```
sdowell, line 459: Instead of this find/replace logic, I think it would be simpler to just remove the verbosity arguments from the manifests and declare it in code. We can assign a default value in code if the override is unset

pkg/reconcilermanager/controllers/rootsync_controller_test.go
```diff
@@ -634,6 +634,85 @@ func TestRootSyncCreateWithNoSSLVerify(t *testing.T) {
...
+	rootDeployment = rootSyncDeployment(rootReconcilerName,
+		setServiceAccountName(rootReconcilerName),
+		secretMutator(rootsyncSSHKey),
+		containerResourcesMutator(overrideAllContainerResources),
+		containerEnvMutator(rootContainerEnvs),
+		setUID("1"), setResourceVersion("2"), setGeneration(2),
+	)
+	wantDeployments[core.IDOf(rootDeployment)] = rootDeployment
+
+	if err := validateDeployments(wantDeployments, fakeDynamicClient); err != nil {
```
sdowell, line 708: Is this actually validating that the log level was set? I don't see anything updating the expected deployment or the validator to assert the log level.

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -689,6 +689,62 @@ func TestUpdateNamespaceReconcilerWithOverride(t *testing.T) {
...
+	repoDeployment := repoSyncDeployment(
+		nsReconcilerName,
+		setServiceAccountName(nsReconcilerName),
+		secretMutator(nsReconcilerName+"-"+reposyncSSHKey),
+		containerResourcesMutator(overrideReconcilerAndGitSyncResourceLimits),
+		containerEnvMutator(repoContainerEnv),
+	)
+	wantDeployments := map[core.ID]*appsv1.Deployment{core.IDOf(repoDeployment): repoDeployment}
+
+	if err := validateDeployments(wantDeployments, fakeDynamicClient); err != nil {
```
sdowell, line 740: Same here regarding the assertion

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -23,7 +23,7 @@ import (
 type OverrideSpec struct {
 	// resources allow one to override the resource requirements for the containers in a reconciler pod.
 	// +optional
-	Resources []ContainerResourcesSpec `json:"resources,omitempty"`
+	Specs []ContainerSpec `json:"resources,omitempty"`
```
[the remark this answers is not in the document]
victorpras: there's no docs (I can create one if needed), I had an offline discussion with Nan last week and the plan was to include loglevel in the API and rename the `resource` related variable to be more generic

pkg/api/configsync/v1alpha1/resource_override.go
```diff
@@ -68,8 +68,8 @@ type OverrideSpec struct {
 	EnableShellInRendering *bool `json:"enableShellInRendering,omitempty"`
 }
 
-// ContainerResourcesSpec allows to override the resource requirements for a container
-type ContainerResourcesSpec struct {
+// ContainerSpec allows to override the resource requirements for a container
+type ContainerSpec struct {
```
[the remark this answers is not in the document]
victorpras: must've missed this one

victorpras:
converting to draft, will follow up when I'm back from PTO

pkg/api/configsync/v1alpha1/resource_override.go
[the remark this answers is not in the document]
karlkfi: probably don't need to add new fields to the old API. Just add it to v1beta1

manifests/templates/reconciler-manager-configmap.yaml
sdowell, line 1: nit: use rebase instead of adding merge commits to your PR

pkg/reconcilermanager/controllers/rootsync_controller_test.go
victorpras: the function `validateDeployments` in the unit test have been updated to validate for args. All other unit test container template are also updated to include default log verbosity

pkg/reconcilermanager/controllers/reposync_controller_test.go
victorpras: same as comment above

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: the logic has been replaced to use default log level value when it's unset.

pkg/api/configsync/v1alpha1/resource_override.go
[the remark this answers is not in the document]
victorpras: removed from v1alpha1, only declared in v1beta1

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -501,6 +503,12 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
 			}
 		}
 	}
+
+	if loglevel, ok := override.LogLevel[c.Name]; ok {
+		c.Args = append(c.Args, fmt.Sprintf("--v=%d", loglevel))
+	} else {
+		c.Args = append(c.Args, fmt.Sprintf("--v=%d", defaultLogLevel))
```
sdowell, line 510: It looks like this changes the default of the `git-sync` container, as it was defaulting to `5` before

pkg/api/configsync/v1beta1/resource_override.go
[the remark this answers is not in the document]
victorpras: changed to use map[string]int in `OverrideSpec` instead of `ContainerResourcesSpec`

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -66,6 +66,10 @@ type OverrideSpec struct {
 	// support pulling remote bases from public repositories.
 	// +optional
 	EnableShellInRendering *bool `json:"enableShellInRendering,omitempty"`
+
+	// logLevel allows one to override the log verbosity of a container
+	// +optional
+	LogLevel map[string]int `json:"logLevel,omitempty"`
```
sdowell, line 72: I think this should cause the reconciler to stall with a validation error if the user provides an unexpected container name

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -66,6 +66,10 @@ type OverrideSpec struct {
 	// support pulling remote bases from public repositories.
 	// +optional
 	EnableShellInRendering *bool `json:"enableShellInRendering,omitempty"`
+
+	// logLevel allows one to override the log verbosity of a container
```
sdowell, line 70: Add more information to description here describing the expected format of the field

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -3558,6 +3558,97 @@ func TestPopulateRepoContainerEnvs(t *testing.T) {
 	}
 }
 
+func TestUpdateLogLevelWithOverride(t *testing.T) {
```
sdowell, line 3561: nit: test name should indicate this is for the reposync controller

pkg/api/configsync/v1beta1/resource_override.go
sdowell: Alternatively, we could enumerate the valid containers as a nested CRD field. Might want to get agreement on the API in a design doc

manifests/templates/reconciler-manager-configmap.yaml
victorpras: fixed by squash&rebase

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: updated to make an exception for git-sync container

pkg/api/configsync/v1beta1/resource_override.go
victorpras: added more description about the intended format of the field

pkg/reconcilermanager/controllers/reposync_controller_test.go
victorpras: updated test name

pkg/api/configsync/v1beta1/resource_override.go
victorpras: from my understanding, the iteration is done in https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/rootsync_controller.go#L917, and it will check for the container name in the override. Unexpected container name should be ignored. 

I tried reproducing this scenario in a cluster with local OSS installation with invalid container name, not seeing any error in yaml validation and reconciler log

sdowell: It is ignored currently, I'm suggesting that the behavior should be added to cause the reconciler to stall if the container name specified for this field is invalid

victorpras: updated to use a struct with containerName string and logLevel int field

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,26 @@ spec:
                     format: int64
                     minimum: 0
                     type: integer
+                  logLevel:
+                    description: logLevel allows one to override the log verbosity
+                      of a container. ContainerLogLevelOverride is a struct with field
```
karlkfi, line 1301: ContainerLogLevelOverride is an internal name. It doesn't need to be in the public description here. How about `Each entry must contain the name of a reconciler deployment container and the desired log level.`

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,26 @@ spec:
                     format: int64
                     minimum: 0
                     type: integer
+                  logLevel:
```
karlkfi, line 1299: As a associative array, this field should have a plural field name, like logLevels, loggingLevels, or containerLoggingLevels.

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,26 @@ spec:
...
+                    description: logLevel allows one to override the log verbosity
+                      of a container. ContainerLogLevelOverride is a struct with field
+                      container name and integer intended log level
+                    items:
+                      description: ContainerLogLevelOverride allows to override the
+                        log level for a container
+                      properties:
+                        containerName:
+                          description: containerName specifies the name of a container
+                            whose log level will be overridden. Must be "reconciler",
```
karlkfi, line 1309: `Must be one of the following:`

I don't know if they all need to be quoted or not. Check for precedent in other field descriptions.

manifests/reposync-crd.yaml
karlkfi: On the topic of naming, whether we use log or logging should be consistent here and in the sub-field.

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,26 @@ spec:
...
+                            "git-sync", "hydration-controller", "oci-sync", or "helm-sync".
+                          pattern: ^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync)$
+                          type: string
+                        logLevel:
```
karlkfi, line 1313: Would "level" be better or do we like "logLevel" here? It seems a little redundant with the parent field named "logLevels". I'm on the fence.

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,26 @@ spec:
                     format: int64
                     minimum: 0
                     type: integer
+                  logLevel:
+                    description: logLevel allows one to override the log verbosity
```
karlkfi, line 1300: `logLevel allows one to override the log verbosity of a reconciler deployment container.`

We need to be explicit that these are for the reconciler deployment containers, not just any old containers.

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -111,7 +109,7 @@ data:
            imagePullPolicy: IfNotPresent
          - name: git-sync
            image: gcr.io/config-management-release/git-sync:v3.6.6-gke.3__linux_amd64
-           args: ["--root=/repo/source", "--dest=rev", "--max-sync-failures=30", "--error-file=error.json", "--v=5"]
+           args: ["--root=/repo/source", "--dest=rev", "--max-sync-failures=30", "--error-file=error.json"]
```
karlkfi, line 112: do we not want git-sync to default to -v=5 any more? 
That was kinda useful for debugging, and not excessively verbose.
If so, we need different defaults for each container.

karlkfi: Oh, I see this default was moved down to code in the reconciler-base. If it's still true, the defaults should probably be documented in the CRD field description.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -501,6 +503,22 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
 			}
 		}
 	}
+	var found = false
+	for _, logLevel := range override.LogLevel {
+		if logLevel.ContainerName == c.Name {
```
karlkfi, line 508: Looks like there isn't any validation of these entries. We probably need some validation added up near the other reconcile validations to ensure the containerName is unique, one of the allowed names, and that the log level is within a valid range.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -501,6 +503,22 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
...
+			c.Args = append(c.Args, fmt.Sprintf("--v=%d", logLevel.LogLevel))
+			found = true
+		}
+	}
+
+	if !found {
+		if c.Name == "git-sync" {
```
karlkfi, line 515: This string should be a constant. I believe there is one already.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -501,6 +503,22 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
...
+			// git-sync container's default is 5
+			c.Args = append(c.Args, fmt.Sprintf("--v=%d", 5))
```
karlkfi, line 517: Might be nice if these defaults were pulled out somewhere so they could be managed in one place, like maybe a `containerLogLevelDefault(containerName string)` function. That would slightly simplify the logic here.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -501,6 +503,22 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
 			}
 		}
 	}
+	var found = false
```
karlkfi, line 506: This whole added code chunk could probably be extracted to a `containerLogLevel(containerName string)` function and then just always append the -v flag with the value, even if it's zero.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -501,6 +503,22 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
...
+	for _, logLevel := range override.LogLevel {
+		if logLevel.ContainerName == c.Name {
+			c.Args = append(c.Args, fmt.Sprintf("--v=%d", logLevel.LogLevel))
```
karlkfi, line 509: Is the double -- required or is one enough? Do all the container entrypoints support --v?

karlkfi (changes requested):

manifests/reposync-crd.yaml
victorpras: done, changed to `LogLevels`

manifests/reposync-crd.yaml
victorpras: changed

manifests/templates/reconciler-manager-configmap.yaml
victorpras: done, updated CRD with description of default value

manifests/reposync-crd.yaml
victorpras: changed

manifests/reposync-crd.yaml
victorpras: updated. Looks like the other field in https://github.com/GoogleContainerTools/kpt-config-sync/blob/8973d806e4bfdd3e15a5ef4817656c5e0ae2a558/pkg/api/configsync/v1beta1/resource_override.go#L72 use double quotes for all container name

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: the validation is done by the CRD validation in `+kubebuilder:validation:Pattern=^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync)$`

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: done. created and used `containerLogLevelDefault(containerName string)` function

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: done. created `containerLogLevel(containerName string, override []v1beta1.ContainerLogLevelOverride) int` that will determine log level value and simplified the logic

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: done. updated to use `reconcilermanager.GitSync` constant

manifests/reposync-crd.yaml
victorpras: i don't really have any preference either way. I will keep `logLevel` here, but if there's a good reason to change it to just `level` or any other name then I can change it

pkg/reconcilermanager/controllers/reconciler_base.go
karlkfi: Do any of the components error if verbosity is less than 0?
Technically I don't think there's a max level in the tools we use, but we could maybe scan the code and check for `.V(\d+)` and see what the max is and pick that way... maybe just 0-10?
Do we want to allow disabling logs with -1?

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: i tried installing local version of ConfigSync using single dash `-v=0` and didn't notice any different in the container logs, including git-sync (-v=5). All other args are using double dash so it might be good just to keep it consistent with the others

```
      - args:
        - --root=/repo/source
        - --dest=rev
        - --max-sync-failures=30
        - --error-file=error.json
        ```

File: pkg/reconcilermanager/controllers/reconciler_base.go [hunk 21]
```diff
@@ -501,6 +504,29 @@ func mutateContainerResource(c *corev1.Container, override *v1beta1.OverrideSpec
...
+	}
+
+	return containerLogLevelDefault(containerName)
+}
+
+// containerLogLevelDefault will return the default log level value for any reconciler deployment container
+// when override is not specified
+func containerLogLevelDefault(containerName string) int {
+	// git-sync container's default is 5
+	if containerName == reconcilermanager.GitSync {
```
karlkfi, line 526: nit: a switch statement would be prettier, but it's not a big deal.

manifests/templates/reconciler-manager-configmap.yaml
karlkfi: probably better to put the default values on the desc of the logLevel field itself, rather than a parent field.

manifests/reposync-crd.yaml
karlkfi: logLevel is fine.

pkg/reconcilermanager/controllers/reconciler_base.go
sdowell: Yeah I'd agree that a single dash is more conventional for a short option, double dashes for a long option. I'd go with `-v` if the container/binary accepts it

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -89,3 +96,17 @@ type ContainerResourcesSpec struct {
...
 }
+
+// ContainerLogLevelOverride allows to override the log level for a container
+type ContainerLogLevelOverride struct {
+	// containerName specifies the name of a reconciler deployment container whose log level will be overridden.
+	// Must be one of the following: "reconciler", "git-sync", "hydration-controller", "oci-sync", or "helm-sync".
+	//
+	// +kubebuilder:validation:Pattern=^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync)$
+	// +optional
+	ContainerName string `json:"containerName,omitempty"`
```
sdowell, line 107: Is there any guarantee that the containerName is unique in the array? What happens if the same container is specified multiple times?

pkg/reconcilermanager/controllers/reconciler_base.go
sdowell: Agreed that a switch would make sense stylistically here, especially if we need to add additional cases

karlkfi (changes requested):

manifests/rootsync-crd.yaml
```diff
@@ -1310,6 +1310,30 @@ spec:
...
+                            Must be one of the following: "reconciler", "git-sync",
+                            "hydration-controller", "oci-sync", or "helm-sync".'
+                          pattern: ^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync)$
+                          type: string
+                        logLevel:
+                          description: logLevel allows one to override the log verbosity
+                            of a container.
+                          type: integer
+                      type: object
+                    type: array
```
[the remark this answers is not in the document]
karlkfi: kubebuilder has tags for this: https://book.kubebuilder.io/reference/markers/crd-processing.html

pkg/api/configsync/v1beta1/resource_override.go
karlkfi: listType: map and listMergeKey: containerName should enforce uniqueness and help SSA merging

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: it looks like the values that are being used are from 0 to 9. I think restricting to 0 to 10 would be a good idea

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: changed to single dash `-v`

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: changed to use switch

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -66,6 +66,13 @@ type OverrideSpec struct {
 	// support pulling remote bases from public repositories.
 	// +optional
 	EnableShellInRendering *bool `json:"enableShellInRendering,omitempty"`
+
+	// logLevel allows one to override the log verbosity of a reconciler deployment container.
+	// Each entry must contain the name of a reconciler deployment container and the desired log level.
+	// If logLevel is not specified, git-sync container will default to 5,
+	// while all other containers will default to 0
+	// +optional
+	LogLevels []ContainerLogLevelOverride `json:"logLevel,omitempty"`
```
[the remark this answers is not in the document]
victorpras: changed to use `containerOverrides`.

I agree that it's appropriate to also use `resources` in `containerOverrides`. Unfortunate that `resources` is a bit to specific to be properly extended.

manifests/rootsync-crd.yaml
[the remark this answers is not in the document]
victorpras: thanks for the reccommendation! Changed to use listType: map and listMapKey: containerName

pkg/api/configsync/v1beta1/resource_override.go
victorpras: thanks for the reccommendation! Changed to use listType: map and listMapKey: containerName. Confirmed that non unique container name will cause CRD validation error now

manifests/reposync-crd.yaml
```diff
@@ -1280,6 +1280,38 @@ spec:
                       about valid inputs: https://pkg.go.dev/time#ParseDuration. Recommended
                       apiServerTimeout range is from "3s" to "1m".'
                     type: string
+                  containerOverrides:
```
karlkfi, line 1283: wait wait wait... we can't call this containerOverrides if it doesn't include the container resource overrides...

karlkfi: @sdowell it looks you maybe you requested this?

sdowell: I [offered this as a suggestion but acknowledged that this name probably isn't feasible with the presence of the other resource overrides field](https://github.com/GoogleContainerTools/kpt-config-sync/pull/677#discussion_r1275211852). Looks like my other comment got resolved. @victorpras might help to let the OP resolve comments

karlkfi: These are already under `override`. So -Override is redundant in a sub-field... and we already have override.resources and other container-specific overrides that we can't change now due to needing reverse compat.

I think logLevels was fine... I agree having logLevel in each entry is redundant, that's why I suggested just "level" but I';d rather have `spec.override.logLevels[].logLevel` than `spec.override.containerOverrides[].logLevel`... we're not designing this from scratch.

victorpras: reverting back to `logLevels`

pkg/api/configsync/v1beta1/resource_override.go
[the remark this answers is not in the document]
victorpras: after the discussion in https://github.com/GoogleContainerTools/kpt-config-sync/pull/677#discussion_r1275472197, will revert back to `logLevels`

manifests/templates/reconciler-manager-configmap.yaml
victorpras: moved the desc to logLevel field

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,38 @@ spec:
                     format: int64
                     minimum: 0
                     type: integer
+                  logLevels:
+                    description: logLevels allows one to override the log verbosity
+                      of a reconciler deployment container. Each entry must contain
```
karlkfi, line 1301: a reconciler deployment container -> reconciler deployment containers

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,38 @@ spec:
...
+                      the name of a reconciler deployment container and the desired
+                      log level.
+                    items:
+                      description: ContainerLogLevelOverride allows to override the
```
karlkfi, line 1305: Don't use the internal struct name in the description here

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,38 @@ spec:
...
+                        log level for a container
+                      properties:
+                        containerName:
+                          description: 'containerName specifies the name of a reconciler
```
karlkfi, line 1309: can these be made mroe consitent in tense and audience? some say "... allows one to ..." and some say "... specifies the ...".

manifests/reposync-crd.yaml
victorpras: it looks like this is taken from the comment in https://github.com/GoogleContainerTools/kpt-config-sync/pull/677/files#diff-f2a8c540eb1da55be20cd847621833115dd35f755304786bda8a4ff063ce9ad6R100 which follows the same pattern in https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/reposync-crd.yaml#L309. How should I update the description here?

victorpras: I think the comment has to start with the struct name, otherwise it will fail the presubmit

manifests/reposync-crd.yaml
victorpras: changed

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,38 @@ spec:
...
+                          description: 'containerName specifies the name of the reconciler
+                            deployment container which log level will be overridden.
+                            Must be one of the following: "reconciler", "git-sync",
+                            "hydration-controller", "oci-sync", or "helm-sync".'
+                          pattern: ^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync)$
+                          type: string
+                        logLevel:
+                          description: logLevel specifies the log level override value
+                            for the container. The default value for git-sync container
+                            is 5, while all other containers will default to 0 Allowed
```
karlkfi, line 1318: missing a period after "will default to 0"

manifests/reposync-crd.yaml
```diff
@@ -1296,6 +1296,38 @@ spec:
...
+                      override value for the reconciler deployment container. Each
+                      entry must contain the name of the reconciler deployment container
+                      and the desired log level.
+                    items:
+                      description: ContainerLogLevelOverride specifies the container
+                        name and log level override value
+                      properties:
+                        containerName:
+                          description: 'containerName specifies the name of the reconciler
+                            deployment container which log level will be overridden.
```
karlkfi, line 1310: s/which/for which/

manifests/reposync-crd.yaml
victorpras: oops, fixed

manifests/reposync-crd.yaml
victorpras: switched

karlkfi (approved):
/lgtm

karlkfi (changes requested):
oops, needs conversion updates, after this merges: https://github.com/GoogleContainerTools/kpt-config-sync/pull/765

victorpras:
looks like `make clientgen` is not giving the expected result gpaste/4785936289234944, need to investigate why (potentially different go version?)

victorpras:
looking at the result of `make clientgen` from other workstation, the difference is mostly function ordering. To unblock this issue, the changes are modified manually. Also added impl of `Convert_v1beta1_OverrideSpec_To_v1alpha1_OverrideSpec` because of lossy conv

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -3710,6 +3810,11 @@ func validateDeployments(wants map[core.ID]*appsv1.Deployment, fakeDynamicClient
 					if diff := cmp.Diff(i.Resources, j.Resources); diff != "" {
 						return errors.Errorf("Unexpected resources found for the %q container of %q, diff %s", i.Name, id, diff)
 					}
+
+					// Compare Args
+					if diff := cmp.Diff(i.Args, j.Args); diff != "" {
```
karlkfi, line 3815: nit: this could use testutils.AssertEqual, which prints a diff for you automatically.

pkg/api/configsync/v1alpha1/zz_generated.conversion.go
```diff
@@ -617,21 +617,22 @@ func Convert_v1alpha1_OverrideSpec_To_v1beta1_OverrideSpec(in *OverrideSpec, out
 	return autoConvert_v1alpha1_OverrideSpec_To_v1beta1_OverrideSpec(in, out, s)
 }
 
+// Convert_v1beta1_OverrideSpec_To_v1alpha1_OverrideSpec is written manually because of lossy conversion
+func Convert_v1beta1_OverrideSpec_To_v1alpha1_OverrideSpec(in *v1beta1.OverrideSpec, out *OverrideSpec, s conversion.Scope) error {
```
karlkfi, line 621: This custom func needs to be in the conversion.go file because it's not generated. This whole file gets replaced when re-generated.

victorpras: moved to conversion.go

pkg/reconcilermanager/controllers/reposync_controller_test.go
victorpras: it looks like using `testutil.AssertEqual` will require refactoring `validateDeployment` func to take `t *testing.T` as param, which might be worth a separate refactoring PR to change all `cmp.Diff` usage within `validateDeployment`. Unless you think it's really important to include this refactor in this PR, i think it'll be better to be done in separate changes. Thoughts?

karlkfi: not a big deal.

karlkfi (approved):
/lgtm

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -3558,6 +3558,106 @@ func TestPopulateRepoContainerEnvs(t *testing.T) {
...
+	wantRs := fake.RepoSyncObjectV1Beta1(reposyncNs, reposyncName)
+	wantRs.Spec = rs.Spec
+	wantRs.Status.Reconciler = nsReconcilerName
+	reposync.SetReconciling(wantRs, "Deployment",
+		fmt.Sprintf("Deployment (config-management-system/%s) InProgress: Replicas: 0/1", nsReconcilerName))
+	controllerutil.AddFinalizer(wantRs, metadata.ReconcilerManagerFinalizer)
+	validateRepoSyncStatus(t, wantRs, fakeClient)
+
+	repoContainerEnv := testReconciler.populateContainerEnvs(ctx, rs, nsReconcilerName)
+	repoDeployment := repoSyncDeployment(
```
karlkfi, line 3584: these unit tests broke after rebasing. the new generation metric tag modified these tests, so you'll need to do the same.

https://github.com/GoogleContainerTools/kpt-config-sync/pull/763/files#diff-4e452cd48e11abd5a08fa09505577300d00a536c7c57d7ebc492a9b166be3f55

basically, you'll need to update the container env vars each time to match the generation.

victorpras: thanks, fixed now!

karlkfi (approved):
/lgtm

pkg/api/configsync/v1alpha1/resource_override.go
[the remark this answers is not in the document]
janetkuo: When adding new fields, it needs to be added to both api versions so that api conversions work. We need to continue doing that until we remove v1alpha1.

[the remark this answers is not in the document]
karlkfi: I don't think this is actually true. The conversions use the v1beta1 as the storage version. So they convert from v1alpha1 -> v1beta1, which is not lossy. There shouldn't be any use case that requires round trip the other way.

PR #678, opened 2023-06-14
test: add conditional for invalid autopilot flags

sdowell:
Some of these flags cannot be provided when creating autopilot clusters, such as machine type.


nan-yu (approved):
/lgtm

PR #679, opened 2023-06-14
add junit cli to gke-e2e image

sdowell:
This binary is needed for the e2e prow script to generate a junit report


karlkfi (approved):
/lgtm

PR #681, opened 2023-06-14
test: retry deletion of gke clusters

sdowell:
It is possible an operation is in progress on the cluster at the time that the deletion request is sent. This causes the deletion request to error. Retry for a brief period to increase the likelihood that the cluster can be deleted.

### DIFF
--- a/e2e/nomostest/clusters/gke.go
+++ b/e2e/nomostest/clusters/gke.go
@@ -19,9 +19,11 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"kpt.dev/configsync/e2e"
+	"kpt.dev/configsync/e2e/nomostest/retry"
 	"kpt.dev/configsync/e2e/nomostest/testing"
 )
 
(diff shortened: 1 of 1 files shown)


e2e/nomostest/clusters/gke.go
```diff
@@ -68,12 +70,19 @@ func deleteGKECluster(t testing.NTB, name string) error {
 		args = append(args, "--region", *e2e.GCPRegion)
 	}
 	t.Logf("gcloud %s", strings.Join(args, " "))
-	cmd := exec.Command("gcloud", args...)
-	out, err := cmd.CombinedOutput()
-	if err != nil {
-		return errors.Errorf("failed to delete cluster %s: %v\nstdout/stderr:\n%s", name, err, string(out))
-	}
-	return nil
+	// Sometimes an operation may be happening at the time the deletion request is
```
karlkfi, line 73: Can we list cluster operations and wait until they're done instead?

sdowell: I think we would need to wrap that in a retry anyways, as other operations could begin after the initial list

karlkfi: yes, but then we wouldn't be spamming deletes when we know they won't work.

sdowell: I'm not super opposed to implementing this, but doing this client side feels like it's just overcomplicating things. If there's an in progress operation then the backend can handle it and will give us an error.

The retry has a backoff strategy so this shouldn't be too spammy, although we could make it so that the backoff is configurable and use a more conservative backoff on this retry.

karlkfi (approved):
/lgtm

PR #683, opened 2023-06-14
test: add list/wait for operations on gke clusters

sdowell:
This updates the delete gke cluster logic to first list and wait for all running operation before attemption to delete the cluster. This should minimize spam from the delete cluster retry logic.

Also wait for operations to complete on cluster creation.


e2e/nomostest/clusters/gke.go
```diff
@@ -58,6 +59,71 @@ func withKubeConfig(cmd *exec.Cmd, kubeconfig string) *exec.Cmd {
...
+			name, err, string(out))
+	}
+	operations := strings.Fields(string(out))
+	return operations, nil
+}
+
+// waitOperation waits for the provided operation to complete
+func waitOperation(t testing.NTB, operation string) error {
+	args := []string{
+		"container", "operations", "wait",
```
karlkfi, line 90: does wait have a timeout?

sdowell: It doesn't look like it, but we could use a context with a timeout

sdowell: added a timeout context

karlkfi: most gke operations take much longer than 1m, some even longer than 5m. like say any upgrade or node pool replacement or repair or even apiserver scale up.

sdowell: bumped the duration

e2e/nomostest/clusters/gke.go
```diff
@@ -58,6 +60,70 @@ func withKubeConfig(cmd *exec.Cmd, kubeconfig string) *exec.Cmd {
...
+		"--project", *e2e.GCPProject,
+	}
+	if *e2e.GCPZone != "" {
+		args = append(args, "--zone", *e2e.GCPZone)
+	}
+	if *e2e.GCPRegion != "" {
+		args = append(args, "--region", *e2e.GCPRegion)
+	}
+	t.Logf("gcloud %s", strings.Join(args, " "))
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
karlkfi, line 102: suggest pulling ctx out as a func var so you can define the timeout or cancel externally

e2e/nomostest/clusters/gke.go
```diff
@@ -73,7 +139,10 @@ func deleteGKECluster(t testing.NTB, name string) error {
 	// Sometimes an operation may be happening at the time the deletion request is
 	// sent, causing delete to error. Retry for a brief period to increase the
 	// chances for the delete operation.
-	took, err := retry.Retry(1*time.Minute, func() error {
+	took, err := retry.Retry(10*time.Minute, func() error {
```
karlkfi, line 142: if you use wait.BackoffUntil you can pass it the ctx.Done and cancel this with the same context that cancels the operation wait.

e2e/nomostest/clusters/gke.go
sdowell: will do

e2e/nomostest/clusters/gke.go
sdowell: Looked into this but I don't think it will be that simple, because `BackOffUntil` doesn't accept a function that returns an error. Thus it continues retrying regardless of whether the function succeeds, which is a bit of a different use case.

sdowell: We would need to implement some sort of hybrid approach that uses the `wait.ConditionFunc` of `wait.ExponentialBackoff` and the `stopCh` of `wait.BackoffUntil`

e2e/nomostest/clusters/gke.go
sdowell: done

sdowell:
gonna experiment with the retry logic
/hold

e2e/nomostest/clusters/gke.go
sdowell: As I mentioned `BackOffUntil` doesn't really provide the right interface for this use case. I added a new function which fits this pattern. I played with a few different ideas but didn't want to go too overkill on async/channels so I think this is a good middle ground. PTAL

/unhold

karlkfi (approved):
/lgtm

PR #684, opened 2023-06-15
test: fix for reconcile timeout test on autopilot

sdowell:
Explicitly set the namespace for the pause-deployment object. When running this in a Pod the applied object defaults to the namespace of the Pod.

Also move the cleanup instantiation to before the apply, so the cleanup runs in the case that the apply errors.

### DIFF
--- a/e2e/testdata/low-priority-pause-deployment.yaml
+++ b/e2e/testdata/low-priority-pause-deployment.yaml
@@ -16,6 +16,8 @@ apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
   name: low-priority
+  labels:
+    nomos-test: enabled
 value: -1
 globalDefault: false
 ---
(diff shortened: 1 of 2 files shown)


e2e/testdata/low-priority-pause-deployment.yaml
```diff
@@ -23,6 +23,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: pause-deployment
+  namespace: default
   labels:
```
karlkfi, line 27: Might want to add the test labels on here so normal cleanup catches it. i don't think MustKubectl applies them the same way KubeClient.Create does. Or switch to using KubeClient.

sdowell: Added the test label

karlkfi (approved):
/lgtm

PR #685, opened 2023-06-15
only run `helm show chart` for version range

natasha41575:
- reverts https://github.com/GoogleContainerTools/kpt-config-sync/pull/674 (as it was just experimental and didn't end up helping at all)
- changes it so that we only run `helm show chart` if the version is provided to us as a range (using the same range processing library that helm uses). If it's not provided as a range, we skip the expensive polling.
- bumps the cpu req for the version test

b/286870867


pkg/helm/helm.go
```diff
@@ -300,3 +309,16 @@ func (h *Hydrator) appendAuthArgs(ctx context.Context, args []string) ([]string,
...
+// we determine if a version is a valid range by checking that (a) it is not
+// valid semver on its own and (b) that it can be parsed correctly as a version range
+func isRange(version string) bool {
+	if version == "" {
+		return true
+	}
+	if semver.IsValid("v" + version) {
+		return false
+	}
+	_, err := semverrange.NewConstraint(version)
```
natasha41575, line 322: This is the same semver library that [helm uses](https://github.com/helm/helm/blob/4e447d87cd3222aa5e5ea2b38170e9d80b99c0b4/pkg/registry/util.go#L75) for range detection

PR #686, opened 2023-06-15
Use foreground delete on RootSyncs

karlkfi:
Now that the reconciler-manager uses a finalizer to delete managed resources, it's safe to delete RootSyncs with foreground deletion propagation.

### DIFF
--- a/e2e/nomostest/reset.go
+++ b/e2e/nomostest/reset.go
@@ -22,7 +22,6 @@ import (
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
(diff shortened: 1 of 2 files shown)


e2e/nomostest/reset.go
```diff
@@ -187,13 +187,8 @@ func ResetRootSyncs(nt *NT, rsList []v1beta1.RootSync) error {
 		defer cancel()
 		go TailReconcilerLogs(ctx, nt, RootReconcilerObjectKey(rsNN.Name))
 
-		// DeletePropagationBackground is required when deleting RSyncs with
```
sdowell, line 190: What about using foreground for RepoSync cleanup? Does that come later?

karlkfi: RepoSync already tolerated foreground because it didn't use owner references

sdowell: The cleanup code below is still using background by the looks of it

karlkfi: weird. fixed.

sdowell (approved):

PR #689, opened 2023-06-15
test: use new logger for each sharedNT

sdowell:
Using the logger from the testing.T ensures proper interleaving of logs. Without this change shared test environments used the logger from the original FreshTestEnv.

### DIFF
--- a/e2e/nomostest/testing/ntb.go
+++ b/e2e/nomostest/testing/ntb.go
@@ -46,6 +46,14 @@ type FakeNTB struct {
 	mu       sync.RWMutex
 	failed   bool
 	cleanups []func()
+	name     string
+}
+
+// NewFakeNTB returns a new FakeNTB
+func NewFakeNTB(name string) *FakeNTB {
+	return &FakeNTB{
+		name: name,
+	}
 }
 
 // Error is equivalent to Log followed by Fail.
@@ -94,7 +102,7 @@ func (t *FakeNTB) Fatalf(format string, args ...interface{}) {
 
 // Name returns an empty string.
 func (t *FakeNTB) Name() string {
-	return ""
+	return t.name
 }
 
 // Helper is a no-op function.
@@ -168,12 +176,11 @@ func (t *FakeNTB) Skipped() bool {
 
 // Log generates the output. It's always at the same stack depth.
 func (t *FakeNTB) Log(args ...interface{}) {
-	fmt.Println(args...)
+	fmt.Println(fmt.Sprintf("[%s]", t.name), fmt.Sprint(args...))
 }
 
 // Logf formats its arguments according to the format, analogous to Printf, and
 // records the text in the error log.
 func (t *FakeNTB) Logf(format string, args ...interface{}) {
-	fmt.Printf(format, args...)
-	fmt.Println()
+	fmt.Println(fmt.Sprintf("[%s]", t.name), fmt.Sprintf(format, args...))
 }
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -139,6 +139,9 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	t.Helper()
 
 	sharedNt := SharedNT(t)
+	// Set t on the logger to ensure proper interleaving of logs
+	sharedNt.Logger.SetNTBForTest(t)
+
 	nt := &NT{
 		Context:                 sharedNt.Context,
 		T:                       t,
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -329,7 +329,7 @@ func newSharedNT(name string) error {
 	if err := os.RemoveAll(tmpDir); err != nil {
 		return errors.Wrap(err, "failed to remove the shared test directory")
 	}
-	fakeNTB := &testing.FakeNTB{}
+	fakeNTB := testing.NewFakeNTB(name)
 	// Register a new SharedNT immediately with the fakeNTB. This way in the case
 	// of an error (e.g. during cluster creation) the cleanup can still run.
 	newSNT := mySharedNTs.newNT(fakeNTB)
--- a/e2e/nomostest/testlogger/testlogger.go
+++ b/e2e/nomostest/testlogger/testlogger.go
@@ -33,6 +33,18 @@ func New(t testing.NTB, debugEnabled bool) *TestLogger {
 	}
 }
 
+// SetNTBForTest sets the loggers NTB for the scope of a test. This ensures that
+// for shared test environments, the actual test's logger is used for the scope
+// of the test. This ensures proper interleaving of logs when running tests in
+// parallel. Reset to the original logger after the test for post-test logging.
+func (tl *TestLogger) SetNTBForTest(t testing.NTB) {
+	originalNTB := tl.t
+	t.Cleanup(func() {
+		tl.t = originalNTB
+	})
+	tl.t = t
+}
+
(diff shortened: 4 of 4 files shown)


e2e/nomostest/gitproviders/git-provider.go
```diff
@@ -63,3 +63,29 @@ func NewGitProvider(t testing.NTB, provider string, logger *testlogger.TestLogge
...
+	case *LocalProvider:
+		// The PortForwarder for LocalProvider is reset for each test in setupTestCase.
+		// Therefore, we don't need to reset the PortForwarder's logger here.
+		return &LocalProvider{
+			PortForwarder: p.PortForwarder,
+		}
+	default:
+		t.Fatalf("unrecognized GitProvider: %v", p)
+	}
+	return nil
```
karlkfi, line 90: this is unreachable. if you need it to compile, move it to the default after the fatal

e2e/nomostest/reset.go
```diff
@@ -451,6 +451,7 @@ func initRepository(nt *NT, repoType gitproviders.RepoType, nn types.NamespacedN
 		}
 		nt.RemoteRepositories[nn] = repo
 	}
+	repo.WithLogger(nt.T, nt.Logger)
```
karlkfi, line 454: how does this work? does it set? what happens to the new one returned?

e2e/nomostest/gitproviders/git-provider.go
sdowell: yeah it's just to compile. I can move it to the default, figured it doesn't make much difference

e2e/nomostest/reset.go
sdowell: See the new method defined in `repository.go` - it uses the provided logger for the scope of the test and switches back to the original logger in cleanup

e2e/nomostest/gitproviders/git-provider.go
sdowell: done

e2e/nomostest/reset.go
sdowell: Updated each of these to behave a bit more consistently

karlkfi: This With- pattern doesn't follow the usual conventions. And the two different impls her have different semantics.

Usually I would expect a With- func to be a method on a builder struct with a Build() after that would return a new instance with the specified settings. 

The Repository one looks like a setter: Set-

The GitProvider one looks kind of like a an aspect oriented programming constructor: New-

But they both include cleanup unsetter, which is also weird.

e2e/nomostest/gitproviders/git-provider.go
```diff
@@ -63,3 +63,17 @@ func NewGitProvider(t testing.NTB, provider string, logger *testlogger.TestLogge
 		return &LocalProvider{}
 	}
 }
+
+// WithLogger updates an existing GitProvider to use the provider TestLogger
+func WithLogger(t testing.NTB, provider GitProvider, logger *testlogger.TestLogger) GitProvider {
+	switch p := provider.(type) {
+	case *BitbucketClient:
```
karlkfi, line 70: why only change the logger for bitbucket?

e2e/nomostest/gitproviders/git-provider.go
```diff
@@ -63,3 +63,17 @@ func NewGitProvider(t testing.NTB, provider string, logger *testlogger.TestLogge
...
+		prevLogger := p.logger
+		t.Cleanup(func() {
```
karlkfi, line 72: why revert the logger on cleanup?

e2e/nomostest/testing/ntb.go
```diff
@@ -168,12 +176,13 @@ func (t *FakeNTB) Skipped() bool {
 
 // Log generates the output. It's always at the same stack depth.
 func (t *FakeNTB) Log(args ...interface{}) {
-	fmt.Println(args...)
+	out := fmt.Sprintf("[%s] ", t.name) + fmt.Sprint(args...)
```
karlkfi, line 179: having the out var doesnt add much here. you can just inline both of these Printlns

e2e/nomostest/gitproviders/git-provider.go
sdowell: it's the only one that configures a logger at the moment

e2e/nomostest/gitproviders/git-provider.go
sdowell: so that it uses the original logger after the test is finished, e.g. in the case of a shared test env if there's any cleanup that's done at the end of the test suite

e2e/nomostest/reset.go
sdowell: I'm going for a fixture type pattern here, where we set something for the context of the test and then restore it to the original state after the test is finished.

The difference in signatures was partly just for convenience, e.g. returning the pointer for GitProvider so that it can be inlined with the NT struct instantiation. I'm open minded on the name/signature if you have a suggestion

e2e/nomostest/testing/ntb.go
sdowell: done

e2e/nomostest/reset.go
sdowell: I updated it to use a consistent `UseLogger` method for setting a logger for the scope of the test.

e2e/nomostest/testing/ntb.go
```diff
@@ -168,12 +176,11 @@ func (t *FakeNTB) Skipped() bool {
 
 // Log generates the output. It's always at the same stack depth.
 func (t *FakeNTB) Log(args ...interface{}) {
-	fmt.Println(args...)
+	fmt.Println(fmt.Sprintf("[%s] ", t.name) + fmt.Sprint(args...))
```
karlkfi, line 179: i think you can replace the + with a comma and remove the space in the quotes. Println will join args with spaces, right?

sdowell: Yeah I think you're right, will do

e2e/nomostest/reset.go
karlkfi: Ok. I like Use- better than With-, just because it doesn't evoke a known pattern. But I think it's missing the detail that the setter has a scope: the duration of the current test. So what about just `SetTestLogger` or `UseLoggerForTest`? Then the method needs a comment about why it needs to be set and why it needs to be unset.

e2e/nomostest/testing/ntb.go
sdowell: done

e2e/nomostest/reset.go
sdowell: SGTM

karlkfi: The other idea that popped to mind was using AOP to wrap the struct with another struct that delegates to the first but uses a different logger. Then it could be unwrapped in cleanup, and the original object wouldn't have been modified. The problem with that is that I don't think Go's embedding allows overriding child methods the way Java inheritance does.

sdowell: done

sdowell: Hmm, since everything is using pointers we could probably just update the `Logger` itself to use the new `t` rather than assigning a new logger to everything

sdowell: That would be a lot cleaner if it works, let me give it a try

sdowell: yep looks like that works and is a lot simpler. PTAL

karlkfi: much cleaner, thanks

karlkfi (approved):
/lgtm

PR #694, opened 2023-06-16
test: longer timeout for gke cluster creation

sdowell:
Cluster creation via gcloud waits for the operation but has been occasionally timing out in CI. This change uses --async with a manual wait for the operation with our own timeout.

### DIFF
--- a/e2e/nomostest/clusters/gke.go
+++ b/e2e/nomostest/clusters/gke.go
@@ -98,6 +98,10 @@ func waitOperation(ctx context.Context, t testing.NTB, operation string) error {
 	if *e2e.GCPRegion != "" {
 		args = append(args, "--region", *e2e.GCPRegion)
 	}
+	start := time.Now()
+	defer func() {
+		t.Logf("took %v to wait for operation %s", time.Since(start), operation)
+	}()
 	t.Logf("gcloud %s", strings.Join(args, " "))
 	cmd := exec.CommandContext(ctx, "gcloud", args...)
 	out, err := cmd.CombinedOutput()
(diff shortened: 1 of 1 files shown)


e2e/nomostest/clusters/gke.go
```diff
@@ -108,8 +112,25 @@ func waitOperation(ctx context.Context, t testing.NTB, operation string) error {
...
+type operationOption func(opts *operationOptions)
+
+func withOperationTimeout(timeout time.Duration) operationOption {
+	return func(opts *operationOptions) {
+		opts.timeout = timeout
+	}
+}
+
+func listAndWaitForOperations(ctx context.Context, t testing.NTB, name string, opts ...operationOption) error {
+	options := &operationOptions{
```
karlkfi, line 127: do we need all this operationOptions complexity? why not just pass in a context with a timeout defined by the caller?

karlkfi: If you want to define a default timeout you can use ctx.Deadline() to see if the context has a timeout or not.

sdowell: done

e2e/nomostest/clusters/gke.go
```diff
@@ -109,8 +113,14 @@ func waitOperation(ctx context.Context, t testing.NTB, operation string) error {
 }
 
 func listAndWaitForOperations(ctx context.Context, t testing.NTB, name string) error {
-	ctxWithTimeout, cancel := context.WithTimeout(ctx, 5*time.Minute)
-	defer cancel()
+	var ctxWithTimeout context.Context
```
karlkfi, line 116: you could just replace the ctx var. that would avoid the else branch

sdowell: done

karlkfi (approved):
/lgtm

PR #695, opened 2023-06-16
test: do not treat ECONNREFUSED as terminal err

sdowell:
If a connection error occurs here, the portforwarder is most likely in the process of restarting the local port forward. If the portforwarded successfully restarts then ValidateMetrics can succeed on a subsequent retry.


karlkfi (approved):
/lgtm

PR #696, opened 2023-06-16
test: wait for failed status for invalid pod

sdowell:
This assertion seems to occasionally fail in CI, most likely due to a race condition with the InProgress status. Update the assertion to check the Failed status instead.


sdowell:
/hold
Gonna do some more local testing, I may have misinterpreted the root cause of the test flake

sdowell:
/unhold
It looks like the root cause here was more likely a transient Failed state with the Pod on autopilot - I realize now the TODO is still valid.

haiyanmeng (approved):

PR #700, opened 2023-06-20
test: treat reset git repos error as non-fatal

sdowell:
Errors can occur during this callback, e.g. if the Pod gets replaced while the callback is running. This happens occasionally on autopilot, which causes the test to fail. Treating this as a non-fatal error will allow the callback to re-run once the port forwarder discovers the error and attempts to recover (e.g. Pod not found).

### DIFF
--- a/e2e/testcases/multi_sync_test.go
+++ b/e2e/testcases/multi_sync_test.go
@@ -120,7 +120,9 @@ func TestMultiSyncs_Unstructured_MixedControl(t *testing.T) {
 	newRepos = append(newRepos, nn5)
 
 	if nt.GitProvider.Type() == e2e.Local {
-		nomostest.InitGitRepos(nt, newRepos...)
+		if err := nomostest.InitGitRepos(nt, newRepos...); err != nil {
+			nt.T.Fatal(err)
+		}
 	}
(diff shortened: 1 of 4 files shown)


e2e/testcases/multi_sync_test.go
```diff
@@ -120,7 +120,9 @@ func TestMultiSyncs_Unstructured_MixedControl(t *testing.T) {
 	newRepos = append(newRepos, nn5)
 
 	if nt.GitProvider.Type() == e2e.Local {
-		nomostest.InitGitRepos(nt, newRepos...)
+		if err := nomostest.InitGitRepos(nt, newRepos...); err != nil {
+			nt.T.Fatal(err)
```
haiyanmeng, line 124: Fatal -> Log ?

sdowell: This is a bit of a special case since it's done in a test case. I haven't looked too closely at this test but I expect the test would not be able to recover if this fails. The PortForwarder on the other hand will re-run the callback asynchronously whenever the port forward is restarted (e.g. if the Deployment's Pod is deleted/recreated)

karlkfi:
it might be better to convert these methods to return error so the caller can decide how to handle them

sdowell:
> it might be better to convert these methods to return error so the caller can decide how to handle them

Done

PR #702, opened 2023-06-21
helm: support reading valuesfiles from ConfigMaps

natasha41575:
This is very close to complete, I'm marking it as a draft for now as I still need to add validation for the new helm-related fields and run some tests. 

- Add support for reading valuesFiles from in-cluster ConfigMaps.
- Watch ConfigMaps in the RSync namespace, and rehydrate/resync when modifications are detected.
- Change the 'commit' to helm charts to specify just the `version`, rather than `chart:version`. 

Design and context: go/cs-helm-values-from-cm


cmd/helm-sync/watch.go
```diff
@@ -0,0 +1,145 @@
+// Copyright 2022 Google LLC
```
janetkuo, line 1: 2023

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +137,131 @@ func TestPublicHelm(t *testing.T) {
 	}
 }
 
+// TestHelmWatchConfigMap can run on both Kind and GKE clusters.
+// It tests that helm-sync properly watches ConfigMaps in the RSync namespace and pulls updated values.
+func TestHelmWatchConfigMap(t *testing.T) {
```
sdowell, line 142: Looks like this only tests RootSync, we need test cases for RepoSync as well.

Side note: I expected to see the ns-reconciler ClusterRole updated with permission to read ConfigMaps. Is the helm-sync container for RepoSync able to read the ConfigMap currently?

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +137,131 @@ func TestPublicHelm(t *testing.T) {
...
+		nt.T.FailNow()
+	}
+
+	configMapFilePath = "../testdata/root-sync-helm-cm-update/configmap-2.yaml"
+	nt.T.Logf("Update the ConfigMap with values to the cluster defined in %s", configMapFilePath)
+	nt.MustKubectl("apply", "-f", configMapFilePath)
+
+	// Config Sync takes about 4 minutes to pick up the update, rehydrate, and sync the resources to the cluster.
+	// Since we cannot use the RootSync status to detect the sync, we wait 7 minutes as a buffer.
+	time.Sleep(7 * time.Minute)
```
sdowell, line 197: Use `WatchObject` (with an increased timeout, if needed) instead of sleep + validate

cmd/helm-sync/main.go
```diff
@@ -112,13 +154,21 @@ func main() {
...
 			IncludeCRDs:     *flIncludeCRDs,
 			Auth:            configsync.AuthType(*flAuth),
 			HydrateRoot:     *flRoot,
 			Dest:            *flDest,
 			UserName:        *flUsername,
 			Password:        *flPassword,
+
+			RefreshCM:   refreshCM,
+			ValuesFrom:  valuesFrom,
+			CMValuesMap: cmValuesMap,
```
sdowell, line 167: This is wrapped in a mutex, but then we're still passing in a pointer to the map here. The struct is going to use this pointer without synchronization.

Related to the other comment, I suggest rewriting this to communicate using channels.

cmd/helm-sync/watch.go
```diff
@@ -0,0 +1,145 @@
...
+	apiwatch "k8s.io/apimachinery/pkg/watch"
+	"k8s.io/klog/v2"
+	api "kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/client/restconfig"
+	"kpt.dev/configsync/pkg/remediator/watch"
+	"kpt.dev/configsync/pkg/util"
+	utillog "kpt.dev/configsync/pkg/util/log"
+)
+
+func watchCM(valuesFrom []*api.ValuesFrom, cmValuesMap map[string]string, refreshCM *bool, log *utillog.Logger) {
```
sdowell, line 35: This function is problematic, IMO. I don't like that we're invoking this function as a goroutine and communicating using `os.Exit` and pointers with a shared mutex.

Can you rewrite this to communicate with the main goroutine using channels?

e2e/testcases/helm_sync_test.go
natasha41575: Will add tests for RepoSync. I think you're right that we'll need to add ConfigMaps to the ns-reconciler ClusterRole. I'm also working on adding tests for error cases

cmd/helm-sync/watch.go
natasha41575: Thanks for the catch! Fixed

manifests/ns-reconciler-cluster-role.yaml
```diff
@@ -26,6 +26,9 @@ rules:
 - apiGroups: ["configsync.gke.io"]
   resources: ["reposyncs/status"]
   verbs: ["get","list","watch","update","patch"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get","list","watch","create","update","patch"]
```
sdowell, line 31: Why does ns-reconciler need create/update/patch?

e2e/testdata/repo-sync-helm-cm-update/configmap-1.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Google LLC
```
sdowell, line 1: I would advise declaring resources inline in the test as the default choice. It generally makes tests easier to read (no need to go find the files where the resources are actually declared). Our go-based k8s client also applies test labels for cleanup, generally has a better UX, and generally has better error reporting compared to shelling out to kubectl

I think declaring test resources as files should be more of a special case for more complex circumstances

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +470,86 @@ func TestHelmNamespaceRepo(t *testing.T) {
 	}
 }
 
+// TestHelmNamespaceRepoWithConfigMap verifies RepoSync can pick up values and updates from ConfigMap references.
+// This test will work only with following pre-requisites:
+// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for accessing images in Artifact Registry.
+func TestHelmConfigMapNamespaceRepo(t *testing.T) {
+	repoSyncNN := nomostest.RepoSyncNN(testNs, configsync.RepoSyncName)
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.RequireGKE(t),
+		ntopts.RepoSyncPermissions(policy.AllAdmin()), // NS reconciler manages a bunch of resources.
```
sdowell, line 479: Can this be more restrictive than `AllAdmin`? What resources are being managed here? Using overkill permissions here can mask other issues (e.g. the base ClusterRole has insufficient permissions)

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +470,86 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
+		nt.T.Error(err)
+	}
+
+	configMapFilePath = "../testdata/repo-sync-helm-cm-update/configmap-2.yaml"
+	nt.T.Logf("Apply the ConfigMap with values to the cluster defined in %s", configMapFilePath)
+	nt.MustKubectl("apply", "-f", configMapFilePath)
+
+	if err = nt.Watcher.WatchObject(kinds.Deployment(), rs.Spec.Helm.ReleaseName+"-"+remoteHelmChart.ChartName, testNs,
+		[]testpredicates.Predicate{testpredicates.HasLabel("labelsTest", "second")},
+		testwatcher.WatchTimeout(7*time.Minute)); err != nil {
```
sdowell, line 520: I suggest using a dynamic timeout here, I expect we need a shorter timeout for kind and a longer timeout for autopilot etc

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +138,185 @@ func TestPublicHelm(t *testing.T) {
...
 }
 
+// TestHelmWatchConfigMap can run on both Kind and GKE clusters.
+// It tests that helm-sync properly watches ConfigMaps in the RSync namespace and pulls updated values.
+func TestHelmWatchConfigMap(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.Unstructured)
+
+	configMapFilePath := "../testdata/root-sync-helm-cm-update/configmap-1.yaml"
+	nt.T.Logf("Apply the ConfigMap with values to the cluster defined in %s", configMapFilePath)
+	nt.MustKubectl("apply", "-f", configMapFilePath)
```
sdowell, line 148: Echoing my comment on the testdata files, it would be preferred to use the go client for the types of operations in these tests as opposed to shelling out to kubectl. It also makes it hard to follow what's happening in the test since the contents of the configmaps are declared elsewhere

manifests/ns-reconciler-cluster-role.yaml
natasha41575: In order to mount the ConfigMaps (which are in the reposync's namespace) to the pod (which is in the config-management-namespace), it needs to copy the ConfigMap over.

sdowell: Doesn't the reconciler-manager upsert the ConfigMap to c-m-s?

sdowell: Actually, with the approach of mounting the ConfigMaps into the Deployment - I don't think the ns-reconciler should even need get/list/watch

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -421,6 +422,33 @@ func (r *reconcilerBase) deployment(ctx context.Context, dRef client.ObjectKey)
...
+	}
+
+	var valuesFiles []string
+	for _, vf := range valuesFrom {
+		envVarName := reconcilermanager.HelmConfigMapRef + "_" + vf.Name
+		valuesFiles = append(valuesFiles, envVarName)
+		c.Env = append(c.Env, corev1.EnvVar{
+			Name: envVarName,
+			ValueFrom: &corev1.EnvVarSource{
+				ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
```
natasha41575, line 437: Here I'm mounting the valuesFile as an environment variable into helm-sync, but I think it will actually be better to just pass to helm-sync the names/keys of the ConfigMaps and have helm-sync read the ConfigMaps to get the values itself. 

I realized that environment variables have a length limit of about 32,000 characters, and valuesFiles can be quite large.

sdowell: Can you mount them as files in the container?

natasha41575: I can! That will probably be better. Thanks

manifests/ns-reconciler-cluster-role.yaml
natasha41575: Ah that's right, I think I confused the roles of the ns-reconciler and the reposync controller which is part of the reconciler-manager. I think we can take out the extra permissions I added, I will double check with some testing tomorrow

natasha41575:
@sdowell Thanks for the feedback, I'm still working on incorporating your comments - will continue tomorrow. Just wanted to push the ConfigMap mounting change so that you and @janetkuo aren't looking at a stale approach if you happen to take a look at that code before that.

pkg/reconcilermanager/controllers/reconciler_base.go
natasha41575: Done.

e2e/testcases/helm_sync_test.go
natasha41575: It needs to manage the resources defined in the [test chart](https://github.com/GoogleContainerTools/kpt-config-sync/tree/main/e2e/testdata/helm-charts/ns-chart/templates). I didn't write the original chart but it seems to me that the HPA doesn't need to be there necessarily. I don't see this chart used anywhere else besides the helm-sync tests, so I took out the HPA. 

It does also need `CoreAdmin`; otherwise it gets this error: 
```polling for status failed: failed to list /v1, Kind=Pod: pods is forbidden: User "system:serviceaccount:config-management-system:ns-reconciler-test-ns" cannot list resource "pods" in API group "" in the namespace "test-ns"
```

So I changed it from `AllAdmin` to just `CoreAdmin`, `AppsAdmin`, and `RepoSyncAdmin`. 

I did some manual testing (applied a RepoSync directly to my cluster) to check that the base ClusterRole has sufficient permissions and it is working fine - helm-sync got the ConfigMap values and successfully rendered the chart.

e2e/testdata/repo-sync-helm-cm-update/configmap-1.yaml
natasha41575: Done

manifests/ns-reconciler-cluster-role.yaml
natasha41575: Did some manual testing and yup, these permissions aren't necessary. Took them out.

e2e/testcases/helm_sync_test.go
natasha41575: done! Now using 2*the default timeout which looks like is dynamic based on whether the cluster is autopilot or not

e2e/testcases/helm_sync_test.go
natasha41575: Done

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +640,121 @@ func TestHelmNamespaceRepo(t *testing.T) {
 	}
 }
 
+// TestHelmNamespaceRepoWithConfigMap verifies RepoSync can pick up values and updates from ConfigMap references.
+// This test will work only with following pre-requisites:
+// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for accessing images in Artifact Registry.
+func TestHelmConfigMapNamespaceRepo(t *testing.T) {
+	repoSyncNN := nomostest.RepoSyncNN(testNs, configsync.RepoSyncName)
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.RequireGKE(t),
+		ntopts.RepoSyncPermissions(policy.RepoSyncAdmin(), policy.AppsAdmin(), policy.CoreAdmin()),
```
sdowell, line 649: Why does the ns reconciler need `RepoSyncAdmin`?

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -486,7 +509,34 @@ func (r *RepoSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
...
+	ctrlr, err := controllerBuilder.Build(r)
+	r.controller = &ctrlr
+
+	return err
+}
+
+func (r *RepoSyncReconciler) watchConfigMaps(rs *v1beta1.RepoSync) error {
+	// We add watches dynamically at runtime based on the RepoSync namespace
+	// in order to avoid watching ConfigMaps in the entire cluster.
+	if rs.Spec.SourceType != string(v1beta1.HelmSource) || rs.Spec.Helm == nil ||
```
janetkuo: Do we remove watches when it's no longer needed (e.g. when the RSync in that namespace is removed or updated to not use ConfigMaps)?

natasha41575: This PR does not do that, but I can add that. I think we'll have to keep a map of RSync to watched namespaces to keep track of that. I agree that removing unneeded watches is for the best, so I will make the change.

e2e/testcases/helm_sync_test.go
natasha41575: You're right, it doesn't. Took it out, thanks for the catch!

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+				resource.MustParse(expectedCPULimit),
+				resource.MustParse(expectedMemoryRequest),
+				resource.MustParse(expectedMemoryLimit)),
+			testpredicates.HasExactlyImage("wordpress", "bitnami/wordpress", "", "sha256:362cb642db481ebf6f14eb0244fbfb17d531a84ecfe099cd3bba6810db56694e"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_USERNAME", "test-user-2"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "TEST_1", "val1"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "TEST_2", "val2"),
+		},
+		testwatcher.WatchTimeout(nt.DefaultWaitTimeout*2)); err != nil {
```
sdowell, line 293: Why does it take so long (2x default) to pickup the change?

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_1", "val1"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_2", "val2")); err != nil {
+		nt.T.Error(err)
+	}
+	if nt.T.Failed() {
+		nt.T.FailNow()
+	}
+
+	nt.T.Log("Apply the ConfigMap with values to the cluster")
```
sdowell, line 251: Can you update the log string here (and elsewhere) to be a bit more descriptive? From looking at the diff in the values file I assume the important bit in this case is updating pullPolicy from `always` -> `never`. Can save the reader some time by making the message more descriptive

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+image:
+  digest: sha256:362cb642db481ebf6f14eb0244fbfb17d531a84ecfe099cd3bba6810db56694e
+  pullPolicy: Never
+`},
+	}
+	if err := nt.KubeClient.Update(&cm3); err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.WaitForRootSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RootSyncs must reference ConfigMaps with valid spec.helm.valuesFrom.key")
+	if nt.T.Failed() {
```
sdowell, line 314: `WaitForRootSyncStalledError` calls `Fatal` by default, so this check is redundant

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+			resource.MustParse(expectedCPURequest),
+			resource.MustParse(expectedCPULimit),
+			resource.MustParse(expectedMemoryRequest),
+			resource.MustParse(expectedMemoryLimit)),
+		testpredicates.HasExactlyImage("wordpress", "bitnami/wordpress", "", "sha256:362cb642db481ebf6f14eb0244fbfb17d531a84ecfe099cd3bba6810db56694e"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_USERNAME", "test-user-2"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_1", "val1"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_2", "val2")); err != nil {
+		nt.T.Error(err)
```
sdowell, line 330: use `Fatal` and remove `Failed` check below

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+	if nt.T.Failed() {
+		nt.T.FailNow()
+	}
+
+	nt.T.Log("Delete the referenced ConfigMap")
+	if err := nt.KubeClient.Delete(&cm3); err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.WaitForRootSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RootSyncs must reference valid ConfigMaps in spec.helm.valuesFrom: ConfigMap \"foo\" not found")
+	if nt.T.Failed() {
```
sdowell, line 341: ditto

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+			resource.MustParse(expectedCPURequest),
+			resource.MustParse(expectedCPULimit),
+			resource.MustParse(expectedMemoryRequest),
+			resource.MustParse(expectedMemoryLimit)),
+		testpredicates.HasExactlyImage("wordpress", "bitnami/wordpress", "", "sha256:362cb642db481ebf6f14eb0244fbfb17d531a84ecfe099cd3bba6810db56694e"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_USERNAME", "test-user-2"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_1", "val1"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_2", "val2")); err != nil {
+		nt.T.Error(err)
```
sdowell, line 357: ditto

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+				resource.MustParse(expectedMemoryRequest),
+				resource.MustParse(expectedMemoryLimit)),
+			testpredicates.HasExactlyImage("wordpress", "bitnami/wordpress", "", "sha256:362cb642db481ebf6f14eb0244fbfb17d531a84ecfe099cd3bba6810db56694e"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_USERNAME", "test-user-1"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "TEST_1", "val1"),
+			testpredicates.DeploymentHasEnvVar("wordpress", "TEST_2", "val2"),
+		},
+		testwatcher.WatchTimeout(nt.DefaultWaitTimeout*2)); err != nil {
+		nt.T.Error(err)
```
sdowell, line 383: ditto

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+		nt.T.Fatal(err)
+	}
+
+	if err := nt.Validate("my-wordpress", "wordpress", &appsv1.Deployment{},
+		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_USERNAME", "test-user-2"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_INLINE", "inline"),
+		testpredicates.DeploymentMissingEnvVar("wordpress", "TEST_CM_1"),
+		testpredicates.DeploymentMissingEnvVar("wordpress", "TEST_CM_2")); err != nil {
+		nt.T.Error(err)
```
sdowell, line 466: `Fatal` for this test as well

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: I looked into it a bit more and it seems that there is no current way to stop a watch once you've started one. There's an open PR and an open issue in controller-runtime, with updates as recent as 3 days ago:

https://github.com/kubernetes-sigs/controller-runtime/pull/2159
https://github.com/kubernetes-sigs/controller-runtime/issues/1884

I added a TODO to this function instead to remind is to keep an eye on that. I found some other similar issues (e.g. https://github.com/opensergo/opensergo-control-plane/issues/23) who are also waiting for the fix.

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +640,121 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
+	}}
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add(nomostest.StructuredNSPath(repoSyncNN.Namespace, repoSyncNN.Name), rs))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update RepoSync to sync from a private Helm Chart without cluster scoped resources"))
+	err = nt.WatchForAllSyncs(nomostest.WithRepoSha1Func(helmChartVersion(privateNSHelmChartVersion)), nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{repoSyncNN: remoteHelmChart.ChartName}))
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	if err := nt.Validate(rs.Spec.Helm.ReleaseName+"-"+remoteHelmChart.ChartName, testNs, &appsv1.Deployment{},
+		testpredicates.HasLabel("labelsTest", "first")); err != nil {
+		nt.T.Error(err)
```
sdowell, line 690: `Fatal`

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +640,121 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
+		},
+	}
+	if err := nt.KubeClient.Update(&cm2); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	if err = nt.Watcher.WatchObject(kinds.Deployment(), rs.Spec.Helm.ReleaseName+"-"+remoteHelmChart.ChartName, testNs,
+		[]testpredicates.Predicate{testpredicates.HasLabel("labelsTest", "second")},
+		testwatcher.WatchTimeout(nt.DefaultWaitTimeout*2)); err != nil {
+		nt.T.Error(err)
```
sdowell, line 708: `Fatal`

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +640,121 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
+			"values.yaml": `label: second`,
+		},
+	}
+	if err := nt.KubeClient.Update(&cm3); err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.WaitForRepoSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RepoSyncs must reference ConfigMaps with valid spec.helm.valuesFrom.key")
+	if err := nt.Validate(rs.Spec.Helm.ReleaseName+"-"+remoteHelmChart.ChartName, testNs, &appsv1.Deployment{},
+		testpredicates.HasLabel("labelsTest", "second")); err != nil {
+		nt.T.Error(err)
```
sdowell, line 725: `Fatal`

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +640,121 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
+	}
+
+	nt.T.Logf("Delete the ConfigMap with values")
+	if err := nt.KubeClient.Delete(&cm3); err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.WaitForRepoSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RepoSyncs must reference valid ConfigMaps in spec.helm.valuesFrom: ConfigMap \"foo\" not found")
+	if err := nt.Validate(rs.Spec.Helm.ReleaseName+"-"+remoteHelmChart.ChartName, testNs, &appsv1.Deployment{},
+		testpredicates.HasLabel("labelsTest", "second")); err != nil {
+		nt.T.Error(err)
```
sdowell, line 738: `Fatal`

e2e/testcases/helm_sync_test.go
```diff
@@ -289,6 +640,121 @@ func TestHelmNamespaceRepo(t *testing.T) {
...
+	}
+
+	nt.T.Log("Reapply the ConfigMap with values")
+	if err := nt.KubeClient.Create(cm1Copy); err != nil {
+		nt.T.Fatal(err)
+	}
+	if err = nt.Watcher.WatchObject(kinds.Deployment(), rs.Spec.Helm.ReleaseName+"-"+remoteHelmChart.ChartName, testNs,
+		[]testpredicates.Predicate{testpredicates.HasLabel("labelsTest", "first")},
+		testwatcher.WatchTimeout(nt.DefaultWaitTimeout*2)); err != nil {
+		nt.T.Error(err)
```
sdowell, line 751: `Fatal`

pkg/helm/helm.go
```diff
@@ -104,14 +104,79 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
...
+		}
+		if len(h.Values) != 0 {
+			klog.Infof("inline values %s\n", h.Values)
+			args, err = writeValuesPath([]byte(h.Values), "", args)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+	case "merge":
```
sdowell, line 134: These modes should be defined as constants somewhere

Also I find the name `merge` a bit misleading here, as the other mode is still technically doing a merge but with a different strategy. Can we come up with a more precise name for this mode?

natasha41575: Hmm... perhaps the ideal name would give the connotation that the files are being combined into one file before being passed to helm. Something like `combine`? Or `mergeFiles`? Do either of those sound more intuitive?

pkg/helm/merge.go
```diff
@@ -0,0 +1,125 @@
...
+//	```
+//	namespaces:
+//	  - name: foo
+//	    rolebindings:
+//	      name: foo-rb
+//	  - name: bar
+//	    rolebindings:
+//	      name: bar-rb
+//	```
+func merge(valuesToMerge [][]byte) ([]byte, error) {
```
sdowell, line 54: Can you add unit tests for each of these merge functions?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -421,6 +422,43 @@ func (r *reconcilerBase) deployment(ctx context.Context, dRef client.ObjectKey)
 	return deployObj, nil
 }
 
+func mountConfigMapValuesFiles(templateSpec *corev1.PodSpec, c *corev1.Container, valuesFrom []v1beta1.ValuesFrom) {
```
sdowell, line 425: Can you add some unit tests which verify the shape of the resulting spec?

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -911,6 +1059,7 @@ func (r *RepoSyncReconciler) mutationsFor(ctx context.Context, rs *v1beta1.RepoS
 					if authTypeToken(rs.Spec.Helm.Auth) {
 						container.Env = append(container.Env, helmSyncTokenAuthEnv(secretName)...)
 					}
+					mountConfigMapValuesFiles(templateSpec, &container, newValuesFrom)
```
sdowell, line 1062: needs unit tests

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -845,6 +935,7 @@ func (r *RootSyncReconciler) mutationsFor(ctx context.Context, rs *v1beta1.RootS
 					if authTypeToken(rs.Spec.Helm.Auth) {
 						container.Env = append(container.Env, helmSyncTokenAuthEnv(secretRefName)...)
 					}
+					mountConfigMapValuesFiles(templateSpec, &container, rs.Spec.Helm.ValuesFrom)
```
sdowell, line 938: needs unit tests

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: As discussed offline, I looked at [the custom watcher in Config Sync](https://github.com/GoogleContainerTools/kpt-config-sync/blob/21100ec59b8419fbe8fbd05b21933c72aebf6d8e/pkg/remediator/watch/manager.go#L79), but it looks pretty specific to the reconciler so I don't think it supports this use case. I think the best option is to leave the watches for now and wait for the fix in controller-runtime for the future.

e2e/testcases/helm_sync_test.go
natasha41575: Fixed

e2e/testcases/helm_sync_test.go

e2e/testcases/helm_sync_test.go
```diff
@@ -136,6 +139,354 @@ func TestPublicHelm(t *testing.T) {
...
+// TestHelmWatchConfigMap can run on both Kind and GKE clusters.
+// It tests that helm-sync properly watches ConfigMaps in the RSync namespace and pulls updated values.
+func TestHelmWatchConfigMap(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.Unstructured)
+
+	configMapTypeMeta := metav1.TypeMeta{Kind: "ConfigMap"}
+	configMapObjMeta := metav1.ObjectMeta{Name: "foo", Namespace: configsync.ControllerNamespace}
+
+	nt.T.Log("Apply the ConfigMap with values to the cluster")
+	cm1 := corev1.ConfigMap{
```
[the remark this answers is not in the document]
natasha41575: Done, thanks for the tip!

pkg/helm/merge.go
natasha41575: done

pkg/reconcilermanager/controllers/reconciler_base.go
natasha41575: done

pkg/helm/helm.go
janetkuo: The difference is how it treats the same key found in different values files. `override` overrides the values of the key with the one comes latter, whereas `merge` merges the values of the same key from different values files.

janetkuo: We could update the CRD doc (i.e. comment in type) to make the behavior more clear (in relation to how keys are handled which users would understand better vs. how we implement it).

sdowell: Documenting the field in CRD helps, but I don't think it's sufficient. IMO the name itself should more clearly distinguish the difference between the modes.

As an example, I got confused when reading the e2e test case because the test behaves like:
- set up a `RootSync` with `valuesMergeMode` unset and with two values files
- verify that the two values files get merged
- set `valuesMergeMode=merge`
- verify that the two values files get merged, but this time merged with a different strategy

With a name like `valuesMergeMode=merge`, I would expect the alternative is that the values files don't get merged at all - for example the reconciler will stall/error if you define multiple values files.

natasha41575: I think part of your confusion with the e2e was my fault, for not clarifying in the test that the default for `valuesMergeMode` is `override` (and I can add that). 

I do think the name `valuesMergeMode` already implies that the values are going to be merged in some way - and it wouldn't make sense for a user to put multiple values files in their RSync unless they were expecting them to be merged. I think `override` will be used in the vast majority of cases because it is equivalent to using multiple values files with helm CLI . 

I'm OK to rename `merge` to something else though, but I don't know if I can think of a great alternative. So unless we have other name candidates, and document properly what each option means and what the default is.

pkg/reconcilermanager/controllers/reposync_controller.go
janetkuo: Sam, what I'm trying to say is that the we're merging vs. overriding values of the same keys in different values files, so the mode types makes sense to me.

pkg/helm/helm.go
sdowell: > Sam, what I'm trying to say is that the we're merging vs. overriding values of the same keys in different values files, so the mode types makes sense to me.

The point I'm trying to make is that `override` (the default behavior) as defined here is just another strategy for merging two objects together in the conventional sense. When that's being contrasted against the other mode, `merge`, I see a user scratching their head wondering what `valuesMergeMode: merge` means in this context since `override` was also performing a different type of conventional merge. Perhaps from a different angle, merely observing that the value `merge` is a substring of the field `valuesMergeMode` seems like a bit of a smell to me. The alternative of `concatenate` that Natasha mentioned seems closer to a intuitive name to me, although she mentioned that isn't precise either.

I'm fine with conceding on this point if I'm getting overruled here

pkg/reconcilermanager/controllers/reposync_controller.go
janetkuo: FYI I came across this issue that talks about a feature request for merge instead of override https://github.com/helm/helm/issues/3486

sdowell: Maybe the merge strategies could be named `shallow`/`deep`?

pkg/helm/helm.go
natasha41575: > These modes should be defined as constants somewhere

Done.

I also updated the godoc/crd doc and added a comment to the test to make it more clear what the behavior is.

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: As discussed offline, we will live it as `override` and `merge`, but rename the field to `valuesKeyMergeMode`.

pkg/helm/helm.go
natasha41575: As discussed offline, we will live it as `override` and `merge`, but rename the field to `valuesKeyMergeMode`.

cmd/helm-sync/main.go
```diff
@@ -38,6 +39,10 @@ var (
 		"the version of the helm chart being synced")
 	flValues = flag.String("values", os.Getenv(reconcilermanager.HelmValues),
 		"set the helm chart values, will be used to override the default values")
+	flValuesFrom = flag.String("values-from", os.Getenv(reconcilermanager.HelmValuesFrom),
+		"set the helm chart values, will be used to override the default values")
+	flValuesKeyMergeMode = flag.String("values-key-merge-mode", os.Getenv(reconcilermanager.HelmValuesKeyMergeMode),
```
karlkfi, line 44: "value-merge-strategy" might be preferable

cmd/helm-sync/main.go
```diff
@@ -38,6 +39,10 @@ var (
 		"the version of the helm chart being synced")
 	flValues = flag.String("values", os.Getenv(reconcilermanager.HelmValues),
 		"set the helm chart values, will be used to override the default values")
+	flValuesFrom = flag.String("values-from", os.Getenv(reconcilermanager.HelmValuesFrom),
```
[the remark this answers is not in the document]
karlkfi: "values-files" might be preferable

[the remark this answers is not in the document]
natasha41575: It's from files, I'll update the flag name

manifests/reposync-crd.yaml
```diff
@@ -1186,8 +1186,52 @@ spec:
...
+                            Required.
+                          enum:
+                          - ConfigMap
+                          type: string
+                        name:
+                          description: name represents the Object name. Required.
+                          type: string
+                      type: object
+                    type: array
+                  valuesKeyMergeMode:
```
karlkfi, line 1220: "valuesFileMergeStrategy" might be preferable

natasha41575: I'm happy to update the flag name in helm-sync, but we had a really long discussion and a lot of back and forth on this field name (and its options), and I'd prefer not to open that up again

manifests/reposync-crd.yaml
```diff
@@ -1186,8 +1186,52 @@ spec:
...
+                    default: override
+                    description: valuesKeyMergeMode specifies the strategy for handling
+                      multiple valueFiles. It refers to how values are merged together
+                      if multiple valuesFile define the same key. Can be 'override'
+                      or 'merge'. 'override' (default) results in the duplicated keys
+                      in later files to override the value from earlier files. This
+                      is equivalent to passing in multiple valuesFiles to Helm CLI.
+                      'merge' results in the duplicated keys being merged together
```
karlkfi, line 1228: This description of merging could be more clear... 
Is it just strings that are merged? how are lists and maps merged?
Is this a helm feature or a CS-specific feature?
Did a customer ask for it?

It feels like "merge" is a less than ideal name if not all field types are merged...

natasha41575: This is a requirement for fleet tenancy integration. I can update the description more clearly. Lists are concatenated, maps are merged, scalars are overridden (since there's no real way to "merge" those).

e2e/testdata/helm-charts/ns-chart/templates/hpa.yaml
```diff
@@ -1,30 +0,0 @@
-{{- if .Values.autoscaling.enabled }}
```
janetkuo, line 1: Not critical, but why's this removed?

manifests/reposync-crd.yaml
natasha41575: Also echoing my above comment, we had a really, really long discussion on this name and I'd prefer not to open that up again

pkg/api/configsync/v1beta1/helmconfig.go
```diff
@@ -93,3 +117,19 @@ type HelmBase struct {
...
+	// +kubebuilder:validation:Enum=ConfigMap
+	Kind string `json:"kind,omitempty"`
+
+	// name represents the Object name. Required.
+	Name string `json:"name,omitempty"`
+
+	// key represents the object data key to read the value from.
+	// +kubebuilder:default:=values.yaml
+	// +optional
+	Key string `json:"key,omitempty"`
```
karlkfi, line 134: The name "Key" seems specific to the Kind ConfigMap.
If this were any other resource type, Key wouldn't make any sense.
Should it be "Field" instead?
Should it support JSONPath to support other resource types?

karlkfi: If we only support ConfigMaps, do we need this complexity of specifying Kind and Key?

pkg/reconcilermanager/constants.go
```diff
@@ -136,8 +136,14 @@ const (
...
-	//HelmValuesFiles is the OS env variable key for Helm values files.
-	HelmValuesFiles = "HELM_VALUES_FILES"
+	// HelmConfigMapRef is the path where valuesFiles from ConfigMaps are mounted.
+	HelmConfigMapRef = "HELM_CONFIGMAP_REF"
+
+	// HelmValuesFrom is the OS env variable key for the names of ConfigMaps holding valuesFiles.
+	HelmValuesFrom = "HELM_VALUES_FROM"
+
+	// HelmValuesKeyMergeMode is the OS env variable key for Helm values files merge strategy.
+	HelmValuesKeyMergeMode = "HELM_VALUES_MERGE_MODE"
```
janetkuo, line 146: nit: should it be HELM_VALUES_KEY_MERGE_MODE?

janetkuo (approved):
Looks good

e2e/testdata/helm-charts/ns-chart/templates/hpa.yaml
natasha41575: I'm struggling to find the related comment thread above, but I removed it so that we wouldn't have to grant the ns-reconciler permission to manage all resource types in the RepoSync test. Without this, that test can just grant permissions for core and app kinds.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -492,6 +495,90 @@ func (r *RootSyncReconciler) mapMembershipToRootSyncs() func(client.Object) []re
...
+				// if the root-reconciler doesn't exist, an updated ConfigMap should trigger a reconcile
+				// to create one
+				result = append(result, requeueRootSyncRequest(obj, &rs)...)
+			}
+			continue
+		}
+
+		// A rereconcile doesn't necessarily update an existing root-reconciler deployment, so
+		// the root-reconciler pods need to be explicitly restarted to pick up the ConfigMap update.
+		reconciler.Spec.Template.ObjectMeta.Annotations["kubectl.kubernetes.io/restartedAt"] = time.Now().Format(time.RFC3339)
```
karlkfi, line 569: What the heck is this?

Is this trying to evict the pods or just restart them?

pkg/reconcilermanager/constants.go
natasha41575: Ah I missed that one. Thanks for the catch, I'll update that

karlkfi (dismissed):
Not a fan of the naming decisions here. See inline comments.

pkg/reconcilermanager/controllers/rootsync_controller.go
natasha41575: It just restarts them, otherwise the pod doesn't pick up the change in the ConfigMap. Inspiration came from this: https://stackoverflow.com/questions/61335318/how-to-restart-a-deployment-in-kubernetes-using-go-client, as I was looking for the equivalent of `kubectl rollout restart`.

karlkfi:
I've been informed that many of these naming decisions were documented in the design doc and approved by other people. So I will summarize my concerns here for posterity.

1. I think copying the valuesFrom syntax from Flux is a mistake. We're not support values from Secrets, so we don't need the complexity of specifying the Kind or Key. We could provide a much simpler API instead. IF we DO plan to add support for Secrets, then we should just do that now. It's not that much work and it makes the API decisions make more sense.
2. I generally prefer "Strategy" over "Mode". I did some research and found examples of both in the K8s API, but the design pattern is called the "strategy pattern". "Mode" is better used when selecting between multiple models.
3. The reason Kubernetes uses the strategy pattern so often (instead of booleans) is to allow extension after initial implementation. With this in mind, the "merge" value is a bad choice of value, because it makes the API `mergeMode: merge`, which i both redundant and all encompassing. You look at it and think "what other possible values could there be?".
4. Not only is "merge" bad for the above reasons, but it's also confusing because it's only actually merging some values types. Scalars are not being merged. Lists are being concatonated/joined. Only maps/objects are actually being merged... We need a better strategy name.
5. If we are going to copy flux's API for valuesFrom, we should at least copy the whole thing and use valuesKey. Why rename it "key"?
6. If we're not going to copy flux's API exactly, I think "key" is an inaccurate name. Maps have keys. Resource objects have fields. If we really want to allow future extension, with other resources than just ConfigMap and Secret, we should use "field" and support JSONPath or at least period separated field path syntax. For example, if using a ConfigMap, you might set `valuesField: data["values.yaml"]` (since you can't use dot-syntax with a key that contains a dot).
7. And if we really want to allow support from other resources as inputs, we might also want to avoid hard coding Secret kind values as needing base64 decoding. That could be done by adding an optional `valuesEncoding: base64` field.
8. Whether we do or don't keep the valueFrom syntax, we should make the command line flag more clear. It's not obvious from the flag name or the description how to use it or what the values should be.
9. If we don't need to copy Flux, then I would recommend not using `-From` in a field name. The name of a list field should in most cases include the type of the entries it expects, for readability and discoverability. `-Sources` or `-Objects` might be more clear. Then the list field can be plural and then entry object can be singlular. That avoids the plural and singular having the same name (e.g. `valuesFrom = []ValuesFrom{}`).

janetkuo:
The goal isn't to replicate Flux's API, so we don't need to follow them. We also don't need to support Secrets now given that no users ask for it. I prefer strategy over mode as well. 

What should the second strategy be called, if not "merge"? It's already a term that's used by Helm users in the feature request for this functionality. How do you expect Scalars and Lists to be merged? If the concern is about the redundant mention of "merge", we can call it something like `valuesFilesApplyStrategy` (the strategy to apply multiple values files).

natasha41575:
Based on the above comments, discussed with @janetkuo and we are going to make the following naming changes:

`valuesKeyMergeMode` -> `valuesKeyApplyStrategy`
`valuesFrom` -> `valuesSources`
`valuesSources.Kind` will be optional, and default to ConfigMap 
`valuesSources.Key` -> `valuesSources.valuesFile`.

karlkfi:
Can we just leave out the kind and key? 
Do they need to be configurable?
If that capability wasn’t asked for by the CX can we punt on how to configure it?
Having a source just me an object with a name field seems nice and simple. Tho I’m also ok with having a kind that defaults to ConfigMap, because that makes the API slightly more readable without needing to see the reference docs. 
However, leaving out the Kind lets us punt on whether we only support resource objects as sources. We might eventually want to add a URL to a values.yaml, for example. 

Other alternative names:
- valuesKeyMergeMode -> valuesFileMergeStrategy (keys aren’t merging, files are merging)
- valuesFrom -> valuesFileSources (The ConfigMap doesn’t contain values. There’s only one data key. It contains a values file encoded as a yaml string.)

karlkfi:
Another consideration:
Can I use values & valuesFrom? What order do they merge/override?

natasha41575:
> Can we just leave out the kind and key?
> Do they need to be configurable?
> If that capability wasn’t asked for by the CX can we punt on how to configure it?

The customer didn't ask for it, so we could leave it out for now if we want to punt on the design. The `key` allows the user to use the same ConfigMap for multiple valuesFiles, which seems convenient to me but it is something we could theoretically add later. Same with kind, we could add it later if we don't want to do it now. I very slightly prefer keeping them but am OK either way. I'll defer to @janetkuo to be the tiebreaker on this. 

> valuesKeyMergeMode -> valuesFileMergeStrategy (keys aren’t merging, files are merging)

This still has the same issue that you mentioned above with `merge` being duplicated in the field name and the value. I made a typo above (in case you saw the comment before I edited it), Janet actually suggested `valuesFilesApplyStrategy` as it is the strategy of how we apply the valuesFiles to the chart; either they "override" each other or we "merge" them together. I think of the options discussed so far I still prefer this. 

> valuesFrom -> valuesFileSources

This SGTM. I think it is more accurate than the above `valuesSources` that I suggested.

> Can I use values & valuesFrom? What order do they merge/override?

Yes you can. `values` is treated as the last valuesFile (so will override any valuesFile from valuesFileSource). There's an explanation in the CRD description and godoc, would appreciate feedback if the wording is clear from there.

karlkfi:
As for the “merge” strategy enum… I think there are a few approaches.
1. Consider what other types of possible strategies we might want to add later and how you might name them all to be unique. 
2. Look for other yaml/json merge tools and see what options they expose. 

For example, in the future, we may want to allow lists to merge using a unique key field, because that’s a very common pattern in k8s config: unordered lists of maps (e.g. containers). Concatenating is only one way to merge lists and probably not the most useful. 

Another consideration is that current merge strategy is actually using “override” on maps. So it’s not super clear what the distinction is between those modes. 

As for prior art, consider CRD merge strategies: https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

So maybe what we really need is type-specified strategy options… I have a couple ideas:
1. mapMergeStrategy, listMergeStrategy, stringMergeStrategy, scalarMergeStrategy
2.  a list of type strategy maps, like fieldMergeStrategies: [{type: map, strategy: overwrite}, {type: list, strategy: listMapKey, key: blah}]
3. a list of default and/or field-specific strategy maps, like fieldMergeStrategies: [{type: list, strategy: atomic}, {type: list, strategy: map, listMapKey: blah, fieldPath: “a.b.c”}]

Option 3 would allow for specifying default and path specific merge strategies for the same type, in case that’s desired. That would more closely mirror the flexibility provided by CRD schema.

janetkuo:
Good point that we can add `kind` later, so it's fine to remove that field now and we can add it when we need to support more types (and keep the default as ConfigMap). 

We're handling values.yaml instead of KRM yaml. Does it  allow those merge strategies? The merge strategy proposed in this PR seems good for our use case.

karlkfi:
> We're handling values.yaml instead of KRM yaml. Does it allow those merge strategies? The merge strategy proposed in this PR seems good for our use case.

AFAIK, helm values can be any data structure supported by yaml. You could have a value named `spec` that was the entire multi-layer resource spec map. It wouldn't even surprise me if it supports non-string keys and non-json-compatible fields, even tho those aren't allowed in KRM yaml. Helm templates use Go-templating to translate values for interpolation, so you can do almost anything with them.

pkg/reconcilermanager/controllers/rootsync_controller.go
janetkuo: We shouldn't need this given that mounted ConfigMap is updated automatically https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically There might be some delay before it detects the change.

janetkuo:
I don't think we need to support other types of merge strategies - I would rather it to be supported by Helm. Natasha implemented this for our own use case. I'm fine with renaming the current strategy from "merge" to something more descriptive.

karlkfi:
Are we certain that the merge behavior, as described, meets the needs of the Fleet Tenancy team?

I'd doubtful that list concatenation is at all useful.

I'm also unsure if "override" works like "set" or "atomic" or some new behavior unlike any existing CRD merge strategy.

pkg/reconcilermanager/controllers/rootsync_controller.go
natasha41575: Discussed offline, another reason to restart the pods is that currently helm-sync only compares version to trigger refresh and rerender. If we don't restart the pod, we will have to compare all the valuesFiles for changes as well.

We will recommend to the best practice of treating ConfigMaps as immutable, and leave the restart as a workaround for now.

natasha41575:
I have changed `merge` to just do list concatenation, and changed its name to `listConcatenate` as that's all that is required for now.

janetkuo (approved):
/lgtm

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: One of the reasons we stopped using ConfigMaps to configure the reconciler was that changing the ConfigMap doesn’t automatically update the Pods. But worse, it also doesn’t update the Deployment, which means the generation doesn’t change. So there’s no way to know if the Deployment status reflects the latest ConfigMap change. 

This will probably be a problem for customers who want to wait until their config change takes effect. 

Do we have a way to communicate that reconciliation?

sdowell: Could we delete the Pod instead to force a restart?

PR #706, opened 2023-06-23
remove the render-helm-chart kpt function from ConfigSync

natasha41575:
b/284029611

This support is undocumented, still alpha in kustomize, and we can take it out now that first party helm support is available.


janetkuo (approved):
/lgtm

PR #707, opened 2023-06-24
perf: limit helm-sync polling to every hour when a version range is specified

natasha41575:
When a version range is detected in the RSync `spec.helm.version`, we currently continuously poll to get the latest version from that range, causing performance degradation. 

This PR does the following:
- by default, we poll every hour for the latest version
- make the poll configurable in helm-sync for testing purposes, but we don't make it configurable through the RSync API
- use 1m as the poll period in tests

b/288478227

### DIFF
--- a/cmd/helm-sync/main.go
+++ b/cmd/helm-sync/main.go
@@ -128,7 +128,7 @@ func main() {
 			Password:        *flPassword,
 		}
 
-		var refreshVersion bool
+		refreshVersion := false
 		if initialSync || time.Now().After(start.Add(duration)) {
 			refreshVersion = true
 			start = time.Now()
--- a/cmd/helm-sync/main.go
+++ b/cmd/helm-sync/main.go
@@ -67,6 +67,8 @@ var (
 		"the username to use for helm authantication")
 	flPassword = flag.String("password", util.EnvString("HELM_SYNC_PASSWORD", ""),
 		"the password or personal access token to use for helm authantication")
+	flVersionPollPeriod = flag.String("version-poll-period", util.EnvString(reconcilermanager.HelmSyncVersionPollingPeriod, "1h"),
+		"the amount of time between version polling when a version range is specified")
 )
 
 func main() {
@@ -76,7 +78,7 @@ func main() {
 		"--chart", *flChart, "--version", *flVersion, "--root", *flRoot,
 		"--values", *flValues, "--include-crds", *flIncludeCRDs, "--dest", *flDest, "--wait", *flWait,
 		"--error-file", *flErrorFile, "--timeout", *flSyncTimeout,
-		"--one-time", *flOneTime, "--max-sync-failures", *flMaxSyncFailures)
+		"--one-time", *flOneTime, "--max-sync-failures", *flMaxSyncFailures, "--version-poll-period", *flVersionPollPeriod)
 
 	if *flRepo == "" {
 		utillog.HandleError(log, true, "ERROR: --repo must be specified")
@@ -100,8 +102,14 @@ func main() {
 		}
 	}
 
+	duration, err := time.ParseDuration(*flVersionPollPeriod)
+	if err != nil {
+		utillog.HandleError(log, true, "ERROR: --version-poll-period must be a valid duration")
+	}
+
 	initialSync := true
 	failCount := 0
+	start := time.Now()
 	for {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(*flSyncTimeout))
 		hydrator := &helm.Hydrator{
@@ -119,7 +127,14 @@ func main() {
 			UserName:        *flUsername,
 			Password:        *flPassword,
 		}
-		if err := hydrator.HelmTemplate(ctx); err != nil {
+
+		var refreshVersion bool
+		if initialSync || time.Now().After(start.Add(duration)) {
+			refreshVersion = true
+			start = time.Now()
+		}
+
+		if err := hydrator.HelmTemplate(ctx, refreshVersion); err != nil {
 			if *flMaxSyncFailures != -1 && failCount >= *flMaxSyncFailures {
 				// Exit after too many retries, maybe the error is not recoverable.
 				log.Error(err, "too many failures, aborting", "failCount", failCount)
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -88,6 +88,8 @@ var (
 	reconcilerPollingPeriod time.Duration
 	// hydrationPollingPeriod specifies hydration-controller polling period as time.Duration
 	hydrationPollingPeriod time.Duration
+	// helmSyncVersionPollingPeriod specifies the helm-sync version polling period as time.Duration.
+	helmSyncVersionPollingPeriod time.Duration
 )
 
 // IsReconcilerTemplateConfigMap returns true if passed obj is the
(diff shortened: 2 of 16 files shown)


e2e/testcases/helm_sync_test.go
```diff
@@ -200,6 +202,18 @@ func TestHelmLatestVersion(t *testing.T) {
 		nt.T.Fatalf("failed to push helm chart: %v", err)
 	}
 
+	nt.T.Log("Update reconciler-manager-cm to override the helm-sync default poll period")
```
sdowell, line 205: Similar test-related alterations to things like polling periods are done [here](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/config_sync.go#L142). Suggest doing it there for consistency and so that it is set for all tests by the test scaffolding

natasha41575: Thanks for the tip! Will update

natasha41575: So I discovered that the current way that polling periods are overridden in tests is actually not working, because it is updating the data in [reconciler-manager-cm](https://github.com/GoogleContainerTools/kpt-config-sync/blob/5ffda55c6c928d234206af6e9233451969151eff/e2e/nomostest/config_sync.go#L578), but the environment is mounted from [reconciler-manger](https://github.com/GoogleContainerTools/kpt-config-sync/blob/5ffda55c6c928d234206af6e9233451969151eff/manifests/templates/reconciler-manager.yaml#L57) (which AFAIK is a ConfigMap that is only used for e2e tests, so it makes sense to put the poll overrides there). I added some logging and saw that yup, during e2e tests, the reconciler and hydration polling periods are still currently set to their default, instead of the having override that they're supposed to have.

I included a fix to the `setPollingPeriod` function in `nomostest` to fix that. And updated this helm-sync test accordingly, to follow a similar model.

e2e/nomostest/config_sync.go
```diff
@@ -88,16 +88,26 @@ var (
...
-func IsReconcilerManagerConfigMap(obj client.Object) bool {
+func IsReconcilerManagerCMConfigMap(obj client.Object) bool {
 	return obj.GetName() == "reconciler-manager-cm" &&
 		obj.GetNamespace() == "config-management-system" &&
 		obj.GetObjectKind().GroupVersionKind() == kinds.ConfigMap()
 }
 
+// IsReconcilerManagerConfigMap returns true if passed obj is the
+// reconciler-manager ConfigMap reconciler-manager in config-management namespace.
+func IsReconcilerManagerConfigMap(obj client.Object) bool {
```
natasha41575, line 105: See the explanation in https://github.com/GoogleContainerTools/kpt-config-sync/pull/707#discussion_r1241009308

e2e/testcases/helm_sync_test.go
natasha41575: Looks like using the sped up reconciler and hydration polling periods caused some presubmits to fail, I'm increasing them to the default to see if it fixes the presubmit

natasha41575: It looks like increasing the reconciler/hydration polling periods back to their default fixed the test failures. So this PR is essentially leaving the hydration/reconciler poling delays as it was before (since the override wasn't working). It is just modifying the helm-sync-version polling period.

I think we can play around with modifying the hydration/reconciler polling periods as a followup, but I don't think it belongs in this PR since it is unrelated to the helm-sync changes.

e2e/nomostest/new.go
```diff
@@ -252,9 +253,11 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	// Speed up the delay between sync attempts to speed up testing
 	// (default: 15s), but leave the remediator (paused while syncing) and retry
 	// code (1s delay) time to execute between sync attempts.
-	nt.ReconcilerPollingPeriod = 2 * time.Second
+	nt.ReconcilerPollingPeriod = 15 * time.Second
 	// Speed up the delay between rendering attempts to speed up testing (default: 5s)
-	nt.HydrationPollingPeriod = 1 * time.Second
+	nt.HydrationPollingPeriod = 5 * time.Second
```
natasha41575, line 258: Had to bump these back to the default to address test failures, see https://github.com/GoogleContainerTools/kpt-config-sync/pull/707#discussion_r1241011085

e2e/testcases/helm_sync_test.go
sdowell: Good catch, sounds like it was bugged before. Thank you for fixing, agreed that you should just leave the other overrided values at their default and it can be investigated separately

e2e/nomostest/new.go
```diff
@@ -252,9 +253,11 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	// Speed up the delay between sync attempts to speed up testing
 	// (default: 15s), but leave the remediator (paused while syncing) and retry
 	// code (1s delay) time to execute between sync attempts.
-	nt.ReconcilerPollingPeriod = 2 * time.Second
+	nt.ReconcilerPollingPeriod = 15 * time.Second
```
sdowell, line 256: Can you add a `TODO` to revisit this?

natasha41575: Added!

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -706,7 +707,7 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 	case v1beta1.OciSource:
 		result[reconcilermanager.OciSync] = ociSyncEnvs(rs.Spec.Oci.Image, rs.Spec.Oci.Auth, v1beta1.GetPeriodSecs(rs.Spec.Oci.Period))
 	case v1beta1.HelmSource:
-		result[reconcilermanager.HelmSync] = helmSyncEnvs(&rs.Spec.Helm.HelmBase, rs.Namespace, "")
+		result[reconcilermanager.HelmSync] = helmSyncEnvs(&rs.Spec.Helm.HelmBase, rs.Namespace, "", r.helmSyncVersionPollingPeriod.String())
```
sdowell, line 710: please add unit tests

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -643,7 +644,7 @@ func (r *RootSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 	case v1beta1.OciSource:
 		result[reconcilermanager.OciSync] = ociSyncEnvs(rs.Spec.Oci.Image, rs.Spec.Oci.Auth, v1beta1.GetPeriodSecs(rs.Spec.Oci.Period))
 	case v1beta1.HelmSource:
-		result[reconcilermanager.HelmSync] = helmSyncEnvs(&rs.Spec.Helm.HelmBase, rs.Spec.Helm.Namespace, rs.Spec.Helm.DeployNamespace)
+		result[reconcilermanager.HelmSync] = helmSyncEnvs(&rs.Spec.Helm.HelmBase, rs.Spec.Helm.Namespace, rs.Spec.Helm.DeployNamespace, r.helmSyncVersionPollingPeriod.String())
```
sdowell, line 647: please add unit tests

cmd/helm-sync/main.go
```diff
@@ -119,7 +127,14 @@ func main() {
 			UserName:        *flUsername,
 			Password:        *flPassword,
 		}
-		if err := hydrator.HelmTemplate(ctx); err != nil {
+
+		var refreshVersion bool
```
sdowell, line 131: minor nit: assign this as `refreshVersion := false`. It's functionally the same but makes the intent a bit clearer IMO

natasha41575:
Rebased and addressed comments!

cmd/helm-sync/main.go
natasha41575: done

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: added

sdowell (approved):

PR #708, opened 2023-06-24
feat: dynamically disable hydration-controller

sdowell:
The hydration-controller is used for last mile hydration when the source of truth contains kustomization files, and is a no-op otherwise. This change optimizes the resource footprint of reconciler Pods by dynamically detecting whether the source of truth requires hydration, and only enabling the hydration-controller container if it does.

This is accomplished by setting an annotation with the reconciler, and having the reconciler-manager read that annotation to determine whether to include hydration-controller in the Deployment.

### DIFF
--- a/e2e/testcases/hydration_test.go
+++ b/e2e/testcases/hydration_test.go
@@ -29,11 +29,13 @@ import (
 	nomosstatus "kpt.dev/configsync/cmd/nomos/status"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/declared"
 	"kpt.dev/configsync/pkg/importer/analyzer/validation/nonhierarchical"
 	"kpt.dev/configsync/pkg/metadata"
@@ -52,6 +54,33 @@ func TestHydrateKustomizeComponents(t *testing.T) {
 		ntopts.Unstructured,
 	)
 
+	syncDirMap := map[types.NamespacedName]string{
+		nomostest.DefaultRootRepoNamespacedName: "kustomize-components",
+	}
+
+	// Dry configs not yet added to repo, assert that hydration is disabled
+	tg := taskgroup.New()
+	tg.Go(func() error {
+		return nt.Validate(
+			configsync.RootSyncName,
+			configsync.ControllerNamespace,
+			&v1beta1.RootSync{},
+			testpredicates.MissingAnnotation(metadata.RequiresRenderingAnnotationKey),
+		)
+	})
+	tg.Go(func() error {
+		return nt.Validate(
+			core.RootReconcilerName(configsync.RootSyncName),
+			configsync.ControllerNamespace,
+			&appsv1.Deployment{},
+			testpredicates.DeploymentMissingContainer(reconcilermanager.HydrationController),
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.Reconciler, reconcilermanager.RenderingEnabled, "false"),
+		)
+	})
+	if err := tg.Wait(); err != nil {
+		nt.T.Fatal(err)
+	}
+
 	nt.T.Log("Add the kustomize components root directory")
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Copy("../testdata/hydration/kustomize-components", "."))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add DRY configs to the repository"))
@@ -59,8 +88,30 @@ func TestHydrateKustomizeComponents(t *testing.T) {
 	nt.T.Log("Update RootSync to sync from the kustomize-components directory")
 	rs := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
 	nt.MustMergePatch(rs, `{"spec": {"git": {"dir": "kustomize-components"}}}`)
-	err := nt.WatchForAllSyncs(nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: "kustomize-components"}))
-	if err != nil {
+	if err := nt.WatchForAllSyncs(nomostest.WithSyncDirectoryMap(syncDirMap)); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// Dry configs added to repo, assert that hydration is enabled
+	tg = taskgroup.New()
+	tg.Go(func() error {
+		return nt.Validate(
+			configsync.RootSyncName,
+			configsync.ControllerNamespace,
+			&v1beta1.RootSync{},
+			testpredicates.HasAnnotation(metadata.RequiresRenderingAnnotationKey, "true"),
+		)
+	})
+	tg.Go(func() error {
+		return nt.Validate(
+			core.RootReconcilerName(configsync.RootSyncName),
+			configsync.ControllerNamespace,
+			&appsv1.Deployment{},
+			testpredicates.DeploymentHasContainer(reconcilermanager.HydrationController),
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.Reconciler, reconcilermanager.RenderingEnabled, "true"),
+		)
+	})
+	if err := tg.Wait(); err != nil {
 		nt.T.Fatal(err)
 	}
 
@@ -90,8 +141,8 @@ func TestHydrateKustomizeComponents(t *testing.T) {
 	nt.T.Log("Add kustomization.yaml back")
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Copy("../testdata/hydration/kustomize-components/kustomization.yml", "./kustomize-components/kustomization.yml"))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add kustomization.yml back"))
-	err = nt.WatchForAllSyncs(nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: "kustomize-components"}))
-	if err != nil {
+
+	if err := nt.WatchForAllSyncs(nomostest.WithSyncDirectoryMap(syncDirMap)); err != nil {
 		nt.T.Fatal(err)
 	}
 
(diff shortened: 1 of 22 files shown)


sdowell:
Pulled out these changes:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/710
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/709

sdowell:
all pulled out changes have been submitted, removing hold
/unhold

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -910,13 +929,19 @@ func (r *RootSyncReconciler) mutationsFor(ctx context.Context, rs *v1beta1.RootS
...
+					// this minimizes the resource footprint of the reconciler
+					addContainer = false
 				} else {
-					container.Image = strings.ReplaceAll(container.Image, reconcilermanager.HydrationController+":", reconcilermanager.HydrationControllerWithShell+":")
+					container.Env = append(container.Env, containerEnvs[container.Name]...)
+					if rs.Spec.SafeOverride().EnableShellInRendering == nil || !*rs.Spec.SafeOverride().EnableShellInRendering {
+						container.Image = strings.ReplaceAll(container.Image, reconcilermanager.HydrationControllerWithShell, reconcilermanager.HydrationController)
+					} else {
+						container.Image = strings.ReplaceAll(container.Image, reconcilermanager.HydrationController+":", reconcilermanager.HydrationControllerWithShell+":")
+					}
```
janetkuo, line 942: nit: This adds another layer to the already deeply nested conditionals. Can we improve it to make it more readable? For example, maybe the shell part can be wrapped as a function.

sdowell: fair nit, updated

pkg/parse/run.go
```diff
@@ -280,63 +301,98 @@ func read(ctx context.Context, p Parser, trigger string, state *reconcilerState,
...
 	} else if !os.IsNotExist(err) {
 		hydrationStatus.message = RenderingFailed
 		hydrationStatus.errs = status.InternalHydrationError(err, "unable to evaluate the hydrated path %s", absHydratedRoot.OSPath())
-		return hydrationStatus, sourceStatus
+		return srcState, hydrationStatus
+	} else {
+		// Source of truth does not require hydration, but hydration-controller is running
+		hydrationStatus.message = RenderingNotRequired
+		hydrationStatus.requiresRendering = false
+		err := hydrate.NewTransientError(fmt.Errorf("rendering enabled but not required"))
```
tiffanny29631, line 335: nit: Use RenderingNotRequired?

e2e/testcases/hydration_test.go
```diff
@@ -52,6 +54,29 @@ func TestHydrateKustomizeComponents(t *testing.T) {
 		ntopts.Unstructured,
 	)
 
+	// Dry configs not yet added to repo, assert that hydration is disabled
```
tiffanny29631, line 57: Is it necessary to assert hydration controller disabled  later in the same test, when DRY configs are removed and hydration controller was known enabled?

pkg/parse/run.go
```diff
@@ -280,63 +301,98 @@ func read(ctx context.Context, p Parser, trigger string, state *reconcilerState,
...
-		return hydrationStatus, sourceStatus
+	if srcState.syncDir == recState.cache.source.syncDir {
+		return hydrationStatus, srcStatus
 	}
 
-	klog.Infof("New source changes (%s) detected, reset the cache", sourceState.syncDir.OSPath())
+	// Read all the files under state.syncDir
+	srcStatus.errs = options.readConfigFiles(&srcState)
+
+	if !options.renderingEnabled {
```
janetkuo, line 373: Can this be combined with the `if options.renderingEnabled ... else` above?

sdowell: No it cannot, because this relies on `srcState.files` being populated by `readConfigFiles`. In a previous iteration I called `readConfigFiles` twice (once before `readHydrationStatus` and once after) so that I could consolidate this logic into a single if/else, but it was a bit less efficient because the filesystem tree was being walked multiple times

e2e/testcases/hydration_test.go
sdowell: Just to be clear, the hydration controller is:
- enabled when dry configs are present
- disabled when dry configs are absent (i.e. only wet configs)

That being said, these before/after assertions are not strictly necessary but accomplish a few things:
- Verify the assumptions/correctness of the reconciler/reconciler-manager logic that is dynamically toggling the hydration-controller
- Provides a clear signal in the case of failure that the reason the test failed is because of a bug in the aforementioned logic

pkg/parse/run.go
sdowell: Yeah I'm not really sure what is best to populate here, as normally we insert an actual `error` that was returned from a function into the status. In this case, it's a contrived error because the error condition is defined by us. It seems a bit redundant to use the same string in `message` and `errors` but I don't currently have a better alternative

sdowell: I updated the error to provide a bit more technical detail about the error, while the message is intended to remain a bit more user-friendly. PTAL

pkg/parse/run.go
janetkuo: Side note -`readConfigFiles()` comments might be outdated. I didn't see `rendered` in the function. 
```go
// readConfigFiles reads all the files under state.syncDir and sets state.files.
// - if rendered is true, state.syncDir contains the hydrated files.
// - if rendered is false, state.syncDir contains the source files.
```

pkg/parse/run.go
```diff
@@ -280,63 +301,98 @@ func read(ctx context.Context, p Parser, trigger string, state *reconcilerState,
...
-		commit: sourceState.commit,
-	}
-	sourceStatus := sourceStatus{
-		commit: sourceState.commit,
-	}
-
+// readHydrationStatus reads from the file path which the hydration-controller
+// container writes to. It checks if the hydrated files are ready and returns
+// a renderingStatus.
+func readHydrationStatus(p Parser, srcState sourceState, hydrationStatus renderingStatus) (sourceState, renderingStatus) {
```
janetkuo, line 307: This function assumes the caller only calls it when hydration controller is running. We should make it clear or move the options.renderingEnabled check logic inside (because someone else might want to use this function later and they should be aware). Also, this function doesn't just read hydrationStatus (it reads files and updates hydrationStatus, so suggest renaming the function.

pkg/parse/run.go
sdowell: The description is still somewhat accurate - there isn't a `rendered` boolean but the files that it ingests are either hydrated or source files depending on the list of file paths that are passed in (I'm not aware of this changing recently). I can update the description to make this more clear

pkg/parse/run.go
```diff
@@ -238,6 +251,14 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
 // It also updates the .status.rendering and .status.source fields.
 func read(ctx context.Context, p Parser, trigger string, state *reconcilerState, sourceState sourceState) status.MultiError {
 	hydrationStatus, sourceStatus := readFromSource(ctx, p, trigger, state, sourceState)
+	if p.options().renderingEnabled != hydrationStatus.requiresRendering {
+		// the reconciler is misconfigured. set the annotation so that the reconciler-manager
+		// will recreate this reconciler with the correct configuration.
```
janetkuo, line 256: Why do we need both `status.requiresRendering` and the requiresRendering annotation? Can we just keep one?

pkg/parse/run.go
```diff
@@ -238,6 +251,14 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
...
+		if err := p.setRenderingRequired(ctx, hydrationStatus.requiresRendering); err != nil {
```
janetkuo, line 257: Can we unify and call this setRequiresRendering instead of setRenderingRequired, given the name of the annotation and the status field? Especially when there's `RenderingRequired` const that has slightly different meaning.

pkg/parse/run_test.go
```diff
@@ -158,6 +159,8 @@ func TestRun(t *testing.T) {
 		name                       string
 		commit                     string
 		sourceRootExist            bool
+		hydrationEnabled           bool
```
janetkuo, line 162: nit: Can we unify and call it `renderingEnabled`

janetkuo (commented):
Have some comments and look good in general. Nits are optional.

pkg/parse/run.go
sdowell: There is no `status.requiresRendering`, at least on the `RSync`. This status is just used internally to compare the rendering status in the SoT to the current configuration/annotation

pkg/parse/run.go
sdowell: Yeah seems reasonable

pkg/parse/run_test.go
sdowell: done

pkg/parse/run.go
sdowell: done

pkg/parse/run.go
sdowell: Fair point, refactored this to push the options check into the function and renamed to `parseHydrationState`. I think `parseHydrationState` captures the spirit of the function more precisely than `readHydrationStatus` - note the actual hydrated files are still read by `readConfigFiles`.

I'm a bit concerned that moving the conditional a level down might impact readability of `readFromSource`, but I think it's a fair middle ground.

pkg/parse/state.go
```diff
@@ -39,10 +39,11 @@ func (gs sourceStatus) equal(other sourceStatus) bool {
...
 type renderingStatus struct {
-	commit     string
-	message    string
-	errs       status.MultiError
-	lastUpdate metav1.Time
+	commit            string
+	message           string
+	errs              status.MultiError
+	lastUpdate        metav1.Time
+	requiresRendering bool
```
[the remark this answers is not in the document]
sdowell: added a comment

pkg/parse/run.go
```diff
@@ -280,63 +301,98 @@ func read(ctx context.Context, p Parser, trigger string, state *reconcilerState,
...
 	}
 
-	if sourceState.syncDir == state.cache.source.syncDir {
-		return hydrationStatus, sourceStatus
+	if srcState.syncDir == recState.cache.source.syncDir {
+		return hydrationStatus, srcStatus
 	}
 
-	klog.Infof("New source changes (%s) detected, reset the cache", sourceState.syncDir.OSPath())
+	// Read all the files under state.syncDir
```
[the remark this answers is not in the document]
sdowell: updated

pkg/parse/run.go
janetkuo: The refactor looks good

janetkuo (approved):
/approve

Approving the change. @tiffanny29631 do you want to take another look?

sdowell:
> Approving the change. @tiffanny29631 do you want to take another look?

jfyi @janetkuo using the approve button in the github UI will also add the lgtm label (since we've enabled the [lgtm](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/oss/plugins.yaml#L57) plugin - see [catalog](https://oss.gprow.dev/plugins))

You can still separately apply lgtm/approve labels by just using the lgtm and approve commands

sdowell:
@tiffanny29631 feel free to leave any comments and I'll be happy to address in a followup

PR #709, opened 2023-06-24
fix: check that symlink is unchanged for hydration

sdowell:
We make the same check when reading the source files, this adds the check when reading the hydration dir for consistency.

Also fixes some test assertions which were passing erroneously.

### DIFF
--- a/pkg/parse/source_test.go
+++ b/pkg/parse/source_test.go
@@ -23,8 +23,10 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/declared"
+	"kpt.dev/configsync/pkg/hydrate"
 	"kpt.dev/configsync/pkg/importer/filesystem"
 	"kpt.dev/configsync/pkg/importer/filesystem/cmpath"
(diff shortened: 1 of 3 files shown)


natasha41575:
I'm not an approver so you'll need someone else to take a look, but I figured I could do a quick peer review.

/lgtm

pkg/parse/source_test.go
```diff
@@ -123,8 +124,89 @@ func TestReadConfigFiles(t *testing.T) {
...
+			wantState := sourceState{
+				commit:  tc.commit,
+				syncDir: cmpath.Absolute(parserCommitDir),
+			}
+
+			hydrationState, hydrationErr := parser.readHydratedDir(
+				cmpath.Absolute(hydratedRoot), parser.reconcilerName, *srcState)
+
+			assert.Equal(t, tc.wantedErr, hydrationErr)
+			if hydrationErr == nil {
```
tiffanny29631, line 207: LGTM, nit: assert hydrationError equals nil?

sdowell: nil assertion for `hydrationError` would be captured on the above line when `wantedErr = nil`

tiffanny29631: Didn't know why I missed that :)

/lgtm
/approve

PR #710, opened 2023-06-24
refactor: consolidate rsync helper functions

sdowell:
Declare a common rootSync/repoSync helper that is reused by the various source types, so that it is easier to define common logic.


pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -222,43 +222,49 @@ func reposyncCACert(caCertSecretRef string) func(sync *v1beta1.RepoSync) {
 
 func repoSync(ns, name string, opts ...func(*v1beta1.RepoSync)) *v1beta1.RepoSync {
```
tiffanny29631, line 223: Is this function still in use after consolidation?

PR #712, opened 2023-06-27
test: fix poll override logic in `nomostest`

natasha41575:
Pulled out from https://github.com/GoogleContainerTools/kpt-config-sync/pull/707, see discussion in https://github.com/GoogleContainerTools/kpt-config-sync/pull/707#discussion_r1240911054

I separated the rename and logic fix into separate commits to ease review.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -90,9 +90,9 @@ var (
 	hydrationPollingPeriod time.Duration
 )
 
-// IsReconcilerManagerCmConfigMap returns true if passed obj is the
+// IsReconcilerTemplateConfigMap returns true if passed obj is the
 // reconciler-manager ConfigMap reconciler-manager-cm in config-management namespace.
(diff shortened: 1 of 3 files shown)


e2e/nomostest/config_sync.go
```diff
@@ -90,14 +90,22 @@ var (
 	hydrationPollingPeriod time.Duration
 )
 
-// IsReconcilerManagerConfigMap returns true if passed obj is the
+// IsReconcilerManagerCmConfigMap returns true if passed obj is the
 // reconciler-manager ConfigMap reconciler-manager-cm in config-management namespace.
-func IsReconcilerManagerConfigMap(obj client.Object) bool {
+func IsReconcilerManagerCmConfigMap(obj client.Object) bool {
```
sdowell, line 95: perhaps let's call this something like `IsReconcilerTemplateConfigMap` to reduce future confusion. WDYT?

natasha41575: SGTM

PR #713, opened 2023-06-27
test: add cluster-names e2e arg

sdowell:
This argument accepts a list of cluster names. This can either be used
with the create-clusters arg to create clusters with the specific name,
or without create-clusters to assume pre-provisioned clusters. If
create-clusters is provided for a cluster that already exists, it will
reuse the existing cluster.

This is written in a way that works for both KinD and GKE clusters, which
reconciles one of the final differences between the interfaces for
interacting with the two cluster types.

Also adds a delete-clusters option, which can be used to indicate
whether to delete clusters after the tests. This can be used in
conjunction with create-clusters to create clusters in the test
execution, and leave them for a subsequent execution.


e2e/flags.go
```diff
@@ -141,9 +169,13 @@ var GKENumNodes = flag.Int("gke-num-nodes", util.EnvInt("GKE_NUM_NODES", Default
 var GKEAutopilot = flag.Bool("gke-autopilot", util.EnvBool("GKE_AUTOPILOT", false),
 	"Whether to create GKE clusters with autopilot enabled.")
 
-// CreateClusters indicates the test framework should create GKE clusters.
+// CreateClusters indicates the test framework should create clusters.
 var CreateClusters = flag.Bool("create-clusters", util.EnvBool("CREATE_CLUSTERS", false),
-	"Whether to create GKE clusters, otherwise will assume pre-provisioned GKE clusters.")
+	"Whether to create clusters, otherwise will assume pre-provisioned clusters.")
```
karlkfi, line 174: Add a "lazy" value option (non-bool) to check if the cluster exists and adopt it if it does. Then "true" should continue to error if  cluster already exists.

sdowell: done

docs/development.md
```diff
@@ -122,7 +122,7 @@ deleted after the test execution.
 ```shell
 # GKE cluster options can be provided as command line flags
 go test ./e2e/... --e2e --test.v --share-test-env --gcp-project=<PROJECT_ID> --test-cluster=gke \
-  --create-clusters --num-clusters=5 --gcp-zone=us-central1-a \
```
karlkfi, line 125: is true not the default?

e2e/flags.go
```diff
@@ -19,13 +19,36 @@ package e2e
...
+func (i *listFlag) String() string {
+	return strings.Join(i.arr, ",")
+}
+
+func (i *listFlag) Set(value string) error {
+	i.arr = strings.Split(value, ",")
+	return nil
+}
+
+func newListFlag(name string, def []string, usage string) *[]string {
```
karlkfi, line 44: suggest: newStringListFlag

e2e/flags.go
```diff
@@ -141,9 +169,13 @@ var GKENumNodes = flag.Int("gke-num-nodes", util.EnvInt("GKE_NUM_NODES", Default
 var GKEAutopilot = flag.Bool("gke-autopilot", util.EnvBool("GKE_AUTOPILOT", false),
 	"Whether to create GKE clusters with autopilot enabled.")
 
-// CreateClusters indicates the test framework should create GKE clusters.
-var CreateClusters = flag.Bool("create-clusters", util.EnvBool("CREATE_CLUSTERS", false),
-	"Whether to create GKE clusters, otherwise will assume pre-provisioned GKE clusters.")
+// CreateClusters indicates the test framework should create clusters.
+var CreateClusters = flag.String("create-clusters", util.EnvString("E2E_CREATE_CLUSTERS", CreateClustersDisabled),
+	"Whether to create clusters, otherwise will assume pre-provisioned clusters.")
```
karlkfi, line 174: Suggest: adding the allowed enum values to the description

docs/development.md
sdowell: `false` is the default for `create-clusters`

e2e/nomostest/client.go
```diff
@@ -110,16 +112,60 @@ type Cluster interface {
...
 
+func upsertCluster(t testing.NTB, cluster Cluster, opts *ntopts.New) {
+	var adoptCluster bool
+	switch *e2e.CreateClusters {
+	case e2e.CreateClustersEnabled:
+		adoptCluster = false
+	case e2e.CreateClustersLazy:
+		adoptCluster = true
+	default:
+		t.Fatalf("unrecognized value for create-cluster: %s", *e2e.CreateClusters)
```
karlkfi, line 123: can we do flag validation earlier?

sdowell: Yeah, should be doable. I'll move this earlier

e2e/nomostest/client.go
```diff
@@ -110,16 +112,60 @@ type Cluster interface {
...
+	exists, err := cluster.Exists()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if exists && !adoptCluster {
+		t.Fatalf("cluster %s already exists and create-clusters=%s", opts.ClusterName, *e2e.CreateClusters)
+	}
+	t.Cleanup(func() {
+		if !*e2e.DestroyClusters || (t.Failed() && *e2e.Debug) {
+			t.Logf("[WARNING] Skipping deletion of %s cluster %s", *e2e.TestCluster, opts.ClusterName)
```
karlkfi, line 134: Maybe add (--delete-clusters=false) here to indicate that this was asked for behavior and not actually a problem/error.

e2e/flags.go
sdowell: will do

e2e/flags.go
sdowell: sgtm

e2e/nomostest/new.go
```diff
@@ -286,20 +286,28 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		// a single test.
 		connectToLocalRegistry(nt)
 	}
-	if !opts.IsEphemeralCluster {
-		// We aren't using an ephemeral cluster, so make sure the cluster is
-		// cleaned before and after running the test.
+	t.Cleanup(func() {
+		if *e2e.DestroyClusters {
+			nt.T.Logf("Skipping cleanup because cluster will be destroyed")
```
karlkfi, line 291: Maybe add (--destroy-clusters=true) here as well by way of explaination

sdowell: agreed, done

e2e/nomostest/client.go
sdowell: agreed, done

e2e/nomostest/client.go
sdowell: done

e2e/flags.go
sdowell: done

e2e/flags.go
sdowell: done

karlkfi (approved):
/lgtm

PR #714, opened 2023-06-28
test: add cluster-prefix parameter

sdowell:
This parameter can be used to specify a prefix to use when generating cluster names. The default behavior remains to generate a prefix using cs-e2e-<timestamp>.

### DIFF
--- a/e2e/flags.go
+++ b/e2e/flags.go
@@ -116,6 +116,12 @@ var TestFeatures = flag.String("test-features", "",
 var NumClusters = flag.Int("num-clusters", util.EnvInt("E2E_NUM_CLUSTERS", 1),
 	"Number of parallel test threads to run. Also dictates the number of clusters which will be created in parallel. Overrides the -test.parallel flag.")
 
+// ClusterPrefix is the prefix to use when naming clusters. An index is appended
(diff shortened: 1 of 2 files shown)


e2e/flags.go
```diff
@@ -116,6 +116,11 @@ var TestFeatures = flag.String("test-features", "",
 var NumClusters = flag.Int("num-clusters", util.EnvInt("E2E_NUM_CLUSTERS", 1),
 	"Number of parallel test threads to run. Also dictates the number of clusters which will be created in parallel. Overrides the -test.parallel flag.")
 
+// ClusterPrefix is the prefix to use when creating clusters. If unset, defaults
+// to cs-e2e-<UNIX_TIME>. cluster-names takes precedence, if provided.
+var ClusterPrefix = flag.String("cluster-prefix", util.EnvString("E2E_CLUSTER_PREFIX", ""),
+	"Prefix to use when creating clusters. If unset, defaults to cs-e2e-<UNIX_TIME>. cluster-names takes precedence, if provided.")
```
karlkfi, line 122: "Prefix to use when naming clusters. An index is appended to the prefix with the pattern '<PREFIX>-<INDEX>', starting at 0."

sdowell: updated

karlkfi (approved):
/lgtm

PR #715, opened 2023-06-29
test: enhance cleanup logic for destroy-clusters

sdowell:
This adds several options for destroying clusters:
- true always destroys clusters
- auto destroys clusters if created (default)
- false never destroys clusters

This streamlines the logic for destroying clusters and decouples it from
the cluster creation logic, while maintaining backwards compatible
defaults.


karlkfi (approved):
/lgtm

PR #717, opened 2023-06-29
helm fix: volume names will use the valuesFileRef index rather than ConfigMap name

natasha41575:
Volume names have a length limit of 63 or we get an error, this fixes it so that we just use the valuesFileRef index rather than the ConifgMap name

b/289378438

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_base.go
+++ b/pkg/reconcilermanager/controllers/reconciler_base.go
@@ -433,7 +433,7 @@ func mountConfigMapValuesFiles(templateSpec *corev1.PodSpec, c *corev1.Container
 		fileName := reconcilermanager.HelmConfigMapRef
 		mountPath := filepath.Join("/etc/config", vf.Name, vf.ValuesFile)
 		valuesFiles = append(valuesFiles, filepath.Join(mountPath, fileName))
-		volumeName := fmt.Sprintf("vol-%d-", i) + strings.ToLower(vf.Name)
+		volumeName := fmt.Sprintf("valuesfile-vol-%d", i)
 
 		if len(volumeName) > volumeNameLengthLimit {
 			// trim the volume name as we only have a length limit of 63, even though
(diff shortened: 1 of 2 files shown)


pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -431,7 +433,14 @@ func mountConfigMapValuesFiles(templateSpec *corev1.PodSpec, c *corev1.Container
...
 		mountPath := filepath.Join("/etc/config", vf.Name, vf.ValuesFile)
 		valuesFiles = append(valuesFiles, filepath.Join(mountPath, fileName))
-		volumeName := "configmap-vol-" + strings.ToLower(vf.Name) + fmt.Sprintf("-%d", i)
+		volumeName := fmt.Sprintf("vol-%d-", i) + strings.ToLower(vf.Name)
+
+		if len(volumeName) > volumeNameLengthLimit {
+			// trim the volume name as we only have a length limit of 63, even though
+			// the ConfigMap name can be much longer.  The prefix
+			// containing the index i should be sufficient to uniquify the value.
+			volumeName = volumeName[:volumeNameLengthLimit]
```
janetkuo, line 442: What happens if the name collide after trimming?

natasha41575: The volume name is prefixed by fmt.Sprintf(“vol-%d-“, i), where i is the index of the volumeFileSource that we are creating this volume from, so there shouldn’t be any name collisions.

natasha41575: I added a test as well so that we can be confident

janetkuo: Do users care about volume names? Should we just name it `vol-<INDEX>`?

natasha41575: SGTM. I updated it to `valuesFile-vol-<INDEX>` to be slightly more descriptive

janetkuo:
/lgtm

Would you update the PR title?

PR #718, opened 2023-06-30
test: enhancements to reconciler finalizer tests

sdowell:
This change makes a few enhancements to the reconciler-finalizer tests to help with debugging, namely:
- Wrap the logging of reconciler logs in a cleanup. This ensures that the logs get printed in the case of a failure, even if kubectl returns before the test fails (e.g. Pod eviction)
- Replace usages of Wait with taskgroup/watch. This provides more useful logging as each watch event gets logged.

### DIFF
--- a/e2e/nomostest/reset.go
+++ b/e2e/nomostest/reset.go
@@ -490,16 +490,23 @@ func TailReconcilerLogs(ctx context.Context, nt *NT, reconcilerNN types.Namespac
 		"-n", reconcilerNN.Namespace,
 		"-c", reconcilermanager.Reconciler,
 		"-f")
+	nt.T.Logf("finished tailing logs for %s", reconcilerNN.String())
 	// Expect the logs to tail until the context is cancelled, or exit early if
 	// the reconciler container exited.
 	if err != nil && err.Error() != "signal: killed" {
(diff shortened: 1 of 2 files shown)


e2e/nomostest/reset.go
```diff
@@ -490,16 +490,23 @@ func TailReconcilerLogs(ctx context.Context, nt *NT, reconcilerNN types.Namespac
...
 	}
-	// Only print the logs if the test has failed
-	if nt.T.Failed() {
-		nt.T.Logf("Reconciler deployment logs (%s):\n%s", reconcilerNN, string(out))
-	}
+	// When TailReconcilerLogs is called async, it's possible that the test fails
+	// after the kubectl calls returns. For example, when the Pod is evicted the
+	// kubectl call will return early. Register a cleanup so that we always log in
+	// the case of a failed test.
+	nt.T.Cleanup(func() {
```
sdowell, line 504: I think we might be able to do more here. For example if the Pod restarted, we can restart the log-tailing and concatenate the logs for each subsequent Pod (still terminated by the same context). Might consider this as a followup

tiffanny29631 (approved):
/lgtm

PR #719, opened 2023-07-05
test: add e2e options for disk size/type

sdowell:
This adds options to configure the disk size/type for GKE standard clusters. This enables customizing the shape of persistent disks when creating clusters with the test framework.

### DIFF
--- a/e2e/flags.go
+++ b/e2e/flags.go
@@ -167,6 +167,14 @@ var GKEClusterVersion = flag.String("gke-cluster-version", util.EnvString("GKE_C
 var GKEMachineType = flag.String("gke-machine-type", util.EnvString("GKE_MACHINE_TYPE", DefaultGKEMachineType),
 	"GKE machine type to use when creating GKE clusters. Defaults to GKE_MACHINE_TYPE env var.")
 
+// GKEDiskSize is the GKE disk size to use when creating GKE clusters.
+var GKEDiskSize = flag.String("gke-disk-size", util.EnvString("GKE_DISK_SIZE", ""),
+	"GKE disk size to use when creating GKE clusters. Defaults to GKE_DISK_SIZE env var.")
+
+// GKEDiskType is the GKE disk type to use when creating GKE clusters.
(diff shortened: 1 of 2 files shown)


e2e/flags.go
```diff
@@ -167,6 +167,14 @@ var GKEClusterVersion = flag.String("gke-cluster-version", util.EnvString("GKE_C
 var GKEMachineType = flag.String("gke-machine-type", util.EnvString("GKE_MACHINE_TYPE", DefaultGKEMachineType),
 	"GKE machine type to use when creating GKE clusters. Defaults to GKE_MACHINE_TYPE env var.")
 
+// GKEDiskSize is the GKE disk size to use when creating GKE clusters.
+var GKEDiskSize = flag.String("gke-disk-size", util.EnvString("GKE_DISK_SIZE", ""),
```
karlkfi, line 171: feels weird to not have defaults specified in the code.  I think we probably want SSDs (so much faster) and some default disk size that's big enough for all the tests...

sdowell: My intent is to experiment with SSDs in a new prowjob, but without changing the defaults for now. We could potentially change the defaults as a subsequent change, but my intent was just to provide the option with this change

karlkfi: I can say categorically that we should always use SSDs with K8s. The only reason ever not to is to minimize cost at the expense of performance, which is almost never a good trade-off. There are no workloads that don't need faster disk, and it greatly speeds up normal lifecycle operations.

karlkfi: Making size configurable could make sense for specific test types, but for that I'm ok leaving the default up to GKE.

sdowell: I don't disagree, but we need to be careful to not exceed GCP quota limits (which is pretty easy with the large number of clusters we use in CI). My intent with this change is to experiment with a new cluster shape such that we can fit SSDs into our quota limits

karlkfi (approved):
/lgtm

PR #722, opened 2023-07-06
test: add namespace yaml to reset error logging

sdowell:
This prints the full namespace yaml for namespaces which fail to be cleaned up in Reset. This is intended to help with diagnosing errors related to Reset, e.g. if the Namespace has a finalizer or deletion timestamp.


karlkfi (approved):
/lgtm

PR #723, opened 2023-07-06
chore: bump k8s patch version

sdowell:
This incorporates various security fixes from upstream k8s


go.mod
```diff
@@ -45,7 +45,7 @@ require (
 	k8s.io/kube-aggregator v0.25.9
 	k8s.io/kube-openapi v0.0.0-20230109183929-3758b55a6596
 	k8s.io/kubectl v0.25.9
-	k8s.io/kubernetes v1.25.9
+	k8s.io/kubernetes v1.25.11
```
karlkfi, line 48: you'll want to update all the other k8s libs to the same patch. This one is ONLY used to have access to the generated conversion and deepcopy code that isn't in the exported api package.

sdowell: done

karlkfi (approved):
/lgtm

PR #726, opened 2023-07-10
remove spec.helm.valuesFileApplyStrategy

natasha41575:
b/289364115


manifests/reposync-crd.yaml
```diff
@@ -1188,29 +1188,12 @@ spec:
...
-                    - override
-                    - listConcatenate
-                    type: string
                   valuesFileSources:
                     description: valuesFileSources holds references to objects in
                       the cluster that repremakesent values to use instead of default
                       values that accompany the chart. Currently, only ConfigMaps
-                      are supported. Objects listed later will take precedence. How
-                      to handle multiple valuesFiles is determined by `valuesFileApplyStrategy`.
+                      are supported. Objects listed later will take precedence.
```
karlkfi, line 1196: maybe describe how override works with inline values as well

pkg/api/configsync/v1beta1/helmconfig.go
```diff
@@ -59,31 +59,16 @@ type HelmBase struct {
...
-	// How to handle multiple valuesFiles is determined by `valuesFileApplyStrategy`.
 	// +optional
 	ValuesFileSources []ValuesFileSources `json:"valuesFileSources,omitempty"`
 
-	// valuesFileApplyStrategy specifies the strategy for handling multiple valueFiles. It
-	// refers to how valuesFiles are applied if multiple valuesFile define the
-	// same key. Can be 'override' or 'listConcatenate'.
-	// 'override' (default) results in the duplicated keys in later files to
-	// override the value from earlier files. This is equivalent to passing in
-	// multiple valuesFiles to Helm CLI.
```
janetkuo, line 79: Move the part about the override/default behavior to `valuesFileSources` (but don't mention override).

karlkfi (changes requested):
see inline comments

pkg/helm/helm.go
```diff
@@ -60,21 +49,20 @@ var (
...
-	UserName                string
-	Password                string
-	ValuesFileApplyStrategy string
+	Chart             string
+	Repo              string
+	Version           string
+	ReleaseName       string
+	Namespace         string
+	DeployNamespace   string
+	Values            string
```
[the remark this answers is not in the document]
natasha41575: Values is not a path, it is the literal yaml string with the inline values so it can't be appended or treated the same as ValuesFileSources without some larger implementation changes.

pkg/api/configsync/v1beta1/helmconfig.go
natasha41575: done

manifests/reposync-crd.yaml
natasha41575: done

e2e/testcases/helm_sync_test.go
```diff
@@ -405,11 +404,11 @@ wordpressEmail: override-this@example.com`,
 			},
 			"valuesFileSources": [
 			  {
-				"name": "helm-config-map-merge",
+				"name": "helm-config-map-override",
```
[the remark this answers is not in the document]
natasha41575: I added an additional map (the image pull policy) so we now are testing list, scalar, and map.

In terms if more complex values, this is a public chart that we are using, so we are limited for now to the values that the chart supports (https://github.com/bitnami/charts/blob/main/bitnami/wordpress/values.yaml) and that our test harness can validate.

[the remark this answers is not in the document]
karlkfi: This is probably fine now. It shows that fields are fully replaced, not merged.

manifests/reposync-crd.yaml
```diff
@@ -1187,30 +1187,17 @@ spec:
...
                       cluster that represent values to use instead of default values
-                      that accompany the chart. Currently, only ConfigMaps are supported.
-                      Objects listed later will take precedence. How to handle multiple
-                      valuesFiles is determined by `valuesFileApplyStrategy`.
+                      that accompany the chart. Currently, only ConfigMaps are supported,
+                      and the ConfigMap must be in the same namespace as the RootSync
+                      or RepoSync. When multiple valuesFiles are used, duplicated
+                      keys in later files will override the value from earlier files.
+                      This is equivalent to passing in multiple valuesFiles to Helm
+                      CLI. If `values` is also specified, `values` will take precedence.
```
karlkfi, line 1200: Suggest:
If `values` is also specified, fields from `values` will override fields from `valuesFileRefs`.

"precedence" is mildly confusing. This should make it clear that it uses the same override mechanism, not a full replacement.

manifests/rootsync-crd.yaml
```diff
@@ -1202,30 +1202,17 @@ spec:
                   values:
                     description: values to use instead of default values that accompany
                       the chart. Format values the same as default values.yaml. These
-                      values will take precedence if valuesFileRefs is also specified.
-                      How to handle multiple valuesFiles is determined by `valuesFileApplyStrategy`.
+                      values will take precedence if `valuesFileRefs` is also specified.
```
karlkfi, line 1205: Suggest:
If `valuesFileRefs` is also specified, fields from `values` will override fields from `valuesFileRefs`.

manifests/reposync-crd.yaml
```diff
@@ -1187,30 +1187,17 @@ spec:
                   values:
                     description: values to use instead of default values that accompany
                       the chart. Format values the same as default values.yaml. These
-                      values will take precedence if valuesFileRefs is also specified.
-                      How to handle multiple valuesFiles is determined by `valuesFileApplyStrategy`.
+                      values will take precedence if `valuesFileRefs` is also specified.
```
karlkfi, line 1190: Suggest:
If valuesFileRefs is also specified, fields from values will override fields from valuesFileRefs.

manifests/rootsync-crd.yaml
```diff
@@ -1202,30 +1202,17 @@ spec:
...
                       cluster that represent values to use instead of default values
-                      that accompany the chart. Currently, only ConfigMaps are supported.
-                      Objects listed later will take precedence. How to handle multiple
-                      valuesFiles is determined by `valuesFileApplyStrategy`.
+                      that accompany the chart. Currently, only ConfigMaps are supported,
+                      and the ConfigMap must be in the same namespace as the RootSync
+                      or RepoSync. When multiple valuesFiles are used, duplicated
+                      keys in later files will override the value from earlier files.
+                      This is equivalent to passing in multiple valuesFiles to Helm
+                      CLI. If `values` is also specified, `values` will take precedence.
```
karlkfi, line 1215: Suggest:
If values is also specified, fields from values will override fields from valuesFileRefs.

pkg/helm/helm.go
```diff
@@ -115,61 +104,23 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
...
-
-		valuesPath, err := writeValuesPath(merged)
+	if len(h.ValuesYAML) != 0 {
+		// inline yaml values are passed in as a literal string, so we
+		// must write them out to a file
+		valuesPath, err := writeValuesPath([]byte(h.ValuesYAML))
 		if err != nil {
 			return nil, err
 		}
 		if valuesPath != "" {
```
karlkfi, line 121: will valuesPath ever be empty if err was nil?

pkg/helm/helm.go
```diff
@@ -115,61 +104,23 @@ func (h *Hydrator) templateArgs(ctx context.Context, destDir string) ([]string,
...
-	switch h.ValuesFileApplyStrategy {
-
-	case "", ValuesFileApplyStrategyOverride:
-		for _, vs := range h.ValuesFilePaths {
-			if vs == "" {
-				continue
-			}
-			args = append(args, "--values", vs)
+	for _, vs := range h.ValuesFilePaths {
+		if vs == "" {
```
karlkfi, line 108: seems like this should be validated in to disallow empty paths. silently skipping them probably indicates a problem

pkg/helm/helm.go
natasha41575: In the current implementation, if the inline values on the RSync are empty, then h.ValuesYAML will end up as an empty string, and `writeValuesPath` will return `"", nil`.

karlkfi: aren't you protected from that with `if len(h.ValuesYAML) != 0 {` already?

natasha41575: oh you're right, I can take out the duplicated check

pkg/helm/helm.go
natasha41575: sg, returning an error now

pkg/helm/helm.go
natasha41575: done

manifests/rootsync-crd.yaml
natasha41575: done

manifests/reposync-crd.yaml
natasha41575: done

manifests/rootsync-crd.yaml
natasha41575: done

manifests/reposync-crd.yaml
natasha41575: done

karlkfi (approved):
/lgtm

PR #728, opened 2023-07-10
Remove unneeded TestHydrateHelmComponents

janetkuo:
`TestHydrateHelmComponents` has been flaky in the 1.15 release branch CI. The Helm function that's being tested in `TestHydrateHelmComponents` has been removed from `main`. However, we don't want to cherrypick that breaking change to release branch. Instead, remove `TestHydrateHelmComponents` directly.

b/290467703
b/290491033
b/290521459


e2e/testcases/hydration_test.go
```diff
@@ -118,102 +118,6 @@ func TestHydrateKustomizeComponents(t *testing.T) {
 	}
 }
 
-func TestHydrateHelmComponents(t *testing.T) {
```
sdowell, line 121: Another option is to use `t.Skip()` so that the test still shows up in the test reports as skipped. Also would make this diff a bit smaller. I'm okay with either option

janetkuo: I thought about `t.Skip()` but given that it's a release branch and it's a test we no longer need. I just removed it.

haiyanmeng (approved):

sdowell:
Looks like the source branch was created in this repo (GoogleContainerTools/kpt-config-sync), please clean up the branch after the change is submitted

PR #730, opened 2023-07-10
test: add support for creating kcc clusters

sdowell:
This change adds support for enabling the ConfigConnector cluster addon when creating standard GKE clusters. Also moves the creation of the ConfigConnector CR to the test scaffolding, so that it can be created in a cluster that was provisioned by the tests.


e2e/nomostest/nt.go
```diff
@@ -784,6 +792,34 @@ func (nt *NT) detectGKEAutopilot(skipAutopilot bool) {
 	}
 }
 
+// setupKCC creates the ConfigConnector CR on the KCC cluster.
+// The target project and associated resources (e.g. GSAs) are currently hard
+// coded due to lack of provisioning automation. Eventually we may want to make
+// this more flexible/repeatable for different environments.
+func (nt *NT) setupKCC() {
```
karlkfi, line 799: suggest avoiding the kcc acronym wherever possible. the acronym has been officially deprecated. tho it's probably ok to leave it in the command flag just so it's short.

karlkfi: Use ConfigConnector

e2e/nomostest/new.go
```diff
@@ -310,6 +310,10 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 
 	}
 
+	if *e2e.Kcc {
```
karlkfi, line 313: This acronym is both deprecated and against Go best practices for field names. If it needs to be short, use KCC, not Kcc.

sdowell: Renamed

e2e/nomostest/nt.go
sdowell: Done

karlkfi (approved):
/lgtm

PR #731, opened 2023-07-11
test: set GKE cidr range when creating clusters

sdowell:
The default value is larger than what is required. This sets the default CIDR range to a more reasonable value. This helps fit a greater number of clusters into a network, particularly for CI use cases.

### DIFF
--- a/e2e/nomostest/clusters/gke.go
+++ b/e2e/nomostest/clusters/gke.go
@@ -193,6 +193,7 @@ func createGKECluster(t testing.NTB, name string) error {
 	args = append(args, name,
 		"--project", *e2e.GCPProject,
 		"--async",
+		"--cluster-ipv4-cidr", "/19", // use smaller than default CIDR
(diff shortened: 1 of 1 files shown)


e2e/nomostest/clusters/gke.go
```diff
@@ -193,6 +193,7 @@ func createGKECluster(t testing.NTB, name string) error {
 	args = append(args, name,
 		"--project", *e2e.GCPProject,
 		"--async",
+		"--cluster-ipv4-cidr", "/19", // use smaller than default CIDR
```
sdowell, line 196: Note this is the same value as is set by the [current terraform](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testinfra/terraform/modules/e2ecluster/e2ecluster.tf#L25)

tiffanny29631 (approved):
/lgtm

PR #734, opened 2023-07-12
test: set oauth scopes for gcenode e2e clusters

sdowell:
The e2e clusters which are used for gcenode auth must have oauth scopes specified to enable auth to CSR/GAR.


karlkfi (approved):
/lgtm

PR #735, opened 2023-07-12
helm: configmaps should be immutable

natasha41575:
- b/290634802: ConfigMaps must be immutable, with enforcement
- b/289374743: Mount ConfigMap as optional (validation checks are elsewhere) to ensure that the reconciler pod stays healthy if the ConfigMap is deleted before the RSync.

Note: If we decide we need/want to support non-immutable ConfigMaps later, we can remove this constraint then, but it will require some additional status improvements to be able to communicate that the latest ConfigMap state has been used for syncing. At the moment, neither RSync generation nor ConfigMap generation is included in the Syncing status condition.


pkg/reconcilermanager/controllers/reconciler_base_test.go
```diff
@@ -652,6 +653,7 @@ func TestMountConfigMapValuesFiles(t *testing.T) {
 								LocalObjectReference: corev1.LocalObjectReference{
 									Name: "foo",
 								},
+								Optional: &tr,
```
natasha41575, line 656: See the comment description in `mountConfigMapValuesFiles`:

```
					// The ConfigMap may be deleted before the RSync. To prevent the reconciler
					// pod from going into an error state when that happens, we must mark
					// this ConfigMap mount as optional and have our validation checks elsewhere.
```

manifests/reposync-crd.yaml
```diff
@@ -1201,7 +1201,7 @@ spec:
                       CLI. If `values` is also specified, fields from `values` will
                       override fields from valuesFileRefs.
                     items:
-                      description: ValuesFileRefs holds references to ConfigMap objects
+                      description: ValuesFileRef holds references to ConfigMap objects
```
karlkfi, line 1204: why singular? it's talking about the list still, right?

natasha41575: This is referring to the singular struct: https://github.com/GoogleContainerTools/kpt-config-sync/blob/547b4aa8f835147d078d2a5e944b903683097a9e/pkg/api/configsync/v1beta1/helmconfig.go#L111

The list field is still plural, so `valuesFileRefs` is now type `[]valuesFileRef`, which seemed like it made more sense.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -191,6 +191,23 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+		// unless the status update also fails.
+		_, updateErr := r.updateStatus(ctx, currentRS, rs)
+		// Use the validation error for metric tagging.
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+
+		// If the RSync is being deleted, we still need to run the teardown function,
+		// as we want the user to be able to delete the ConfigMap before the RSync.
+		// Therefore, we only return here if the RSync is not being deleted.
+		if rs.GetDeletionTimestamp().IsZero() {
+			return controllerruntime.Result{}, updateErr
```
karlkfi, line 207: else... log the error?

manifests/reposync-crd.yaml
natasha41575: The full CRD description reads like this:

```
valuesFileRefs:
                    description: valuesFileRefs holds references to objects in the
                      cluster that represent values to use instead of default values
                      that accompany the chart. Currently, only ConfigMaps are supported,
                      and the ConfigMap must be in the same namespace as the RootSync
                      or RepoSync. When multiple valuesFiles are used, duplicated
                      keys in later files will override the value from earlier files.
                      This is equivalent to passing in multiple valuesFiles to Helm
                      CLI. If `values` is also specified, fields from `values` will
                      override fields from valuesFileRefs.
                    items:
                      description: ValuesFileRef holds references to ConfigMap objects
                        in the cluster that represent values to use instead of default
                        values that accompany the chart.
                      properties:
                        name:
                          description: name represents the Object name. Required.
                          type: string
                        valuesFile:
                          default: values.yaml
                          description: valuesFile represents the object data key to
                            read the value from.
                          type: string
                      type: object
                    type: array
```

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi, line 207: are there any validation errors that might need to stop progress here that are not NotFound?

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -759,6 +736,25 @@ func (r *RootSyncReconciler) validateSpec(ctx context.Context, rs *v1beta1.RootS
 	}
 }
 
+// checkValuesFileSourcesRefs validates that the ConfigMaps specified in the RSync ValuesFileSources exist and have the
+// specified data key.
+func (r *RootSyncReconciler) checkValuesFileSourcesRefs(ctx context.Context, cl client.Client, rs *v1beta1.RootSync) status.Error {
```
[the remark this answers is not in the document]
karlkfi: also, when it's on the struct, we dont need to pass in the client.

pkg/validate/raw/validate/source_spec_validator.go
```diff
@@ -432,3 +419,12 @@ func MissingConfigMapKey(o client.Object, name, key string) status.Error {
 		Sprintf("%s field spec.helm.valuesFileRefs.valuesFile must specify an existing data key in the referenced ConfigMap; data key %q not found in ConfigMap %q", kind, key, name).
 		BuildWithResources(o)
 }
+
+// ConfigMapMustBeImmutable reports that a referenced ConfigMap from RSync spec.helm.valuesFileSources is
+// not immutable.
+func ConfigMapMustBeImmutable(o client.Object) status.Error {
+	kind := o.GetObjectKind().GroupVersionKind().Kind
+	return invalidSyncBuilder.
+		Sprintf("%ss must only reference ConfigMaps that are immutable", kind).
```
karlkfi, line 428: can we make this message more specific to helm valuesFileRefs? and include the name/ns of the COnfigMap that wasn't immutable

manifests/reposync-crd.yaml
karlkfi: if this is a description of a single item then it needs some work...

ValuesFileRef references a ConfigMap object that contains a values file to use for helm rendering. The ConfigMap must be in the same namespace as the RootSync/RepoSync.

natasha41575: updated the description

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: done

natasha41575: The errors that we might encounter from this validation check are that we can't fetch the ConfigMap, the ConfigMap isn't immutable, or the referenced ConfigMap doesn't have the specified data key. I don't think any of these errors need to halt teardown or deletion of the RSync.

pkg/reconcilermanager/controllers/rootsync_controller.go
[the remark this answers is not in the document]
natasha41575: I moved as much of this check as I could into the validation function to avoid the repetition, and am no longer passing in the client as a separate argument

pkg/validate/raw/validate/source_spec_validator.go
natasha41575: done

e2e/testcases/helm_sync_test.go
```diff
@@ -205,47 +177,58 @@ service:
...
-		expectedCPURequest = "250m"
-		expectedCPULimit = expectedCPURequest
-		expectedMemoryRequest = "512Mi"
-		expectedMemoryLimit = expectedMemoryRequest
-	} else {
-		expectedCPURequest = "150m"
-		expectedCPULimit = "1"
-		expectedMemoryRequest = "250Mi"
-		expectedMemoryLimit = "300Mi"
+	nt.WaitForRootSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RootSyncs must only reference ConfigMaps that are immutable; ConfigMap \"helm-watch-config-map\" in namespace \"config-management-system\" is not immutable")
```
karlkfi, line 201: Can we make this more consistent with the other error message and say "RootSyncs must reference valid ConfigMaps in spec.helm.valuesFileRefs: ConfigMap \"helm-watch-config-map\" in namespace \"config-management-system\" is not immutable"

e2e/testcases/helm_sync_test.go
```diff
@@ -205,47 +177,58 @@ service:
...
-			resource.MustParse(expectedMemoryLimit)),
-		testpredicates.HasExactlyImage("wordpress", "bitnami/wordpress", "", "sha256:362cb642db481ebf6f14eb0244fbfb17d531a84ecfe099cd3bba6810db56694e"),
-		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_USERNAME", "test-user-1"),
-		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
-		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_1", "val1"),
-		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_2", "val2")); err != nil {
+	if err := nt.KubeClient.Update(cm1); err != nil {
 		nt.T.Fatal(err)
 	}
+	nt.WaitForRootSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RootSync field spec.helm.valuesFileRefs.valuesFile must specify an existing data key in the referenced ConfigMap; data key \"values.yaml\" not found in ConfigMap \"helm-watch-config-map\"")
```
karlkfi, line 216: `RootSyncs must reference valid ConfigMaps in spec.helm.valuesFileRefs: ConfigMap \"helm-watch-config-map\" in namespace \"config-management-system\" does not contain the data key \"values.yaml\".

e2e/testcases/helm_sync_test.go
```diff
@@ -205,47 +177,58 @@ service:
...
+	// delete the ConfigMap
+	if err := nt.KubeClient.Delete(cm1); err != nil {
+		nt.T.Error(err)
+	}
+	nt.WaitForRootSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RootSyncs must reference valid ConfigMaps in spec.helm.valuesFileRefs: ConfigMap \"helm-watch-config-map\" not found")
 
-	nt.T.Log("Apply the ConfigMap with values: imagePullPolicy: Never; wordpressUserName: test-user-2")
+	nt.T.Log("Apply valid ConfigMap with values: imagePullPolicy: Always; wordpressUserName: test-user-1")
 	cm2 := fake.ConfigMapObject(core.Name(cmName), core.Namespace(configsync.ControllerNamespace))
+	cm2.Immutable = &tr
```
karlkfi, line 226: pointer.Bool(true)

manifests/rootsync-crd.yaml
```diff
@@ -1209,16 +1209,16 @@ spec:
...
                       cluster that represent values to use instead of default values
                       that accompany the chart. Currently, only ConfigMaps are supported,
-                      and the ConfigMap must be in the same namespace as the RootSync
-                      or RepoSync. When multiple valuesFiles are used, duplicated
-                      keys in later files will override the value from earlier files.
-                      This is equivalent to passing in multiple valuesFiles to Helm
-                      CLI. If `values` is also specified, fields from `values` will
-                      override fields from valuesFileRefs.
+                      and the ConfigMap must be in the same namespace as the RootSync/RepoSync.
+                      When multiple valuesFiles are used, duplicated keys in later
```
karlkfi, line 1213: s/valuesFiles are used/values files are specified/

karlkfi: same for valuesFiles below

pkg/reconcilermanager/controllers/reconciler_base_test.go
karlkfi: What use cases require this behavior?

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -191,6 +191,27 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+		r.logger(ctx).Error(err, "Sync spec invalid")
+		reposync.SetStalled(rs, "Validation", err)
+		// Validation errors should not trigger retry (return error),
+		// unless the status update also fails.
+		_, updateErr := r.updateStatus(ctx, currentRS, rs)
+		// Use the validation error for metric tagging.
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+
+		// If the RSync is being deleted, we still need to run the teardown function,
+		// as we want the user to be able to delete the ConfigMap before the RSync.
```
karlkfi, line 204: > we want the user to be able to delete the ConfigMap before the RSync.

Why do we want this?

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -191,6 +191,27 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+		// Therefore, we only return here if the RSync is not being deleted.
+		if rs.GetDeletionTimestamp().IsZero() {
+			return controllerruntime.Result{}, updateErr
+		}
+		if updateErr != nil {
+			r.logger(ctx).Error(updateErr, "Failed to update status")
```
karlkfi, line 210: Failed to update sync status

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -864,15 +844,21 @@ func (r *RepoSyncReconciler) validateSpec(ctx context.Context, rs *v1beta1.RepoS
...
-		return validate.CheckValuesFileRefs(ctx, r.client, rs.Spec.Helm.ValuesFileRefs, rs)
+		return validate.HelmSpec(reposync.GetHelmBase(rs.Spec.Helm), rs)
 	default:
 		return validate.InvalidSourceType(rs)
 	}
 }
 
+// checkValuesFileSourcesRefs validates that the ConfigMaps specified in the RSync ValuesFileSources exist and have the
+// specified data key.
+func (r *RepoSyncReconciler) checkValuesFileSourcesRefs(ctx context.Context, rs *v1beta1.RepoSync) status.Error {
```
karlkfi, line 855: rename validate- instead of check- for consistency with other validation methods

pkg/validate/raw/validate/source_spec_validator.go
```diff
@@ -193,31 +193,23 @@ func HelmSpec(helm *v1beta1.HelmBase, rs client.Object) status.Error {
 	return nil
 }
 
-// CheckValuesFileRefs validates that the ConfigMaps specified in the RSync ValuesFileRefs exist and have the
-// specified data key.
-func CheckValuesFileRefs(ctx context.Context, cl client.Client, valuesFileRefs []v1beta1.ValuesFileRefs, rs client.Object) status.Error {
+// CheckValuesFileRefs checks that the ConfigMaps specified by valuesFileRefs exist, are immutable, and have the provided data key.
+func CheckValuesFileRefs(ctx context.Context, cl client.Client, rs client.Object, valuesFileRefs []v1beta1.ValuesFileRef) status.Error {
```
karlkfi, line 197: with validate in the package name, "check" is redundant here

pkg/validate/raw/validate/source_spec_validator.go
```diff
@@ -432,3 +424,12 @@ func MissingConfigMapKey(o client.Object, name, key string) status.Error {
 		Sprintf("%s field spec.helm.valuesFileRefs.valuesFile must specify an existing data key in the referenced ConfigMap; data key %q not found in ConfigMap %q", kind, key, name).
 		BuildWithResources(o)
 }
+
+// ConfigMapMustBeImmutable reports that a referenced ConfigMap from RSync spec.helm.valuesFileSources is
+// not immutable.
+func ConfigMapMustBeImmutable(o client.Object, name string) status.Error {
```
karlkfi, line 430: HelmValuesConfigMapMustBeImmutable

karlkfi: prefix the other error funcs too. they're helm specific here

karlkfi (changes requested):

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: Are you referring to the behavior of allowing the ConfigMap to be deleted before the RootSync? I think there are two different issues here, so I will do my best to explain:

1. If the ConfigMap is deleted before the RSync and this mount is _not_ set to optional, the reconciler pod enters an error state. I think this is a bug, we should try to keep the reconciler healthy.

2. If the ConfigMap is deleted before the RSync, the reconciler-manager loop halts and the RSync can no longer be deleted (because it still has the reconciler-manager finalizer). The use case is just that I think the user should be allowed to delete the ConfigMap/RSync in whatever order they want to, since they may be independently managed. It seems like a bug to require that they be deleted in a particular order (especially because they can be created in any order).

pkg/validate/raw/validate/source_spec_validator.go
natasha41575: done

pkg/validate/raw/validate/source_spec_validator.go
natasha41575: fixed

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: done

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: done

manifests/rootsync-crd.yaml
natasha41575: done

e2e/testcases/helm_sync_test.go
natasha41575: thanks for the tip! updated

e2e/testcases/helm_sync_test.go
natasha41575: updated the error messages

e2e/testcases/helm_sync_test.go

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: gotcha. so validation errors only block setup, not teardown. makes sense.

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: It seems like a lot of extra code and complexity to fight against a constraint that k8s has for all deployments: the config must be specified before the deployment can become healthy, and the config must exist until after the deployment is deleted. This is an inherent dependency built into k8s primitives. It's also why we built the depends-on feature to handle creation and deletion ordering.

I do see what you're saying that it feels like a bug. It is eventually consistent, but it requires ordering to remain in a healthy state.

Thing is, if there's no helm config, the reconciler doesn't actually need to be healthy because the reconciler-manager will report on the RSync the error that the configmap is missing or invalid.

The only case that seems like it might need this feature is for deletion, so the reconciler can perform finalization based on the recorded inventory even if the ConfigMap was already deleted. Otherwise, something like namespace deletion could cause ConfigMap and RepoSync to be deleted in random order, potentially causing deadlock.

That said, we already have this problem with RoleBinding... When a namespace is deleted, sometimes the RoleBinding is deleted before the RepoSync, causing the reconciler finalizer to deadlock because it no longer has permission to delete objects in that namespace.

e2e/testcases/helm_sync_test.go
```diff
@@ -205,47 +178,56 @@ service:
...
-		expectedCPURequest = "250m"
-		expectedCPULimit = expectedCPURequest
-		expectedMemoryRequest = "512Mi"
-		expectedMemoryLimit = expectedMemoryRequest
-	} else {
-		expectedCPURequest = "150m"
-		expectedCPULimit = "1"
-		expectedMemoryRequest = "250Mi"
-		expectedMemoryLimit = "300Mi"
+	nt.WaitForRootSyncStalledError(rs.Namespace, rs.Name, "Validation", "KNV1061: RootSyncs must reference valid ConfigMaps in spec.helm.valuesFileRefs; ConfigMap \"helm-watch-config-map\" in namespace \"config-management-system\" is not immutable")
```
karlkfi, line 201: got a semi colon that should be a colon here. I've noticed this in a few of your errors in the past. The standard go pattern is to join error messages with colons. The error on the left provides more context about the error on the right, recursively.

That said, the configsync K- errors are kind of rule breakers, since they start with a capital letter. I'm not sure if there's some other convention I don't know about for semicolon use.

pkg/reconcilermanager/controllers/reconciler_base_test.go
karlkfi, line 656: pointer.Bool(true)

natasha41575: 🤦‍♀️ Ah my bad, missed that one, I'll update it

natasha41575: done

e2e/testcases/helm_sync_test.go
natasha41575: I changed these all to colons. I think you're right that we should try to follow the standard go pattern here where we can.

pkg/reconcilermanager/controllers/reposync_controller.go
natasha41575: SG. I added a comment that hopefully clarifies why this is needed.

karlkfi (approved):
/lgtm

PR #738, opened 2023-07-14

haiyanmeng:
Patch the OS-level packages and remove unneeded dependencies to minimize the vulnerabilities.

Note this does not address all the fixable vulnerabilities.


haiyanmeng:
Let us wait for the base image to pick up these vulnerability fixes.

PR #739, opened 2023-07-14

haiyanmeng:
Patch the OS-level packages and remove unneeded dependencies to minimize the vulnerabilities


karlkfi (approved):
/lgtm

PR #740, opened 2023-07-14

haiyanmeng:
This is to reduce the vulnerabilities. 

Note this does not address all the fixable vulnerabilities.

### DIFF
--- a/build/prow/vulnerability-scanner/Dockerfile
+++ b/build/prow/vulnerability-scanner/Dockerfile
@@ -16,9 +16,13 @@ FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:431.0.0 as gcloud-install
 
 FROM golang:1.19
 
-RUN apt-get update && apt-get install -y \
-    jq \
-    bsdmainutils # needed for "column" util
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get install -y \
+        jq \
+        bsdmainutils # needed for "column" util \
+    && apt-get -y autoremove \
+    && rm -rf /var/lib/apt/lists/*
 
 COPY --from=gcloud-install /usr/lib/google-cloud-sdk /opt/gcloud/google-cloud-sdk
 ENV PATH /opt/gcloud/google-cloud-sdk/bin:$PATH


build/prow/vulnerability-scanner/Dockerfile
```diff
@@ -16,9 +16,13 @@ FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:431.0.0 as gcloud-install
 
 FROM golang:1.19
 
-RUN apt-get update && apt-get install -y \
-    jq \
-    bsdmainutils # needed for "column" util
+RUN apt-get update \
+    && apt-get -y upgrade \
```
karlkfi, line 20: wait... why are we doing an upgrade? shouldn't we just bump the base image to fix vulns? Or does Go not publish enough to do this?

Doing upgrade gives you a different OS than what is in the base image...

If we really need to upgrade from Go, we should probably make that it's own base image, so it's consistent, and it will make these build faster.

haiyanmeng: I tested bumping the base image, but it doesn't reduce the fixable vulnerabilities.
Doing an upgrade is to follow the step 4 listed in [go/autovm-container-remediation](https://goto.google.com/autovm-container-remediation).

karlkfi: upgrades should really only be done in OS-level images not the higher level images that depend on them, otherwise the higher level image builds are slower and not reproducable. We want to make sure re-building gets you the exact same bits, as much as possible, otherwise you can't reliably test and promote the images.

If we need to do an upgrade, we need to start hosing our own Go base image with the upgrade already done. Then we can point to that versioned image in all the child images that need it.

karlkfi: Another option could be to switch to using chainguard images for base images. They patch them to fix vulns.

https://edu.chainguard.dev/chainguard/chainguard-images/reference/go/overview/

haiyanmeng: The chianguard golang images does not have `apt-get`:
```
$ docker buildx build build/prow/vulnerability-scanner/ -t gcr.io/haiyanmeng-gke-322517/vulnerability-scanner:chain-1.19
[+] Building 18.3s (7/8)                                                                                                                                                                                                                                                                  
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                                 0.0s
 => => transferring dockerfile: 955B                                                                                                                                                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                      0.0s
 => [internal] load metadata for cgr.dev/chainguard/go:1.19                                                                                                                                                                                                                          0.7s
 => [internal] load metadata for gcr.io/google.com/cloudsdktool/google-cloud-cli:431.0.0                                                                                                                                                                                             0.1s
 => CACHED [gcloud-install 1/1] FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:431.0.0@sha256:6a8a85c6c68b1d85621f61d46c5a8e49cb2420193ad1f4f12dd766f75ee30d87                                                                                                                 0.0s
 => [stage-1 1/3] FROM cgr.dev/chainguard/go:1.19@sha256:df0f29f3e6443cdd67440734b395b3ec6ca3c6391b66a29f6b56b8217c7f9d8b                                                                                                                                                           16.0s
 => => resolve cgr.dev/chainguard/go:1.19@sha256:df0f29f3e6443cdd67440734b395b3ec6ca3c6391b66a29f6b56b8217c7f9d8b                                                                                                                                                                    0.0s
 => => sha256:de912e8bb05f516298f9b0a9fd7edc92da5d982c7b413f010881f747fa926489 707B / 707B                                                                                                                                                                                           0.0s
 => => sha256:e7554c23d1b8805016c623ca1d00c6b362136bdae4459b85ce3ddc8033cc1c50 549B / 549B                                                                                                                                                                                           0.0s
 => => sha256:3b7ebf97a83f5326904d6216f10d0cc33b354f33212ae49db0b4047805b7833f 241.39MB / 241.39MB                                                                                                                                                                                  10.3s
 => => sha256:df0f29f3e6443cdd67440734b395b3ec6ca3c6391b66a29f6b56b8217c7f9d8b 791B / 791B                                                                                                                                                                                           0.0s
 => => extracting sha256:3b7ebf97a83f5326904d6216f10d0cc33b354f33212ae49db0b4047805b7833f                                                                                                                                                                                            5.5s
 => ERROR [stage-1 2/3] RUN apt-get update && apt-get install -y     jq     bsdmainutils # needed for "column" util                                                                                                                                                                  1.4s
------                                                                                                                                                                                                                                                                                    
 > [stage-1 2/3] RUN apt-get update && apt-get install -y     jq     bsdmainutils # needed for "column" util:
#0 0.458 /bin/sh: apt-get: not found
------
ERROR: failed to solve: executor failed running [/bin/sh -c apt-get update && apt-get install -y     jq     bsdmainutils # needed for "column" util]: exit code: 127
```

haiyanmeng: Let us wait for the base image to pick up these vulnerability fixes.

haiyanmeng:
Let us wait for the base image to pick up these vulnerability fixes.

build/prow/vulnerability-scanner/Dockerfile
karlkfi: chainguard has it's own package manager. we would need to switch to that to get packages we need.

PR #747, opened 2023-07-17
Avoid creating LoadBalancer in wordpress Helm chart

janetkuo:
This removes the need of LoadBalancer and avoids the problem of firewall rules cleanup.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -408,7 +408,7 @@ image:
 			  "wordpressEmail": "test-user@example.com",
 			  "service": {
 				"type": "ClusterIP"
-			  },
+			  }
 			},
 			"valuesFileRefs": [
 			  {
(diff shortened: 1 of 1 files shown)


karlkfi (approved):
/lgtm

e2e/testcases/helm_sync_test.go
```diff
@@ -405,7 +405,10 @@ image:
 				  "value": "inline"
 				}
 			  ],
-			  "wordpressEmail": "test-user@example.com"
+			  "wordpressEmail": "test-user@example.com",
+			  "service": {
+				"type": "ClusterIP"
```
karlkfi, line 410: missing a trailing comma

sdowell: This is JSON rather than Go, so I think the validation error may actually be coming from the superfluous trailing comma on line 411

janetkuo: Right, it's the last item in values so shouldn't need to trailing comma. Updated.

PR #748, opened 2023-07-17
Fix otel-collector logging repeat error

tiffanny29631:
The otel-collector pod log shows repeat entries of duplicate timeseries error that takes up logging quota. Although the metris are correctly being sent, the error message can be confusing to users.

The removal of `type` tag is causing one timeseries being processed multiple times in `batch`, both with and without the tag, which causes the above error.

This change uses aggregation in metricstransfrom processor to eliminate the `type` tag from Monarch pipeline. Format follows otel-collector-contrib 0.54.0.

b/290678742


pkg/metrics/otel.go
```diff
@@ -163,10 +163,15 @@ processors:
...
-      - key: type
-        action: delete
   metricstransform/kubernetes:
     transforms:
+      - include: api_duration_seconds
+        action: update
+        operations:
+          # Eliminate 'type' tag from metric using aggregation
+          - action: aggregate_labels
+            label_set: [ status, operation ]
```
karlkfi, line 173: remove leading and trailing space in the array

tiffanny29631: Done!

pkg/metrics/otel.go
sdowell, line 1: Can we add some e2e tests with this change? I'd expect to see some tests with a fix like this, e.g. that the error that was being produced before is no longer being produced now

tiffanny29631: A scanning of the otel-collector log would be able to check this, let me see what I can do

sdowell: Are there any other signals we could verify? e.g. status/metrics? Logs would be better than nothing but would most likely be fragile

tiffanny29631: The error in this scenario is specifically in the log - metrics are all exporting fine and pod status is healthy, only log is printing repeat error and growing into gigantic file.

pkg/metrics/otel.go
```diff
@@ -163,10 +163,15 @@ processors:
...
-      - key: type
-        action: delete
   metricstransform/kubernetes:
     transforms:
+      - include: api_duration_seconds
+        action: update
+        operations:
+          # Eliminate 'type' tag from metric using aggregation
+          - action: aggregate_labels
+            label_set: [status, operation]
```
karlkfi, line 173: This removes all labels except the ones listed, right? 
Does is also remove resource attributes?

tiffanny29631: This lists out the remaining labels according to doc https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.54.0/processor/metricstransformprocessor#aggregate-labels.

tiffanny29631: This is one of the downside of using metricstransform to aggregate labels, you need to call out the metric, and the labels that you want to *keep*, not *aggregated away*

tiffanny29631: FYI the resource attributes remain, i.e. the `configsync_sync_*` tags

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+			nt.PodLogs("config-management-monitoring", ocmetrics.OtelCollectorName, "", false)
+		}
+	})
+
+	// If Workload Identity enabled on cluster, setup KSA to GSA annotation
+	if workloadPool, err := getWorkloadPool(nt); err != nil {
+		nt.T.Fatal(err)
+	} else if workloadPool != "" {
+		nt.T.Log("Workload identity enabled, adding KSA annotation to use %s service account", MonitorGSA)
+		nt.MustKubectl("annotate", "serviceaccount", "--namespace", ocmetrics.MonitoringNamespace, DefaultMonitorKSA,
```
karlkfi, line 51: Is this necessary with the new feature @haiyanmeng added? 
The reconciler-manager automatically copies the annotation now?
Don't remember from where to where tho.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+			fmt.Sprintf("iam.gke.io/gcp-service-account=%s@%s.iam.gserviceaccount.com", MonitorGSA, *e2e.GCPProject))
+	}
+
+	nt.T.Log("Restart otel-collector pod to refresh the ConfigMap, log and IAM")
+	nomostest.DeletePodByLabel(nt, "app", "opentelemetry", false)
+	if err := nt.WatchForAllSyncs(); err != nil {
```
karlkfi, line 57: Killing the collector pod will drop the metrics database, right?
Or is it persisted now?
Do we need to wait for sync? Or can we just wait for the otel-collector to be Current?

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+		nt.T.Fatal(err)
+	}
+	if err := validateOtelCollectorStatusCurrent(nt, false); err != nil {
+		nt.T.Fatal()
+	}
+
+	nt.T.Log("Wait 1 minute for metric exporter to settle")
+	time.Sleep(1 * time.Minute)
```
karlkfi, line 65: What does settle mean here?

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+	nt.T.Log("Checking the otel-collector log contains no failure...")
+	err := CheckDeploymentLogHasNoFailure(nt, ocmetrics.OtelCollectorName, ocmetrics.MonitoringNamespace)
```
karlkfi, line 67: What kind of failure are we expecting not to be here?

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+	return nil
+}
+
+// func validateDeploymentstatuscurrent
+func validateOtelCollectorStatusCurrent(nt *nomostest.NT, customConfigMap bool) error {
+	tg := taskgroup.New()
+	tg.Go(func() error {
+		return nt.Watcher.WatchForCurrentStatus(kinds.Deployment(), ocmetrics.OtelCollectorName, ocmetrics.MonitoringNamespace)
```
karlkfi, line 135: Should we maybe wait for the Deployment to be current after all the ConfigMaps are current? That way we know it has the latest ConfigMap?

Do ConfigMaps even have status? What exactly are we waiting for here?

pkg/metrics/otel.go
```diff
@@ -165,10 +165,15 @@ processors:
...
       - key: commit
         action: delete
-      - key: type
-        action: delete
   metricstransform/kubernetes:
     transforms:
+      - include: api_duration_seconds
+        action: update
+        operations:
+          # Eliminate 'type' tag from metric using aggregation
```
karlkfi, line 173: Did you also remove commit or just type?

e2e/testcases/otel_collector_test.go
sdowell: I suspect `WatchForAllSyncs` will just return immediately in this case, I don't think it's going to wait for the otel Pod to get restarted.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	nt.T.Log("wait 1 minute for metric exporter to settle")
+	time.Sleep(1 * time.Minute)
+	nt.T.Log("Checking the otel-collector log contains failure...")
+	if err := CheckDeploymentLogHasFailure(nt, ocmetrics.OtelCollectorName, ocmetrics.MonitoringNamespace); err != nil {
+		nt.T.Fatal(err)
+	}
+}
+
+func CheckDeploymentLogHasFailure(nt *nomostest.NT, deployment, namespace string) error {
```
sdowell, line 91: nit: this function is named pretty generically, so I'd suggest surfacing the log string (used by `strings.Contains`) as a parameter

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+		nt.T.Fatal(err)
+	}
+	if err := validateOtelCollectorStatusCurrent(nt, true); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("wait 1 minute for metric exporter to settle")
+	time.Sleep(1 * time.Minute)
+	nt.T.Log("Checking the otel-collector log contains failure...")
+	if err := CheckDeploymentLogHasFailure(nt, ocmetrics.OtelCollectorName, ocmetrics.MonitoringNamespace); err != nil {
```
sdowell, line 86: Using `time.Sleep` and then checking for this seems like a potential source of flakiness. Can we instead retry until we see the failure mode?

e2e/testcases/otel_collector_test.go
sdowell: Is there a success signal that we can validate instead of checking the absence of the log?

e2e/testcases/otel_collector_test.go
tiffanny29631: For the ConfigMap I just want to check it's existence.

It's a good point to wait for the ConfigMap before deployment. For the test ConfigMap (the fault one), it's applied by mustKubectl, could it provide some guarantee that the resource will be applied otherwise error out?

pkg/metrics/otel.go
tiffanny29631: Just `type`, commit is not triggering the duplicate error so it can remain the same way with other resource tags.

e2e/testcases/otel_collector_test.go
tiffanny29631: Here we are expecting no failure at all. A healthy unclogged pipeline should be running with `..everything ready..`, clean of error.

e2e/testcases/otel_collector_test.go
tiffanny29631: That when the exporters are either shipping the metrics out, or emitting errors in the log. From the past test there is about 30 to 40 seconds of delay between `otel-collector ready` and `metrics showing up in GCM / errors showing up in pod log`.

e2e/testcases/otel_collector_test.go
tiffanny29631: Killing the pod should drop the metrics database. We are not validating any metric or value in this test, just checking if the exporting is clogged, so persistence is not hard requirement.

The repo should be already synced when the test started. We are just waiting for the new otel-collector deployment to be current, with the validateOtelCollectorStatusCurrent.

e2e/testcases/otel_collector_test.go
tiffanny29631: Quick involving @haiyanmeng to confirm :]

I see annotation is still a necessary step in the public doc https://cloud.google.com/anthos-config-management/docs/how-to/monitor-config-sync-cloud-monitoring#custom-monitoring, is the copying of annotation in OSS or in CDCS?

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,150 @@
...
+	for _, m := range entry {
+		if strings.Contains(m, "failed to export time series to GCM") {
+			return fmt.Errorf("failure found in the log of deployment %s, namespace %s: %s", deployment, namespace, m)
+		}
+	}
+	return nil
+}
+
+// func validateDeploymentstatuscurrent
+func validateOtelCollectorStatusCurrent(nt *nomostest.NT, customConfigMap bool) error {
```
sdowell, line 132: The `Validate` functions in the test scaffolding typically don't wait for the condition to be satisfied - rather they just check that the condition is currently satisfied.

Suggest renaming this to something like `waitForOtelCollectorStatusCurrent` or `watchForOtelCollectorStatusCurrent`

e2e/testcases/otel_collector_test.go
sdowell: The issue with checking for the absence of a failure signal is that we aren't really sure that things are working as expected. For example, if we had waited an extra 10 seconds would we have seen the failure log?

e2e/testcases/otel_collector_test.go
karlkfi: There's a reconciler-manager part and a nomos-operator part and a pending hub integration.
AFAIK, the public docs are not updated yet because it's a 1.16 change.

e2e/testcases/otel_collector_test.go
karlkfi: Is this because of the otel-agent buffer?
I have a pending PR to disable it for e2e to get more responsive metrics.
https://github.com/GoogleContainerTools/kpt-config-sync/pull/784

e2e/testcases/otel_collector_test.go
karlkfi: I think after applying a ConfigMap, as long as the apply didn't error, you don't need to check that it's current.

However, I would recommend using the KubeClient instead of MustKubectl. The KubeClient will give is much better errors that we can catch if necessary.

karlkfi: The problem with mutable ConfigMaps is that there's no way to know after updating them that the Pods have actually picked up the change. You might want to patch the otel-collector Deployment to add a spec.template.spec.metadata.label with a hash of the ConfigMap. That way you can check if the label has been updated on the pod to ensure it's been replaced since the ConfigMap was updated.

karlkfi: Alternatively, you could just change the Deployment pod spec to point to a new ConfigMap (new name). That's the immutable ConfigMap strategy.

e2e/testcases/otel_collector_test.go
tiffanny29631: 🤔 Then we will have to ping the listmetric API then - get the ack for metrics, then check for error free log. That's a different monster, I'll see what I can do.

e2e/testcases/otel_collector_test.go
tiffanny29631: It's likely the buffer, or batch processing. I'm thinking of using more interactive to decide on ok/error status instead of a hardcoded 1min sleep so that any buffering wouldn't bring flakiness. Might take a few moment.

e2e/testcases/otel_collector_test.go
karlkfi: Would be awesome if this test performed a GCM metrics query and validated the result.

e2e/testcases/otel_collector_test.go
tiffanny29631: The test ConfigMap is with a different name (using otel-collector-custom). Problem is otel-collector deployment is using the projected volume for the CMs and the bottom one will take precedence, and the ConfigMap would just overrides the config data. For now there is no way but just assuming if both `otel-collcetor-googlecloud` and `otel-collector-custom` exist under monitoring namespace, the later one will be effective according to k8s design. https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/templates/otel-collector.yaml#L163

e2e/testcases/otel_collector_test.go
tiffanny29631: Name changed to `watchFor...`

e2e/testcases/otel_collector_test.go
tiffanny29631: Changed to use retry with 1min timeout

e2e/testcases/otel_collector_test.go
tiffanny29631: Surfaced failure string to args.

e2e/testcases/otel_collector_test.go
tiffanny29631: Added list request to verify that metrics have reached GCM. Right now I'm only using the apply_operations_total and checking the cluster name & metric name to match.

e2e/testcases/otel_collector_test.go
tiffanny29631: The sleep is replaced by the GCM list request, if request successes within 2min timeout then test move on to check pod log. The buffer is no longer a blocker.

e2e/testcases/otel_collector_test.go
tiffanny29631: For now what I see is test would fail if the annotation is not added.

e2e/testcases/otel_collector_test.go
sdowell: In that case I think this `WatchForAllSyncs` can be removed

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,215 @@
...
+// Make a ListTimeSeries request of a specific metric to GCM and look for the
+// exact metric by metricType and clusterName in response.
+// Note: metricType in this context is the metric descriptor name, for example
+// "custom.googleapis.com/opencensus/config_sync/apply_operations_total".
+func watchForMetricInGCM(ctx context.Context, nt *nomostest.NT, client *monitoringv2.MetricClient, timeDiff time.Duration, metricType string) error {
+	startTime := time.Now().UTC().Add(timeDiff)
+	endTime := time.Now().UTC()
+	req := &monitoringpb.ListTimeSeriesRequest{
+		Name:   "projects/" + *e2e.GCPProject,
+		Filter: `metric.type="` + metricType + `" AND resource.labels.cluster_name="` + *e2e.GCPCluster + `"`,
```
sdowell, line 184: Use `nt.ClusterName` instead of `*e2e.GCPCluster`. `GCPCluster` only works if the test suite is running sequentially against a single cluster.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,215 @@
...
+		return nil, err
+	}
+	return client, nil
+}
+
+// Make a ListTimeSeries request of a specific metric to GCM and look for the
+// exact metric by metricType and clusterName in response.
+// Note: metricType in this context is the metric descriptor name, for example
+// "custom.googleapis.com/opencensus/config_sync/apply_operations_total".
+func watchForMetricInGCM(ctx context.Context, nt *nomostest.NT, client *monitoringv2.MetricClient, timeDiff time.Duration, metricType string) error {
```
sdowell, line 179: This function doesn't appear to be performing a `watch`, suggest renaming

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,215 @@
...
+	nt.T.Log("Checking the otel-collector log contains failure...")
+	_, err = retry.Retry(60*time.Second, func() error {
+		return CheckDeploymentLogHasFailure(nt, ocmetrics.OtelCollectorName, ocmetrics.MonitoringNamespace, GCMExportErrorCaption)
+	})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+}
+
+func CheckDeploymentLogHasFailure(nt *nomostest.NT, deployment, namespace, errorString string) error {
```
sdowell, line 117: Suggest renaming each of these `Check*` functions to `Validate*` for consistency with similar functions

tiffanny29631: Done!

e2e/testcases/otel_collector_test.go
tiffanny29631: Split to a list function and a validate function

e2e/testcases/otel_collector_test.go
tiffanny29631: Done!

tiffanny29631:
/hold  

Not merging until RC is promoted.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,216 @@
...
+		if strings.Contains(m, errorString) {
+			return fmt.Errorf("failure found in the log of deployment %s, namespace %s: %s", deployment, namespace, m)
+		}
+	}
+	return nil
+}
+
+// func validateDeploymentstatuscurrent
+func watchForOtelCollectorStatusCurrent(nt *nomostest.NT) error {
+	tg := taskgroup.New()
```
karlkfi, line 162: this taskgroup is unnecessary with only one task. in fact, you can probably just inline the WatchForCurrentStatus call.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,216 @@
...
+	ctx := context.Background()
+	client, err := createGCMClient(ctx)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	// retry for 2 minutes until metric is accessible from GCM
+	_, err = retry.Retry(120*time.Second, func() error {
+		for _, metricType := range GCMMetricTypes {
+			descriptor := fmt.Sprintf("%s/%s", GCMMetricPrefix, metricType)
+			it := listMetricInGCM(ctx, nt, client, -2*time.Minute, descriptor)
```
karlkfi, line 81: Is 2m always going to be enough?

Do we expect to see all the metrics in the list in this time frame if we didn't trigger any actions?

pkg/metrics/otel.go
karlkfi, line 173: nit: maybe just "Remove" instead of "Eliminate"

karlkfi: maybe also note that it's high cardinality as the reason

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	return client, nil
+}
+
+// Make a ListTimeSeries request of a specific metric to GCM and look for the
+// exact metric by metricType and clusterName in response.
+// Note: metricType in this context is the metric descriptor name, for example
+// "custom.googleapis.com/opencensus/config_sync/apply_operations_total".
+func listMetricInGCM(ctx context.Context, nt *nomostest.NT, client *monitoringv2.MetricClient, timeDiff time.Duration, metricType string) *monitoringv2.TimeSeriesIterator {
+	startTime := time.Now().UTC().Add(timeDiff)
```
karlkfi, line 183: nit: might be slightly clearer if this just uses Sub with a positive value, maybe called "range"

e2e/testcases/otel_collector_test.go
tiffanny29631: Done!

e2e/testcases/otel_collector_test.go
tiffanny29631: They use Sub() to get the duration between t-u.. not quite aligned with Add where a time of t+diff is returned. https://github.com/golang/go/blob/master/src/time/time.go#L893

pkg/metrics/otel.go
tiffanny29631: Done!

e2e/testcases/otel_collector_test.go
tiffanny29631: From what I've seen so far (in freshtestenv) 2m is good enough, the initial commit should be syncing so no additional trigger actions are needed.  The metrics listed in GCMMetricTypes can all be detected once the exporter is up running.

e2e/testcases/otel_collector_test.go
tiffanny29631: Done!

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,208 @@
...
+		nt.MustKubectl("delete", "-f", "../testdata/otel-collector/otel-cm-duplicate-timeseries.yaml", "--ignore-not-found")
+		if t.Failed() {
+			nt.PodLogs("config-management-monitoring", ocmetrics.OtelCollectorName, "", false)
+		}
+	})
+
+	// If Workload Identity enabled on cluster, setup KSA to GSA annotation
+	if workloadPool, err := getWorkloadPool(nt); err != nil {
+		nt.T.Fatal(err)
+	} else if workloadPool != "" {
```
sdowell, line 58: Will this test case work if workload identity is disabled? If not, perhaps we should fail fast here if it's disabled

e2e/testcases/otel_collector_test.go
sdowell: Instead of subtracting two minutes, could we record the time right after we restart the Pod and use that as the start time?

e2e/testcases/otel_collector_test.go
sdowell: Perhaps we could just provide a start time as an argument instead of using a delta? (see my comment above)

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,208 @@
...
+		nt.T.Log("Workload identity enabled, adding KSA annotation to use %s service account", MonitorGSA)
+		nt.MustKubectl("annotate", "serviceaccount", "--namespace", ocmetrics.MonitoringNamespace, DefaultMonitorKSA,
```
sdowell, line 60: Looks like this KSA needs a cleanup to be removed after the test. Also suggest using the golang `KubeClient` instead of shelling to kubectl when possible

sdowell (commented):
close to LGTM, but a few more comments

e2e/testcases/otel_collector_test.go
tiffanny29631: Good point, just did, using the time right after otel-collector became current as startTime.

e2e/testcases/otel_collector_test.go
tiffanny29631: Done

e2e/testcases/otel_collector_test.go
tiffanny29631: The test would work if WI is disabled, it just doesn't need the annotation. Tested with both clusters locally.

sdowell: I see, I guess in that case the compute GSA needs permission to write metrics?

tiffanny29631: Yes it's looking for the GSA account added in https://github.com/GoogleContainerTools/kpt-config-sync/pull/773 if WI is enabled. For WI disabled cluster we don't need to do anything special.

e2e/testcases/otel_collector_test.go
tiffanny29631: I can see the apply being done by KubeClient,  but Annotation as a new helper function to be added in the KubeClient? https://github.com/GoogleContainerTools/kpt-config-sync/blob/3dde5af1eda0b5477a28efccd63680c0646cf34a/e2e/nomostest/testkubeclient/client.go#L38

tiffanny29631: Added annotate and deannotate function in the test using kubeClient. Do we want to do the same to the test yaml file?

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+		if resp.GetMetric().GetType() == metricType &&
+			resp.GetResource().GetLabels()["cluster_name"] == clusterName {
+			return nil
+		}
+	}
+	return fmt.Errorf("did not find target time series in GCM: type %s, cluster name %s", metricType, nt.ClusterName)
+}
+
+func annotateKSAwithGSA(nt *nomostest.NT, annotationKey, annotationValue, ksaName, ksaNamespace string) error {
```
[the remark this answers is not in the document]
tiffanny29631: Done! Now called setAnnotation and unsetAnnotation with according arguments and return types.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+var GCMMetricTypes = []string{ocmetrics.ReconcilerErrors.Name(), ocmetrics.PipelineError.Name(), ocmetrics.ReconcileDuration.Name(), ocmetrics.ParserDuration.Name(), ocmetrics.InternalErrors.Name()}
+
+func TestOtelCollectorDeployment(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.Reconciliation1, ntopts.RequireGKE(t))
+	nt.T.Cleanup(func() {
+		nt.MustKubectl("delete", "-f", "../testdata/otel-collector/otel-cm-duplicate-timeseries.yaml", "--ignore-not-found")
+		err := deannotateKSAwithGSA(nt, "iam.gke.io/gcp-service-account", DefaultMonitorKSA, ocmetrics.MonitoringNamespace)
+		if err != nil {
```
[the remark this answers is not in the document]
tiffanny29631: I put them in a separate cleanup in the same test, let me know if it works.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// The ConfigMap that is expected to trigger duplicate time series error has
+	// name 'otel-collector-custom', which by the setup in otel-collector deployment
+	// will take precedence over 'otel-collector-googlecloud' that was deployed
+	// by the test.
+	nt.T.Log("Apply custom otel-collector ConfigMap that could cause duplicate time series error")
+	nt.MustKubectl("apply", "-f", "../testdata/otel-collector/otel-cm-duplicate-timeseries.yaml")
+	nt.T.Log("Restart otel-collector pod to refresh the ConfigMap and log")
+	nomostest.DeletePodByLabel(nt, "app", ocmetrics.OpenTelemetry, false)
+	if err := nt.WatchForAllSyncs(); err != nil {
```
sdowell, line 124: nit: I think this `WatchForAllSyncs` is most likely a no-op and can be removed, since `WatchForAllSyncs` won't wait for the otel Pod to restart. The `WatchForCurrentStatus` below should be sufficient AFAICT.

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,260 @@
...
+	obj.SetAnnotations(annotations)
+	if err := client.Update(obj); err != nil {
+		return false, fmt.Errorf("failed to apply updated object to cluster: %v", err)
+	}
+	return true, nil
+}
+
+func unsetAnnotation(client *testkubeclient.KubeClient, annotationKey string, obj client.Object) (bool, error) {
+	annotations := obj.GetAnnotations()
+	if annotations != nil {
```
sdowell, line 248: nit: could reduce the layers of nested logic by slightly refactoring this (for readability). e.g.
```golang
func unsetAnnotation(client *testkubeclient.KubeClient, annotationKey string, obj client.Object) (bool, error) {
	annotations := obj.GetAnnotations()
    if annotations == nil {
        return false, nil
    }
    if _, exists := annotations[annotationKey]; !exists {
        return false, nil
    }
    delete(annotations, annotationKey)
	obj.SetAnnotations(annotations)
	if err := client.Update(obj); err != nil {
	    return false, fmt.Errorf("failed to apply updated object to cluster: %v", err)
	}
	return true, nil
}
```

sdowell (commented):
couple nits but mostly LGTM

e2e/testcases/otel_collector_test.go
tiffanny29631: Removed!

e2e/testcases/otel_collector_test.go
tiffanny29631: Done! Thanks for the suggestion.

e2e/testcases/otel_collector_test.go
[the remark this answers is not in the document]
tiffanny29631: Also added otel-collector pod restart to reset the ConfigMap back to otel-collector-googlecloud, as by the time test finishes, the exporters are in error status.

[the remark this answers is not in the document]
sdowell: Good catch! Might also want to wait for a current status on the deployment after deleting the Pod

e2e/testcases/otel_collector_test.go
sdowell: Ideally I think yes it would be preferred to inline the object with `KubeClient`, however I can see an argument for using the yaml file since it's fairly verbose

e2e/testcases/otel_collector_test.go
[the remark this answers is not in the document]
tiffanny29631: Done!

e2e/testcases/otel_collector_test.go
tiffanny29631: Keep it for now?

sdowell: Yeah that's okay with me

e2e/testcases/otel_collector_test.go
```diff
@@ -0,0 +1,262 @@
...
+			nt.PodLogs("config-management-monitoring", ocmetrics.OtelCollectorName, "", false)
+		}
+	})
+
+	nt.T.Cleanup(func() {
+		nt.MustKubectl("delete", "-f", "../testdata/otel-collector/otel-cm-duplicate-timeseries.yaml", "--ignore-not-found")
+		nt.T.Log("Restart otel-collector pod to reset the ConfigMap and log")
+		nomostest.DeletePodByLabel(nt, "app", ocmetrics.OpenTelemetry, false)
+		if err := nt.Watcher.WatchForCurrentStatus(kinds.Deployment(), ocmetrics.OtelCollectorName, ocmetrics.MonitoringNamespace); err != nil {
+			nt.T.Log(fmt.Sprintf("otel-collector pod failed to come up after a restart: %v", err))
```
sdowell, line 62: Final nit: Replace `nt.T.Log(fmt.Sprintf(...` with `nt.T.Errorf(...`. Ditto below

tiffanny29631: Done!

sdowell: I don't see any new changes, you may have forgotten to push :)

tiffanny29631: Forgot to amend the commit, now good

sdowell (approved):

karlkfi:
This merged during code freeze... is it needed for v1.16.0?

tiffanny29631:
It is no longer targeting for 1.16. It was merged after the projected preview date, in theory the release candidate is confirmed by then.

PR #750, opened 2023-07-19
test: update default GKE shape

sdowell:
This updates the default shape of GKE clusters created by the e2e test scaffolding. The default values here are chosen to be identical to the cluster shape used by the CI clusters, which is optimized for speed and minimal resource allocation.


karlkfi (approved):
/lgtm

PR #753, opened 2023-07-20
Use a patch go version

haiyanmeng:
Using a minor go version, different builds may generate different images.


sdowell (approved):
We might want to make a similar change for the other docker images (e.g. those used for tests), but I think this is sufficient for now

nan-yu (approved):
/lgtm

PR #756, opened 2023-07-20
Bump askpass image (#755)

haiyanmeng:
This is to pick up the vulnerability fix in http://tg/1958866.

b/291841151


nan-yu (approved):
/lgtm

PR #759, opened 2023-07-21
Avoid resetting the retryTimer to re-enable the smart retry logic

haiyanmeng:
The smart-retry logic was added in http://tg/983432, but was disabled in https://github.com/GoogleContainerTools/kpt-config-sync/pull/279.

The retry behaves in the following way:
 * For the first 5 retries for reconciling a git change which fail with
   the same errors, the reconciler waits 1 second before retrying;
 * For the remaining retries for reconciling a git change which fail
   with the same errors, the reconciler does exponential backoff retry
   up to 5 minutes (i.e., 1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s, 5m, 5m, ...).

### DIFF
--- a/pkg/parse/run.go
+++ b/pkg/parse/run.go
@@ -130,7 +130,6 @@ func Run(ctx context.Context, p Parser) {
 			}
 			run(ctx, p, trigger, state)
 
-			retryTimer.Reset(opts.retryPeriod)               // Schedule retry attempt
 			statusUpdateTimer.Reset(opts.statusUpdatePeriod) // Schedule status update attempt
 
 		// Update the sync status to report management conflicts (from the remediator).


pkg/parse/run.go
```diff
@@ -130,7 +130,6 @@ func Run(ctx context.Context, p Parser) {
 			}
 			run(ctx, p, trigger, state)
 
-			retryTimer.Reset(opts.retryPeriod)               // Schedule retry attempt
```
sdowell, line 133: How does this change the interaction with the `readyToRetry` logic? It seems like that's independent from this timer

haiyanmeng: You are right. I just tested our main branch, it does still support the smart retry. 

Let me go back to Morten's complaint to see whether I missed anything. Thanks!

PR #691, opened 2023-06-16
[WIP] Configure VPA for reconciler, if enabled

karlkfi:
- Configure VPA for reconciler deployment using annotation on
  RootSync/RepoSync:
  `configsync.gke.io/reconciler-autoscaling-strategy: Auto`
  - Auto - evict and recreate pods to apply recommended resource
    values, as needed.
  - Recommend - monitor and record recommended resource values for
    each reconciler, but don't automatically apply them.
  - Disabled - Do not apply any VPA config, and delete it if it
    exists with the same name as the reconciler.
- VPA disabled by default (opt-in for preview and testing)
- When VPA is enabled, set smaller resource requests/limits for
  smaller footprint on initial install. Adding limits helps
  hasten VPA adjustments by causing OOMKills, instead of waiting
  for the VPA to evict the pod.
- Move regular (non-VPA) defaults out of a ConfigMap and into the
  reconciler-manager code, next to the new VPA resource defaults.
  This should make them easier to keep in sync.
- test: Install VPA on kind when --vpa is specified
- test: Enable the VPA addon in GKE when creating clusters when
  --vpa is specified
- test: Rewrite some e2e tests to handle resource defaults
- test: Log reconciler pod resources on test failure to help debug
  VPA.

Design: [go/config-sync-reconciler-autoscaling](https://goto.google.com/config-sync-reconciler-autoscaling)

Bug: [b/289388701](https://buganizer.corp.google.com/issues/289388701)

Depends On:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/763
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/776
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/777
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/778
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/780
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/872
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/876

Notes:
- The helm-sync container doesn't seem to handle OOMKills very well. The helm CLI is being executed and exits with a killed message, but it doesn't seem to recover and scale up fast enough to avoid e2e test timeout.

PR #763, opened 2023-07-25
Add configsync_sync_generation metric resource tag

karlkfi:
- Add `configsync.gke.io/sync-generation` reconciler deployment/pod label, populated by the reconciler-manager
- Add `CONFIGSYNC_SYNC_GENERATION` env var on the otel-agent container, populated by the k8s downward API from the pod label
- Add `configsync.sync.generation` metric resource attribute on the otel-agent config, populated from the env var
- Add attribute filter to delete the `configsync.sync.generation` label in the otel-collector config, when sending to Monarch
- Change the e2e tests to use the sync metric labels, instead of the pod name. This allows the metrics validation to tolerate pod replacement due to rescheduling, oomkill, or autoscaling

This PR blocks VPA: https://github.com/GoogleContainerTools/kpt-config-sync/pull/691

### Motivation
The e2e tests have been using pod name + source commit to filter metrics queries. But pod name was really just a proxy for RSync generation, so that we could match expected metrics with the most recent RSync spec and Git/OCI/Helm source changes. However, it turns out using pod name is still occasionally flaky, because the pod might be replaced or rescheduled even when the RSync spec didn't change (e.g. node replacement, rescheduling, oomkill, or autoscaling). Using RSync generation instead of pod name makes the e2e tests less flaky and makes it possible to add VPA and memory limits, which would otherwise cause most e2e tests to fail metrics validation.

Note: `configsync.sync.generation` is a metric resource attribute, NOT a normal metric label. So it will not be sent to either Monarch or Cloud Monitoring, because those providers do not support custom resource attributes without hard-coded support and a custom resource type registered with Google. However, custom resources attributes ARE being sent to Prometheus, because of the `resource_to_telemetry_conversion: { enabled: true }` setting, which converts them to metric labels. So `configsync.sync.generation` becomes `configsync_sync_generation` in Prometheus metrics. This is how the e2e tests use them for query filtering.

### DIFF
--- a/pkg/metrics/otel.go
+++ b/pkg/metrics/otel.go
@@ -159,6 +159,8 @@ processors:
         action: delete
       - key: configsync.sync.namespace
         action: delete
+      - key: configsync.sync.generation
+        action: delete
       # Remove high cardinality configsync metric labels when sending to Monarch.
       # These labels are useful to users, but too noisy for global aggregation.
       - key: commit
--- a/pkg/metrics/tagkeys.go
+++ b/pkg/metrics/tagkeys.go
@@ -73,7 +73,7 @@ var (
 	// TODO: replace with k8s.container.name resource attribute
 	KeyContainer, _ = tag.NewKey("container")
 
-	// KeyResourceType groups metris by their resource types. Possible values: cpu, memory.
+	// KeyResourceType groups metrics by their resource types. Possible values: cpu, memory.
 	KeyResourceType, _ = tag.NewKey("resource")
 )
 
@@ -89,6 +89,10 @@ var (
 
 	// ResourceKeySyncNamespace groups metrics by the Sync namespace.
 	ResourceKeySyncNamespace, _ = tag.NewKey("configsync_sync_namespace")
+
+	// ResourceKeySyncGeneration groups metrics by the Sync metadata.generation.
+	// This allows matching metrics to a specific Sync configuration.
+	ResourceKeySyncGeneration, _ = tag.NewKey("configsync_sync_generation")
 
 	// ResourceKeyDeploymentName groups metrics by k8s deployment name.
 	ResourceKeyDeploymentName, _ = tag.NewKey("k8s_deployment_name")
--- a/e2e/nomostest/prometheus_metrics.go
+++ b/e2e/nomostest/prometheus_metrics.go
@@ -24,12 +24,13 @@ import (
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	prometheusmodel "github.com/prometheus/common/model"
 	"go.uber.org/multierr"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	testmetrics "kpt.dev/configsync/e2e/nomostest/metrics"
 	"kpt.dev/configsync/e2e/nomostest/retry"
 	"kpt.dev/configsync/pkg/api/configmanagement"
 	"kpt.dev/configsync/pkg/api/configsync"
-	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metrics"
 	ocmetrics "kpt.dev/configsync/pkg/metrics"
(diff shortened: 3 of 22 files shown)


karlkfi:
We could aggregate generation to hide it from GCM, but I'm waiting for https://github.com/GoogleContainerTools/kpt-config-sync/pull/748 to implement that for `type` first. I think we can merge this and add that aggregation later.

pkg/metrics/otel.go
```diff
@@ -159,6 +159,8 @@ processors:
         action: delete
       - key: configsync.sync.namespace
         action: delete
+      - key: configsync.sync.generation
```
tiffanny29631, line 162: This only removes the `generation` tag from the Monarch pipeline. The e2e test for otel-collector is still in progress, currently you need to build and install Config Sync, let the metrics flow and check if deleting it would cause duplicate time series error in the batch.

pkg/metrics/tagkeys.go
```diff
@@ -90,6 +90,10 @@ var (
 	// ResourceKeySyncNamespace groups metrics by the Sync namespace.
 	ResourceKeySyncNamespace, _ = tag.NewKey("configsync_sync_namespace")
 
+	// ResourceKeySyncGeneration groups metrics by the Sync metadata.generation.
+	// This allows matching metrics to a specific Sync configuration.
```
tiffanny29631, line 94: Add a brief description of the possible values of this label, so we can have the estimation of the cardinality.

pkg/metrics/otel.go
tiffanny29631: I also suggest adding an attribute processor for GCM pipeline, to temporarily remove this tag from GCM until we have better assessment of the impact. 

Some of the customers are having pretty frequent commit pushes that brings up the cardinality (on record ~300 commits within 30 minutes). IIUC the `generation` is not overlapping with label `commit`, including it in GCM will bump it even higher.

pkg/metrics/tagkeys.go
karlkfi: It's an int32 that increases any time the sync spec is updated. But since it's included in the reconciler pod spec, it doesn't increase cardinality any more than pod name already does.

pkg/metrics/otel.go
karlkfi: I'm waiting for you to merge https://github.com/GoogleContainerTools/kpt-config-sync/pull/748 before I add an aggregation to filter `generation`. The delete is consistent with the previous way of doing it. We can add aggregation as a follow up PR after feature freeze, but before code freeze, or in 1./16.1 if your PR is delayed further.

Yes, generation and commit are non-overlapping, but sync_generation fully overlaps with pod_name. For every pod_name there will be only one sync_generation value. In fact, pod_name is higher cardinality than sync_generation, because the pod can be replaced even when the sync generation has not changed. So adding sync_generation should not increase the cardinality, even if you aggregate out type and commit.

karlkfi: It occurred to me that these are custom resource attributes, not normal metrics labels. They're being added with OTEL_RESOURCE_ATTRIBUTES on the otel-agent. I'm pretty sure the provider for both monarch and cloud monitoring will just ignore them, because they're not registered with the server and not hard-coded into the provider.

So they will only show up in prometheus, because we have `resource_to_telemetry_conversion: { enabled: true }`, which converts all resource attributes into labels.

karlkfi: I've removed the config in the otel-agent to inject the generation as a metric label. So now it's only being added as a resource attribute, which are ignored by Monarch and Cloud Monitoring otel providers. So the generation should only show up as a label in Prometheus.

tiffanny29631: Not adding this tag in otel-agent attribute processor resolves the concern, confirmed label is not carried into GCM or Monarch.

pkg/metrics/tagkeys.go
tiffanny29631: Ack, thanks for explanation.

tiffanny29631 (approved):
/lgtm

janetkuo:
@karlkfi would you update https://cloud.google.com/anthos-config-management/docs/reference/labels-and-annotations to include `configsync.gke.io/sync-generation`?

PR #764, opened 2023-07-25
Update the vulnerabitility scan script to check all fixable CVEs

nan-yu:
b/293187300


karlkfi (approved):
/lgtm

nomos-operator likely needs a CL too.

PR #766, opened 2023-07-26
chore: bump kpt, cli-utils, and k8s

sdowell:
The primary intention here is to upgrade cli-utils to pick up some
recent fixes; however this requires also bumping kpt and k8s libs.

The k8s upgrade required re-running the auto-generated conversion and
updating the fake client interfaces.


sdowell:
Broke out https://github.com/GoogleContainerTools/kpt-config-sync/pull/765 to be reviewed separately, which this change is based on

cmd/nomos/util/configmanagement.go
```diff
@@ -49,7 +49,9 @@ const (
 )
 
 // DynamicClient obtains a client based on the supplied REST config.  Can be overridden in tests.
-var DynamicClient = dynamic.NewForConfig
+var DynamicClient = func(config *rest.Config) (dynamic.Interface, error) {
```
karlkfi, line 52: this doesn't change the signature... why is it necessary?

pkg/syncer/syncertest/fake/client.go
```diff
@@ -214,3 +219,8 @@ func (c *Client) RESTMapper() meta.RESTMapper {
 func (c *Client) Storage() *MemoryStorage {
 	return c.storage
 }
+
+// SubResource implements client.Client
+func (c *Client) SubResource(_ string) client.SubResourceClient {
+	panic("SubResource not implemented for FakeClient")
```
karlkfi, line 225: this should be trivial to impl. it's the same as Status() just without a hard coded subresource name.

pkg/syncer/syncertest/fake/error_client.go
```diff
@@ -81,3 +86,30 @@ func (e ErrorClient) Scheme() *runtime.Scheme {
 func (e ErrorClient) RESTMapper() meta.RESTMapper {
 	panic("fake.ErrorClient does not support RESTMapper()")
 }
+
+// SubResourceErrorClient is a StatusWriter that always returns a specified error.
+type SubResourceErrorClient struct {
```
karlkfi, line 91: This shouldn't be necessary. You can just rename the StatusWriter to be a SubResourceWriter and use it for all subresources.

cmd/nomos/util/configmanagement.go
sdowell: It's subtle but it does change the signature. `dynamic.NewForConfig` has the following signature:
```
(inConfig *rest.Config) (*DynamicClient, error)
```
ref: https://github.com/kubernetes/client-go/blob/master/dynamic/simple.go#L71

This changes the signature to use `dynamic.Interface` (the `interface` rather than the impl) for compatibility with the unit tests that override this value with the `fake.NewSimpleDynamicClient`.

ref: https://github.com/kubernetes/client-go/blob/master/dynamic/fake/simple.go#L37

pkg/syncer/syncertest/fake/subresource_storage.go
```diff
@@ -44,11 +44,11 @@ func (ss *SubresourceStorage) setSubresourceInterface(uObj *unstructured.Unstruc
 }
 
 // Update the sub-resource field. All other fields are ignored.
-func (ss *SubresourceStorage) Update(ctx context.Context, obj client.Object, opts *client.UpdateOptions) error {
+func (ss *SubresourceStorage) Update(ctx context.Context, obj client.Object, opts *client.SubResourceUpdateOptions) error {
 	ss.Storage.lock.Lock()
 	defer ss.Storage.lock.Unlock()
 
-	err := ss.Storage.validateUpdateOptions(opts)
+	err := ss.Storage.validateUpdateOptions(&opts.UpdateOptions)
```
karlkfi, line 51: Might be slightly cleaner to have a Storage.validateSubresourceUpdateOptions that takes the whole opts and handles delegating the validation of UpdateOptions to validateUpdateOptions

pkg/syncer/syncertest/fake/error_client.go
sdowell: I updated the comment here but it looks like github is slow to update. This actually implements `SubResourceClient` (writer+reader) which has a different signature than `StatusWriter`. The interface of `client.SubResourceClient` is different from `client.Client`, which is what `ErrorClient` currently implements.

pkg/syncer/syncertest/fake/client.go
sdowell: Was hoping I could avoid fully refactoring all of these interfaces since these aren't used currently, but I guess I'll take a stab at it. Probably not difficult to do but properly refactoring this will be a larger change.

sdowell: Refactored

pkg/syncer/syncertest/fake/subresource_storage.go
sdowell: done

sdowell:

unit test flake

cmd/nomos/util/configmanagement.go
karlkfi: oi. ok. add a comment explaining.

pkg/syncer/syncertest/fake/error_client.go
karlkfi: oh, i see. this isn't the primary impl, just one that returns a specific error for testing. I guess that makes sense. It could be much more complex, but it's probably fine for now.

karlkfi (commented):
Looks good. Just need rebase after this merges: https://github.com/GoogleContainerTools/kpt-config-sync/pull/765

sdowell:
@karlkfi rebased

karlkfi (approved):
/lgtm

PR #768, opened 2023-07-27
fix: wait for deleted namespaces in Reset

sdowell:
The stress tests have been failing due to the Reset code not being able to clean up the large number of resources with the finalizer. With the recent fix in cli-utils the finalizer is now properly deleting all of the resources, but it is still possible that the namespaces have not been garbage collected at the end of the tests. This change adds a wait for deleted namespaces to be garbage collected instead of erroring.


sdowell:
/hold
This change may be replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/769
fyi @karlkfi

PR #769, opened 2023-07-27
fix: reconciler fail/timeout as error for destroy

sdowell:
This change updates the WaitEvent processing to treat a reconcile failure or timeout as an error for Destroyer. This ensures that the finalizer waits until all objects are NotFound before exiting.

The behavior is left unchanged for the Applier.

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -212,7 +212,14 @@ func wrapInventoryObj(obj *unstructured.Unstructured) (*live.InventoryResourceGr
 	return inv, nil
 }
 
-func processApplyEvent(ctx context.Context, e event.ApplyEvent, s *stats.ApplyEventStats, objectStatusMap ObjectStatusMap, unknownTypeResources map[core.ID]struct{}) status.Error {
+type eventHandler struct {
+	// isDestroy indicates whether the events being processed are for a Destroy
+	isDestroy bool
+	// clientSet is used for accessing various k8s clients
+	clientSet *ClientSet
+}
+
+func (h *eventHandler) processApplyEvent(ctx context.Context, e event.ApplyEvent, s *stats.ApplyEventStats, objectStatusMap ObjectStatusMap, unknownTypeResources map[core.ID]struct{}) status.Error {
 	id := idFrom(e.Identifier)
 	s.Add(e.Status)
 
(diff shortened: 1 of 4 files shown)


pkg/applier/applier.go
```diff
@@ -560,7 +568,7 @@ func (a *supervisor) applyInner(ctx context.Context, objs []client.Object) (map[
 			} else {
 				klog.V(1).Info(e.WaitEvent)
 			}
-			a.addError(processWaitEvent(e.WaitEvent, s.WaitEvent, objStatusMap))
+			a.addError(processWaitEvent(e.WaitEvent, s.WaitEvent, objStatusMap, false))
```
karlkfi, line 571: Passing bools as function args is usually a smell, IMO. It makes it unclear what the bool means when you call them, without reading the func def. Maybe convert processWaitEvent into a struct method?

sdowell: fair point, I wasn't super pleased with this either. I'll refactor

sdowell: refactored

pkg/applier/applier_test.go
```diff
@@ -573,29 +573,84 @@ func TestProcessWaitEvent(t *testing.T) {
...
-	expectedObjStatusMap := ObjectStatusMap{
-		idFrom(deploymentID): {
-			Reconcile: actuation.ReconcileFailed,
+	type eventStatus struct {
+		status         event.WaitEventStatus
+		id             *object.ObjMetadata
+		expectedErr    error
+		expectedStatus actuation.ReconcileStatus
+	}
+	testCases := []struct {
```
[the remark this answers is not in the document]
sdowell: added a test case

pkg/applier/applier.go
```diff
@@ -254,7 +254,12 @@ func processApplyEvent(ctx context.Context, e event.ApplyEvent, s *stats.ApplyEv
 	}
 }
 
-func processWaitEvent(e event.WaitEvent, s *stats.WaitEventStats, objectStatusMap ObjectStatusMap) error {
+type waitEventProcessor struct {
```
karlkfi, line 257: I don't love how this struct is new and unique for this method... other similar methods are on the supervisor directly, which won't work for this method, because it's specific to apply/destroy... 

What do you think about making an eventProcessor or eventHandler struct and putting all the handler methods on it? It would need to also hold the abandon method which needs the clientset. But it might make things a little cleaner. Then just make a new eventHandler every time apply or destroy is called.

karlkfi (changes requested):

pkg/applier/applier.go
sdowell: Fair point, done

karlkfi (approved):
/lgtm

Please add an e2e test in a follow up PR that validates that deletion propagation waits for all objects to reconcile. You can use a finalizer and a short reconcile timeout to simulate something being blocked deleting.

PR #773, opened 2023-07-28
[Terraform] Add e2e-metric-writer-sa to test infra

tiffanny29631:
Otel Collector e2e test requires IAM permission on WI enabled clusters.

This change sets up a Google Service Account with metricWriter role and binds it with the default Kubernetes Service Account under config-management-monitoring namespace.

b/255608560

### DIFF
--- a/e2e/testinfra/terraform/common/service_accounts.tf
+++ b/e2e/testinfra/terraform/common/service_accounts.tf
@@ -60,3 +60,21 @@ resource "google_project_iam_member" "gce-default-sa-iam" {
   member  = "serviceAccount:${data.google_compute_default_service_account.default.email}"
   project = data.google_project.project.id
 }
+
+resource "google_service_account" "e2e_metric_writer_sa" {
+  account_id = "e2e-test-metric-writer"
+  display_name = "Test Metric Writer"
+  description = "Service account used to write to Google Cloud Monitoring"
+}
+
+resource "google_project_iam_member" "e2e_metric_writer_gcp_role" {
+  role = "roles/monitoring.metricWriter"
+  member = "serviceAccount:${google_service_account.e2e_metric_writer_sa.email}"
+  project = data.google_project.project.id
+}
+
+resource "google_service_account_iam_member" "e2e_metric_writer_gke_binding" {
+  service_account_id = google_service_account.e2e_metric_writer_sa.name
+  role               = "roles/iam.workloadIdentityUser"
+  member             = "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-monitoring/default]"
+}
--- a/e2e/testinfra/terraform/common/services.tf
+++ b/e2e/testinfra/terraform/common/services.tf
@@ -23,6 +23,7 @@ resource "google_project_service" "services" {
     "secretmanager.googleapis.com",
     "container.googleapis.com",
     "compute.googleapis.com",
+    "monitoring.googleapis.com"
   ])
   service = each.value
   disable_on_destroy = false


e2e/testinfra/terraform/modules/service_account/service_account.tf
```diff
@@ -35,6 +35,12 @@ resource "google_service_account_iam_member" "k8s_sa_ns_binding" {
   member             = "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns]"
 }
 
+resource "google_service_account_iam_member" "k8s_sa_monitoring_binding" {
+  service_account_id = google_service_account.gcp_sa.name
+  role               = "roles/iam.workloadIdentityUser"
+  member             = "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-monitoring/default]"
```
sdowell, line 41: This is a different principal than the RootSync/RepoSync reconciler KSA. I don't think the intent is to create a binding for this principal to the other GSAs that use this module.

For this use case it might make sense to create a separate module or just create the binding directly since this is a one-off. The reusable module was defined for the other bindings since we needed to create multiple different bindings for the same KSA.

tiffanny29631: A one-off setup should be good enough, can it be moved to common/terraform.tf?

sdowell: `common/service_accounts.tf` would be a good place imo

e2e/testinfra/terraform/common/service_accounts.tf
```diff
@@ -38,6 +38,14 @@ module "e2e-gcr-reader-sa" {
   role = "roles/storage.objectViewer"
 }
 
+module "e2e-metric-writer-sa" {
+  source = "../modules/service_account"
+  gcp_sa_id = "e2e-test-metric-writer"
+  gcp_sa_display_name = "Test Metric Writer"
+  gcp_sa_description = "Service account used to write to Google Cloud Monitoring"
+  role = "roles/monitoring.metricWriter"
```
sdowell, line 46: What are the metrics going to be written to? Do any other GCP resources need to be provisioned?

scripts/set-up-workload-identity-configs.sh
```diff
@@ -72,6 +72,12 @@ gcloud iam service-accounts add-iam-policy-binding --project="${GCP_PROJECT}" \
   --member="serviceAccount:${FLEET_HOST_PROJECT}.svc.id.goog[config-management-system/root-reconciler]" \
   "e2e-test-gcr-reader@${GCP_PROJECT}.iam.gserviceaccount.com"
 
+echo "Granting the e2e-metric-writer service account the 'roles/iam.workloadIdentityUser' role for the '${FLEET_HOST_PROJECT}' project"
+gcloud iam service-accounts add-iam-policy-binding --project="${GCP_PROJECT}" \
```
sdowell, line 76: Why do we need a binding to the fleet project? Are you intending to test fleet integration?

e2e/testinfra/terraform/common/service_accounts.tf
tiffanny29631: They will be written to the cloud monitoring API, no other GCP resources are needed.

e2e/testinfra/terraform/modules/service_account/service_account.tf
tiffanny29631: Changed to a one-off setup in `common/service_accounts.tf`

e2e/testinfra/terraform/common/service_accounts.tf
sdowell: Does the API need to be enabled? If so, API enablement is currently handled here: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testinfra/terraform/common/services.tf

e2e/testinfra/terraform/modules/service_account/service_account.tf
tiffanny29631: Ah wait the role was left out

tiffanny29631: Added role.

scripts/set-up-workload-identity-configs.sh
tiffanny29631: I was adding it here for consistency, the test does not require fleet integration.

e2e/testinfra/terraform/common/service_accounts.tf
tiffanny29631: Good point, added monitoring service.

scripts/set-up-workload-identity-configs.sh
sdowell: I'm struggling to come up with a use case for this, I expect this isn't needed

e2e/testinfra/terraform/common/service_accounts.tf
```diff
@@ -60,3 +60,21 @@ resource "google_project_iam_member" "gce-default-sa-iam" {
...
   project = data.google_project.project.id
 }
+
+resource "google_service_account" "e2e_metric_writer_sa" {
+  account_id = "e2e-test-metric-writer"
+  display_name = "Test Metric Writer"
+  description = "Service account used to write to Google Cloud Monitoring"
+}
+
+resource "google_project_iam_member" "metric_writer" {
```
sdowell, line 70: Suggest naming this something like `e2e_metric_writer_gcp_role`

e2e/testinfra/terraform/common/service_accounts.tf
```diff
@@ -60,3 +60,21 @@ resource "google_project_iam_member" "gce-default-sa-iam" {
...
+  role = "roles/monitoring.metricWriter"
+  member = "serviceAccount:${google_service_account.e2e_metric_writer_sa.email}"
+  project = data.google_project.project.id
+}
+
+resource "google_service_account_iam_member" "iam_binding" {
```
sdowell, line 76: Suggest naming this something like `e2e_metric_writer_gke_binding`

e2e/testinfra/terraform/common/service_accounts.tf
```diff
@@ -60,3 +60,21 @@ resource "google_project_iam_member" "gce-default-sa-iam" {
...
+  service_account_id = google_service_account.e2e_metric_writer_sa.name
+  role               = "roles/iam.workloadIdentityUser"
+  member             = "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-monitoring/default]"
```
sdowell, line 79: Is it right that this is using `default` service account? Does we not create a service account for monitoring?

sdowell (commented):
Mostly LGTM, with a few comments on naming

e2e/testinfra/terraform/common/service_accounts.tf
tiffanny29631: Done

e2e/testinfra/terraform/common/service_accounts.tf
tiffanny29631: Done

e2e/testinfra/terraform/common/service_accounts.tf
tiffanny29631: It is correct, the process follows https://cloud.google.com/anthos-config-management/docs/how-to/monitor-config-sync-cloud-monitoring#custom-monitoring. `default` K8s Service Account  comes with GKE clusters by default.

sdowell (approved):

PR #774, opened 2023-07-28
test: add finalizer reconcile timeout e2e test

sdowell:
With recent changes to cli-utils and the finalizer, the finalizer will properly wait for managed objects to be NotFound before completing. This adds an e2e test to document/verify the behavior.

### DIFF
--- a/e2e/nomostest/testutils/validate.go
+++ b/e2e/nomostest/testutils/validate.go
@@ -20,6 +20,7 @@ import (
 	"github.com/pkg/errors"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/util/log"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ValidateError returns true if the specified errors contain an error
@@ -39,4 +40,16 @@ func ValidateError(errs []v1beta1.ConfigSyncError, code, message string) error {
 		return errors.Errorf("error %s not present with message %q: %s", code, message, log.AsJSON(errs))
 	}
 	return errors.Errorf("error %s not present: %s", code, log.AsJSON(errs))
+}
+
+// AppendFinalizer adds a finalizer to the object
+func AppendFinalizer(obj client.Object, finalizer string) {
+	finalizers := obj.GetFinalizers()
+	finalizers = append(finalizers, finalizer)
+	obj.SetFinalizers(finalizers)
+}
+
+// RemoveFinalizers removes all finalizers from the object
+func RemoveFinalizers(obj client.Object) {
+	obj.SetFinalizers(nil)
 }
--- a/e2e/testcases/reconciler_finalizer_test.go
+++ b/e2e/testcases/reconciler_finalizer_test.go
@@ -16,11 +16,14 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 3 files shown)


e2e/nomostest/testutils/validate.go
```diff
@@ -40,3 +41,15 @@ func ValidateError(errs []v1beta1.ConfigSyncError, code, message string) error {
 	}
 	return errors.Errorf("error %s not present: %s", code, log.AsJSON(errs))
 }
+
+// WithFinalizer adds a finalizer to the object
+func WithFinalizer(obj client.Object, finalizer string) {
```
karlkfi, line 46: this With- pattern is more for builder options... perhaps AppendFinalizer & RemoveFinalizers/ClearFinalizers/DeleteFinalizers?

e2e/nomostest/testutils/validate.go
```diff
@@ -40,3 +41,15 @@ func ValidateError(errs []v1beta1.ConfigSyncError, code, message string) error {
...
+	finalizers := obj.GetFinalizers()
+	finalizers = append(finalizers, finalizer)
+	obj.SetFinalizers(finalizers)
+}
+
+// WithoutFinalizers removes all finalizers from the object
+func WithoutFinalizers(obj client.Object) {
+	obj.SetFinalizers([]string{})
```
karlkfi, line 54: Just nil it out, IMO.

karlkfi: Or does that not work for SSA when you don't own the field?

e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -492,6 +498,88 @@ func TestReconcilerFinalizer_MultiLevelMixed(t *testing.T) {
...
+		nt.T.Fatal(err)
+	}
+	if err := nt.Validate(namespaceNN.Name, namespaceNN.Namespace, &corev1.Namespace{}); err != nil {
+		nt.T.Fatal(err)
+	}
+	// Remove the fake finalizer
+	namespace = fake.NamespaceObject(namespaceNN.Name)
+	testutils.WithoutFinalizers(namespace)
+	nt.T.Logf("Remove the fake finalizer named %s from Namespace %s", contrivedFinalizer, namespaceNN.Name)
+	if err := nt.KubeClient.Apply(namespace); err != nil {
```
karlkfi, line 565: SSA probably works here, but only because you're setting the finalizers to an empty list, instead of nil. If you set it to nil, you'd have to be using an Unstructured object to get it to pass the nil instead of skipping the field (due to omitempty on the Namespace struct field).

It might be clearer just to use a merge Patch with JSON. There's lots of examples in other tests.

karlkfi: Actually... with SSA, since you already appended a finalizer, it should remove it simply by not passing it. So it should be safe to just replace the empty list with a nil.

e2e/nomostest/testutils/validate.go
karlkfi: Since you're using SSA to append a finalizer, yes, nilling it out will remove it with SSA, even if it's just omitted.

e2e/nomostest/testutils/validate.go
sdowell: done

e2e/nomostest/testutils/validate.go
sdowell: done

karlkfi (approved):
/lgtm

PR #775, opened 2023-07-28
test: use HasGenerationAtLeast for autopilot

sdowell:
Using an exact generation predicate is very flaky on Autopilot, because the generation can often change due to resource adjustments. HasGenerationAtLeast is a less precise assertion, but passes more reliably on Autopilot.

Also replaces usage of nomostest.Wait with the more current Watcher.

### DIFF
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -255,6 +255,7 @@ func TestManagingReconciler(t *testing.T) {
 	if err != nil {
 		nt.T.Fatal(err)
 	}
+	generation = getDeploymentGeneration(nt, nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem)
 
 	// test case 3:  the reconciler-manager should not revert the change to the fields that are not owned by reconciler-manager
 	nt.T.Log("Manually update fields that are not owned by reconciler-manager")
@@ -275,6 +276,7 @@ func TestManagingReconciler(t *testing.T) {
 	if err != nil {
 		nt.T.Fatal(err)
 	}
+	generation = getDeploymentGeneration(nt, nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem)
 	// change the fields back to default values
 	mustUpdateRootReconciler(nt, func(d *appsv1.Deployment) {
 		d.Spec.Template.Spec.Containers[0].TerminationMessagePath = "dev/termination-log"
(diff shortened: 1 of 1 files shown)


e2e/testcases/reconciler_manager_test.go
```diff
@@ -766,9 +770,16 @@ func TestAutopilotReconcilerAdjustment(t *testing.T) {
 	expectedTotalMemory = initialTotalMemory
 
 	nt.T.Log("Validating initial container request value")
-	if err := nt.Validate(nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem, &appsv1.Deployment{},
-		hasGeneration(generation), totalContainerCPURequestIs(expectedTotalCPU),
-		totalContainerMemoryRequestIs(expectedTotalMemory), firstContainerCPURequestIs(expectedFirstContainerCPU)); err != nil {
+	err := nt.Watcher.WatchObject(kinds.Deployment(),
+		nomostest.DefaultRootReconcilerName, v1.NSConfigManagementSystem,
+		[]testpredicates.Predicate{
+			testpredicates.HasGenerationAtLeast(generation),
```
karlkfi, line 776: I've actually rewritten this test significantly in the VPA PR, but it still uses these generation things. It turns out using HasGenerationAtLeast is kind of pointless after the first one without re-getting the generation before each validation. Otherwise a single generation bump early in the test by Autopilot throws off all the subsequent generation checks.

karlkfi: Maybe add a `Get` after each `Watch` and set `generation = obj.GetGeneration()`

sdowell: SGTM, I'll update the generation between each watch. Still a weaker assertion than an exact generation check, but an exact generation check isn't feasible for autopilot

karlkfi: Wait... doesn't autopilot adjust with a mutating webhook? Why would that affect generation?

karlkfi: The override adjustment code in the reconciler-manager is really complex. I don't really know when it should perform a patch and when it shouldn't. But I assume that it should be predictable. So we should be able to just do `if autopilot { generation++ }` after each update, right?

sdowell: The spec is updated to make resource adjustments - does that not change generation? I don't have much context on how autopilot makes the adjustments.

I don't have another explanation for the current flakiness, e.g. https://testgrid.k8s.io/googleoss-kpt-config-sync-main#autopilot-regular

sdowell: Would be nice to get the flakiness improved before 1.16.0

karlkfi (approved):
/lgtm

PR #777, opened 2023-07-28
test: Clean up ReconcileTimeout config

karlkfi:
- Add WithoutReconcileTimeout to remove the setting, if needed.
- Move default ReconcileTimeout into the options constructor, so it can be overridden by test options.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -458,28 +458,6 @@ func noOOMKilledContainer(nt *NT) testpredicates.Predicate {
 	}
 }
 
-func setupRootSync(nt *NT, rsName string, reconcileTimeout *time.Duration) {
-	// create RootSync to initialize the root reconciler.
-	rs := RootSyncObjectV1Beta1FromRootRepo(nt, rsName)
-	if reconcileTimeout != nil {
-		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(*reconcileTimeout)
-	}
-	if err := nt.KubeClient.Apply(rs); err != nil {
-		nt.T.Fatal(err)
-	}
-}
-
-func setupRepoSync(nt *NT, nn types.NamespacedName, reconcileTimeout *time.Duration) {
-	// create RepoSync to initialize the Namespace reconciler.
-	rs := RepoSyncObjectV1Beta1FromNonRootRepo(nt, nn)
-	if reconcileTimeout != nil {
-		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(*reconcileTimeout)
-	}
-	if err := nt.KubeClient.Create(rs); err != nil {
-		nt.T.Fatal(err)
-	}
-}
-
 // RepoSyncRoleBinding returns rolebinding that grants service account
 // permission to manage resources in the namespace.
 func RepoSyncRoleBinding(nn types.NamespacedName) *rbacv1.RoleBinding {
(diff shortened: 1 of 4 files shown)


e2e/nomostest/nt.go
```diff
@@ -119,8 +119,8 @@ type NT struct {
 	DefaultWaitTimeout time.Duration
 
 	// DefaultReconcileTimeout is the default timeout for the applier to wait
-	// for object reconcilition.
-	DefaultReconcileTimeout time.Duration
+	// for object reconciliation.
+	DefaultReconcileTimeout *time.Duration
```
sdowell, line 123: Why make it a pointer? Is this to enable the `nil` check when it's unset?

karlkfi: Has to allow nil to unset it, to replace the default value.

karlkfi: I didn't add a test that does unset it, but I imagine it might be useful for the stress test, where 1m is too short, and we might as well stress test the default config.

e2e/nomostest/config_sync.go
```diff
@@ -699,8 +683,10 @@ func RootSyncObjectV1Alpha1FromRootRepo(nt *NT, name string) *v1alpha1.RootSync
 	repoURL := nt.GitProvider.SyncURL(repo.RemoteRepoName)
 	sourceFormat := repo.Format
 	rs := RootSyncObjectV1Alpha1(name, repoURL, sourceFormat)
-	if nt.DefaultReconcileTimeout != 0 {
-		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(nt.DefaultReconcileTimeout)
+	if nt.DefaultReconcileTimeout != nil {
+		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(*nt.DefaultReconcileTimeout)
+	} else if rs.Spec.Override != nil {
+		rs.Spec.Override.ReconcileTimeout = nil
```
sdowell, line 689: I feel like this might be better form if we get rid of the `rs.Spec.Override != nil` check and do:
```golang
	} else {
		rs.Spec.SafeOverride().ReconcileTimeout = nil
	}
```
WDYT?

karlkfi: I did it this way to avoid populating the override if I don't have to.
The way it is now, if override is nil o missing, it doesn't replace that with `override = { }`.

karlkfi: For now, since the default is 1m, that case will never happen. but if we changed the default to nil, it would.

e2e/nomostest/config_sync.go
```diff
@@ -699,8 +683,10 @@ func RootSyncObjectV1Alpha1FromRootRepo(nt *NT, name string) *v1alpha1.RootSync
 	repoURL := nt.GitProvider.SyncURL(repo.RemoteRepoName)
 	sourceFormat := repo.Format
 	rs := RootSyncObjectV1Alpha1(name, repoURL, sourceFormat)
-	if nt.DefaultReconcileTimeout != 0 {
-		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(nt.DefaultReconcileTimeout)
+	if nt.DefaultReconcileTimeout != nil {
```
sdowell, line 686: Should these defaults be moved into the `R*SyncObjectV1Alpha`/`R*SyncObjectV1Beta` to reduce duplication?

karlkfi: Those don't have access to NT.
I have considered making them depend on NT to force the defaults into tests that use them directly, but that's a bigger refactor.

sdowell (approved):

PR #780, opened 2023-07-29
Fix reconciler-manager cmd to accept log flags

karlkfi:
Previously, -v=# and other log flags were not being registered. Not it sets up log flags the sam as other component binaries.

### DIFF
--- a/cmd/reconciler-manager/main.go
+++ b/cmd/reconciler-manager/main.go
@@ -33,6 +33,7 @@ import (
 	"kpt.dev/configsync/pkg/profiler"
 	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
+	"kpt.dev/configsync/pkg/util/log"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	// +kubebuilder:scaffold:imports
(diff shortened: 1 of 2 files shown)


cmd/reconciler-manager/main.go
```diff
@@ -70,8 +71,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.Parse()
 
+	log.Setup()
```
sdowell, line 75: It looks like the `init` function above hard codes the `v` flag to `1`. I expect we need to remove that so that we use the actual verbosity flag that is passed in to the binary

karlkfi: good catch. I guess I'll just set it to -v=1 in the manifest

sdowell: IIRC it doesn't need to be set in the manifest any more, it gets set in code now with the new logLevel api (getting a default value from code)

karlkfi: nope. that's just for reconciler containers.

sdowell: oh I guess this is reconciler-manager, nevermind

karlkfi: done

sdowell (approved):

PR #781, opened 2023-07-29
test: install kind at runtime

sdowell:
Assuming that kind is pre-installed on the prow image makes it very difficult to upgrade the kind version, because the kind version of the binary must be the same as the go import. Installing it at runtime makes it easier to change the version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,9 @@ GO_DIR := $(OUTPUT_DIR)/go
 BIN_DIR := $(GO_DIR)/bin
 KUSTOMIZE_VERSION := v5.1.0
 HELM_VERSION := v3.12.2
+# Keep KIND_VERSION in sync with the version defined in go.mod
+# When upgrading, update the node image versions at e2e/nomostest/clusters/kind.go
+KIND_VERSION := v0.14.0
 
 # Directory used for staging Docker contexts.
 STAGING_DIR := $(OUTPUT_DIR)/staging
@@ -300,6 +303,15 @@ lint-license: pull-buildenv buildenv-dirs
 
 .PHONY: install-kustomize
 install-kustomize: "$(BIN_DIR)/kustomize"
+
+"$(GOBIN)/kind":
+	go install sigs.k8s.io/kind@$(KIND_VERSION)
+
+"$(BIN_DIR)/kind": "$(GOBIN)/kind" buildenv-dirs
+	cp $(GOBIN)/kind $(BIN_DIR)/kind
+
+.PHONY: install-kind
+install-kind: "$(BIN_DIR)/kind"
 
 .PHONY: license-headers
 license-headers: "$(GOBIN)/addlicense"
--- a/Makefile.build
+++ b/Makefile.build
@@ -247,8 +247,8 @@ config-sync-manifest: config-sync-manifest-no-push push-images
 # config-sync-manifest-local builds config sync for local testing in kind.
 # starts local docker registry and pushes images to the local registry
 .PHONY: config-sync-manifest-local
-config-sync-manifest-local:
-	@bash scripts/docker-registry.sh
+config-sync-manifest-local: install-kind
+	@KIND_VERSION=$(KIND_VERSION) bash scripts/docker-registry.sh
 	$(MAKE) config-sync-manifest \
 		REGISTRY=localhost:5000 \
 		IMAGE_TAG=$(IMAGE_TAG)
--- a/Makefile.e2e
+++ b/Makefile.e2e
@@ -9,6 +9,8 @@ GOTOPT2_BINARY := docker run -i -u $(UID):$(GID) $(BUILDENV_IMAGE) /bin/gotopt2
 
 GCP_ZONE ?= us-central1-a
 GKE_E2E_TIMEOUT ?= 6h
+KIND_E2E_TIMEOUT ?= 60m
+
 E2E_CREATE_CLUSTERS ?= "true"
 E2E_DESTROY_CLUSTERS ?= "true"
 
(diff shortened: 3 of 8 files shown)


Makefile
```diff
@@ -301,6 +304,16 @@ lint-license: pull-buildenv buildenv-dirs
...
 
+"$(GOBIN)/kind":
+	go install sigs.k8s.io/kind@$(KIND_VERSION)
+
+# install kind binary for containerized testing
+"$(BIN_DIR)/kind": "$(GOBIN)/kind" buildenv-dirs
+	cp $(GOBIN)/kind $(BIN_DIR)/kind
+
+.PHONY: install-kind
+install-kind: "$(BIN_DIR)/kind"
```
karlkfi, line 315: this target seems like it might be confusing if used locally. how about just specifying one of the above?

Makefile.build
```diff
@@ -247,8 +247,8 @@ config-sync-manifest: config-sync-manifest-no-push push-images
 # config-sync-manifest-local builds config sync for local testing in kind.
 # starts local docker registry and pushes images to the local registry
 .PHONY: config-sync-manifest-local
-config-sync-manifest-local:
-	@bash scripts/docker-registry.sh
+config-sync-manifest-local: install-kind
```
karlkfi, line 250: i use this target all the time locally. do i really want it to install kind for CI?

Makefile
karlkfi: Alternatively: make install-kind (GOBIN) & install-kind-ci (BIN_DIR)

Makefile.build
karlkfi: config-sync-manifest-local-ci could depend on install-kind-ci, while config-sync-manifest-local could depend on install-kind.

karlkfi: or... is there a way to tell if we're in CI with an env var or something and use that to decide where to install kind?

sdowell: Main goal with this is to install kind in the CI job at runtime, so that it's possible to change the kind version and have presubmits pass.

With the preinstalled kind version, we basically have to:
- publish a new image with a bumped kind version
- update CI to use the new image, which will break our presubmits until we bump the kind version used in the repo
- create a PR that bumps the kind version, and submit to fix presubmits

This requires putting the CI in a broken state in the interim of an upgrade, which is undesirable. Using the dynamic kind version like this makes the upgrade look like:
- create a PR that bumps the kind version

That being said, I could move this as a dependency of a different target so that it isn't called every time you run `config-sync-manifest-local` locally.

Makefile
sdowell: This follows the pattern of other similar make targets that install local binaries. The `install-kind` target is a phony target to make it easier to call from the command line. The other targets aren't phony and point to the actual path of the binary

Makefile.build
sdowell: Moved the dependency to another make target (the e2e-kind entrypoint used by CI)

Makefile
karlkfi: I understand. My complaint is that this install target is a phony target for use by CI, not developers locally. 

If we only have one, I think it should detect CI or not CI. Otherwise, have two, one with a -ci suffix.

karlkfi: That said, I see kustomize has the same problem...

Makefile.build
sdowell: I don't think there's a need to treat CI/local differently in this case

Makefile
sdowell: The intent of this make target is to also be used by developers locally

sdowell: The reason it gets installed both in `GOBIN` and `BIN_DIR` is for legacy reasons at this point, not necessarily for local vs CI use case. Some of the make targets for running tests used to mount `BIN_DIR` as a path when running the targets inside a docker container - this would be the same locally and in CI.

sdowell: I updated this to add a separate CI target, although I don't really feel like this is necessary. The same `install-kind` target should be usable in both CI and local use cases.

I suppose with how finicky `kind` can be with changing versions, it's okay to force the developer to separately call `install-kind`. I was thinking it would be a convenience to have the testing targets ensure `kind` is installed before running the tests

karlkfi: I think the other make targets all use `"$(BIN_DIR)/kustomize"` when executing kustomize commands, which uses what's installed in the the output dir, regardless of what's in your PATH.

But for kind, I think they all just call kind, which uses what's in your PATH, regardless of what's in your output dir.

For kustomize, I almost never need to run it manually. I just use the make targets to build the manifests.

But for kind, I use the kind command in my path all the time, to list clusters, delete clusters, and export the kubeconfig. So I'd much rather use the kind in my PATH, rather than have to specify the exact path to the kind binary in the output dir...

So it only makes sense to treat kind and kustomize the same if I never have to call kind directly, which isn't the case today.

sdowell: All of the make target entrypoints will use the `kind` binary in BIN_DIR, because it's [injected in the PATH](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/Makefile#L175). 

Why do you need to specify the exact for the kind binary in the output dir? For local command line I do what's described in the [dev docs](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/docs/development.md#e2e-tests-kind), and put `GOPATH/bin` in my `PATH`. `install-kind` puts the kind binary in both `GOPATH/bin` and `.output/go/bin`

karlkfi: Ah, I see it's installing to GOBIN and then copying to BIN_DIR. 

I assumed it was installing to BIN_DIR using a modified GOPATH, which is how it originally worked, IIRC.

It looks like it's all be changed and my understanding of how it works is out of date.

So it's probably fine. Sorry for the confusion.

karlkfi: You can move install-kind back to be a dependency of config-sync-manifest-local.
If it's installing in GOBIN and copying to BIN_DIR it should work fine locally and in CI.

sdowell: No worries, moved it back to `config-sync-manifest-local`

karlkfi (approved):
/lgtm

PR #782, opened 2023-07-30
test: fix memory request predicate

sdowell:
Reassigning the value in the nested function causes the value to keep increasing on each invocation of the predicate.


e2e/testcases/reconciler_manager_test.go
```diff
@@ -705,11 +705,11 @@ func firstContainerCPURequestIs(cpuRequest int64) testpredicates.Predicate {
 }
 
 func firstContainerMemoryRequestIs(memoryRequest int64) testpredicates.Predicate {
+	memoryRequest *= memoryMB
```
karlkfi, line 708: This is fine, but I rewrote these in the VPA PR.

karlkfi (approved):
/lgtm

PR #783, opened 2023-07-31
fix: Increase otel-agent CPU request to 20m

karlkfi:
The limits are being set to request values on autopilot, which has been causing the otel-agent to fail liveness probes and get restarted. This increase should help prevent that.

### DIFF
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -51,7 +51,7 @@ import (
 const (
 	initialFirstCPU                    = 10
 	initialFirstMemory                 = 100
-	initialTotalCPU                    = 80
+	initialTotalCPU                    = 90
 	initialAdjustedTotalCPU            = 250
 	initialTotalMemory                 = 600
 	autopilotCPUIncrements             = 250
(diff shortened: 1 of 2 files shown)


janetkuo:
The cpu request hasn't changed for a few releases. Did we introduce something new in this release that causes the need to increase cpu request/limit on Autopilot?

karlkfi:
> The cpu request hasn't changed for a few releases. Did we introduce something new in this release that causes the need to increase cpu request/limit on Autopilot?

I don't know. All I know is that it's failing on autopilot-rapid, because it's setting CPULimit = CPURequest.

karlkfi:
This PR may have increased resource usage of the otel-agent somewhat: https://github.com/GoogleContainerTools/kpt-config-sync/pull/763

The generation metrics label isn't going to Cloud Monitoring, but it does go through the otel-agent and otel-collector.

janetkuo:
Given that this number will multiply by the number of RSyncs in the cluster, I'd like to avoid increasing it if possible (or keep it to a minimum). Here's our current number: https://cloud.google.com/anthos-config-management/docs/how-to/installing-config-sync#detailed_resource_requests

karlkfi:
I think risking breakage to keep requirements low is a bad trade-off, but now that the liveness probes are removed, it at least won't cause endless restarts.

PR #785, opened 2023-07-31
test: remove test finalizer in cleanup

sdowell:
If an error occurs mid-test, it is possible that the finalizer removal is not reached and the object cannot be deleted. This adds a test-specific cleanup and generic suite cleanup for removing the test finalizer.

### DIFF
--- a/e2e/nomostest/clean.go
+++ b/e2e/nomostest/clean.go
@@ -49,6 +49,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// ConfigSyncE2EFinalizer is a contrived finalizer to block deletion of
+	// objects created by the e2e tests.
+	ConfigSyncE2EFinalizer = "config-sync/e2e-test"
+)
+
 // Clean removes all objects of types registered in the Scheme, with the above
 // caveats. It should be run before and after a test is run against any
 // non-ephemeral cluster.
--- a/e2e/testcases/reconciler_finalizer_test.go
+++ b/e2e/testcases/reconciler_finalizer_test.go
@@ -522,8 +522,26 @@ func TestReconcileFinalizerReconcileTimeout(t *testing.T) {
 	if err := nt.WatchForAllSyncs(); err != nil {
 		nt.T.Fatal(err)
 	}
+	nt.T.Cleanup(func() {
+		namespace := &corev1.Namespace{}
+		if err := nt.KubeClient.Get(namespaceNN.Name, namespaceNN.Namespace, namespace); err != nil {
+			if apierrors.IsNotFound(err) { // Happy path - exit
+				return
+			}
+			nt.T.Error(err) // unexpected error
+			return
+		}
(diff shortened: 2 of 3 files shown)


e2e/nomostest/clean.go
```diff
@@ -383,6 +390,13 @@ func DeleteObjectsAndWait(nt *NT, objs ...client.Object) error {
 		if err != nil {
 			return err
 		}
+		if testutils.RemoveFinalizer(obj, ConfigSyncE2EFinalizer) {
```
karlkfi, line 393: I don't know if we want it here in the generic delete func...
Having it in the test cleanup should be good enough to prevent it from leaking again.

karlkfi: If you really want to add it to the general cleanup, I'd add a RemoveTestFinalizers func and call it before the deletes.

sdowell: Agreed it shouldn't be necessary here. My thinking was that having it in Clean as well would help with recovering previously wedged clusters, but clusters shouldn't get wedged any more with the test-specific cleanup.

I'm not too attached to the generic cleanup, so I'll remove

sdowell: Done

e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -522,8 +522,18 @@ func TestReconcileFinalizerReconcileTimeout(t *testing.T) {
 	if err := nt.WatchForAllSyncs(); err != nil {
 		nt.T.Fatal(err)
 	}
+	nt.T.Cleanup(func() {
+		// Remove the fake finalizer. This should be already handled in the happy
+		// path, but remove in case there was an error.
+		namespace := fake.NamespaceObject(namespaceNN.Name)
```
karlkfi, line 528: If the test succeeded, this would re-create the namespace. I don't think Apply returns NotFound. You could use Update tho. I don't think it needs SSA particularly.

sdowell: Ah good point, I'll fix this

sdowell: Done

e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -522,8 +522,26 @@ func TestReconcileFinalizerReconcileTimeout(t *testing.T) {
 	if err := nt.WatchForAllSyncs(); err != nil {
 		nt.T.Fatal(err)
 	}
+	nt.T.Cleanup(func() {
+		namespace := &corev1.Namespace{}
+		if err := nt.KubeClient.Get(namespaceNN.Name, namespaceNN.Namespace, namespace); err != nil {
+			if apierrors.IsNotFound(err) { // Happy path - exit
+				return
+			}
+			nt.T.Error(err) // unexpected error
```
karlkfi, line 531: this is fine, but fwiw, its safe to use Fatal in Cleanup. The caller catches panics and makes sure other cleanups get called, plus then you don't need the return.

karlkfi (approved):
/lgtm

PR #788, opened 2023-08-01
fix: Remove otel agent & collector liveness probe

karlkfi:
This should prevent the otel-agent and otel-collector from being restarted due to CPU throttling.

### DIFF
--- a/manifests/templates/otel-collector.yaml
+++ b/manifests/templates/otel-collector.yaml
@@ -152,10 +152,6 @@ spec:
         volumeMounts:
         - name: otel-collector-config-vol
           mountPath: /conf
-        livenessProbe:
-          httpGet:
-            path: /
-            port: 13133 # Health Check extension default port.
         readinessProbe:
           httpGet:
             path: /
--- a/manifests/third_party/resourcegroup-manifest.yaml
+++ b/manifests/third_party/resourcegroup-manifest.yaml
@@ -556,10 +556,6 @@ spec:
         - name: OTEL_RESOURCE_ATTRIBUTES
           value: k8s.pod.name=$(KUBE_POD_NAME),k8s.pod.namespace=$(KUBE_POD_NAMESPACE),k8s.pod.uid=$(KUBE_POD_UID),k8s.pod.ip=$(KUBE_POD_IP),k8s.node.name=$(KUBE_NODE_NAME),k8s.deployment.name=$(KUBE_DEPLOYMENT_NAME)
         image: gcr.io/config-management-release/otelcontribcol:v0.54.0-gke.1
-        livenessProbe:
-          httpGet:
-            path: /
-            port: 13133
         name: otel-agent
         ports:
         - containerPort: 55678
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -195,11 +195,6 @@ data:
            volumeMounts:
            - name: otel-agent-config-vol
              mountPath: /conf
-           livenessProbe:
-             httpGet:
-               path: /
-               port: 13133 # Health Check extension default port.
-               scheme: HTTP
            readinessProbe:
              httpGet:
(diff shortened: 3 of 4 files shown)


manifests/third_party/resourcegroup-manifest.yaml
sdowell, line 1: This manifest comes from upstream, so we should probably make this update in the upstream repo and cut a new release.

karlkfi: https://gke-internal-review.git.corp.google.com/c/GoogleCloudPlatform/resource-group/+/775501

karlkfi: I don't think pushing a new release is necessary for just a manifest change.

manifests/templates/otel-collector.yaml
```diff
@@ -152,10 +152,6 @@ spec:
         volumeMounts:
         - name: otel-collector-config-vol
           mountPath: /conf
-        livenessProbe:
```
tiffanny29631, line 155: This one is the collector in monitoring namespace, we could keep the liveness probe here as a deadman switch? I just realized it's given 200m CPU, whereas the otel-agent that now uses same image as otel-collector is only given 10m as a container in reconciler.

manifests/third_party/resourcegroup-manifest.yaml
```diff
@@ -556,10 +556,6 @@ spec:
         - name: OTEL_RESOURCE_ATTRIBUTES
           value: k8s.pod.name=$(KUBE_POD_NAME),k8s.pod.namespace=$(KUBE_POD_NAMESPACE),k8s.pod.uid=$(KUBE_POD_UID),k8s.pod.ip=$(KUBE_POD_IP),k8s.node.name=$(KUBE_NODE_NAME),k8s.deployment.name=$(KUBE_DEPLOYMENT_NAME)
         image: gcr.io/config-management-release/otelcontribcol:v0.54.0-gke.1
-        livenessProbe:
```
tiffanny29631, line 559: Need a change from resource group controller as well (or no?)

manifests/templates/otel-collector.yaml
karlkfi: I don't think it's necessary here either. Like I said in chat, livenessProbe is really only recommended to be added to work around problems with apps you don't control that can get wedged and don't recover automatically.

manifests/third_party/resourcegroup-manifest.yaml
karlkfi: https://gke-internal-review.git.corp.google.com/c/GoogleCloudPlatform/resource-group/+/775501

tiffanny29631: Saw the CL, thanks.

manifests/third_party/resourcegroup-manifest.yaml
sdowell: The manifest comes directly from upstream, so IMO making manual changes just to the version in this repo will lead to confusion down the line

manifests/templates/otel-collector.yaml
tiffanny29631: 👌 sgtm.

tiffanny29631 (commented):
/lgtm

See if Janet has more input.

manifests/third_party/resourcegroup-manifest.yaml
karlkfi: The next time someone wants to update it, they're gonna need to cut a release first, right?

sdowell: Yeah. I'm okay if you don't want to cut a release, as long as the manifests are consistent at each HEAD

karlkfi: They are now. The above CL was merged.

janetkuo (approved):
/approve

PR #790, opened 2023-08-03
refactor: move the askpass declaration to manifest

sdowell:
The askpass sidecar is a container in the reconciler Pod akin to the other containers, but is treated differently and declared in code rather than the manifests. This inconsistency can be a gotcha and requires our other tooling to have special exceptions for parsing the image tag from code. This change makes the askpass sidecar consistent with the other reconciler containers.

Note this is intended as a non-functional refactor, so the overall behavior should be unchanged.

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -129,6 +129,18 @@ data:
              requests:
                cpu: "10m"
                memory: "200Mi"
+         - name: gcenode-askpass-sidecar
+           image: gcr.io/config-management-release/gcenode-askpass-sidecar:v1.0.3
+           args: ["--port=9102", "--logtostderr"]
+           imagePullPolicy: IfNotPresent
+           terminationMessagePolicy: File
+           terminationMessagePath: /dev/termination-log
+           securityContext:
+             allowPrivilegeEscalation: false
+             readOnlyRootFilesystem: false
+             capabilities:
+               drop:
+               - NET_RAW
          - name: oci-sync
            image: OCI_SYNC_IMAGE_NAME
            args: ["--root=/repo/source", "--dest=rev", "--max-sync-failures=30", "--error-file=error.json"]
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -886,6 +886,9 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 			noSSLVerify:     rs.Spec.Git.NoSSLVerify,
 			caCertSecretRef: v1beta1.GetSecretName(rs.Spec.Git.CACertSecretRef),
 		})
+		if enableAskpassSidecar(rs.Spec.SourceType, rs.Spec.Git.Auth) {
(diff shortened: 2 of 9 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -1131,6 +1134,14 @@ func (r *RepoSyncReconciler) mutationsFor(ctx context.Context, rs *v1beta1.RepoS
 					container.Env = append(container.Env, gitSyncHTTPSProxyEnv(secretName, keys)...)
 					mutateContainerResource(&container, rs.Spec.Override)
 				}
+			case reconcilermanager.GCENodeAskpassSidecar:
+				if !enableAskpassSidecar(rs.Spec.SourceType, auth) {
+					addContainer = false
+				} else {
+					container.Env = append(container.Env, containerEnvs[container.Name]...)
+					injectFWICredsToContainer(&container, injectFWICreds)
+					// TODO: enable resource/logLevel overrides for gcenode-askpass-sidecar
```
karlkfi, line 1143: want to do this TODO as part of this PR?

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -129,6 +129,18 @@ data:
              requests:
                cpu: "10m"
                memory: "200Mi"
+         - name: gcenode-askpass-sidecar
```
karlkfi, line 132: I wonder if there're any way we can use this process to delay git-sync errors until the askpass sidecar is up and ready... We don't have a good ordering solution yet for this and it can cause git-sync to error every time the reconciler pod its rescheduled.

pkg/reconcilermanager/controllers/reposync_controller.go
sdowell: I intentionally did not because that would be a functional change. I'd like for this PR to just be a refactor, and functional changes made as subsequent commits

manifests/templates/reconciler-manager-configmap.yaml
sdowell: I can think of a few mitigations but not a real solution to the problem.

For example we could have the reconciler container treat the askpass credential error as a transient error (potentially only during the first minute or so after startup), such that it gets logged but not surfaced on the RSync API.

I think a real solution would require changes to `git-sync` to enable waiting gracefully for the askpass endpoint to be ready.

tiffanny29631 (commented):
/lgtm

There might not be a good way to test this besides comparing the generated manifest before and after the change?

sdowell:
> There might not be a good way to test this besides comparing the generated manifest before and after the change?

Yeah the best way to test would probably be to compare a `reconciler` Deployment that includes the gcenode-askpass-sidecar before and after the change. For reviewing, If you compare the new template in the manifest against what was previously declared in code it should look the same

tiffanny29631:
> > There might not be a good way to test this besides comparing the generated manifest before and after the change?
> 
> Yeah the best way to test would probably be to compare a `reconciler` Deployment that includes the gcenode-askpass-sidecar before and after the change. For reviewing, If you compare the new template in the manifest against what was previously declared in code it should look the same

Cool. Can approve unless @karlkfi has more comment.

PR #791, opened 2023-08-07
test: reduce flakiness of reconcile timeout e2e

sdowell:
This test has been relatively flaky, particularly on autopilot. The condition of the flake appears to frequently coincide with a node replacement that causes Pod eviction. This presumably causes the reconciler Pod to get evicted and restarted, and by the time it comes back up the goproxy Pod is ready.

This extends the delay of the goproxy readinessProbe in order to reduce the chance of encountering this scenario on autopilot.

[Example flake](https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-rapid/1688646113445285888):
```
override_reconcile_timeout_test.go:106: ERROR: WatchObject(ResourceGroup config-management-system/root-sync): predicates not satisfied: object "pod-1" does not have expected actuation and reconcile status, actual actuation status is "Succeeded", expect "Succeeded"; actual reconcile status is "Succeeded", expect "Timeout": context deadline exceeded
```

### DIFF
--- a/e2e/testcases/override_reconcile_timeout_test.go
+++ b/e2e/testcases/override_reconcile_timeout_test.go
@@ -86,7 +86,7 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 			// Make initial delay longer than current reconcile timeout 30s.
 			// This will cause the applier to exit with a timeout,
 			// but the workload will still become available afterwards
-			InitialDelaySeconds: 60,
+			InitialDelaySeconds: 3600,
 			PeriodSeconds:       10,
 		},
 	}
@@ -115,9 +115,6 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 		nt.Watcher.WatchObject(kinds.RootSyncV1Beta1(), configsync.RootSyncName, configsync.ControllerNamespace, []testpredicates.Predicate{
 			testpredicates.RootSyncHasObservedGenerationNoLessThan(rootSync.Generation),
 		}))
-	if err := nt.WatchForAllSyncs(); err != nil {
-		nt.T.Fatal(err)
-	}
 
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Remove("acme/pod-1.yaml"))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush(fmt.Sprintf("Remove pod/%s", pod1Name)))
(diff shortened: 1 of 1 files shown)


e2e/testcases/override_reconcile_timeout_test.go
```diff
@@ -129,8 +126,9 @@ func TestOverrideReconcileTimeout(t *testing.T) {
 	if err := nt.ValidateNotFound(pod1Name, namespaceName, &corev1.Pod{}); err != nil {
 		nt.T.Fatal(err)
 	}
+	pod1.Spec.Containers[0].ReadinessProbe.InitialDelaySeconds = 60
```
nan-yu, line 129: Since the initial `InitialDelaySeconds` was 3600s, here we can use a shorter delay.

sdowell: I'm not sure the first assignment of `InitialDelaySeconds` is necessarily relevant, because the first set of assertions are actually waiting for `override.reconcileTimeout` for a reconcile timeout.

This set of assertions are made after the Pod is deleted and recreated. That being said, I agree that we could probably use a shorter delay here so the test isn't as long

sdowell: Reduced it to 30s readiness delay. This is a bit arbitrary, so let me know if you'd prefer a different value

nan-yu: I'm actually thinking if we can set `InitialDelaySeconds` to 0, but I guess it doesn't matter that much as we only wait for the Pod to be actuated, rather than reconciled.

nan-yu (approved):
/lgtm

e2e/testcases/override_reconcile_timeout_test.go
sdowell: The assertion on line 138 should still be waiting for the Pod to reconcile

nan-yu: Ah, missed that. If we do check the reconciliation status, maybe a zero delay makes it more efficient.

sdowell: That seems reasonable, there's not much need for the delayed readiness in the case of testing the happy path

PR #792, opened 2023-08-08
fix: Handle watcher context cancel

karlkfi:
Update the remediator's filteredwatcher to handle context cancellation, which is used to shut down the remediator. Previously, the cancel error was treated as any other error and the watcher would restart repeatedly until hitting the retry timeout (~4m). Now it should exit more quickly.

This also avoids logging the confusing watch error, `unable to decode an event from the watch stream`, when the context is cancelled during normal execution.

b/281728670

### DIFF
--- a/pkg/remediator/watch/filteredwatcher.go
+++ b/pkg/remediator/watch/filteredwatcher.go
@@ -16,6 +16,7 @@ package watch
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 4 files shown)


pkg/remediator/watch/filteredwatcher.go
```diff
@@ -258,39 +299,59 @@ func (w *filteredWatcher) Run(ctx context.Context) status.Error {
...
+				if ignoreEvent {
+					ignoredEventCount++
+				}
+				if err != nil {
+					if errors.Is(err, context.Canceled) || isContextCancelledStatusError(err) {
+						// The error wrappers are especially confusing for
+						// users, so just return context.Canceled.
+						runErr = status.InternalWrapf(context.Canceled, "remediator watch stopped for %s", w.gvk)
+						// Stop the watcher & return the status error
+						break Watcher
```
sdowell, line 327: Is there a reason for using these labeled breaks with an extra variable to track the return value, rather than using `return`? It seems like the latter may be more readable and easier to follow.

karlkfi: It avoids needing to duplicate the logging.

karlkfi: Also even if we use return to escape the function we would still need a named loop to break the event loop, and not the inner select.

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -234,16 +244,47 @@ func waitUntilNextRetry(retries int) {
...
+		if isDecodingErr, message := hasStatusErrorCause(statusErr, ClientWatchDecodingCause); isDecodingErr {
+			if message == ContextCancelledCauseMessage {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func hasStatusErrorCause(statusErr *apierrors.StatusError, causeType metav1.CauseType) (bool, string) {
```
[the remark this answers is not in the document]
karlkfi: done.

sdowell (approved):

PR #794, opened 2023-08-10
Use metricstransfrom to align Monarch monitoring pipeline

tiffanny29631:
The pipeline was previously removing any label that are not defined in the Monarch pipeline go/config-sync-monarch-metrics using attributes processor, the removal has two issues:
- Some labels that could be uniquely defining a timeseries cannot be removed correctly and would cause duplicate timeseries error
- Having a blocklist of labels means any newly added labels in metrics will need to be added to the blocklist, which can be hard to maintain

This change uses metricstransform processor to allowlist the metric name and labels that are defined in the Cloud Monarch, thus isolates Monarch (kubernetes) pipeline from Config Sync internal metric changes.

By using blocklist this change also fixes the duplicate timeseries error that is caused by removing `commit` label when there are multiple commits in the sync repo.

Also cleaning up the metrics that were consolidated and no longer exist.


e2e/testcases/otel_collector_test.go
```diff
@@ -137,6 +146,13 @@ func TestOtelCollectorDeployment(t *testing.T) {
 	if err != nil {
 		nt.T.Fatal(err)
 	}
+
+	nt.T.Log("Test done, removing test resources")
```
sdowell, line 150: If this is just being done to remove test resources then this isn't necessary, the test scaffolding should handle cleanup of anything declared in the git repo

e2e/testcases/otel_collector_test.go
```diff
@@ -92,6 +93,14 @@ func TestOtelCollectorDeployment(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
+	nt.T.Log("Adding test commit after otel-collector is started up so multiple commit hashes are processed in pipelines")
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Copy("../testdata/namespaces/foo.yaml", "foo.yaml"))
```
sdowell, line 97: nit: if we just need a simple object here, suggest declaring the object inline for readability. For example:
```golang
namespace := fake.NamespaceObject("foo")
nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/ns.yaml", namespace))
```

sdowell (commented):
Some comments on the test but otherwise looks okay. Should probably get a different reviewer for the otel config changes as I don't have much context

e2e/testcases/otel_collector_test.go
tiffanny29631: Done!

e2e/testcases/otel_collector_test.go
tiffanny29631: Done!

pkg/metrics/otel.go
```diff
@@ -140,88 +140,109 @@ processors:
           - remediate_duration_seconds
           - resource_conflicts_total
           - internal_errors_total
-          - rendering_count_total
```
karlkfi, line 143: i dont see some of these below. are they being deliberately removed?

tiffanny29631: These metrics no longer exist in Config Sync, so clearing them up from the allowlist.

karlkfi (approved):
/lgtm

PR #802, opened 2023-08-17
Skip validating R*Sync if it has a deletion timestamp

nan-yu:
The reconciler manager validates the R*Sync first before finalizing it. If the R*Sync is invalid (e.g. missing git-creds), the validation error results in the failure to finalize the object.

This commit skips the validatin if the R*Sync object has a deletion timestamp, so that it can still be finalized.

b/296270191

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -149,80 +149,38 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		return controllerruntime.Result{}, status.APIServerError(err, "failed to get RepoSync")
 	}
 
-	if !rs.DeletionTimestamp.IsZero() {
-		// Deletion requested.
-		// Cleanup is handled above, after the RepoSync is NotFound.
-		r.logger(ctx).V(3).Info("Sync deletion timestamp detected")
-	}
-
 	currentRS := rs.DeepCopy()
 
-	if err := r.watchConfigMaps(rs); err != nil {
-		r.logger(ctx).Error(err, "Error watching ConfigMaps")
-		reposync.SetStalled(rs, "ConfigMapWatch", err)
-		_, updateErr := r.updateStatus(ctx, currentRS, rs)
-		if updateErr != nil {
-			r.logger(ctx).Error(updateErr, "Error setting status")
+	if rs.DeletionTimestamp.IsZero() {
+		// Only validate RepoSync if it is not deleting. Otherwise, the validation
+		// error will block the finalizer.
+		if err := r.watchConfigMaps(rs); err != nil {
+			r.logger(ctx).Error(err, "Error watching ConfigMaps")
+			reposync.SetStalled(rs, "ConfigMapWatch", err)
+			_, updateErr := r.updateStatus(ctx, currentRS, rs)
+			if updateErr != nil {
+				r.logger(ctx).Error(updateErr, "Error setting status")
+			}
+			// Use the watch setup error for metric tagging.
+			metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+			// the error may be recoverable, so we retry (return the error)
+			return controllerruntime.Result{}, err
 		}
-		// the error may be recoverable, so we retry (return the error)
-		return controllerruntime.Result{}, err
-	}
 
-	if err := r.validateNamespaceName(rsRef.Namespace); err != nil {
-		r.logger(ctx).Error(err, "Sync namespace invalid")
-		reposync.SetStalled(rs, "Validation", err)
-		// Validation errors should not trigger retry (return error),
-		// unless the status update also fails.
-		_, updateErr := r.updateStatus(ctx, currentRS, rs)
-		// Use the validation error for metric tagging.
-		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
-		return controllerruntime.Result{}, updateErr
-	}
-
-	if errs := validation.IsDNS1123Subdomain(reconcilerRef.Name); errs != nil {
-		err := errors.Errorf("Invalid reconciler name %q: %s.", reconcilerRef.Name, strings.Join(errs, ", "))
-		r.logger(ctx).Error(err, "Sync name or namespace invalid")
-		reposync.SetStalled(rs, "Validation", err)
-		// Validation errors should not trigger retry (return error),
-		// unless the status update also fails.
-		_, updateErr := r.updateStatus(ctx, currentRS, rs)
-		// Use the validation error for metric tagging.
-		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
-		return controllerruntime.Result{}, updateErr
-	}
-
-	if err := r.validateSpec(ctx, rs, reconcilerRef.Name); err != nil {
-		r.logger(ctx).Error(err, "Sync spec invalid")
-		reposync.SetStalled(rs, "Validation", err)
-		// Validation errors should not trigger retry (return error),
-		// unless the status update also fails.
-		_, updateErr := r.updateStatus(ctx, currentRS, rs)
-		// Use the validation error for metric tagging.
-		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
-		return controllerruntime.Result{}, updateErr
-	}
-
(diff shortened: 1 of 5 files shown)


pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -140,69 +140,68 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
-		r.logger(ctx).Error(err, "Sync spec invalid")
-		rootsync.SetStalled(rs, "Validation", err)
-		// Validation errors should not trigger retry (return error),
-		// unless the status update also fails.
-		_, updateErr := r.updateStatus(ctx, currentRS, rs)
-		// Use the validation error for metric tagging.
-		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
-		return controllerruntime.Result{}, updateErr
-	}
+	} else {
```
sdowell, line 148: This is a pretty long nested conditional, could this be refactored for readability?

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -153,95 +153,97 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
-
-	currentRS := rs.DeepCopy()
-
-	if err := r.watchConfigMaps(rs); err != nil {
-		r.logger(ctx).Error(err, "Error watching ConfigMaps")
-		reposync.SetStalled(rs, "ConfigMapWatch", err)
-		_, updateErr := r.updateStatus(ctx, currentRS, rs)
-		if updateErr != nil {
-			r.logger(ctx).Error(updateErr, "Error setting status")
+	} else {
```
karlkfi, line 163: Long else statements are a code smell. Can you extract the else stuff into another method or two?

I'm thinking maybe validateNamespaceName, IsDNS1123Subdomain, and validateSpec can all just go under validateSpec. And then copyConfigMapsToCms should probably move into upsertManagedObjects.

karlkfi: The gotcha here is `r.setDefaultHelmDataKey(rs)`, which seems like a smell. We shouldn't be modifying the rs here. That's what got us into the trouble needing the currentRS. If we avoid modifying the rs here, we can punt the DeepCopy into the upsert method later. The default helm key shouldn't be applied back to the rs anway. It should just be an internal default we use when it's not specified.

karlkfi: I'll make a PR to extract the setDefaultHelmDataKey. It's a little involved.

nan-yu: Thanks. I will extract other validations into a separate method and then we can coordinate with the conflict.

pkg/reconcilermanager/controllers/rootsync_controller.go
nan-yu: refactored

pkg/reconcilermanager/controllers/reposync_controller.go
nan-yu: Refactored to make it less lengthy. As we discussed offline, it's not easy to avoid setting default helm data key in rs, so I ended up moving it above and reorganize the validation and upsert functions.

karlkfi: https://github.com/GoogleContainerTools/kpt-config-sync/pull/803

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -149,80 +149,38 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+		// error will block the finalizer.
+		if err := r.watchConfigMaps(rs); err != nil {
+			r.logger(ctx).Error(err, "Error watching ConfigMaps")
+			reposync.SetStalled(rs, "ConfigMapWatch", err)
+			_, updateErr := r.updateStatus(ctx, currentRS, rs)
+			if updateErr != nil {
+				r.logger(ctx).Error(updateErr, "Error setting status")
+			}
+			// the error may be recoverable, so we retry (return the error)
+			return controllerruntime.Result{}, updateErr
```
karlkfi, line 165: i think we want to return the `err` here, not the `updateErr` that came afterwards. That's why we log the  `updateErr`.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -149,80 +149,38 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+			// Use the validation error for metric tagging.
+			metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
 			return controllerruntime.Result{}, updateErr
 		}
-		if updateErr != nil {
-			r.logger(ctx).Error(updateErr, "Failed to update sync status")
-		}
+	} else {
+		// Deletion requested.
+		// Cleanup is handled above, after the RepoSync is NotFound.
```
karlkfi, line 182: this comment line is no longer accurate and can be removed. yes, we do delete above if not found, but ideally it's already cleaned up by the teardown below.

pkg/reconcilermanager/controllers/reposync_controller.go
nan-yu: Thanks for capturing it. Done.

pkg/reconcilermanager/controllers/reposync_controller.go
nan-yu: done

pkg/reconcilermanager/controllers/reposync_controller.go
nan-yu: rebased

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -149,80 +149,36 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
...
+		// Only validate RepoSync if it is not deleting. Otherwise, the validation
+		// error will block the finalizer.
+		if err := r.watchConfigMaps(rs); err != nil {
+			r.logger(ctx).Error(err, "Error watching ConfigMaps")
+			reposync.SetStalled(rs, "ConfigMapWatch", err)
+			_, updateErr := r.updateStatus(ctx, currentRS, rs)
+			if updateErr != nil {
+				r.logger(ctx).Error(updateErr, "Error setting status")
+			}
+			// the error may be recoverable, so we retry (return the error)
```
karlkfi, line 164: I think we're missing the metric here:
```
// Use the validation error for metric tagging.
metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
```

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -882,7 +838,23 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
...
 }
 
-func (r *RepoSyncReconciler) validateSpec(ctx context.Context, rs *v1beta1.RepoSync, reconcilerName string) error {
+func (r *RepoSyncReconciler) validateRepoSync(ctx context.Context, rs *v1beta1.RepoSync, reconcilerName string) error {
+	if rs.Namespace == configsync.ControllerNamespace {
+		return fmt.Errorf("RepoSync objects are not allowed in the %s namespace", configsync.ControllerNamespace)
+	}
+
+	if err := validation.IsDNS1123Subdomain(reconcilerName); err != nil {
+		return fmt.Errorf("invalid reconciler name %q: %s", reconcilerName, strings.Join(err, ", "))
```
karlkfi, line 847: Normally Go errors need lowercase, but i vaguely recall most of our status errors start with uppercase. I don't know if this gets prefixed or not when set as the error status... Is this going to be a consistency problem?

pkg/reconcilermanager/controllers/reposync_controller.go
nan-yu: done

pkg/reconcilermanager/controllers/reposync_controller.go
nan-yu: Right. The message is exposed in the STALLED condition. Change it back to start with uppercase.

I believe we have other places with this consistency problem, but that can be done separately.

PR #803, opened 2023-08-18
Refactor helm ConfigMap management

karlkfi:
- Make management more consistent with other managed resources
- Use upsertManagedObjects and deleteManageObjects to handle create, update, and delete of ConfigMaps.
- Revert drift to ConfigMap copies for the RepoSync
- Delete unused ConfigMaps in the Reconcile loop, instead of the event filter object mapping.
- Filter ConfigMap list by namespace
- Avoid double-logging reconcile triggers
- Avoid updating the DataKey to the default value when unspecified by the user. Just use the default internally if the value is empty.
- Extract the ConfigMap copy code to a function so it can be re-used, instead of passing the result through multiple layers.
- Add rsync labels to the ConfigMap copy can be found with a List filter. Keep the annotations until 1.16.0 is not longer supported.
- Use kinds.ObjectSummary to fix logs where the object Kind is not set, for typed objects.


pkg/reconcilermanager/controllers/helm_value_files.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// upsertHelmConfigMap creates or updates a config map in the
+// config-management-system namespace using an existing config map.
+func (r *RepoSyncReconciler) upsertHelmConfigMap(ctx context.Context, cmsCMRef, rsRef types.NamespacedName, userCM *corev1.ConfigMap, labelMap map[string]string) (controllerutil.OperationResult, error) {
+	cmsConfigMap := &corev1.ConfigMap{}
+	cmsConfigMap.Name = cmsCMRef.Name
+	cmsConfigMap.Namespace = cmsCMRef.Namespace
+
+	op, err := CreateOrUpdate(ctx, r.client, cmsConfigMap, func() error {
```
nan-yu, line 139: Can we actually do an update? https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.16.0-rc.2/pkg/reconcilermanager/controllers/reposync_controller.go#L1249 mentions that the copied ConfigMap is immutable.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -730,72 +716,73 @@ func (r *RepoSyncReconciler) mapSecretToRepoSyncs(secret client.Object) []reconc
 
 func (r *RepoSyncReconciler) mapConfigMapToRepoSyncs(obj client.Object) []reconcile.Request {
 	objRef := client.ObjectKeyFromObject(obj)
-	if objRef.Namespace == v1.NSConfigManagementSystem {
-		// ignore ConfigMaps in the config-management-system namespace
+
+	// Use annotations/labels to map ConfigMap copies in config-management-system
+	if objRef.Namespace == configsync.ControllerNamespace {
```
nan-yu, line 721: This will add labels and annotations to all ConfigMaps in the c-m-s namespace. What about ConfigMaps not for RepoSync Helm values?

pkg/reconcilermanager/controllers/helm_value_files.go
karlkfi: I guess it depends if the metadata is immutable or just the data. The docs do indicate that watches are closed when immutable, so maybe the metadata is immutable too.

Then again, if something changed such that it's not semantically equal any more and it tries to update it, we would want to see that error, right?  CreateOrUpdate already skips updates if nothing changed.

pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -171,31 +170,71 @@ func (r *RepoSyncReconciler) deleteRoleBinding(ctx context.Context, reconcilerRe
...
 
-func (r *RepoSyncReconciler) deleteHelmValuesFileRefCopies(ctx context.Context, rs *v1beta1.RepoSync) error {
-	if rs == nil || rs.Spec.SourceType != string(v1beta1.HelmSource) || rs.Spec.Helm == nil || len(rs.Spec.Helm.ValuesFileRefs) == 0 {
-		return nil
+// deleteUnusedHelmConfigMapCopies deletes helm values file ConfigMap copies
+// that are no longer required.
+func (r *RepoSyncReconciler) deleteUnusedHelmConfigMapCopies(ctx context.Context, rs *v1beta1.RepoSync) error {
+	rsRef := client.ObjectKeyFromObject(rs)
+	var cmNamesToKeep map[string]struct{}
+	if rs.Spec.SourceType == string(v1beta1.HelmSource) && rs.Spec.Helm != nil {
```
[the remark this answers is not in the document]
karlkfi: I was just being thorough to avoid nil pointer errors. If we do call it from somewhere else later, it will still work.

pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -171,31 +170,71 @@ func (r *RepoSyncReconciler) deleteRoleBinding(ctx context.Context, reconcilerRe
...
+		cmNamesToKeep := make(map[string]struct{}, len(rs.Spec.Helm.ValuesFileRefs))
```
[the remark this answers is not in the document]
karlkfi: I deliberately extracted getHelmConfigMapCopyRef so that we don't have to pass the computed []ValuesFileRefs through multiple layers of abstraction, so we can have cleaner, more consistent code. Yes, it re-hashes the names, but it's not that expensive.

pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -171,31 +170,71 @@ func (r *RepoSyncReconciler) deleteRoleBinding(ctx context.Context, reconcilerRe
...
+			continue
 		}
-
-		var existingConfigMapCopy corev1.ConfigMap
-		if err = r.client.Get(ctx, types.NamespacedName{Name: sourceConfigMapCopy.Name, Namespace: sourceConfigMapCopy.Namespace}, &existingConfigMapCopy); err != nil {
-			if apierrors.IsNotFound(err) {
-				// the copied ConfigMap doesn't exist, so we don't need to do anything
+		labels := cm.GetLabels()
+		if labels != nil {
+			if value, found := labels[repoSyncNameAnnotationKey]; !found || value != rsRef.Name {
```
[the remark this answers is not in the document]
karlkfi: I think I'm gonna fix this to use the existing labels instead:
		metadata.SyncNamespaceLabel: rs.Namespace,
		metadata.SyncNameLabel:      rs.Name,
		metadata.SyncKindLabel:      r.syncKind,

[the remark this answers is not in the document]
karlkfi: done

pkg/reconcilermanager/controllers/garbage_collector.go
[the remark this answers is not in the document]
karlkfi: I guess I could in-line the deleteUnusedHelmConfigMapCopies code into upsertHelmConfigMaps to avoid hashing them twice. I just figured this was easier to read.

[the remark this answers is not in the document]
karlkfi: Done.
This also fixes the garbage collection when SourceType changed or ValuesFileRefs was removed.

This makes me think... are other resources being garbage collected when the config changes? Like when someone switches from one Secret to another? Or are the copies being left behind?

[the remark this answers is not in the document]
karlkfi: upsertServiceAccount and upsertRoleBinding probably don't need to garbage collect, because they're always required and don't change name based on user input.

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: mapConfigMapToRepoSyncs doesn't add any labels. It just uses the labels to look up which RepoSync to trigger a Reconcile for. If the labels are not present, it doesn't trigger a Reconcile.

karlkfi: Ideally, this will never happen, because the objects are immutable, but it's still possible for them to be deleted and/or re-created. If they're deleted and the last known object had our labels, then this will catch it and trigger Reconcile to re-create it.

pkg/reconcilermanager/controllers/garbage_collector.go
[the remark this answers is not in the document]
karlkfi: deleteUnusedHelmConfigMapCopies removed

pkg/reconcilermanager/controllers/helm_value_files.go
nan-yu: `cmsConfigMap.Immutable` only sets the data to be immutable. Its metadata can still be modified.
If the copiedConfigMap's metadata is accidentally updated, this will revert the change, which is the correct behavior.

pkg/reconcilermanager/controllers/reposync_controller.go
nan-yu: I see. It enqueued ConfigMaps with the SyncName and SyncNamespace labels and annotations.

nan-yu (commented):
/lgtm

Looks good to me. I would like @natasha41575 to take a look as well in case I missed anything.

pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -171,31 +170,56 @@ func (r *RepoSyncReconciler) deleteRoleBinding(ctx context.Context, reconcilerRe
...
 
-func (r *RepoSyncReconciler) deleteHelmValuesFileRefCopies(ctx context.Context, rs *v1beta1.RepoSync) error {
-	if rs == nil || rs.Spec.SourceType != string(v1beta1.HelmSource) || rs.Spec.Helm == nil || len(rs.Spec.Helm.ValuesFileRefs) == 0 {
-		return nil
+// deleteHelmConfigMapCopies deletes helm values file ConfigMap copies,
+// except those specified in the cmNamesToKeep set.
+func (r *RepoSyncReconciler) deleteHelmConfigMapCopies(ctx context.Context, rsRef types.NamespacedName, cmNamesToKeep map[string]struct{}) error {
+	// cleanup old copied ConfigMaps that we don't need anymore, e.g. if the list in valuesFileRefs changes
+	// TODO: Eventually remove the annotations and use the labels for list filtering, to optimize cleanup.
+	// We can't remove the annotations until v1.16.0 is no longer supported.
```
natasha41575, line 178: Why is this? Would removal of the annotations be considered a breaking change?

natasha41575 (commented):
/lgtm

Thanks for the cleanup, this looks much better than what I had!

pkg/reconcilermanager/controllers/garbage_collector.go
karlkfi: The problem is mostly around garbage collection. We need to make sure any CMs made with v1.16.0 are deleted if the config changes, namely between v1.16.0 and v1.16.1+. We could stop adding the annotation, but it's not really hurting anything, and might help if there's a downgrade. Plus, not everyone upgrades every revision. So by v1.19.0 (or whatever the +3 version is after synchronizing with GKE versions) we can drop the annotation and stop depending on it for GC.

nan-yu (approved):
/lgtm

PR #804, opened 2023-08-18
Add a stress test with 100 CRDs

nan-yu:
We've found OOMKills in ACM 1.16.0 on Autopilot clusters with 100+ CRDs. This commit adds a stress test to make sure Config Sync can still work with those CRDs.

### DIFF
--- a/e2e/testcases/stress_test.go
+++ b/e2e/testcases/stress_test.go
@@ -18,15 +18,18 @@ import (
 	"fmt"
 	"io/ioutil"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
@@ -37,6 +40,7 @@ import (
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
+	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/testing/fake"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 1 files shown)


e2e/testcases/stress_test.go
```diff
@@ -261,6 +265,86 @@ func TestStressLargeRequest(t *testing.T) {
 	}
 }
 
+func TestStress100CRDs(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.Reconciliation1, ntopts.Unstructured, ntopts.StressTest,
+		ntopts.WithReconcileTimeout(configsync.DefaultReconcileTimeout))
+
+	// Apply 100 small CRDs on the cluster to make sure Config Sync can still work.
+	for i := 1; i <= 100; i++ {
+		crd := fakeCRD("anvil", fmt.Sprintf("acme-%d.com", i))
```
karlkfi, line 274: FWIW, this creates 100 API groups too, not just 100 resources. This will cause API Discovery to make 100 different API calls on older k8s clusters (one per group), but I assume that's intentional. Using the same group might test the memory usage without the api calls, but I don't know if that's necessary. This way is more stress.

e2e/testcases/stress_test.go
```diff
@@ -261,6 +265,86 @@ func TestStressLargeRequest(t *testing.T) {
...
+		if err := nt.KubeClient.Create(crd); err != nil {
+			nt.T.Fatal(err)
+		}
+	}
+
+	nt.T.Log("restarting the reconciler Pod to discovery the new 100 CRDs")
+	nomostest.DeletePodByLabel(nt, "app", reconcilermanager.Reconciler, true)
+	nt.T.Logf("Wait for the sync to complete")
+	if err := nt.WatchForAllSyncs(); err != nil {
```
karlkfi, line 283: I'm not sure this will actually do anything. It's probably already synced. Replacing the reconciler may not cause it to become Synced=False. It will cause it to become Reconciling=True, but I don't know if WatchForAllSyncs checks for that. It might be good to make a commit afterwards to ensure it syncs something.

nan-yu: Yes, you're right, `WatchForAllSyncs` doesn't check the reconciling condition.

Added a dummy commit to trigger the sync status change.

e2e/testcases/stress_test.go
```diff
@@ -261,6 +265,91 @@ func TestStressLargeRequest(t *testing.T) {
...
+
+	// Apply 100 small CRDs on the cluster to make sure Config Sync can still work.
+	for i := 1; i <= 100; i++ {
+		crd := fakeCRD("anvil", fmt.Sprintf("acme-%d.com", i))
+		if err := nt.KubeClient.Create(crd); err != nil {
+			nt.T.Fatal(err)
+		}
+	}
+
+	nt.T.Log("restarting the reconciler Pod to discovery the new 100 CRDs in the controller startup time")
```
sdowell, line 280: Are we concerned about stress testing the `reconciler-manager` pod as well, or just the `reconciler` pod?

e2e/testcases/stress_test.go
```diff
@@ -261,6 +265,91 @@ func TestStressLargeRequest(t *testing.T) {
...
+	nomostest.DeletePodByLabel(nt, "app", reconcilermanager.Reconciler, true)
+
+	nt.T.Log("adding a dummy file to the repo to trigger a new sync")
+	nt.Must(nt.RootRepos[configsync.RootSyncName].AddFile("stress-text.txt", []byte("stress-test")))
```
sdowell, line 284: I feel we may as well add a simple k8s resource such as a namespace

e2e/testcases/stress_test.go
nan-yu: We only want to stress the reconciler Pod. Restarting the reconciler Pod here is because the discovery client pulls the fresh schemas from the openapi when creating the reconciler in the startup time.

We can also wait until it refreshes the schema at the validation point, but the OOMKills we observed is at the startup time, so I feel it's better to simulate the same scenario.

e2e/testcases/stress_test.go
nan-yu: I guess we don't have to. The new commit will trigger the entire fetch-parse-apply loop even though the resource list is the same.

sdowell: Agreed it shouldn't make much difference, I just figure the test complexity is similar and it ensures that the parse-watch-apply loop has some apply work to do

e2e/testcases/stress_test.go
sdowell: Ack, SGTM. I have no issue with restarting the `reconciler` Pod, was mainly curious if we might want to restart the `reconciler-manager` Pod as well

e2e/testcases/stress_test.go
nan-yu: done

sdowell:
LGTM, I'll let Karl approve in case he has any final comments
/lgtm

karlkfi (approved):
/lgtm

PR #805, opened 2023-08-21
helm: use spec.helm.period to configure repulling charts

natasha41575:
Per offline discussion and b/295905345:

- Re-syncing a helm chart now also includes re-fetching, if the version is specified as latest or is a range of values from which we need to fetch the latest. Static versions still do not get re-fetched.
- Previously, spec.helm.period was used as the “re-sync” timer. Since re-syncing didn’t include re-fetching, it didn’t really do much. Now it is both the “re-sync/re-fetch” timer. This means that we don’t need a separate “re-fetch timer”, which we previously had as `—helm-sync-version-polling-period`. Removed this extra timer. 
- Changed spec.helm.period default to 1 hour to preserve previous behavior of refetching every 1 hour. 
- The literal tag “latest” will be treated as a special case, and will be re-fetched every sync.
- Updated CRD descriptions accordingly.


pkg/api/configsync/v1beta1/helmconfig.go
```diff
@@ -79,11 +85,12 @@ type HelmBase struct {
 	// +optional
 	IncludeCRDs bool `json:"includeCRDs,omitempty"`
 
-	// period is the time duration between consecutive syncs. Default: 15s.
+	// period is the time duration between consecutive syncs. Default: 1hour.
```
nan-yu, line 88: Per offline discussion, `consecutive syncs` seems confusing. We can update the comment to  `refetching chart` specifically.

nan-yu (commented):
Only one comment on the CRD description, otherwise looks good.

pkg/api/configsync/v1beta1/helmconfig.go
natasha41575: Updated the crd description here to be more specific

manifests/reposync-crd.yaml
```diff
@@ -1163,9 +1163,10 @@ spec:
                       false.'
                     type: boolean
                   period:
-                    description: 'period is the time duration between consecutive
-                      syncs. Default: 1hour. Use string to specify this field value,
-                      like "30s", "5m". More details about valid inputs: https://pkg.go.dev/time#ParseDuration.
+                    description: 'period is the time duration that Config Sync waits
+                      before refetching the chart and resyncing. Default: 1 hour.
```
nan-yu, line 1167: Suggest deleting `and resyncing`. If the period is configured to be less than 15s, internally, the reconciler still waits for 15s to resync.

natasha41575: Done, took out `and resyncing` and also changed the language below not to reference `syncs`.

nan-yu (approved):
/lgtm

PR #806, opened 2023-08-21
fetch helm from patched GKE version

sdowell:
This version is similar to the upstream release, but with several CVE fixes.


nan-yu (approved):
/lgtm

PR #807, opened 2023-08-22
ci: use common distro for test images

sdowell:
The test images were using different distributions, which could lead to inconsistent behaviors in things like GNU core utilities. This updates each of the test images to use a consistent debian based distro.


Makefile
```diff
@@ -36,6 +36,15 @@ OUTPUT_DIR := $(abspath .output)
 # Self-contained GOPATH dir.
 GO_DIR := $(OUTPUT_DIR)/go
 
+# Base image used for all golang containers
+GOLANG_IMAGE := golang:1.20.8-bullseye
```
nan-yu, line 40: The base image is going to change to debian 12 bookworm. For consistency, we can use golang:1.20.8-bookworm as well.

nan-yu, line 40: Can we have docker build args for other images (e.g. gcr.io/gke-release/debian-base:bullseye-v1.4.3-gke.9)?
Then we just need to update this single file for image update.

Makefile.e2e
```diff
@@ -59,17 +59,34 @@ test-e2e-gke-nobuild:
 .PHONY: test-e2e-gke
 test-e2e-gke: config-sync-manifest test-e2e-gke-nobuild
 
+KIND_IMAGE := "kind-image"
+
+.PHONY: build-kind-image
+build-e2e-kind: config-sync-manifest-local build-gke-e2e
```
nan-yu, line 65: Why does it need `build-gke-e2e`?

nit: Can we use the same style, either `build-gke-e2e|build-kind-e2e`, or `build-e2e-gke|build-e2e-kind`?

nan-yu, line 65: `build-e2e-kind` doesn't need to depend on `config-sync-manifest-local`. We can make `config-sync-manifest-local` a dependency of `test-e2e-kind-multi-repo`.

nan-yu: > Why does it need build-gke-e2e?

Got it answered. The kind dockerfile depends on the e2e image, which is built by build-gke-e3e.

Makefile
sdowell: What is the reason we are changing to `bookworm`? This should be doable, however one issue I ran into when trying bookworm is that the [upx](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/build/buildenv/Dockerfile#L78) package is not available on bookworm.
```
 > [stage-1 2/3] RUN apt-get update   && apt-get install -y   bash   gcc   git   musl-dev   upx   wget:                                                                     
#0 0.566 Get:1 http://deb.debian.org/debian bookworm InRelease [151 kB]                                                                                                     
#0 0.589 Get:2 http://deb.debian.org/debian bookworm-updates InRelease [52.1 kB]                                                                                            
#0 0.598 Get:3 http://deb.debian.org/debian-security bookworm-security InRelease [48.0 kB]                                                                                  
#0 0.660 Get:4 http://deb.debian.org/debian bookworm/main amd64 Packages [8906 kB]                                                                                          
#0 0.729 Get:5 http://deb.debian.org/debian bookworm-updates/main amd64 Packages [6408 B]
#0 0.767 Get:6 http://deb.debian.org/debian-security bookworm-security/main amd64 Packages [74.9 kB]
#0 1.474 Fetched 9239 kB in 1s (9563 kB/s)
#0 1.474 Reading package lists...
#0 1.901 Reading package lists...
#0 2.309 Building dependency tree...
#0 2.440 Reading state information...
#0 2.446 E: Unable to locate package upx
------
ERROR: failed to solve: executor failed running [/bin/sh -c apt-get update   && apt-get install -y   bash   gcc   git   musl-dev   upx   wget]: exit code: 100
make: *** [Makefile.build:19: build-buildenv] Error 1

```

This means we may have to either:
- find an alternate installation method for upx
- remove our [dependency on upx](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/scripts/build.sh#L64)

sdowell: done

Makefile.e2e
sdowell: done

sdowell: Done. To accomplish this I did some refactoring to allow running `config-sync-manifest-local` without a recursive make call. This preserves the dependency tree so certain targets don't get run multiple times (e.g. `install-helm`)

Makefile
sdowell: At any rate, we can change both base images together with a subsequent change

nan-yu (approved):
/lgtm

PR #808, opened 2023-08-22
fix: use short option for sha256sum

sdowell:
Some distributions of sha256sum only support the short options. This updates the call to use the short option for broader compatibility. This check is currently broken on our alpine-based images used in some test environments. We may want to switch these alpine-based images to debian as a followup change so that our build images are more similar.


karlkfi (approved):
/lgtm

PR #811, opened 2023-08-22
Add log level override API to v1alpha1

victorpras:
we need to continue to support both v1alpha1 and v1beta1 API until v1alpha1 is removed. https://kubernetes.io/docs/reference/using-api/deprecation-policy/


pkg/api/configsync/v1alpha1/resource_override.go
```diff
@@ -66,6 +66,13 @@ type OverrideSpec struct {
 	// support pulling remote bases from public repositories.
 	// +optional
 	EnableShellInRendering *bool `json:"enableShellInRendering,omitempty"`
+
+	// logLevels specify the container name and log level override value for the reconciler deployment container.
+	// Each entry must contain the name of the reconciler deployment container and the desired log level.
+	// +listType=map
+	// +listMapKey=containerName
+	// +optional
+	LogLevels []ContainerLogLevelOverride `json:"logLevels,omitempty"`
```
sdowell, line 75: Should we have e2e tests that cover each api version, or is it sufficient to just have test coverage for a field on one api version?

cc @janetkuo @nan-yu @karlkfi

pkg/api/configsync/v1alpha1/zz_generated.conversion.go
sdowell, line 1: You should also update the manual conversion code

karlkfi:
Why are we backporting this to v1alpha1?

pkg/api/configsync/v1alpha1/zz_generated.conversion.go
victorpras: removed the previously manually added conversion function and regenerated

victorpras:
> Why are we backporting this to v1alpha1?

From my understanding, we want to keep API changes to be consistent between v1alpha1 and v1beta1 until we remove v1alpha1

karlkfi:
> From my understanding, we want to keep API changes to be consistent between v1alpha1 and v1beta1 until we remove v1alpha1

But why? What's the technical constraint making this necessary?

I thought the point of the versioning system was to enable change.

victorpras:
After the internal discussion, we agree that this is needed to follow rule2 on https://kubernetes.io/docs/reference/using-api/deprecation-policy/

pkg/api/configsync/v1alpha1/resource_override.go
nan-yu: We have a few tests that cover both v1apha1 and v1beta1, especially the overrides fields. If we test both api versions for all fields, a lot of tests need to be added.

I'm okay that we're super cautious when adding new fields to both APIs and only test the v1beta1.

nan-yu (approved):
/lgtm

PR #812, opened 2023-08-23
Improve the retry behavior

haiyanmeng:
Currently, the retry behavior for different triggers are different. When the trigger is managementConflict, it waits 5 seconds between retries. When the trigger is watchUpdate, it retries every second. When the trigger is the fact that the last reconciliation failed, it retries every second for the first 5 retires, and then does exponential backoff retry up to 5 minutes.

This PR unifies the retry behavior for these three triggers with backoff + jitter (inital duration: 1s; factor: 2; jitter: 0.1). There can be up to 12 retries for a given commit.

b/289254423


pkg/parse/state.go
```diff
@@ -104,42 +118,19 @@ func (s *reconcilerState) reset() {
...
 	klog.Errorf("Invalidating reconciler checkpoint: %v", status.FormatSingleLine(errs))
-	oldErrs := s.cache.errs
 	s.cache.errs = errs
 	// Invalidate state on error since this could be the result of switching
 	// branches or some other operation where inverting the operation would
 	// result in repeating a previous state that was checkpointed.
 	s.lastApplied = ""
 	s.cache.needToRetry = true
-	if status.DeepEqual(oldErrs, s.cache.errs) {
-		s.cache.reconciliationWithSameErrs++
```
nan-yu, line 122: The retry attempt was based on the same set of errors, but now it is based on all types of errors. Is it an intentional change?

haiyanmeng: Yes. It is intentional. The current retry logic is a little aggressive.

pkg/parse/run.go
```diff
@@ -116,18 +120,21 @@ func Run(ctx context.Context, p Parser) {
...
 				trigger = triggerRetry
 			} else if opts.needToUpdateWatch() {
-				klog.Infof("Some watches need to be updated")
 				trigger = triggerWatchUpdate
 			} else {
 				// Don't reset the retry timer if there's nothing to retry.
 				continue
 			}
+			if state.backoff.Steps == 0 {
+				// No more retires are allowed
```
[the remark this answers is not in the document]
haiyanmeng: Fixed it.

pkg/parse/run.go
```diff
@@ -116,18 +120,21 @@ func Run(ctx context.Context, p Parser) {
...
+				continue
```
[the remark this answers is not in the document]
haiyanmeng: Good suggestion.

pkg/parse/run.go
```diff
@@ -116,18 +120,21 @@ func Run(ctx context.Context, p Parser) {
...
+			}
+			time.Sleep(state.backoff.Step())
+			retries := retryLimit - state.backoff.Steps
+			klog.Infof("a retry is triggered (trigger type: %v, retries: %v)", trigger, retries)
```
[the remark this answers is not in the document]
haiyanmeng: Done.

pkg/parse/run.go
```diff
@@ -116,18 +120,21 @@ func Run(ctx context.Context, p Parser) {
...
 				trigger = triggerWatchUpdate
 			} else {
 				// Don't reset the retry timer if there's nothing to retry.
 				continue
 			}
+			if state.backoff.Steps == 0 {
+				// No more retires are allowed
+				continue
+			}
+			time.Sleep(state.backoff.Step())
```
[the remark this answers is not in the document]
haiyanmeng: Yeah. I like your suggestion.

pkg/parse/run.go
```diff
@@ -81,54 +81,77 @@ func Run(ctx context.Context, p Parser) {
...
-			run(ctx, p, trigger, state)
+			if state.backoff.Steps == 0 {
+				klog.Infof("Retry limit (%v) has been reached", retryLimit)
+				// Don't reset retryTimer if retry limit has been reached.
+				continue
+			}
+			// retryTimer is reset to `opts.retryPeriod` every time after `run` is executed
+			// to catch triggerManagementConflict/triggerRetry/triggerWatchUpdate in time.
+			// `state.NextRetryTime` is used to check when the next retry should start.
+			if !time.Now().Before(state.nextRetryTime) {
```
nan-yu, line 144: The time check here, along with the timer reset, seems to use two different strategies to manage the retry backoff. Can we do `retryTimer.Reset(retryDuration)`, and get rid of the hard-coded retryPeriod?

haiyanmeng: Unfortunately, we cannot. I added a comment here L141-143 to explain why we need to reset `retryTimer` to `opts.retryPeriod` every time after `run` . Let me know if it is clear enough.

haiyanmeng: You are right. I have been going back and force on the experiments, and was a little overwhelmed.  Adding https://github.com/GoogleContainerTools/kpt-config-sync/pull/812/files#diff-4e49e2db3ba5fd76214acd7bc50a14b99623650c72ca6ffaf51157b2fe1f4f15R130 could avoid introducing `nextRetryTime`.

nan-yu (approved):
/lgtm

PR #813, opened 2023-08-23
fix: nil pointer in upsertHelmConfigMap

karlkfi:
- Add AddAnnotations and AddLabels funcs to core
- Fix TestHelmConfigMapNamespaceRepo to change to a new ConfigMap instead of re-creating an immutable one.


natasha41575 (approved):
lgtm

PR #814, opened 2023-08-23
Set default resource requests for askpass container

victorpras:
Set default resource requests for askpass container

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -132,6 +132,10 @@ data:
          - name: gcenode-askpass-sidecar
            image: gcr.io/config-management-release/gcenode-askpass-sidecar:v1.0.4
            args: ["--port=9102", "--logtostderr"]
+           resources:
+             requests:
+               cpu: 10m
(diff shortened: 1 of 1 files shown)


haiyanmeng (approved):

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -132,6 +132,10 @@ data:
          - name: gcenode-askpass-sidecar
            image: gcr.io/config-management-release/gcenode-askpass-sidecar:v1.0.4
            args: ["--port=9102", "--logtostderr"]
+           resources:
+             requests:
+               cpu: 10m
+               memory: 100Mi
```
janetkuo, line 138: Does askpass need this much resource? The default memory request before this change was 10Mi.

victorpras: ran a few more benchmark and the highest i've seen is 16Mi. I have created a PR to reduce the default to 20Mi https://github.com/GoogleContainerTools/kpt-config-sync/pull/835

```
CPU(cores)   MEMORY(bytes)
7m           16Mi
```

PR #816, opened 2023-08-24
outputting a proper extension for bugreport files

mikebz:
A small nit for me. When I use bugreport all the files are txt and they are not getting proper syntax highlighting.

This is a way to use log, json and yaml.

I do not believe we make promises about the filenames in bugreport, I could be wrong though
@janetkuo


mikebz:
I am guessing that the test is failing because it's using a pre-built version of `nomos` CLI.  What do we want to do here?  Should we:
1) not use the CLI and use nomos as a library?
2) not run these as a part of the PR CI?  Because there is a circular dependency here.
3) temporarily remove this for this particular PR since maybe the intent here is to use the CLI to verify that other parts of the system didn't change the way CLI extracts bug report.

karlkfi (commented):
seems ok to me. my only concern is whether we consider this an interface we need to remain reverse compatibile.

karlkfi (approved):
/lgtm

PR #817, opened 2023-08-24
fix: Remove 'type' label from all metrics

karlkfi:
The type label was increasing the cardinality of metrics by a significant amount, especially when used with config connector or crossplane, which have hundreds or even thousands of CRDs.

So this removes the type label from all metrics, which should reduce the total number of metrics and the cost, especially when combined with high-commit repositories.


tiffanny29631 (commented):
Looks pretty good to me, are the Prom test running in postsubmits or they are in presubmits?

karlkfi:
> are the Prom test running in postsubmits or they are in presubmits?

Both.

tiffanny29631 (commented):
/lgtm
/approve

PR #818, opened 2023-08-25
helm: add new fields to v1alpha1

natasha41575:
We only added the new helm fields to v1beta1, but turns out we need to also add them to v1alpha1 so that they can be round-tripped.

Generated files were updated using `make configsync-crds`, `make generate`, and `make clientgen`.


nan-yu (approved):
/lgtm

PR #819, opened 2023-08-25
Modify the otel_collector_test error condition

tiffanny29631:
This is a cleanup after https://github.com/GoogleContainerTools/kpt-config-sync/pull/817.

With `type` label removed from the original metrics, the attribute processor that force removes the tag will no longer create the error thus checking for the error condition is not needed.

Changed the error case to metrics fail to export due to the rejected labels by the Cloud Monarch.


karlkfi (approved):
/lgtm

PR #821, opened 2023-08-26
copying nomos from output

mikebz:


mikebz:
I am including this in another branch and PR.

PR #830, opened 2023-08-29
test: reduce flakiness of reconcile timeout e2e (#791)

haiyanmeng:
This test has been relatively flaky, particularly on autopilot. The condition of the flake appears to frequently coincide with a node replacement that causes Pod eviction. This presumably causes the reconciler Pod to get evicted and restarted, and by the time it comes back up the goproxy Pod is ready.

This extends the delay of the goproxy readinessProbe in order to reduce the chance of encountering this scenario on autopilot.


nan-yu (approved):
/lgtm

PR #835, opened 2023-08-29
Reduce askpass container default memory

victorpras:
After discussing with @janetkuo , we will reduce the askpass container default memory

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -135,7 +135,7 @@ data:
            resources:
              requests:
                cpu: 10m
-               memory: 100Mi
+               memory: 20Mi
            imagePullPolicy: IfNotPresent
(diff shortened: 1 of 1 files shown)


karlkfi:
This will reduce the limit in autopilot. Have you verified it won't OOM at the lower limit?

victorpras:
> This will reduce the limit in autopilot. Have you verified it won't OOM at the lower limit?

I've run a few different setup for config sync and the memory consumption never seem to exceed 20M. Before the changes in https://github.com/GoogleContainerTools/kpt-config-sync/pull/814, the container was created with default memory value of 10M and afaict there was no issue with OOM

janetkuo:
This is a follow-up of https://github.com/GoogleContainerTools/kpt-config-sync/pull/814

nan-yu (approved):
/lgtm

@haiyanmeng FYI, this also needs to be cherrypicked into v1.16.

PR #836, opened 2023-08-30
Improve the retry behavior (#812)

haiyanmeng:
Currently, the retry behavior for different triggers are different. When the trigger is managementConflict, it waits 5 seconds between retries. When the trigger is watchUpdate, it retries every second. When the trigger is the fact that the last reconciliation failed, it retries every second for the first 5 retires, and then does exponential backoff retry up to 5 minutes.

This PR unifies the retry behavior for these three triggers with backoff
+ jitter (inital duration: 1s; factor: 2; jitter: 0.1). There can be up to 12 retries for a given commit.

b/289254423

### DIFF
--- a/pkg/parse/cache.go
+++ b/pkg/parse/cache.go
@@ -16,7 +16,6 @@ package parse
 
 import (
 	"sort"
-	"time"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


nan-yu (approved):
/lgtm

PR #837, opened 2023-08-30
refactor: move the askpass declaration to manifest (#790)

haiyanmeng:
The askpass sidecar is a container in the reconciler Pod akin to the other containers, but is treated differently and declared in code rather than the manifests. This inconsistency can be a gotcha and requires our other tooling to have special exceptions for parsing the image tag from code. This change makes the askpass sidecar consistent with the other reconciler containers.

Note this is intended as a non-functional refactor, so the overall behavior should be unchanged.


haiyanmeng:
This PR avoids merge conflicts when cherrypicking https://github.com/GoogleContainerTools/kpt-config-sync/pull/800/files.

nan-yu (approved):
/lgtm

PR #842, opened 2023-08-30
Set default resource requests for askpass container

victorpras:
Set default resource requests for askpass container


nan-yu (approved):
/lgtm

PR #845, opened 2023-08-30
Change label aggregation in Moanrch to MAX

tiffanny29631:
Use max for timestamp to get the latest;

Use max for duration distribution to get the longest;

Use max for count to get the latest.

### DIFF
--- a/pkg/metrics/otel.go
+++ b/pkg/metrics/otel.go
@@ -184,98 +184,98 @@ processors:
           - action: aggregate_labels
             # label_set is the allowlist of labels to keep after aggregation
             label_set: [status, operation]
-            aggregation_type: sum
+            aggregation_type: max
       - include: declared_resources
         action: update
         new_name: current_declared_resources
         operations:
           - action: aggregate_labels
             label_set: []
-            aggregation_type: sum
+            aggregation_type: max
       - include: kcc_resource_count
         action: update
         operations:
           - action: aggregate_labels
             label_set: [resourcegroup]
-            aggregation_type: sum
+            aggregation_type: max
       - include: reconciler_errors
         action: update
         new_name: last_reconciler_errors
         operations:
           - action: aggregate_labels
             label_set: [component, errorclass]
-            aggregation_type: sum
+            aggregation_type: max
       - include: reconcile_duration_seconds
         action: update
         operations:
           - action: aggregate_labels
             label_set: [status]
-            aggregation_type: sum
+            aggregation_type: max
       - include: rg_reconcile_duration_seconds
         action: update
         operations:
           - action: aggregate_labels
             label_set: [stallreason]
-            aggregation_type: sum
+            aggregation_type: max
       - include: last_sync_timestamp
         action: update
         operations:
           - action: aggregate_labels
             label_set: [status]
-            aggregation_type: sum
+            aggregation_type: max
       - include: parser_duration_seconds
         action: update
         operations:
           - action: aggregate_labels
             label_set: [status, source, trigger]
-            aggregation_type: sum
+            aggregation_type: max
       - include: pipeline_error_observed
         action: update
         new_name: last_pipeline_error_observed
         operations:
           - action: aggregate_labels
             label_set: [name, component, reconciler]
-            aggregation_type: sum
+            aggregation_type: max
       - include: apply_operations_total
         action: update
         new_name: apply_operations_count
         operations:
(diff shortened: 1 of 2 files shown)


pkg/metrics/otel.go
```diff
@@ -184,14 +184,14 @@ processors:
...
         action: update
         new_name: current_declared_resources
         operations:
           - action: aggregate_labels
             label_set: []
-            aggregation_type: sum
+            aggregation_type: max
       - include: kcc_resource_count
         action: update
         operations:
```
karlkfi, line 197: should this be max too? it's the count for a single repo. i dont think we want to sum it, otherwise it looks like the repo has twice as many objects

pkg/metrics/otel.go
```diff
@@ -204,38 +204,38 @@ processors:
...
         action: update
         new_name: last_pipeline_error_observed
         operations:
           - action: aggregate_labels
             label_set: [name, component, reconciler]
-            aggregation_type: sum
+            aggregation_type: max
       - include: apply_operations_total
         action: update
         new_name: apply_operations_count
```
karlkfi, line 241: This is interesting.
Every time this is recorded the value is always 1. so summing will change that.
I don't think this is a problem, as long as we graph by sum in our dashboards over a time period or something...

pkg/metrics/otel.go
karlkfi: I can't find where this is recorded in the code... Do we even send this any more?

haiyanmeng: This is a resource group metric: https://gke-internal.googlesource.com/GoogleCloudPlatform/resource-group/+/refs/heads/master/controllers/metrics/record.go#54

karlkfi: huh. might be nice to group the metrics by source so this is more obvious in the future. not important for 1.16.1 tho.

karlkfi: Looks like this is the count for all the KCC objs in the ResourceGroup. So max seems correct.

pkg/metrics/otel.go
tiffanny29631: The pipeline error is an odd one iirc. The value of it is not used as counter but more of a boolean status, where non-zero indicates error. Sum probably works for this case.

pkg/metrics/otel.go
karlkfi: That said... do we even need to aggregate this? It doesn't have commit or type does it?

tiffanny29631: Ah you are right this is after the metric is recorded and counter has already incremented. Changed.

pkg/metrics/otel.go
```diff
@@ -184,51 +184,51 @@ processors:
...
       - include: parser_duration_seconds
         action: update
         operations:
           - action: aggregate_labels
             label_set: [status, source, trigger]
-            aggregation_type: sum
+            aggregation_type: max
       - include: pipeline_error_observed
         action: update
         new_name: last_pipeline_error_observed
```
karlkfi, line 234: this is a timestamp, right? should it be max?

pkg/metrics/otel.go
tiffanny29631: The purpose for this filter is to constrain the labels so that any future labels added will always be excluded from Monarch pipeline.

pkg/metrics/otel.go
tiffanny29631: Not a time stamp, the value is a boolean, records 1 if errlen > 0 https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/metrics/record.go#L73-L74

karlkfi: ah. then yeah, we definitely want max, not sum.

haiyanmeng:
> Use max for timestamp to get the latest;
> 
> Use max for duration distribution to get the longest;
> 
> Use sum for count to get the total of one time serie.

I don't see `sum` is used to aggregate any metrics.

pkg/metrics/otel.go
tiffanny29631: max it is.

tiffanny29631:
> I don't see sum is used to aggregate any metrics.

Fixed description.

karlkfi (approved):
/lgtm

haiyanmeng (approved):

PR #847, opened 2023-08-30
Cherrypick: fix: Handle watcher context cancel (#792)

karlkfi:
Update the remediator's filteredwatcher to handle context cancellation, which is used to shut down the remediator. Previously, the cancel error was treated as any other error and the watcher would restart repeatedly until hitting the retry timeout (~4m). Now it should exit more quickly.

This also avoids logging the confusing watch error, `unable to decode an event from the watch stream`, when the context is cancelled during normal execution.

b/281728670

### DIFF
--- a/pkg/remediator/watch/filteredwatcher.go
+++ b/pkg/remediator/watch/filteredwatcher.go
@@ -16,6 +16,7 @@ package watch
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


nan-yu (approved):
/lgtm

haiyanmeng (approved):

karlkfi:
rebased

PR #848, opened 2023-08-30
fix: race condition in scheme

karlkfi:
- Scheme is not thread-safe.
- Remove all code that was adding types to scheme at runtime.
- Scheme types are managed at init-time in pkg/core/scheme.go


pkg/webhook/configuration/update.go
```diff
@@ -53,12 +53,6 @@ func Update(ctx context.Context, c client.Client, dc discovery.ServerResourcer,
 		return nil
 	}
 
-	// Ensure the scheme used by the client knows about ValidatingWebhookConfiguration.
```
nan-yu, line 56: We can remove it here because the scheme has already been added in https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.16.0-rc.2/pkg/core/scheme.go#L87.

Suggest mentioning it in the commit message. Otherwise, it is not safe to delete without registering at least once.

karlkfi: updated commit msg

nan-yu (approved):
/lgtm

janetkuo (approved):
/lgtm

PR #852, opened 2023-09-01
Fix TestManagingReconciler

victorpras:
With the change in #814, test case 6 in `TestManagingReconciler` will only bump generation by 1 in both standard and autopilot cluster

### DIFF
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -432,11 +432,7 @@ func TestManagingReconciler(t *testing.T) {
 	nt.T.Log("Switch the auth type from none to gcpserviceaccount")
 	nt.MustMergePatch(rs, `{"spec":{"git":{"auth":"gcpserviceaccount","secretRef":{"name":""},"gcpServiceAccountEmail":"test-gcp-sa-email@test-project.iam.gserviceaccount.com"}}}`)
 	nt.T.Log("Verify the gcenode-askpass-sidecar container should be added")
-	if nt.IsGKEAutopilot {
-		generation += 2 // generation bumped by 2 because the sidecar container will first be added and then the resource requirements will be adjusted by autopilot
-	} else {
-		generation++ // generation bumped by 1 to apply the new sidecar container
-	}
+	generation++ // generation bumped by 1 to apply the new sidecar container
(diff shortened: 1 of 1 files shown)


e2e/testcases/reconciler_manager_test.go
```diff
@@ -432,11 +432,7 @@ func TestManagingReconciler(t *testing.T) {
 	nt.T.Log("Switch the auth type from none to gcpserviceaccount")
 	nt.MustMergePatch(rs, `{"spec":{"git":{"auth":"gcpserviceaccount","secretRef":{"name":""},"gcpServiceAccountEmail":"test-gcp-sa-email@test-project.iam.gserviceaccount.com"}}}`)
 	nt.T.Log("Verify the gcenode-askpass-sidecar container should be added")
-	if nt.IsGKEAutopilot {
-		generation += 2 // generation bumped by 2 because the sidecar container will first be added and then the resource requirements will be adjusted by autopilot
-	} else {
-		generation++ // generation bumped by 1 to apply the new sidecar container
-	}
+	generation++ // generation bumped by 1 to apply the new sidecar container
```
karlkfi, line 435: Autopilot won't update the container resources after adding it? Or was it already added? What's the actual expected update?

tiffanny29631: seems like with resource request specified in https://github.com/GoogleContainerTools/kpt-config-sync/pull/814 autopilot won't perform the adjustment, thus saved one generation?

victorpras: I compared how this test works in main and in 1.15 branch (before all the askpass container changes). In 1.15, the askpass initial container input resource was empty, and in later generation autopilot added the resource value, so it took 2 generations to properly add the container. In main, the initial input resource already came with default value from the configmap, therefore it took just 1 generation to add the container

karlkfi: I may be confused. Shouldn't autopilot still update the container resources to add a limit if we only set the request?

victorpras: looking at both scenario logs and deployment generation, it looks like limits are always set at the same time when requests are set, so not specifying limits shouldn't impact generation count. According to the docs in https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests, `Autopilot only considers requests`

karlkfi (approved):
/lgtm

PR #855, opened 2023-09-05
test: wait for deployments before checking pods

sdowell:
The subsequent Pod checks are flaky because the Deployments may still be reconciling. When the Deployment is still reconciling, there may be more than the expected number of Pods - because the old replicas may not have been garbage collected yet.


nan-yu (approved):
/lgtm

PR #860, opened 2023-09-08
fix: Retry status updates

karlkfi:
Status updates should retry if the error is recoverable, especially if it's just a ResourceVersion conflict, as long as the generation hasn't changed. This should reduce unnecessary reconcile loops and API calls in the reconciler-manager and reconciler finalizer.

### DIFF
--- a/pkg/util/mutate/mutate.go
+++ b/pkg/util/mutate/mutate.go
@@ -16,13 +16,12 @@ package mutate
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/util/retry"
+	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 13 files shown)


pkg/util/mutate/mutate.go
```diff
@@ -117,66 +143,66 @@ func WithRetry(ctx context.Context, c client.Client, obj client.Object, mutate F
...
-			obj.GetNamespace(), obj.GetName())
-	}
-	return fmt.Sprintf("%T %s/%s", obj,
-		obj.GetNamespace(), obj.GetName())
+type specClient struct {
+	client client.Client
+}
+
+// Get the current state of the specified object.
+func (c *specClient) Get(ctx context.Context, obj client.Object) error {
```
haiyanmeng, line 175: `state` to `spec`

pkg/util/mutate/mutate.go
```diff
@@ -117,66 +143,66 @@ func WithRetry(ctx context.Context, c client.Client, obj client.Object, mutate F
...
+	return c.client.Get(ctx, client.ObjectKeyFromObject(obj), obj)
+}
+
+// Update the state of the specified object.
```
haiyanmeng, line 179: `state` to `spec`

pkg/util/mutate/mutate.go
```diff
@@ -117,66 +143,66 @@ func WithRetry(ctx context.Context, c client.Client, obj client.Object, mutate F
...
+// to this updater.
+func (c *specClient) WrapError(_ context.Context, obj client.Object, err error) error {
+	return errors.Wrapf(err, "failed to update object: %s", kinds.ObjectSummary(obj))
+}
+
+type statusClient struct {
+	client client.Client
+}
+
+// Get the current state of the specified object.
```
haiyanmeng, line 194: `state` to `status`

pkg/util/mutate/mutate.go
```diff
@@ -117,66 +143,66 @@ func WithRetry(ctx context.Context, c client.Client, obj client.Object, mutate F
...
-			// The error (or an error it wraps) is a known status error from K8s
-			return false, status.APIServerError(retryErr, "failed to update object status", obj)
-		}
-		return false, errors.Wrapf(retryErr, "failed to update object status: %s", objSummary(obj))
-	}
-	return true, nil
+// updater is a simplified client interface for performing reads and writes on
+// an object, without options.
+type updater interface {
+	// Get the current state of the specified object.
```
haiyanmeng, line 159: `Get the current state of the specified object.` to `Get the specified object.` since there are two implementations of this interface: `specClient` and `statusClient`.

pkg/util/mutate/mutate.go
karlkfi: done

pkg/util/mutate/mutate.go
karlkfi: done

pkg/util/mutate/mutate.go
karlkfi: done

pkg/util/mutate/mutate.go
karlkfi: done

haiyanmeng (approved):

PR #861, opened 2023-09-08
Add more logs when the test fails

nan-yu:
Previously, printTestLogs is in its own cleanup block to catch fatal errors from the reset block.
    
Now we decided to reset after `printTestLogs`, so we can combine them into a single block for simplicity.
    
It also prints test logs if reset fails.

The `testPods` function is also renamed to `listAndDescribePods` to print the pod describe info, which shows the reason for pod restarts.

The purpose of the change is to capture pods describe info before RSync is deleted or reset.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -181,23 +181,23 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 
 	nt.detectGKEAutopilot(opts.SkipAutopilot)
 
-	// Print container logs in its own cleanup block to catch fatal errors from
-	// tests and test setup (including resetSyncedRepos).
 	t.Cleanup(func() {
 		if t.Failed() {
 			nt.printTestLogs()
 		}
-	})
-	t.Cleanup(func() {
+		// Reset after `printTestLogs` to catch pods status and logs before resetting.
 		nt.T.Log("[RESET] SharedTestEnv after test")
 		if err := Reset(nt); err != nil {
 			nt.T.Errorf("[RESET] Failed to reset test environment: %v", err)
+			// Print test logs if reset fails.
+			nt.printTestLogs()
 		}
 	})
 
 	nt.T.Log("[RESET] SharedTestEnv before test")
 	if err := Reset(nt); err != nil {
 		nt.T.Fatalf("[RESET] Failed to reset test environment: %v", err)
+		// No need to print test logs here because the cleanup block will print them.
 	}
 	// a previous e2e test may stop the Config Sync webhook, so always call `installWebhook` here to make sure the test starts
 	// with the webhook enabled.
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -467,11 +467,7 @@ func (nt *NT) printTestLogs() {
 	nt.testLogs(true)
 	nt.T.Log("[CLEANUP] Printing test logs for running pods")
 	for _, ns := range CSNamespaces {
-		nt.testPods(ns)
-	}
-	nt.T.Log("[CLEANUP] Describing not-ready pods")
-	for _, ns := range CSNamespaces {
-		nt.describeNotRunningTestPods(ns)
+		nt.listAndDescribePods(ns)
 	}
 }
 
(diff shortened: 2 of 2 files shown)


e2e/nomostest/reset.go
```diff
@@ -478,34 +479,59 @@ func initialCommit(nt *NT, repoType gitproviders.RepoType, nn types.NamespacedNa
...
-	// the reconciler container exited.
-	if err != nil && err.Error() != "signal: killed" {
-		// We're only using this for debugging, so don't trigger test failure.
-		nt.T.Logf("Failed to tail logs from reconciler %s: %v", reconcilerNN, err)
-	}
-	// When TailReconcilerLogs is called async, it's possible that the test fails
+// The pods and logs will only be printed if the test has failed when the
+// command exits.
+// Run in a goroutine to capture logs in the background while deleting RSyncs.
+func TailCSPodsAndReconcilerLogs(ctx context.Context, nt *NT, reconcilerNNs ...types.NamespacedName) {
```
sdowell, line 490: I don't think this will have the desired output since this is done sequentially. The pod get/describe is done immediately at the beginning of the function/goroutine, which I expect won't have the desired information (we probably want to describe after entering an error state). The log tailing is also being done sequentially, which I expect means we won't actually capture the logs for each reconciler. By the time the first iteration of the loop stops tailing logs for the first deployment, the other deployments/pods may have been deleted. To tail logs for each reconciler deployment/pod, I expect we would need a goroutine for each reconciler.

e2e/nomostest/reset.go
```diff
@@ -478,34 +479,59 @@ func initialCommit(nt *NT, repoType gitproviders.RepoType, nn types.NamespacedNa
...
+		if err != nil && err.Error() != "signal: killed" {
+			nt.T.Log("error running `kubectl describe pods`:", err)
+		}
+		logs = append(logs, string(out))
+	}
+	nt.T.Log("finished tailing pods info")
+
+	// Tail reconciler logs
+	for _, reconcilerNN := range reconcilerNNs {
+		logs = append(logs, fmt.Sprintf("kubectl logs deployment/%s -n %s -c %s -f",
```
karlkfi, line 511: suggest putting all the args in an array and joining them for the log. that way you don't need to duplicate building the command. you can do the same above too.

e2e/nomostest/reset.go
```diff
@@ -478,34 +479,59 @@ func initialCommit(nt *NT, repoType gitproviders.RepoType, nn types.NamespacedNa
...
-		// We're only using this for debugging, so don't trigger test failure.
-		nt.T.Logf("Failed to tail logs from reconciler %s: %v", reconcilerNN, err)
-	}
-	// When TailReconcilerLogs is called async, it's possible that the test fails
+// The pods and logs will only be printed if the test has failed when the
+// command exits.
+// Run in a goroutine to capture logs in the background while deleting RSyncs.
+func TailCSPodsAndReconcilerLogs(ctx context.Context, nt *NT, reconcilerNNs ...types.NamespacedName) {
+	var logs []string
+	// Tail CS pods
```
karlkfi, line 492: this is just listing & describing pods, not tailing, them.

e2e/nomostest/reset.go
sdowell: I think the simplest solution for log tailing would probably be to just keep this function scoped to a single reconcilerNN

nan-yu: That makes sense. So for N reconcilers, we'll need N+1 goroutines to get and describe the current pods, and then tail each reconciler log.

e2e/nomostest/reset.go
nan-yu: SG

e2e/nomostest/reset.go
nan-yu: As Sam pointed out, we will need N+1 separate goroutines to list and describe pods and tail reconciler logs. I'll create separate functions with more accurate names.

e2e/nomostest/reset.go
nan-yu: Per offline discussion, we decided to print test logs before resetting by merging the two cleanup blocks into a single one.

e2e/nomostest/new.go
```diff
@@ -181,23 +181,23 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 
 	nt.detectGKEAutopilot(opts.SkipAutopilot)
 
-	// Print container logs in its own cleanup block to catch fatal errors from
-	// tests and test setup (including resetSyncedRepos).
 	t.Cleanup(func() {
 		if t.Failed() {
```
sdowell, line 185: Ah I was thinking the logic would look a bit differently here, such that we wouldn't print the logs again below if they were already printed here. With this implementation we could probably put this `printTestLogs` in a separate cleanup block, but I think this seems good. It seems reasonable to print the logs again if reset failed to help diagnose cleanup errors

e2e/nomostest/reset.go
nan-yu: Renamed `testPods` to `listAndDescribePods`

e2e/nomostest/reset.go
nan-yu: Renamed `testPods` to `listAndDescribePods`

e2e/nomostest/nt.go
```diff
@@ -467,7 +467,7 @@ func (nt *NT) printTestLogs() {
 	nt.testLogs(true)
 	nt.T.Log("[CLEANUP] Printing test logs for running pods")
 	for _, ns := range CSNamespaces {
-		nt.testPods(ns)
+		nt.listAndDescribePods(ns)
 	}
 	nt.T.Log("[CLEANUP] Describing not-ready pods")
 	for _, ns := range CSNamespaces {
```
sdowell, line 473: Looks like the `describeNotRunningTestPods` is mostly redundant now, since we always describe all pods above

sdowell (approved):

e2e/nomostest/new.go
nan-yu: I thought you suggested putting them in the same cleanup block to make the control flow clearer.

The pupose of printing logs for both test failures and cleanup errors is to get the diagnosis in one shot. The reset should be resilient enough to proceed with the next test case even if the current one fails.

e2e/nomostest/nt.go
nan-yu: done

PR #862, opened 2023-09-09
Merge reset and printTestLogs into a single cleanup block (#861)

nan-yu:
Previously, printTestLogs is in its own cleanup block to catch fatal errors from the reset block.

Now we decided to reset after `printTestLogs`, so we can combine them into a single block for simplicity.

It also prints test logs if reset fails.

The `testPods` function is also renamed to `listAndDescribePods` to print the pod describe info, which shows the reason for pod restarts.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -181,23 +181,23 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 
 	nt.detectGKEAutopilot(opts.SkipAutopilot)
 
-	// Print container logs in its own cleanup block to catch fatal errors from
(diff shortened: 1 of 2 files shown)


nan-yu:
Cherrypicking #861 to get more logs for the flaky test on the v1.16 branch:
- kpt-config-sync-autopilot-rapid-release: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-rapid-release/1700247289735417856
- kpt-config-sync-standard-stable-release: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-stable-release/1700269540115484672

haiyanmeng (approved):

PR #865, opened 2023-09-11
test: Fix TestComposition to allow status changes

karlkfi:
TestComposition was testing ResourceVersion, when it should have been testing Generation. Status changes should be allowed and expected when changing the spec. Should make this test less flakey, especially on autopilot where node pool updates are common and can cause eviction & rescheduling of the reconciler and reconciler-manager.


nan-yu (approved):
/lgtm

PR #868, opened 2023-09-12
Nomos CLI: replace APIGROUP with APIVERSION

victorpras:
`APIGROUP` column name has been [replaced](https://github.com/kubernetes/kubernetes/pull/95253) from Kubernetes v1.20 onwards with `APIVERSION`. This has broken the usage of `api-resources.txt` with Nomos command for users using Kubernetes v1.20 or newer. 

[b/299337751](http://b/299337751)

### DIFF
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -29,6 +29,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/klog/v2"
@@ -1216,13 +1217,10 @@ func TestApiResourceFormatting(t *testing.T) {
 
 	header := strings.Fields(strings.Split(string(out), "\n")[0])
 
-	if len(header) != len(columnName) {
-		nt.T.Fatalf("Number of column titles doesn't match expected result")
-	}
+	assert.Equal(t, len(columnName), len(header))
 
 	for i, column := range header {
-		if column != columnName[i] {
-			nt.T.Fatalf("%s not found in the column titles", column)
-		}
+		assert.Equal(t, columnName[i], column)
+
 	}
 }
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -1200,3 +1200,29 @@ func TestNomosStatusNameFilter(t *testing.T) {
 		nt.T.Fatalf("Expected to not find bookinfo:crontab-sync component in output:\n%s\n", string(out4))
 	}
 }
+
+func TestApiResourceFormatting(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI)
+
+	// expected column name
+	var columnName = []string{"NAME", "SHORTNAMES", "APIVERSION", "NAMESPACED", "KIND"}
+
+	cmd := nt.Shell.Command("kubectl", "api-resources")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		nt.T.Log(string(out))
+		nt.T.Fatal(err)
+	}
+
+	header := strings.Fields(strings.Split(string(out), "\n")[0])
+
+	if len(header) != len(columnName) {
+		nt.T.Fatalf("Number of column titles doesn't match expected result")
+	}
+
+	for i, column := range header {
+		if column != columnName[i] {
+			nt.T.Fatalf("%s not found in the column titles", column)
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


pkg/vet/api_resources_test.go
```diff
@@ -48,7 +48,7 @@ rolebindings                                   rbac.authorization.k8s.io      tr
 		},
 		{
 			name: "invalid line",
-			lines: `NAME                              SHORTNAMES   APIGROUP                       NAMESPACED   KIND
+			lines: `NAME                              SHORTNAMES   APIVERSION                       NAMESPACED   KIND
 namespaces                        ns                                          other        Namespace
 deployments                       deploy       apps                           true         Deployment
 clusterroles                                   rbac.authorization.k8s.io      false        ClusterRole
```
janetkuo, line 54: This should have version information, e.g. `rbac.authorization.k8s.io/v1`. It'd also be good to update the test so that we catch issues like this earlier b/299119028#comment13. Might be enough to just check the header line of `kubectl api-resources` to see if it still produces what `nomos` expects.

victorpras: the test case has been updated to include version information to resemble the current api-resource output. The unit test has covered a scenario that checks for the header line, I added another scenario that will closely resemble the APIGROUP/APIVERSION switch

pkg/vet/api_resources_test.go
```diff
@@ -47,23 +47,33 @@ rolebindings                                   rbac.authorization.k8s.io      tr
 			wantErr: nil,
 		},
 		{
-			name: "invalid line",
+			name: "incorrect column name",
 			lines: `NAME                              SHORTNAMES   APIGROUP                       NAMESPACED   KIND
+namespaces                        ns                                          false        Namespace
+deployments                       deploy       apps                           true         Deployment
```
janetkuo, line 53: `apps` is group name instead of group version. Suggest using real output from `kubectl` instead of writing it manually. Suggest having a test to catch future breaking changes from `kubectl api-resources`. The easiest test could be validating the output column names.

victorpras: real output from kubectl is 70+ lines long. I've selected a few of them that are similar to the current test data and added a few more. All test data are now real output from `kubectl api-resoureces`.

Also added e2e test that check the column names of `kubectl api-resoureces`

pkg/vet/api_resources.go
```diff
@@ -91,8 +91,9 @@ func addLine(scoper *discovery.Scoper, path cmpath.Absolute, line string) status
 	kind := fields[len(fields)-1]
 	namespaced := fields[len(fields)-2]
 	group := ""
-	if len(fields) > 2 {
-		group = fields[0]
+	if len(fields) > 2 && len(fields[0]) > 2 {
```
janetkuo, line 94: What's the `len(fields[0]) > 2` checking?

pkg/vet/api_resources.go
```diff
@@ -91,8 +91,9 @@ func addLine(scoper *discovery.Scoper, path cmpath.Absolute, line string) status
...
+		field := strings.Split(fields[0], "/")
+		group = field[0]
```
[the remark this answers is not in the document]
victorpras: if the `APIVERSION` is just `v1`, then the entry for API group will be empty (`""`) as declared in https://github.com/GoogleContainerTools/kpt-config-sync/pull/868/files#diff-d95c9ef5c2ae2bcee59a43bcbb92b0026aef90af3f2b589d71dd883079787699L93. 

What's not covered by the test right now is API version entry, is that something that needs to be added in this changes?

pkg/vet/api_resources.go
victorpras: as described in https://github.com/GoogleContainerTools/kpt-config-sync/pull/868/files#diff-d95c9ef5c2ae2bcee59a43bcbb92b0026aef90af3f2b589d71dd883079787699R83, fields is APIVERSION, NAMESPACED, and KIND. `len(fields[0]) > 2` is intended to check whether APIVERSION is just `v1` or contains group like `apps/v1`. I'll update the code to be more clear and add comments to explain

e2e/testcases/cli_test.go
```diff
@@ -1200,3 +1200,29 @@ func TestNomosStatusNameFilter(t *testing.T) {
...
+
+	if len(header) != len(columnName) {
+		nt.T.Fatalf("Number of column titles doesn't match expected result")
+	}
+
+	for i, column := range header {
+		if column != columnName[i] {
+			nt.T.Fatalf("%s not found in the column titles", column)
+		}
+	}
```
nan-yu, line 1227: nit: You can replace the comparison with `assert.Equal(t, columnName, header)`.

pkg/vet/api_resources.go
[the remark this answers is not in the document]
nan-yu: group will only be set when the field contains `/`. In the case of `Namespace`, group is still empty, so it looks ok.

nan-yu:
only one minor nit, otherwise looks good.

e2e/testcases/cli_test.go
victorpras: replaced using assert

nan-yu: You can compare the list directly, instead of comparing the length and elements.

nan-yu (approved):
/lgtm

Issue #207, opened 2022-10-10
Support for multiple RootSyncs managing resources in the same implicit namespace


tomasaschan:
We have a setup where we have multiple `RootSync`s fighting for control over the same namespace. Simplified, the setup looks something like this:

* A `RootSync` named `sync-a` declares `namespace-a` explicitly
* A `RootSync` named `sync-x` declares resources in `namespace-a`, but not `namespace-a` itself
* A `RootSync` named `sync-y` also declares resources in `namespace-a`, but not `namespace-a` itself
* A `RootSync` named `root` declares all the above `RootSync`s

Since we create all of `sync-{a,x,y}` with git-ops (managed by `root`) we don't have any control of the order in which they are reconciled. _If_ either of `sync-x` or `sync-y` manage to create `namespace-a` before `sync-a` does, then they will assume ownership, and `sync-a` will no longer be allowed to create it (and as a result, all its dependent resources also fail to be created).

(Creating this as a separate issue to have something to refer to from a design doc I'm adding as a companion to #191.)

sdowell:
@tomasaschan Is your setup using the admission webhook?

mikebz:
@tomasaschan thinking out loud here - is it possible to change sync-x and sync-y to be RepoSyncs?  Also we do have a dependsOn annotation that can help with sequencing here.

tomasaschan:
@mikebz I'm on parental leave until after summer, so I'm a bit out of the loop here. I recommend you reach out to the team in the config mgmt SIG.

tomasaschan:
Back from parental leave!

And in short:

* No, changing them to be `RepoSync`s is not an option here. A concrete use case we have is that we install Prometheus in `prometheus-system` using one `RootSync` (which declares the namespace explicitly, corresponding to `sync-x`  in the example), and then we have other infrastructure components which install cluster-scoped components where we also want to add a `ServiceMonitor` in the `prometheus-system` namespace; we'd rather have the `ServicveMonitor` resources declared and managed together with the thing they monitor, rather than with the Prometheus infra.

* No, dependencies doesn't help here since it doesn't support depending on things managed in different `*Sync`s.

sdowell:
Implemented by https://github.com/GoogleContainerTools/kpt-config-sync/pull/883

PR #869, opened 2023-09-12
Proposal: Explicit-only namespace strategy for RootSyncs

tomasaschan:
This proposal suggests adding support for an annotation `configsync.gke.io/namespace-strategy`, with values `implicit` (the default) or `explicit`. By setting the annotation to `explicit`, setups like #207 become possible without resource fights.

### DIFF
--- a/docs/design-docs/01-explicit-only-namespace-strategy.md
+++ b/docs/design-docs/01-explicit-only-namespace-strategy.md
@@ -70,7 +70,7 @@ spec:
   # ...
 ```
 
-Under the explicit strategy, the behavior for a `RootSync` `sync-a` can be understand as follows:
+Under the explicit strategy, the behavior for a `RootSync` `sync-a` can be understood as follows:
 
 | Namespace declared in `sync-a` | Namespace managed elsewhere | Behavior                                                                                                      |
 | ------------------------------ | --------------------------- | ------------------------------------------------------------------------------------------------------------- |
--- a/docs/design-docs/01-explicit-only-namespace-strategy.md
+++ b/docs/design-docs/01-explicit-only-namespace-strategy.md
@@ -41,8 +41,8 @@ the namespace will also fail to be created).
 
 ## Design Overview
 
-The idea is to introduce an annotation that can be set on a `RootSync` which defines how to handle
-creation of namespaces: `configsync.gke.io/namespace-strategy`. The possible values are
+The idea is to introduce a new field on `spec.overrides` for a `RootSync` which defines how to handle
+creation of namespaces. The possible values are
 
 * `implicit` (the default): same behavior as today, including the failure modes triggering this proposal
 * `explicit`: namespaces must either exist in the cluster, or be declared in the configuration managed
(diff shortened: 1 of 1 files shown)


docs/design-docs/01-explicit-only-namespace-strategy.md
```diff
@@ -0,0 +1,101 @@
...
+  managed with a `RootSync`)
+
+If _either_ of `sync-x` or `sync-y` create `namespace-a` first, they will assume ownership of the
+namespace, and `sync-a` will not be allowed to create it (and, as a result, all its resources inside
+the namespace will also fail to be created).
+
+### Objective
+
+* The configuration described above should be supported without resource fights or reconciliation errors
+* Users who depend on implicit namespaces without also declaring them explicitly should need ot make
```
sdowell, line 39: typo: `need ot make` -> `need to make`

sdowell (commented):
/ok-to-test

docs/design-docs/01-explicit-only-namespace-strategy.md
```diff
@@ -0,0 +1,101 @@
+# Explicit-only Namespace Strategy for `RootSync`s
+
+* Author(s): @tomasaschan
+* Approver: TBD
```
sdowell, line 4: LGTM

There was some discussion about potentially implementing this as a spec field rather than an annotation. Assigning @janetkuo as approver so she can weigh in and approve the API decision.

/assign @janetkuo

sdowell: /lgtm

docs/design-docs/01-explicit-only-namespace-strategy.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Design Overview
+
+The idea is to introduce an annotation that can be set on a `RootSync` which defines how to handle
+creation of namespaces: `configsync.gke.io/namespace-strategy`. The possible values are
+
+* `implicit` (the default): same behavior as today, including the failure modes triggering this proposal
+* `explicit`: namespaces must either exist in the cluster, or be declared in the configuration managed
+  by the `RootSync`; no implicit namespace creation happens.
+* (not set: same as `implicit`)
```
janetkuo, line 50: Suggest making it a field to be more discoverable. It can be implemented as RootSync's `spec.override.namespaceStrategy`.

janetkuo (commented):
Thanks! The design looks good in general, except for one comment about making it a field instead of an annotation.

docs/design-docs/01-explicit-only-namespace-strategy.md
tomasaschan: Sure!

I suggested an annotation mainly based on [this comment thread](https://docs.google.com/document/d/1QK-vMQkcjmgKqaqI7fBBpejwr2eJsWu7lzO1q3PQqe4/edit?disco=AAAAu_xRu98) but I don't have any strong opinions either way. I'll update to use a spec field instead.

janetkuo (approved):

karlkfi:
lgtm

Are you planning to implement. @tomasaschan?

tomasaschan:
Yes; we haven't planned the work yet, but I expect to be able to start working on it within a week or two.

It's not _important_ to me to do it myself, so if you or your team would rather do it yourself, that's fine with me too - just let me know. It is, however, important to us that this feature is merged into main (a published release would be nice, but we'll build from source if we must) with enough time for us to roll it out across all our clusters before 1.24 is EOL. I suspect the best way to guarantee that is to do the work myself :)

karlkfi:
Looks like Sam beat you to it. https://github.com/GoogleContainerTools/kpt-config-sync/pull/883

PR #870, opened 2023-09-12
fix: TestCRDDeleteBeforeRemoveCustomResourceV1

karlkfi:
Validate management conflict metric separately from source error. The conflict should be recorded with the first commit, and then not recorded with the second commit, because the remediator will be paused until the source error is resolved.


nan-yu (approved):
/lgtm

PR #874, opened 2023-09-12
test: Use partial metadata when counting lists

karlkfi:
This should slightly speed up listing in stress tests and reduce memory allocation.


nan-yu (approved):
/lgtm

PR #872, opened 2023-09-12
fix: Increase reconciler resources on autopilot

karlkfi:
- Increase requests & limits on autopilot to avoid OOMKill and
  CPU throttling. Leave standard defaults alone, to continue
  using boosting above the minimal requests.
- Clean up createOrPatchDeployment impl
- Use a fake autopilot webhook to test autopilot enablement
- Add TestStressMemoryUsage to test maximum number of supported objects with 100 CRDs and 50 objects each

Depends On:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/876
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/882
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/910

Proposal: [go/config-sync-reconciler-profiling](http://goto.google.com/config-sync-reconciler-profiling)

b/300294367

PR #876, opened 2023-09-13
chore: Configure container resources at runtime

karlkfi:
- Make askpass sidecar resources and log level configurable using overrides, like the other containers
- Hard-code a map of container resources and use it for both impl and tests, to simplify tests and make them less fragile.
- No changes were made to configured resources, just how they are specified.
- Log reconciler container resources on e2e test failure

Fixes: b/295952540, b/295952438

Unblocks: https://github.com/GoogleContainerTools/kpt-config-sync/pull/872


e2e/nomostest/testpredicates/predicates.go
```diff
@@ -250,36 +249,98 @@ func HasExactlyImage(containerName, expectImageName, expectImageTag, expectImage
...
 		}
-		if !equality.Semantic.DeepEqual(container.Resources.Requests[corev1.ResourceMemory], memoryRequest) {
-			return errors.Errorf("The memory request of the %q container should be %v, got %v", container.Name, memoryRequest, container.Resources.Requests[corev1.ResourceMemory])
+		d, ok := o.(*appsv1.Deployment)
+		if !ok {
+			return WrongTypeErr(d, &appsv1.Deployment{})
 		}
-		if !equality.Semantic.DeepEqual(container.Resources.Limits[corev1.ResourceMemory], memoryLimit) {
-			return errors.Errorf("The memory limit of the %q container should be %v, got %v", container.Name, memoryLimit, container.Resources.Limits[corev1.ResourceMemory])
+		for _, container := range d.Spec.Template.Spec.Containers {
```
sdowell, line 299: This would pass if the actual deployment has no containers (or fewer than the expected containers). Is that intentional?

e2e/testcases/override_resource_limits_test.go
```diff
@@ -151,33 +103,44 @@ func TestOverrideReconcilerResourcesV1Alpha1(t *testing.T) {
...
 		rootReconcilerNN.Name, rootReconcilerNN.Namespace,
 		[]testpredicates.Predicate{
 			testpredicates.GenerationEquals(rootReconcilerDeploymentGeneration),
-			testpredicates.HasCorrectResourceRequestsLimits(reconcilermanager.Reconciler, resource.MustParse("500m"), resource.MustParse("800m"), resource.MustParse("400Mi"), resource.MustParse("411Mi")),
-			testpredicates.HasCorrectResourceRequestsLimits(reconcilermanager.GitSync, defaultGitSyncCPURequest, defaultGitSyncCPULimits, defaultGitSyncMemRequest, defaultGitSyncMemLimits),
+			testpredicates.DeploymentContainerResourcesEqual(updatedRootReconcilerResources),
+			testpredicates.DeploymentContainerResourcesEqual(defaultResources[reconcilermanager.GitSync]),
 		},
-		testwatcher.WatchTimeout(30*time.Second))
+		testwatcher.WatchTimeout(30*time.Second),
```
sdowell, line 122: I suspect 30 seconds may not always be long enough for autopilot

e2e/nomostest/testpredicates/predicates.go
karlkfi: The default map includes all the possible containers, so we don't really want to require them all to match, otherwise we have to filter & copy the map being used for every assertion. 

I think it's probably fine to just test the resources and assume the right containers exist and are tested by other tests.

e2e/testcases/override_resource_limits_test.go
```diff
@@ -234,26 +197,39 @@ func TestOverrideReconcilerResourcesV1Alpha1(t *testing.T) {
...
-		testpredicates.HasCorrectResourceRequestsLimits(reconcilermanager.Reconciler, resource.MustParse("500m"), resource.MustParse("800m"), resource.MustParse("400Mi"), resource.MustParse("411Mi")),
-		testpredicates.HasCorrectResourceRequestsLimits(reconcilermanager.GitSync, defaultGitSyncCPURequest, defaultGitSyncCPULimits, defaultGitSyncMemRequest, defaultGitSyncMemLimits))
+		testpredicates.DeploymentContainerResourcesEqual(updatedRootReconcilerResources),
+		testpredicates.DeploymentContainerResourcesEqual(defaultResources[reconcilermanager.GitSync]),
+	)
 	if err != nil {
 		nt.T.Fatal(err)
 	}
 
+	updatedBackendReconcilerResources := v1beta1.ContainerResourcesSpec{}
```
sdowell, line 207: Is this snippet just testing schema conversion? Curious if that needs to be an e2e test or if it could be a unit test

e2e/testcases/override_resource_limits_test.go
karlkfi: This is just the Deployment spec update, not the Pod update. The reconciler Deployment should get updated immediately by the reconciler-manager when the RootSync spec update event is received. So I don't know if being on autopilot matters, unless the reconciler-manager itself is being rescheduled.

e2e/testcases/override_resource_limits_test.go
karlkfi: The default map function returns v1beta1 objects, so we have to convert them to alpha to do the validation in this test.

We could make a function to return v1alpha1 defaults, but I didn't want to make the binary depend on the conversion code. So far, we're only using the conversion code in tests.

e2e/nomostest/testpredicates/predicates.go
karlkfi: Or do you think I should add a helper function to filter the map by a specific set of expected containers before passing it to the predicate?

e2e/testcases/reconciler_manager_test.go
```diff
@@ -772,389 +712,276 @@ func firstContainerCPURequestIs(cpuRequest int64) testpredicates.Predicate {
...
+	expectedResources[firstContainerName] = updated
 	if nt.IsGKEAutopilot {
-		// hydration-controller input: 190m/200Mi
-		// with autopilot adjustment, CPU of container[0]/hydration-controller is increased to 430m
-		// bringing the total to 500m
-		expectedTotalCPU += autopilotCPUIncrements
-		expectedFirstContainerCPU = expectedFirstContainerCPU2
-		expectedFirstContainerMemory = expectedFirstContainerMemory2
+		// Round up to the nearest multiple of 250m.
+		updated := expectedResources[firstContainerName]
```
sdowell, line 835: These snippets where we calculate the expected requests/limits are a bit hard to follow IMO. Assuming the expected value is constant for standard and autopilot, can we inline these rather than doing math in the test to evaluate the expected value? I think this would help with readability and also avoids potential bugs in the test's math. I would suggest inlining the values rather than declaring a const elsewhere like before, as the variables also make it a bit hard to read. WDYT?

e2e/testcases/override_resource_limits_test.go
sdowell: Yeah my main concern is in the case of reconciler-manager being rescheduled. Perhaps unlikely but could be a source of flakiness

e2e/nomostest/testpredicates/predicates.go
sdowell: Yeah I think it would be a stronger assertion if we ensured the expected containers are in the deployment. I like the suggestion of a helper function or a parameter to filter by a list of expected containers

e2e/testcases/reconciler_manager_test.go
karlkfi: > Assuming the expected value is constant for standard and autopilot, can we inline these rather than doing math in the test to evaluate the expected value?

e2e/nomostest/testpredicates/predicates.go
karlkfi: added sorted container name comparison

e2e/testcases/override_resource_limits_test.go
karlkfi: removed

e2e/testcases/override_resource_limits_test.go
karlkfi: added helper

sdowell (approved):

PR #877, opened 2023-09-13
Stop rebuilding the kubekins-e2e image in Config Sync

nan-yu:
Config Sync used to build a customized kubekinds-e2e-with-kind image based on the k8s staging image. Now PR#781 changes it to install kind at runtime, so we can use the k8s image directly without rebuilding it.

It also adds `build-gke-e2e` as a dependency of `push-gke-e2e` so that we only need to run a single command to push the image.


Makefile.oss.prow
```diff
@@ -52,7 +32,7 @@ build-gke-e2e:
 		.
 
 .PHONY: push-gke-e2e
-push-gke-e2e:
+push-gke-e2e: build-gke-e2e
```
sdowell, line 35: Personally I prefer the flexibility of being able to build/push separately, so that I can sanity check before pushing

nan-yu: I have no strong preference. Reverted this part.

sdowell (approved):
I think there are motivations to continue maintaining our own Dockerfile/image, but I don't feel too strongly

PR #878, opened 2023-09-13
Update go version in the test images

nan-yu:
We need to update the go version to fix CVEs.
This commit updates the test images first, so that we can build and push the gke-e2e image from a commit on the main branch.

We'll have a following commit to update go versions used by Config Sync components.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ TEST_INFRA_REGISTRY ?= $(LOCATION)-docker.pkg.dev/$(TEST_INFRA_PROJECT)/test-inf
 
 # Docker image used for build and test. This image does not support CGO.
 # When upgrading this tag, publish the image after the change is submitted.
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -93,7 +93,7 @@ TEST_INFRA_PROJECT ?= kpt-config-sync-ci-artifacts
 TEST_INFRA_REGISTRY ?= $(LOCATION)-docker.pkg.dev/$(TEST_INFRA_PROJECT)/test-infra
 
 # Docker image used for build and test. This image does not support CGO.
-BUILDENV_IMAGE ?= $(TEST_INFRA_REGISTRY)/buildenv:v0.2.12
```
sdowell, line 96: It looks like this image needs to be manually published before the PR is submitted (since it's used by presubmits), which is a bit of an anti-pattern. I can revisit this another time and try to improve this

nan-yu: Yup, I have manually pushed the image, but still seeing presubmit failures. Looking now

sdowell (approved):

PR #879, opened 2023-09-13
ci: enable presubmits on new buildenv before publishing

sdowell:
The buildenv image is used for several presubmit targets, which means that this image needed to be published before the change is approved/submitted. This is a bit of an anti-pattern, as the change should be reviewed/approved before the image is published. This change enables the presubmits to run on a new image version before it is published, so that the image can be published after submission.


nan-yu (approved):
/lgtm

PR #881, opened 2023-09-14
fix: explicitly set makefile shell to bash

sdowell:
Some of our make targets use bash syntax, which requires for the shell used by make to be bash. Make uses sh by default, so the shell should be set explicitly here. This was a gotcha before, because some systems replace sh with bash. Explicitly setting the shell should help avoid future gotchas.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@
 
 REPO := kpt.dev/configsync
 
+# Some of our recipes use bash syntax, so explicitly set the shell to bash.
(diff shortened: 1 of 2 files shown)


sdowell:
/hold

holding temporarily as there are some changes included not intended for submission (to verify presubmit works as expected)

nan-yu (approved):
/lgtm

sdowell:
unit test failure, I suspect it's a flake with the fake client

/retest

PR #883, opened 2023-09-15
feat: add namespaceStrategy API

sdowell:
This change adds a new namespaceStrategy API field which dictates how RootSync reconcilers handle undeclared namespaces.

The default value is implicit, which is the same as the current behavior. This mode creates namespaces implicitly if they are not declared and do not exist.

The new strategy is explicit, in which namespaces will only be created by the reconciler if they are declared in the source.

Issue: https://github.com/GoogleContainerTools/kpt-config-sync/issues/207
Doc PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/869


pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -73,6 +74,19 @@ type OverrideSpec struct {
...
+	// Must be "implicit" or "explicit"
+	// "implicit" means that the reconciler will implicitly create Namespaces
+	// if they do not exist, even if they are not declared in the source.
+	// "explicit" means that the reconciler will not create Namespaces which
+	// are not declared in the source.
+	//
+	// +kubebuilder:validation:Enum=implicit;explicit
+	// +optional
+	NamespaceStrategy configsync.NamespaceStrategy `json:"namespaceStrategy,omitempty"`
+	// TODO: For reviewers - should this be a RootSync-specific field?
```
sdowell, line 89: The `OverrideSpec` is currently shared between both RootSync and RepoSync. Should we make this a RootSync-only field?

WDYT @janetkuo ?

janetkuo: Good point. Yes, this field should be RootSync only. Alternatively we could document this field as "it has no effect on RepoSync".

e2e/testcases/namespace_strategy_test.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2022 Google LLC
```
janetkuo, line 1: nit: 2023

pkg/api/configsync/v1beta1/resource_override.go
sdowell: SGTM. Do you have a preference?

e2e/testcases/namespace_strategy_test.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	// add a resource for which the namespace is not declared/created
+	fooNamespace := fake.NamespaceObject("foo-implicit")
+	cm1 := fake.ConfigMapObject(core.Name("cm1"), core.Namespace(fooNamespace.Name))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/cm1.yaml", cm1))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Add cm1"))
+	// check for error
+	nt.WaitForRootSyncSyncError(rootSyncNN.Name, applier.ApplierErrorCode,
```
karlkfi, line 63: Should we perhaps also test the use case where this error recovers automatically when the namespace is created by another client?

Another possible permutation would be to have one RootSync block/error until a second RootSync provisions the namespace it needs.

I don't know if we need all these cases coved in e2e, but they seem pretty common.

pkg/api/configsync/v1beta1/resource_override.go
karlkfi: RootSync-only, imo

e2e/testcases/namespace_strategy_test.go
sdowell: Seems reasonable - I'm thinking I can add another testcase which covers the following behavior:

- sync-1 creates a resource with namespaceStrategy=explicit, and expects an error
- sync-2 creates the namespace
- assert that both sync-1 and sync-2 successfully sync, and that the namespace is managed by sync-1

pkg/api/configsync/v1beta1/resource_override.go
sdowell: Done. It's a RootSync-only field now. I split up OverrideSpec into a separate RootSync and RepoSync struct

e2e/testcases/namespace_strategy_test.go
sdowell: done

e2e/testcases/namespace_strategy_test.go
sdowell: added another testcase which approximates the scenario described in the original issue

e2e/testcases/namespace_strategy_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/declared"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/metadata"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+	"kpt.dev/configsync/pkg/testing/fake"
+	"sigs.k8s.io/cli-utils/pkg/common"
+)
+
+func TestNamespaceStrategy(t *testing.T) {
```
karlkfi, line 38: can you add some test func descriptions that summarize what the tests do?

pkg/api/configsync/v1alpha1/zz_generated.conversion.go
```diff
@@ -138,6 +138,11 @@ func RegisterConversions(s *runtime.Scheme) error {
 	}); err != nil {
 		return err
 	}
+	if err := s.AddGeneratedConversionFunc((*v1beta1.OverrideSpec)(nil), (*OverrideSpec)(nil), func(a, b interface{}, scope conversion.Scope) error {
```
karlkfi, line 141: why is there still an OverrideSpec here? should it be RootSyncOverrideSpec or RepoSyncOverrideSpec?

karlkfi: oh, i see. you added a shared parent object for embedding. i get it.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -663,7 +663,18 @@ func (r *RootSyncReconciler) mapSecretToRootSyncs(secret client.Object) []reconc
 func (r *RootSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1beta1.RootSync, reconcilerName string) map[string][]corev1.EnvVar {
 	result := map[string][]corev1.EnvVar{
 		reconcilermanager.HydrationController: hydrationEnvs(rs.Spec.SourceType, rs.Spec.Git, rs.Spec.Oci, declared.RootReconciler, reconcilerName, r.hydrationPollingPeriod.String()),
-		reconcilermanager.Reconciler:          append(reconcilerEnvs(r.clusterName, rs.Name, rs.Generation, reconcilerName, declared.RootReconciler, rs.Spec.SourceType, rs.Spec.Git, rs.Spec.Oci, rootsync.GetHelmBase(rs.Spec.Helm), r.reconcilerPollingPeriod.String(), rs.Spec.SafeOverride().StatusMode, v1beta1.GetReconcileTimeout(rs.Spec.SafeOverride().ReconcileTimeout), v1beta1.GetAPIServerTimeout(rs.Spec.SafeOverride().APIServerTimeout), enableRendering(rs.GetAnnotations())), sourceFormatEnv(rs.Spec.SourceFormat)),
+		reconcilermanager.Reconciler: append(
+			reconcilerEnvs(
```
karlkfi, line 667: This reconcilerEnvs func should really be an object with this many fields... but i see you just tried to clean up the call a little. Maybe we can refactor it later.

sdowell: Agreed this function has too many parameters. I suggest we refactor as a separate change

e2e/testcases/namespace_strategy_test.go
sdowell: done

e2e/testcases/namespace_strategy_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"kpt.dev/configsync/pkg/testing/fake"
+	"sigs.k8s.io/cli-utils/pkg/common"
+)
+
+// This test case is primarily focused on the behavior of switching between modes
+// of namespaceStrategy for a single RootSync.
+// - Set namespaceStrategy: explicit
+// - Declare resource in "foo-implicit" namespace but not the namespace itself (should error)
+// - Set namespaceStrategy: implicit - should create the namespace implicitly
+// - Set namespaceStrategy: explicit - namespace still exists but is unmanaged
```
karlkfi, line 43: What do you think about adding a step at the end to take over the unmanaged namespace with explicit and validate that it can be deleted?

e2e/testcases/namespace_strategy_test.go
```diff
@@ -0,0 +1,211 @@
...
+			testpredicates.MissingAnnotation(metadata.ResourceManagementKey),
+			testpredicates.MissingAnnotation(metadata.ResourceManagerKey),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+}
+
+// This test case is focused on the behavior of using namespaceStrategy: explicit
```
karlkfi, line 105: Comment needs to start with TestNamespaceStrategyMultipleRootSyncs

karlkfi: surprised it passed linting without this...

sdowell: I don't think the linter requires for test functions to have the test name in the docstring

e2e/testcases/namespace_strategy_test.go
sdowell: Done. Added subsequent steps to demonstrate how the namespace can be explicitly re-managed and deleted

karlkfi: great. thanks

karlkfi (approved):

cmd/reconciler/main.go
```diff
@@ -95,25 +96,30 @@ var (
 		"Enable debug mode, panicking in many scenarios where normally an InternalError would be logged. "+
 			"Do not use in production.")
 
-	renderingEnabled = flag.Bool("rendering-enabled", util.EnvBool(reconcilermanager.RenderingEnabled, false), "")
+	renderingEnabled  = flag.Bool("rendering-enabled", util.EnvBool(reconcilermanager.RenderingEnabled, false), "")
+	namespaceStrategy = flag.String(flags.namespaceStrategy, util.EnvString(reconcilermanager.NamespaceStrategy, ""),
```
janetkuo, line 100: Why isn't the default set here?

cmd/reconciler/main.go
```diff
@@ -95,25 +96,30 @@ var (
...
+		fmt.Sprintf("Set the namespace strategy for the reconciler. Must be %s or %s.",
```
janetkuo, line 101: Should mention the default value

cmd/reconciler/main.go
```diff
@@ -207,6 +219,10 @@ func main() {
 			klog.Fatalf("Flag %s and Environment variable%q must not be passed to a Namespace reconciler",
 				flags.sourceFormat, filesystem.SourceFormatKey)
 		}
+		if *namespaceStrategy != "" {
+			klog.Fatalf("Flag %s and Environment variable%q must not be passed to a Namespace reconciler",
```
janetkuo, line 223: Shouldn't capitalize Environment, and there should be a space before %q

manifests/rootsync-crd.yaml
```diff
@@ -1427,6 +1439,18 @@ spec:
                     x-kubernetes-list-map-keys:
                     - containerName
                     x-kubernetes-list-type: map
+                  namespaceStrategy:
+                    description: namespaceStrategy controls how the reconciler handles
+                      Namespaces which are used by resources in the source but not
+                      declared. Must be "implicit" or "explicit" "implicit" means
```
janetkuo, line 1445: Missing a period here

pkg/api/configsync/v1alpha1/resource_override.go
```diff
@@ -75,6 +76,28 @@ type OverrideSpec struct {
...
+type RootSyncOverrideSpec struct {
+	OverrideSpec `json:",inline"`
+
+	// namespaceStrategy controls how the reconciler handles Namespaces
+	// which are used by resources in the source but not declared.
+	// Must be "implicit" or "explicit"
+	// "implicit" means that the reconciler will implicitly create Namespaces
+	// if they do not exist, even if they are not declared in the source.
+	// "explicit" means that the reconciler will not create Namespaces which
+	// are not declared in the source.
```
janetkuo, line 89: Need to mention the default behavior.

pkg/parse/root.go
```diff
@@ -141,7 +146,9 @@ func (p *root) parseSource(_ context.Context, state sourceState) ([]ast.FileObje
 	options = OptionsForScope(options, p.scope)
 
 	if p.sourceFormat == filesystem.SourceFormatUnstructured {
-		options.Visitors = append(options.Visitors, p.addImplicitNamespaces)
+		if p.namespaceStrategy == configsync.NamespaceStrategyImplicit {
+			options.Visitors = append(options.Visitors, p.addImplicitNamespaces)
```
janetkuo, line 150: Does this mean that the "explicit" namespace mode isn't supported in hierarchical mode?

cmd/reconciler/main.go
sdowell: I opted to follow the existing precedent in this file and set the default in the main function below. Would you prefer I refactor this to set the default value up here?

sdowell:
@janetkuo addressed your PR feedback in a followup https://github.com/GoogleContainerTools/kpt-config-sync/pull/888

cmd/reconciler/main.go
janetkuo: Yes, it's weird not to set the default directly where the flag is defined but set it separately. Setting default is already supported in the `flag.*` functions.

PR #884, opened 2023-09-15
Bump Kustomize version to fix CVE

victorpras:


nan-yu (commented):
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/hydrate/tool_util.go#L48 also needs to be updated.

victorpras:
> https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/hydrate/tool_util.go#L48 also needs to be updated.

I missed that, updated now.

nan-yu (approved):
/lgtm

PR #885, opened 2023-09-19
Store last hydration error in-memory for rehydrate

nan-yu:
This PR adds commit SHA and syncDir to the hydration error payload, so that the rehydrateOnError function can get them directly without invoking hydrate.SourceCommitAndSyncDir.

### DIFF
--- a/pkg/hydrate/controller.go
+++ b/pkg/hydrate/controller.go
@@ -77,29 +77,28 @@ func (h *Hydrator) Run(ctx context.Context) {
 	rehydrateTimer := time.NewTimer(h.RehydratePeriod)
 	defer rehydrateTimer.Stop()
 	absSourceDir := h.absSourceDir()
+	var hydrateErr HydrationError
+	var srcCommit string
+	var syncDir cmpath.Absolute
+	var err error
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-rehydrateTimer.C:
-			commit, syncDir, err := SourceCommitAndDir(h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
-			if err != nil {
-				klog.Errorf("failed to get the commit hash and sync directory from the source directory %s: %v", absSourceDir.OSPath(), err)
-			} else {
-				h.rehydrateOnError(commit, syncDir.OSPath())
-			}
+			hydrateErr = h.rehydrateOnError(hydrateErr, srcCommit, syncDir)
 			rehydrateTimer.Reset(h.RehydratePeriod) // Schedule rehydrate attempt
 		case <-runTimer.C:
-			commit, syncDir, err := SourceCommitAndDir(h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
+			srcCommit, syncDir, err = SourceCommitAndDir(h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
 			if err != nil {
 				klog.Errorf("failed to get the commit hash and sync directory from the source directory %s: %v", absSourceDir.OSPath(), err)
-			} else if DoneCommit(h.DonePath.OSPath()) != commit {
+			} else if DoneCommit(h.DonePath.OSPath()) != srcCommit {
 				// If the commit has been processed before, regardless of success or failure,
 				// skip the hydration to avoid repeated execution.
 				// The rehydrate ticker will retry on the failed commit.
-				hydrateErr := h.hydrate(commit, syncDir.OSPath())
-				if err := h.complete(commit, hydrateErr); err != nil {
-					klog.Errorf("failed to complete the rendering execution for commit %q: %v", commit, err)
+				hydrateErr = h.hydrate(srcCommit, syncDir)
+				if err := h.complete(srcCommit, hydrateErr); err != nil {
+					klog.Errorf("failed to complete the rendering execution for commit %q: %v", srcCommit, err)
 				}
 			}
 			runTimer.Reset(h.PollingPeriod) // Schedule re-run attempt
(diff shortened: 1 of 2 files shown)


pkg/hydrate/error.go
```diff
@@ -70,8 +80,43 @@ func (e TransientError) Code() string {
...
+	// SyncDir is the sync directory associated with the hydration error.
+	SyncDir string
 	// Code is the error code to indicate if it is a user error or an internal error.
 	Code string
 	// Error is the message of the hydration error.
 	Error string
 }
+
+// GetHydrationErrorPayload returns the payload of the hydration error.
+func GetHydrationErrorPayload(errorFile, reconciler string) (*HydrationErrorPayload, error) {
```
karlkfi, line 94: Maybe Read or Parse, instead of Get in the title?

pkg/hydrate/controller.go
```diff
@@ -177,18 +171,34 @@ func (h *Hydrator) hydrate(sourceCommit, syncDir string) HydrationError {
...
+	var hydrationErr HydrationError
 	errorFile := h.HydratedRoot.Join(cmpath.RelativeSlash(ErrorFile))
-	if _, err := os.Stat(errorFile.OSPath()); err != nil {
-		if !os.IsNotExist(err) {
-			klog.Warningf("unable to check the error file %s: %v", errorFile, err)
-		}
+	_, err := os.Stat(errorFile.OSPath())
+	switch {
+	case err != nil && os.IsNotExist(err):
+		// return directly if the error file doesn't exist.
```
karlkfi, line 183: Returning on error without a log message seem like a smell. Is it because it's not an error if the file doesn't exist? If so, maybe explain that in the comment.

pkg/hydrate/controller.go
```diff
@@ -177,18 +171,34 @@ func (h *Hydrator) hydrate(sourceCommit, syncDir string) HydrationError {
...
 		return
+	case err != nil:
+		// Return a hydration error to retry next time
```
karlkfi, line 186: Return? The error isn't being returned from this function... What's `complete` do with it?

pkg/hydrate/error.go
nan-yu: Changed to Parse.

pkg/hydrate/controller.go
nan-yu: If the error file doesn't exist, it means the previous hydration process succeeded.
Updated the comment.

pkg/hydrate/controller.go
nan-yu: Updated the comment. `complete` creates a done file to indicate the hydration process is done, and also exports any errors to an error file.

pkg/hydrate/controller.go
```diff
@@ -177,18 +171,35 @@ func (h *Hydrator) hydrate(sourceCommit, syncDir string) HydrationError {
...
+		// If the error file doesn't exist, it indicates the previous hydration succeeded,
+		// so return directly without re-hydrate.
 		return
+	case err != nil:
+		// If it fails to check the existence of the error file, reset the error
+		// file and create a done file via `h.complete`, so it can be retried next time.
+		hydrationErr = NewInternalError(errors.Wrapf(err, "unable to check the error file %s", errorFile))
+	default:
+		// former hydration error exists
+		errPayload, err := GetHydrationErrorPayload(errorFile.OSPath(), h.ReconcilerName)
```
sdowell, line 192: The error file here is being consumed by the same controller that is writing the file (namely hydration-controller). The filesystem storage is ephemeral, so is there a reason we need to read the error file back from the disk? It seems like this might be simpler if we stored the previously processed commit/dir/error/etc in memory for the hydration controller to use between iterations.

In other words - We need the file to communicate between containers (e.g. hydration-controller -> reconciler), but do we need to use the file to communicate within the same controller/container?

WDYT?

karlkfi (approved):
/lgtm

sdowell:
/hold

Putting a temporarily hold so that my comment can be addressed

pkg/hydrate/controller.go
nan-yu: That is a good point. We can store the hydration error in memory. I'm going to create a separate PR to address that.

sdowell: Ack. Feel free to remove my hold if you'd prefer to address this in a separate PR

nan-yu: Looking more into this PR. If we read the previous error from memory, we don't need to add sourceCommit and syncDir to the hydration error payload.
Let me address it here.

sdowell: Sounds good 👍

nan-yu: done

pkg/hydrate/controller.go
```diff
@@ -77,29 +77,28 @@ func (h *Hydrator) Run(ctx context.Context) {
...
-			if err != nil {
-				klog.Errorf("failed to get the commit hash and sync directory from the source directory %s: %v", absSourceDir.OSPath(), err)
-			} else {
-				h.rehydrateOnError(commit, syncDir.OSPath())
-			}
+			hydrateErr = h.rehydrateOnError(hydrateErr, srcCommit, syncDir)
 			rehydrateTimer.Reset(h.RehydratePeriod) // Schedule rehydrate attempt
 		case <-runTimer.C:
-			commit, syncDir, err := SourceCommitAndDir(h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
+			srcCommit, syncDir, err = SourceCommitAndDir(h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
```
sdowell, line 92: Do we want to update `srcCommit` and `syncDir` if err != nil here? I'm thinking we might only want to set the value if we successfully read a new source commit/dir.

In other words, the alternative I'm suggesting is assigning `srcCommit` and `syncDir` inside the if statement starting on line `95` and assign a temporary/scoped variable on line 92 here.

nan-yu: If we don't reset `srcCommit` and `syncDir` when err != nil, they may still point to older source for rehydrate.

I think we do need to reset them when err != nil, but we need to handle empty values in rehydrateOnError.

nan-yu: Done. Added a check for `if len(prevSrcCommit) == 0 || len(prevSyncDir) == 0` in the `rehydrateOnError` function.

sdowell: LGTM, thank you!

sdowell (approved):

PR #886, opened 2023-09-19
Add internal retry for SourceCommitAndDir and readHydratedDir

nan-yu:
If the internal retry fails eventually, it returns an internal error and surfaces the error in the R*Sync status field.

### DIFF
--- a/pkg/hydrate/controller.go
+++ b/pkg/hydrate/controller.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
@@ -32,6 +33,7 @@ import (
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/status"
+	"kpt.dev/configsync/pkg/util"
 )
 
 const (
@@ -89,9 +91,16 @@ func (h *Hydrator) Run(ctx context.Context) {
 			hydrateErr = h.rehydrateOnError(hydrateErr, srcCommit, syncDir)
 			rehydrateTimer.Reset(h.RehydratePeriod) // Schedule rehydrate attempt
 		case <-runTimer.C:
-			srcCommit, syncDir, err = SourceCommitAndDir(h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
+			// pull the source commit and directory with retries within 5 minutes.
+			srcCommit, syncDir, err = SourceCommitAndDirWithRetry(util.SourceRetryBackoff, h.SourceType, absSourceDir, h.SyncDir, h.ReconcilerName)
 			if err != nil {
-				klog.Errorf("failed to get the commit hash and sync directory from the source directory %s: %v", absSourceDir.OSPath(), err)
+				hydrateErr = NewInternalError(errors.Wrapf(err,
+					"failed to get the commit hash and sync directory from the source directory %s",
+					absSourceDir.OSPath()))
+				if err := h.complete(srcCommit, hydrateErr); err != nil {
+					klog.Errorf("failed to complete the rendering execution for commit %q: %v",
+						srcCommit, err)
+				}
 			} else if DoneCommit(h.DonePath.OSPath()) != srcCommit {
 				// If the commit has been processed before, regardless of success or failure,
 				// skip the hydration to avoid repeated execution.
@@ -326,14 +335,34 @@ func deleteErrorFile(file string) error {
 	return nil
 }
 
-// SourceCommitAndDir returns the source hash (a git commit hash or an OCI image digest or a helm chart version), the absolute path of the sync directory, and source errors.
-func SourceCommitAndDir(sourceType v1beta1.SourceType, sourceRevDir cmpath.Absolute, syncDir cmpath.Relative, reconcilerName string) (string, cmpath.Absolute, status.Error) {
-	// Check if the source root directory is mounted
+// SourceCommitAndDirWithRetry returns the source hash (a git commit hash or an
+// OCI image digest or a helm chart version), the absolute path of the sync
+// directory, and source errors.
+// It retries with the provided backoff.
+func SourceCommitAndDirWithRetry(backoff wait.Backoff, sourceType v1beta1.SourceType, sourceRevDir cmpath.Absolute, syncDir cmpath.Relative, reconcilerName string) (commit string, sourceDir cmpath.Absolute, _ status.Error) {
+	err := util.RetryWithBackoff(backoff, func() error {
+		var err error
+		commit, sourceDir, err = SourceCommitAndDir(sourceType, sourceRevDir, syncDir, reconcilerName)
+		return err
+	})
+	// If a retriable error can't be addressed with retry, it is identified as a
+	// source error, and will be exposed in the R*Sync status.
+	return commit, sourceDir, status.SourceError.Wrap(err).Build()
+}
+
+// SourceCommitAndDir returns the source hash (a git commit hash or an OCI image
+// digest or a helm chart version), the absolute path of the sync directory,
+// and source errors.
+func SourceCommitAndDir(sourceType v1beta1.SourceType, sourceRevDir cmpath.Absolute, syncDir cmpath.Relative, reconcilerName string) (string, cmpath.Absolute, error) {
 	sourceRoot := path.Dir(sourceRevDir.OSPath())
-	if _, err := os.Stat(sourceRoot); err != nil && os.IsNotExist(err) {
-		return "", "", status.TransientError(err)
+	if _, err := os.Stat(sourceRoot); err != nil {
+		// It fails to check the source root directory status, either because of
+		// the path doesn't exist, or other OS failures. The root cause is
+		// probably because the *-sync container is not ready yet, so retry until
+		// it becomes ready.
+		return "", "", util.NewRetriableError(fmt.Errorf("failed to check the status of the source root directory %q: %v", sourceRoot, err))
 	}
-	// Check if the source configs are synced successfully.
+	// Check if the source configs are pulled successfully.
 	errFilePath := filepath.Join(sourceRoot, ErrorFile)
 
 	var containerName string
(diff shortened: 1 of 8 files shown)


nan-yu:
The first 3 commits are extracted out in #885.

pkg/parse/run.go
```diff
@@ -63,6 +64,8 @@ const (
 	RenderingNotRequired string = "Rendering not required but is currently enabled"
 )
 
+var retryBackoff = util.BackoffWithDurationLimit(10 * time.Minute)
```
sdowell, line 67: 10 minutes seems like a long time to retry. I feel like the wait time we choose here should be a reasonable amount of time for retriable errors to resolve themselves, but if we wait too long then we delay:
- the error being surfaced on the RSync status
- the `Run` loop re-entering to trigger on the next timer

Granted it's a bit arbitrary, but I feel something like 1 minute might be closer to the right value. WDYT?

pkg/hydrate/controller.go
```diff
@@ -327,60 +337,82 @@ func deleteErrorFile(file string) error {
...
+		// If the commit hash is not changed, the hydration will be skipped.
+		// Therefore, it is relatively safe to keep the Helm charts local in the source directory.
+		relSyncDir := gitDir.Join(syncDir)
+		sourceDir, err = relSyncDir.EvalSymlinks()
+		if err != nil {
+			// `relSyncDir` points to the source config directory. Now the commit SHA
+			// has been checked out, but it fails to evaluate the symlink with the
+			// config directory, which indicates a misconfiguration of `spec.git.dir`,
+			// `spec.oci.dir`, or `spec.helm.dir`, so return the error without retry.
+			return fmt.Errorf("failed to evaluate the config directory %q: %w", relSyncDir.OSPath(), err)
```
sdowell, line 410: It looks like this error is changed from a `PathError` to a `SourceError`. Is that intentional?

pkg/parse/source.go
```diff
@@ -116,33 +118,55 @@ func (o *files) sourceContext() sourceContext {
...
-		return result, hydratedError(errorFile.OSPath(),
-			fmt.Sprintf("%s=%s", metadata.ReconcilerLabel, reconciler))
-	} else if !os.IsNotExist(err) {
-		return result, hydrate.NewTransientError(errors.Wrapf(err, "failed to check the error file: %s", errorFile.OSPath()))
-	}
-	hydratedDir, err := hydratedRoot.Join(cmpath.RelativeSlash(o.HydratedLink)).EvalSymlinks()
-	if err != nil {
-		return result, hydrate.NewInternalError(errors.Wrapf(err, "unable to load the hydrated configs under %s", hydratedRoot.OSPath()))
-	}
+	err := util.RetryWithBackoff(backoff, func() error {
```
sdowell, line 123: `result` should be reset to nil values at the start of each loop, otherwise it could have partial values set by multiple retries.

I think it may help avoid this issue and improve testability/readability by moving this snippet to a dedicated function. e.g.
```golang
// readHydratedDirWithRetry returns a sourceState object whose `commit` and `syncDir` fields are set if succeeded.
func (o *files) readHydratedDirWithRetry(backoff wait.Backoff, hydratedRoot cmpath.Absolute, reconciler string, srcState sourceState) (sourceState, hydrate.HydrationError) {
	result := sourceState{}
	err := util.RetryWithBackoff(backoff, func() error {
        var error
        result, err = readHydratedDir(hydratedRoot, reconciler, srcState)
        return err
    })
...
```

pkg/hydrate/controller.go
```diff
@@ -327,60 +337,82 @@ func deleteErrorFile(file string) error {
...
-	case v1beta1.OciSource:
-		containerName = reconcilermanager.OciSync
-	case v1beta1.GitSource:
-		containerName = reconcilermanager.GitSync
-	case v1beta1.HelmSource:
-		containerName = reconcilermanager.HelmSync
-	}
+// It retries with the provided backoff.
+func SourceCommitAndDir(backoff wait.Backoff, sourceType v1beta1.SourceType, sourceRevDir cmpath.Absolute, syncDir cmpath.Relative, reconcilerName string) (commit string, sourceDir cmpath.Absolute, _ status.Error) {
+	err := util.RetryWithBackoff(backoff, func() error {
```
sdowell, line 342: I suggest a similar approach here to avoid partial values and improve readability/testability. For example:
```golang
func SourceCommitAndDirWithRetry(backoff wait.Backoff, sourceType v1beta1.SourceType, sourceRevDir cmpath.Absolute, syncDir cmpath.Relative, reconcilerName string) (commit string, sourceDir cmpath.Absolute, _ status.Error) {
	err := util.RetryWithBackoff(backoff, func() error {
        var err error
        commit, sourceDir, err = SourceCommitAndDir(sourceType, sourceRevDir, syncDir, reconcilerName)
        return err
    })
...
```

pkg/parse/run.go
```diff
@@ -185,7 +188,8 @@ func Run(ctx context.Context, p Parser) {
 func run(ctx context.Context, p Parser, trigger string, state *reconcilerState) {
```
sdowell, line 188: `run` is used by several different timers - `resyncTimer`, `runTimer`, `retryTimer`, etc. I'm curious if we want to use this same 10 minute retry logic for all of these entrypoints? It seems this could delay re-entry into the loop so that the next time can be reached. WDYT?

As a side note - this function reaches the retry logic for both `SourceCommitAndDir` and `readHydratedDir`. In the unlikely worst case, that means we could spend ~20 minutes in this function doing retries.

pkg/parse/run.go
nan-yu: Agreed. My original thought is to allow 10 internal retries before claiming it is not transient, but maybe too long. 

The retryBackoff is used twice, one for `SourceCommitAndDir` and the other one is for `readHydratedDir`.
I'm thinking we can use 5 minutes for `SourceCommitAndDir`, which allows 2 git-sync loops to be executed (based on a 120s timeout for one sync), and use 1 minute for `readHydratedDir`.

pkg/hydrate/controller.go
nan-yu: Yes. That's intentional. I think it was wrong initially. `PathError` is used for invalid YAML configuration and special characters in the path: https://cloud.google.com/anthos-config-management/docs/reference/errors#knv2001.

It makes more sense to classify the misconfiguration of `spec.git.dir` as a source error.

pkg/parse/run.go
nan-yu: `run` is used by all the timers. Theoretically, `resyncTimer` and `retryTimer` should use a shorter timeout than `runTimer` because the source and hydratedDir should already be computed. However, the `run` function with all the timers re-reads the file system, which means the source can change at any time. I think using the same retry logic guarantees the source and hydratedDir can be recomputed.

With a 5 mins retry timeout for `SourceCommitAndDir` and 1 min for `readHydratedDir`, it adds up to 6 mins in the worst case, which seems to be okay.

pkg/parse/source.go
nan-yu: done

pkg/hydrate/controller.go
nan-yu: done

pkg/parse/run.go
nan-yu: changed to 5 mins and 1 min

pkg/parse/run.go
nan-yu: changed to 5 mins and 1 min

pkg/hydrate/controller.go
```diff
@@ -347,24 +376,33 @@ func SourceCommitAndDir(sourceType v1beta1.SourceType, sourceRevDir cmpath.Absol
...
-			metadata.ReconcilerLabel, reconcilerName).Build()
-	} else if err == nil {
-		return "", "", status.SourceError.Sprintf("error in the %s container: %s", containerName, string(content)).Build()
+			metadata.ReconcilerLabel, reconcilerName)
+	case err == nil && len(content) != 0:
+		// The source error file exists, which indicates the *-sync container is
+		// ready, so return the error directly without retry.
+		return "", "", fmt.Errorf("error in the %s container: %s", containerName, string(content))
+	default:
+		// The source error file doesn't exist => *-sync is ready without any errors.
```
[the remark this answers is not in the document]
nan-yu: The `*-sync` container is ready at this point because the `sourceRoot` directory is created. The absence of the error file indicates there is no error **so far**, but it doesn't imply the sync succeeds.

When the error file doesn't exist, it checks the symlink next. If the symlink doesn't exist, it retries. The retry timeout is longer than a single sync timeout (2 mins), so the error file should exist in the second loop.

I'll clarify it in the comment as below:
```
// The sourceRoot directory exists, but the source error file doesn't exist.
// It indicates that *-sync is ready, but no errors so far.
```

Does it look good to you?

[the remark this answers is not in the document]
sdowell: LGTM, thank you!

pkg/parse/run_test.go
```diff
@@ -326,23 +358,27 @@ func TestRun(t *testing.T) {
...
-			testutil.AssertEqual(t, tc.expectedErrors, actualErrs, "[%s] unexpected state.cache.errs return", tc.name)
+
 			testutil.AssertEqual(t, tc.expectedStateSourceErrs, state.sourceStatus.errs, "[%s] unexpected state.sourceStatus.errs return", tc.name)
 			testutil.AssertEqual(t, tc.expectedStateRenderingErrs, state.renderingStatus.errs, "[%s] unexpected state.renderingStatus.errs return", tc.name)
 
 			rs := &v1beta1.RootSync{}
 			if err = parser.options().client.Get(context.Background(), rootsync.ObjectKey(parser.options().syncName), rs); err != nil {
 				t.Fatal(err)
 			}
-			testutil.AssertEqual(t, tc.expectedRSSourceErrs, rs.Status.Source.Errors, "[%s] unexpected source errors in RootSync return", tc.name)
```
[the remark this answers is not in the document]
nan-yu: Before the change, not all errors cached in `state.sourceStatus.errs` or `state.renderingStatus.errs` are surfaced in the R*Sync status, e.g. transient errors are skipped.

Now, transient errors are retried internally, and are stored in `state.***Status.errs` and surfaced in R*Sync status if they eventually failed. 

`testutil.AssertEqual(t, tc.expectedStateSourceErrs, state.sourceStatus.errs, "[%s] unexpected state.sourceStatus.errs return", tc.name)` asserts that the actual cached error is the same as expected. This assertion ensures the surfaced error is the same as cached error.

I can create a separate field for the expected R*Sync status error, but it basically is built on top of `tc.expectedStateSourceErrs`.

pkg/parse/run_test.go
```diff
@@ -271,45 +276,72 @@ func TestRun(t *testing.T) {
...
+			util.HydratedRetryBackoff = util.SourceRetryBackoff
 
 			rootDir := filepath.Join(tempDir, tc.id)
-			if err = os.Mkdir(rootDir, os.ModePerm); err != nil {
-				t.Fatal(err)
-			}
 			sourceRoot := filepath.Join(rootDir, "source")     // /repo/source
 			hydratedRoot := filepath.Join(rootDir, "hydrated") // /repo/hydrated
 			sourceDir := filepath.Join(sourceRoot, symLink)
+			go func() {
```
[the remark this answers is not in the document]
nan-yu: done

pkg/hydrate/controller_test.go
```diff
@@ -35,6 +37,166 @@ const (
...
+				Cap:      tc.retryCap,
+				Jitter:   0.1,
+			}
+
+			srcRootDir := filepath.Join(os.TempDir(), "test-srcCommit-syncDir", rand.String(10))
+			commitDir := filepath.Join(srcRootDir, commit)
+			if len(tc.syncDir) == 0 {
+				tc.syncDir = syncDir
+			}
+			defer func() {
```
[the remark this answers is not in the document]
nan-yu: done

pkg/hydrate/controller_test.go
```diff
@@ -35,6 +37,166 @@ const (
...
+				if err := os.RemoveAll(srcRootDir); err != nil {
+					t.Logf("failed to remove source root directory: %v", err)
+				}
+			}()
+			go func() {
```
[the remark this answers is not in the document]
nan-yu: done

pkg/hydrate/controller_test.go
```diff
@@ -35,6 +37,166 @@ const (
...
+						if err != nil {
+							return fmt.Errorf("failed to create error file %q: %v", errFilePath, err)
+						}
+						t.Logf("error file %q created at %v", errFilePath, time.Now())
+						if len(tc.errFileContent) != 0 {
+							if _, err = errFile.WriteString(tc.errFileContent); err != nil {
+								return fmt.Errorf("failed to write to error file %q: %v", errFilePath, err)
+							}
+						}
+						if err = errFile.Close(); err != nil {
```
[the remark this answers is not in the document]
nan-yu: Updated to use `os.WriteFile`, which handles closing the file handle.

pkg/hydrate/controller_test.go
```diff
@@ -35,6 +37,166 @@ const (
...
+						symDir := filepath.Join(srcRootDir, "rev")
+						if err := os.Symlink(commitDir, symDir); err != nil {
+							return fmt.Errorf("failed to create the symlink %q: %v", symDir, err)
+						}
+						symlinkCreated = true
+						t.Logf("symlink %q created and linked to %q at %v", symDir, commitDir, time.Now())
+					}
+
+					// Create the sync directory
+					if symlinkCreated && tc.syncDir == syncDir {
```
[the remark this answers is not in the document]
nan-yu: Good call, moved above to create syncDir before generating symlink.

pkg/parse/run_test.go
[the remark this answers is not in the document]
sdowell: The test should block for the channel to close before moving to the next test case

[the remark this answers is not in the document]
nan-yu: done

sdowell (approved):

PR #887, opened 2023-09-19
Expose transient errors in the R*Sync status field

nan-yu:
The intent is to indicate the processing status.


nan-yu:
This PR needs rebase after #885 and #886 are merged.

pkg/parse/run.go
```diff
@@ -296,14 +296,6 @@ func read(ctx context.Context, p Parser, trigger string, state *reconcilerState,
 				status.InternalHydrationError(err, "error setting %s annotation", metadata.RequiresRenderingAnnotationKey))
 		}
 	}
-	// Return the transient errors here to avoid surfacing them to the R*Sync status field.
-	// The transient errors might be auto-resolved in the next retry loop, so no need to expose to users.
```
sdowell, line 300: I'm curious about the intent here. Given these should auto-resolve, why do we now want to surface these to users via the RSync status?

nan-yu: It shows the syncing progress, similar as `rendering is still in progress`. Ideally, they should be warnings, but since we don't support surfacing warning at this moment, they're exposed as errors.

nan-yu: A side-effect is to make the message of the Syncing Condition more accurate.
See the change in run_test.go. Without the change, the message will be `Rendering is still in progress`, which could be confusing for WET configs.

sdowell (approved):

PR #890, opened 2023-09-21
ci: fix junit test output parsing

sdowell:
After upgrading to Go 1.20 there was a breaking change in the test output, in which resuming tests print NAME instead of CONT.

https://go-review.git.corp.google.com/c/go/+/443596

The go-junit-report utility does not properly parse NAME, so this hack replaces the output to look how it did before. The upstream go-junit-report should be fixed to properly parse the new gotest output format.

### DIFF
--- a/build/test-e2e-go/e2e.sh
+++ b/build/test-e2e-go/e2e.sh
@@ -34,6 +34,11 @@ echo "Tests took $(( end_time - start_time )) seconds"
 # enables running this script more flexibly, e.g. without docker in docker.
 if [[ -n "${ARTIFACTS}" && -d "${ARTIFACTS}" ]]; then
   echo "Creating junit xml report"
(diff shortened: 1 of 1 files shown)


karlkfi (approved):
/lgtm

nan-yu (commented):
Do we have an upstream issue open? If so, I would suggest adding a TODO to revert the hack when the upstream is fixed.

sdowell:
> Do we have an upstream issue open? If so, I would suggest adding a TODO to revert the hack when the upstream is fixed.

I haven't created the issue yet but was planning to as a followup. Good suggestion, I'll add a TODO so that we have context on when to revert the hack.

/hold

sdowell:
@nan-yu created https://github.com/jstemmer/go-junit-report/issues/169 and added a TODO

/unhold

nan-yu (approved):
/lgtm

PR #891, opened 2023-09-21
fix: add askpass to override container enum

sdowell:
The askpass sidecar is supported by the overrides in code, but is rejected by the RSync API because it is not enumerated in the CRD. This change enables overriding the askpass container via the RSync.


nan-yu (approved):
/lgtm

We may want to add `otel-agent` to the list as well, which can be a separate PR.

PR #892, opened 2023-09-22
fix: increase default cpuRequest for askpass

sdowell:
The askpass sidecar is failing regularly on autopilot due to timing out before the request can be served. This seems to be an issue of insufficient CPU to serve the request within the 1s timeout. The new value seems to pass consistently from experimentation. Intermediate values of 20m and 30m were tested but still flaky.

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_container_resources.go
+++ b/pkg/reconcilermanager/controllers/reconciler_container_resources.go
@@ -55,7 +55,7 @@ func ReconcilerContainerResourceDefaults() map[string]v1beta1.ContainerResources
 		},
 		reconcilermanager.GCENodeAskpassSidecar: {
 			ContainerName: reconcilermanager.GCENodeAskpassSidecar,
-			CPURequest:    resource.MustParse("10m"),
+			CPURequest:    resource.MustParse("50m"),
 			MemoryRequest: resource.MustParse("20Mi"),
 		},
 		metrics.OtelAgentName: {


pkg/reconcilermanager/controllers/reconciler_container_resources.go
```diff
@@ -55,7 +55,7 @@ func ReconcilerContainerResourceDefaults() map[string]v1beta1.ContainerResources
 		},
 		reconcilermanager.GCENodeAskpassSidecar: {
 			ContainerName: reconcilermanager.GCENodeAskpassSidecar,
-			CPURequest:    resource.MustParse("10m"),
+			CPURequest:    resource.MustParse("50m"),
```
karlkfi, line 58: The problem this fixes is autopilot specific. On standard, the askpass container can boost. So I don't know if we want to increase the requirements for standard to fix autopilot. This is why I wrote https://github.com/GoogleContainerTools/kpt-config-sync/pull/872 to split apart the standard and autopilot resources.

/cc @janetkuo

nan-yu: This will fix CI failures and unblock the upcoming 1.16.2 release next week.
I would suggest that we make the change and consolidate with #872 later.

nan-yu (approved):
/lgtm
/hold

It looks good to me, but I would like @janetkuo to make the final call.

PR #894, opened 2023-09-22
fix: add otel-agent to override api

sdowell:
The support for otel-agent overrides has already been implemented in code, but was not allowed on the RSync api since it was not enumerated on the CRD.


nan-yu (approved):
/lgtm

PR #895, opened 2023-09-22
fix: increase default memoryRequest for helm-sync

sdowell:
The helm-sync container is failing regularly on autopilot due to OOMKill. Autopilot sets the limits equal to the requests, and the current value is too low for all of the e2e tests to pass. Standard allows bursting while autopilot does not. We may be able to revert this increase after an upcoming change which will allow separate defaults for standard and autopilot.


nan-yu:
Any links to OOMKill? I saw a few `signal: killed` in the helm-sync container, which usually indicates low CPU.

sdowell:
> Any links to OOMKill? I saw a few `signal: killed` in the helm-sync container, which usually indicates low CPU.

https://oss.gprow.dev/view/gcs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-rapid/1705283066261409792

Was able to reproduce the OOMKill locally as well, and the test passed for me locally after bumping the default to this new value. We may need to bump cpuRequest as well, but this seems to be sufficient for now

nan-yu (approved):
/lgtm

PR #898, opened 2023-09-26
fix: persist deployment fields during allowListing

sdowell:
The deleteDeploymentFields function was mutating the passed in pointer when deleting the allowList fields. This caused the Deployment patch data to be missing any of the allowListed fields if set. The reconciler Deployment does not contain any of the allowListed fields by default, but these fields can be patched in the reconciler-manager-cm ConfigMap by users.

The allow-listed fields are intended to be ignored when comparing the declared to the actual object to allow drift, but we should still apply the Deployment as it is declared.


nan-yu (approved):
/lgtm

janetkuo:
@sdowell let's cherry pick this to 1.16

sdowell:
> @sdowell let's cherry pick this to 1.16

Ack, will do

PR #900, opened 2023-09-26
Migrate to git-sync v4

nan-yu:
`git-sync` v4 has a bunch of improvement, such as efficiency, and closing the race condition where a symbolic name can change after `git ls-remote` but before `git fetch`. It also addresses the missing error file when `git clone` is slow.

Most of the env variables are backward compatible, but this commit updates the env variable names to avoid jamming the git-sync logs.

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -101,8 +101,8 @@ data:
                - NET_RAW
            imagePullPolicy: IfNotPresent
          - name: git-sync
-           image: gcr.io/config-management-release/git-sync:v3.6.9-gke.2__linux_amd64
-           args: ["--root=/repo/source", "--dest=rev", "--max-sync-failures=30", "--error-file=error.json"]
+           image: gcr.io/config-management-release/git-sync:v4.0.0-gke.1__linux_amd64
+           args: ["--root=/repo/source", "--link=rev", "--max-failures=30", "--error-file=error.json"]
            volumeMounts:
            - name: repo
              mountPath: /repo
--- a/pkg/api/configsync/v1alpha1/getters.go
+++ b/pkg/api/configsync/v1alpha1/getters.go
@@ -14,19 +14,6 @@
 
 package v1alpha1
 
-import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"kpt.dev/configsync/pkg/api/configsync"
-)
-
-// GetPeriodSecs returns the sync period defaulting to 15 if empty.
-func GetPeriodSecs(g *Git) float64 {
-	if g.Period.Duration == 0 {
-		return configsync.DefaultReconcilerPollingPeriodSeconds
-	}
-	return g.Period.Duration.Seconds()
-}
-
 // SafeOverride creates an override or returns an existing one
 // use it if you need to ensure that you are assigning
 // to an object, but not to test for nil (current existence)
@@ -45,12 +32,4 @@ func (rs *RootSyncSpec) SafeOverride() *RootSyncOverrideSpec {
 		rs.Override = &RootSyncOverrideSpec{}
 	}
 	return rs.Override
-}
-
-// GetReconcileTimeout returns reconcile timeout in string, defaulting to 5m if empty
-func GetReconcileTimeout(d *metav1.Duration) string {
-	if d == nil || d.Duration == 0 {
-		return configsync.DefaultReconcileTimeout.String()
-	}
-	return d.Duration.String()
 }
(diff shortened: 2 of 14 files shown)


pkg/reconcilermanager/controllers/gitsync_env.go
```diff
@@ -170,48 +194,49 @@ func gitSyncEnvs(_ context.Context, opts options) []corev1.EnvVar {
...
+	// When branch and ref not set in RootSync/RepoSync then dont set GITSYNC_REF,
+	// git-sync will use the default values for them.
 	if opts.branch != "" {
 		result = append(result, corev1.EnvVar{
-			Name:  "GIT_SYNC_BRANCH",
+			Name:  gitSyncRef,
 			Value: opts.branch,
 		})
 	}
+	// opts.ref takes precedence over opts.branch, so overwrite the EnvVar
```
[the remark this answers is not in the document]
nan-yu: done

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -101,8 +101,8 @@ data:
                - NET_RAW
            imagePullPolicy: IfNotPresent
          - name: git-sync
-           image: gcr.io/config-management-release/git-sync:v3.6.9-gke.2__linux_amd64
-           args: ["--root=/repo/source", "--dest=rev", "--max-sync-failures=30", "--error-file=error.json"]
+           image: gcr.io/config-management-release/git-sync:v4.0.0-gke.1__linux_amd64
```
sdowell, line 104: nit: I think it may help to break this up into two changes:
1. refactor the env variables to the constants without changing functionality
2. upgrade git-sync and make associated changes to env vars

This would make it easier to visualize what changes come with the git-sync upgrade, and make for a smaller change in case we need to revert

nan-yu: That's a good point. Extract the non-functional change into https://github.com/GoogleContainerTools/kpt-config-sync/pull/904.

pkg/api/configsync/v1beta1/gitconfig.go
```diff
@@ -25,11 +25,13 @@ type Git struct {
...
 
-	// branch is the git branch to checkout. Default: "master".
+	// branch is the git branch to checkout.
+	// The field is deprecated. Use `revision` instead.
+	// If both `branch` and `revision` are defined, `revision` takes precedence over `branch`.
 	// +optional
 	Branch string `json:"branch,omitempty"`
 
-	// revision is the git revision (tag, ref or commit) to fetch. Default: "HEAD".
+	// revision is the git revision (branch, tag, ref or commit) to fetch. Default: "HEAD".
```
sdowell, line 34: similar update should be made to the field on `v1alpha1`

pkg/reconcilermanager/controllers/gitsync_env.go
```diff
@@ -212,21 +210,21 @@ func gitSyncEnvs(_ context.Context, opts options) []corev1.EnvVar {
 	}
 	result = append(result, corev1.EnvVar{
 		Name:  gitSyncPeriod,
-		Value: fmt.Sprintf("%f", opts.period),
+		Value: fmt.Sprintf("%fs", opts.period),
```
sdowell, line 213: It looks like git-sync expects a `time.Duration` format string here. In `gitSyncEnvs` we convert the `metav1.Duration` -> `float64` (seconds) and then here we are formatting that float into string seconds.

I suggest refactoring this so that the `options` struct accepts a `time.Duration` type for `period` and then we can just use the builtin String conversion

sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/rootsync_controller.go#L686

pkg/api/configsync/v1beta1/gitconfig.go
nan-yu: Good catch! done

pkg/reconcilermanager/controllers/gitsync_env.go
nan-yu: done

pkg/api/configsync/v1alpha1/getters.go
```diff
@@ -22,7 +22,7 @@ import (
 // GetPeriodSecs returns the sync period defaulting to 15 if empty.
 func GetPeriodSecs(g *Git) float64 {
```
sdowell, line 23: Looks like this is unused - do you think it can be removed?

nan-yu: Sure. I was thinking removing unused function seems to belong to a separate CL, but it's okay since it is a small change.
There is another unused function, which is also removed in this PR.

sdowell: I'm okay with removing it as a separate change, just wanted to mention it since I noticed it while reviewing this

sdowell (approved):

PR #903, opened 2023-09-27
[CHERRY-PICK] fix: persist deployment fields during allowListing (#898)

sdowell:
The deleteDeploymentFields function was mutating the passed in pointer when deleting the allowList fields. This caused the Deployment patch data to be missing any of the allowListed fields if set. The reconciler Deployment does not contain any of the allowListed fields by default, but these fields can be patched in the reconciler-manager-cm ConfigMap by users.

The allow-listed fields are intended to be ignored when comparing the declared to the actual object to allow drift, but we should still apply the Deployment as it is declared.


sdowell:
/assign @victorpras 
/cc @janetkuo 

Original PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/898

PR #907, opened 2023-09-29
Tune retryCap used in unit test to fix flakiness

nan-yu:
Some unit tests failed intermittently because the test directory wasn't created in the retryCap. This commit increases the cap. It also makes the backoff step consistent.


sdowell (approved):

sdowell:
looks like it is still flaky

nan-yu:
> looks like it is still flaky

It is flaky because the `updateSyncStatusPeriodically` goroutine sets the syncing condition back to true after the sync is done.
I changed to use the default statusUpdatePeriod (5s) so that `updateSyncStatusPeriodically` won't be triggered.

PR #908, opened 2023-10-02
Remove PSP from installation guide

janetkuo:
PSP has been removed since K8s 1.25


janetkuo:
@sdowell we can remove PSP manifest from installation manifests as well

sdowell:
> @sdowell we can remove PSP manifest from installation manifests as well

SG, I can follow up on that cleanup

sdowell (approved):

PR #910, opened 2023-10-02
fix: TestAutopilotReconcilerAdjustment math

karlkfi:
Extract the math to a simulation method.


sdowell (approved):
thanks for pulling this change out

PR #911, opened 2023-10-02
Revert "fix: increase default cpuRequest for askpass"

karlkfi:
Reverts GoogleContainerTools/kpt-config-sync#892

These changes have been made for autopilot-only in https://github.com/GoogleContainerTools/kpt-config-sync/pull/872

Reverting the change for non-autopilot clusters which support boosting.


karlkfi:
dupe of https://github.com/GoogleContainerTools/kpt-config-sync/pull/912

PR #912, opened 2023-10-02
revert resource increases

sdowell:
Reason for revert: The problems solved by these commits have been fixed
by using autopilot-specific defaults

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_container_resources.go
+++ b/pkg/reconcilermanager/controllers/reconciler_container_resources.go
@@ -46,7 +46,7 @@ func ReconcilerContainerResourceDefaults() map[string]v1beta1.ContainerResources
 		reconcilermanager.HelmSync: {
 			ContainerName: reconcilermanager.HelmSync,
 			CPURequest:    resource.MustParse("50m"),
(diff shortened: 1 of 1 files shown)


karlkfi (approved):
/lgtm

Approved by @janetkuo in chat

pkg/reconcilermanager/controllers/reconciler_container_resources.go
```diff
@@ -46,7 +46,7 @@ func ReconcilerContainerResourceDefaults() map[string]v1beta1.ContainerResources
 		reconcilermanager.HelmSync: {
 			ContainerName: reconcilermanager.HelmSync,
 			CPURequest:    resource.MustParse("50m"),
-			MemoryRequest: resource.MustParse("256Mi"),
```
janetkuo, line 49: @nan-yu I remember we bumped helm memory several times for Autopilot. Should we lower the number here?

nan-yu: We bumped helm cpu request for Autopilot, not the memory: https://github.com/GoogleContainerTools/kpt-config-sync/pull/378.
We could probably lower it from `50m` to `10m`.

PR #919, opened 2023-10-04
Update git-sync version for CVE fixes

nan-yu:
This is not a cherrypick commit from the main branch because the git-sync version diverges from the main branch.


nan-yu:
FYI, this is not a cherrypick because the git-sync version diverges from the main branch.

karlkfi (approved):
/lgtm

PR #925, opened 2023-10-06
fix: remove policy-controller-psp from operator

sdowell:
The policy-controller-psp was removed by a separate commit, but the inclusion in this kustomization file was overlooked.

The manifest was removed in https://github.com/GoogleContainerTools/kpt-config-sync/pull/909


nan-yu:
Any plan to remove other references used in the test?

PR #928, opened 2023-10-09
test: Prevent autopilot scale-to-zero

karlkfi:
- Add a persistent Deployment on GKE Autopilot clusters to prevent scale-to-zero. This prevents metric-server from becoming unhealthy and breaking API discovery. b/209800496

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


e2e/nomostest/new.go
```diff
@@ -269,6 +274,14 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 	// init the Client & WatchClient
 	nt.RenewClient()
 
+	// Create workload to disable scale-to-zero.
+	// Scale-to-zero breaks API discovery on GKE Autopilot: b/209800496S
```
sdowell, line 278: Could you add a bit more context on the commit message for why we need this? e.g. what problems does this cause for the e2e tests

karlkfi: done

sdowell (approved):

PR #930, opened 2023-10-10
ci: remove upx from CLI build

sdowell:
The upx package was being used to compress the nomos binary as part of the build process. upx poses a few problems:
- Compressed binaries cannot be scanned for packages (SBOM)
- The upx package is not available on bookworm package manager

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ TEST_INFRA_REGISTRY ?= $(LOCATION)-docker.pkg.dev/$(TEST_INFRA_PROJECT)/test-inf
 
 # Docker image used for build and test. This image does not support CGO.
 # When upgrading this tag, publish the image after the change is submitted.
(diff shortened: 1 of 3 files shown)


sdowell:
> Did you check the docker image size and the binary size? I'm wondering how much it may increase without the compression.

The docker image size is unchanged. Before and after this change, the docker images use the uncompressed binary. This change impacts the nomos binary that is built standalone.

Here are some comparisons for linux_amd64:
nomos: 70236971
nomos + tar: 30170436 (42.96%)
nomos + upx: 31378588 (44.68%)
nomos + upx + tar: 30437264 (43.34%)

So the binary itself is 44.68% size with upx, however the tarball is actually smaller (negligibly) with the uncompressed binary. Given that the binary is distributed via the tarball, it seems like the benefits of upx compression are negligible.

At any rate, the `upx` compression is preventing the SBOM scanner from discovering go modules from the binary file.

PR #931, opened 2023-10-10
fix: use k8s yaml parser to unmarshal

sdowell:
The keepalive deployment was not being parsed since it was using the wrong yaml parser. This changes the import to use the k8s parser.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -22,7 +22,6 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sdowell:
I made an incorrect assumption on the last fix, looks like the actual problem is that we were using the wrong yaml parser. I verified locally that this fixes the issue.

/assign @nan-yu

karlkfi (approved):
/lgtm

PR #932, opened 2023-10-11
fix: specify registry for auth-docker

sdowell:
This was missed during a rebase, and is required to set up docker auth for the correct registry host (e.g. GAR).

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -22,7 +22,7 @@ build-buildenv: build/buildenv/Dockerfile
 		$(DOCKER_BUILD_ARGS)
 
 push-buildenv: build-buildenv
-	@gcloud $(GCLOUD_QUIET) auth configure-docker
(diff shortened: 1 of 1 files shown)


Makefile.build
```diff
@@ -82,7 +82,7 @@ build-images-multirepo: build-images
 auth-docker:
 	@echo "+++ Using account:"
 	gcloud config get-value account
-	@gcloud $(GCLOUD_QUIET) auth configure-docker
+	@gcloud $(GCLOUD_QUIET) auth configure-docker $(firstword $(subst /, ,$(REGISTRY)))
```
nan-yu, line 85: Line 25 also doesn't have registry specified.

sdowell: fixed

nan-yu (approved):
/lgtm

PR #938, opened 2023-10-13
Custom ClusterRole for RootSync reconcilers

### LINKED ISSUE #935: Make it possible to reconcile RootSyncs without cluster-admin
We've recently started working on slimming down the permissions we grant to `RootSync`s that target repositories we are not in control of, but realized whatever permissions we grant are going to be effectively ignored, as `reconciler-manager` [is hard-coded to grant `cluster-admin` to root reconcilers](https://github.com/GoogleContainerTools/kpt-config-sync/blob/199db6dbfa0b9eb1824925c8c687574de095a294/pkg/reconcilermanager/controllers/rootsync_controller.go#L789).

I could think of a few options for how to configure that would work for us, so would like to hear your thoughts on what you'd prefer before I submit a design proposal and/or PR:

* A boolean property, defaulting to `false`, which just turns off creating the `ClusterRoleBinding` entirely. If a user turns this on, they are responsible for creating a `ClusterRole` and binding it to the KSA.  
  Example: `.spec.overrides.manuallyManagePermissions: true` (name very much pending)

* A string property denoting which role to grant to the KSA. If provided `reconciler-manager` then creates a `ClusterRoleBinding` to that role instead of `cluster-admin`.  
  Example: `.spec.overrides.reconcilerClusterRole: my-config-sync-role`

* A string property denoting a different KSA to use rather than the one currently created by the `reconciler-manager`. If provided, neither the KSA nor the `ClusterRoleBinding` would be created (i.e. [these lines would be skipped](https://github.com/GoogleContainerTools/kpt-config-sync/blob/199db6dbfa0b9eb1824925c8c687574de095a294/pkg/reconcilermanager/controllers/rootsync_controller.go#L205-L212)); the onus to do that would be on the user instead. (In addition, [these lines](https://github.com/GoogleContainerTools/kpt-config-sync/blob/199db6dbfa0b9eb1824925c8c687574de095a294/pkg/reconcilermanager/controllers/rootsync_controller.go#L914-L918) would have to be made aware of the override).
  
  Example: `.spec.overrides.serviceAccountName: my-own-ksa`

Would you accept a feature contribution that makes this configurable? If so, is there one of the approaches above (or something else entirely) you prefer?

tomasaschan:
Closes #935.

This introduces a new override for `RootSync`s, which allows customization of which `ClusterRole` to bind the reconciler's service account to. The default is `cluster-admin`, which is the same as what was done before this change. For more details, see the included design document.

### DIFF
--- a/docs/design-docs/02-custom-root-reconciler-clusterrole.md
+++ b/docs/design-docs/02-custom-root-reconciler-clusterrole.md
@@ -41,18 +41,25 @@ on `spec.overrides` for a `RootSync`:
 ```yaml
 spec:
   overrides:
-    clusterRole: my-custom-role
+    clusterRole:
+      name: my-custom-role
 ```
 
 This role is defaulted to `cluster-admin` in order to stay backwards compatible.
 
 For the case where a single `ClusterRole` is not expressive enough to configure the
-permissions a user want, we also allow the special sentinel value `%none%`, which
-disables creating the `ClusterRoleBinding` entirely. The value is chosen so that it can
-never coincide with the name of an existing `ClusterRole`, since they [do not allow `%`
-signs in their names].
+permissions a user want, you can instead set `disabled: true` on the override object:
 
-[do not allow `%` signs in their names]: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#path-segment-names
+```yaml
+spec:
+  overrides:
+    clusterRole:
+      disabled: true
+```
+This disables creating the `ClusterRoleBinding` entirely.
+
+If both `disabled: true` and a custom name is specified, `disabled` "wins" and no binding
+is created.
 
 ## User Guide
 
@@ -68,7 +75,8 @@ metadata:
   namespace: config-management-system
 spec:
   overrides:
-    clusterRole: my-cluster-role
+    clusterRole:
+      name: my-cluster-role
   # ...
 ---
 apiVersion: rbac.authorization.k8s.io/v1
@@ -79,7 +87,21 @@ rules:
 # ...
 ```
 
-You can also use the special value `%none%` and disable automatic binding to any role.
+You can also disable automatic binding to any role.
+
+```yaml
+apiVersion: configsync.gke.io/v1beta1
+kind: RootSync
+metadata:
+  name: root-sync
+  namespace: config-management-system
+spec:
+  overrides:
+    clusterRole:
+      name: my-cluster-role
+  # ...
+```
+
 The reconciler will then run with effectively no permissions, until you manually create
 some role or cluster role bindings for it.
 
--- a/docs/design-docs/02-custom-root-reconciler-clusterrole.md
+++ b/docs/design-docs/02-custom-root-reconciler-clusterrole.md
@@ -98,7 +98,7 @@ metadata:
 spec:
   overrides:
     clusterRole:
-      name: my-cluster-role
+      disabled: true
   # ...
 ```
 
--- a/docs/design-docs/02-custom-root-reconciler-clusterrole.md
+++ b/docs/design-docs/02-custom-root-reconciler-clusterrole.md
@@ -1,0 +1,111 @@
+# Custom `ClusterRole` assignments for `RootSync`s
+
+* Author(s): @tomasaschan
+* Approver: @sdowell, @janetkuo
+* Status: approved
+
+## Summary
+
+When creating a reconciler deployment for a `RootSync`, Config Sync currently [creates a
+`ClusterRoleBinding`] for the generated `ServiceAccount` granting it `cluster-admin`.
+
+This proposal introduces a mechanism to customize this behavior.
+
+[creates a `ClusterRoleBinding`]: https://github.com/GoogleContainerTools/kpt-config-sync/blob/199db6dbfa0b9eb1824925c8c687574de095a294/pkg/reconcilermanager/controllers/rootsync_controller.go#L789
+
+## Motivation
+
+If one wants to allow deploying resources from a source where one doesn't have full
+control of the contents of the upstream config source, and want to limit what another
+team or third party can accomplish in the cluster, one can imagine limiting what
+permissions are granted to the `ServiceAccount` running the reconciler.
+
+However, as Config Sync is currently granting `cluster-admin`, any custom role bindings
+are effectively ignored; the reconciler will, due to the binding added by Config Sync,
+have access to do _anything_ regardless.
+
+To follow the [principle of least privilege], one should ensure the reconciler only has
+access to deploy the expected resources.
+
+See also [#935].
+
+[principle of least privilege]: https://en.wikipedia.org/wiki/Principle_of_least_privilege
+[#935]: https://github.com/GoogleContainerTools/kpt-config-sync/issues/935
+
+## Design Overview
+
+By providing the name of a user-defined `ClusterRole` for a `RootSync`, a user can
+override which role Config Sync binds to. This configuration is exposed as a new field
+on `spec.overrides` for a `RootSync`:
+
+```yaml
+spec:
+  overrides:
+    clusterRole: my-custom-role
+```
+
+This role is defaulted to `cluster-admin` in order to stay backwards compatible.
+
+For the case where a single `ClusterRole` is not expressive enough to configure the
+permissions a user want, we also allow the special sentinel value `%none%`, which
+disables creating the `ClusterRoleBinding` entirely. The value is chosen so that it can
+never coincide with the name of an existing `ClusterRole`, since they [do not allow `%`
+signs in their names].
+
+[do not allow `%` signs in their names]: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#path-segment-names
+
+## User Guide
+
+To take advantage of this new feature, set `spec.overrides.clusterRole` to the name of a
+`ClusterRole` you wish this `RootSync` to be reconciled under; you must create this role
+yourself, but Config Sync will create the `ClusterRoleBinding` for you.
+
+```yaml
+apiVersion: configsync.gke.io/v1beta1
+kind: RootSync
+metadata:
+  name: root-sync
+  namespace: config-management-system
+spec:
+  overrides:
+    clusterRole: my-cluster-role
+  # ...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: my-cluster-role
+rules:
+# ...
+```
+
+You can also use the special value `%none%` and disable automatic binding to any role.
+The reconciler will then run with effectively no permissions, until you manually create
+some role or cluster role bindings for it.
+
+**Note** that while the name of the relevant service account will _often_ be predictable
+as `root-reconciler-<root-reconciler-name>`, this is not guaranteed by Config Sync. You
+can always find the service account by its labels; in particular, the labels
+`configsync.gke.io/sync-kind` and `configsync.gke.io/sync-name` should be useful.
+
+## Risks and Mitigations
+
+This feature is entirely backwards-compatible, by means of being opt-in.
+
+## Test Plan
+
+Implementation of this feature can includes automated tests that verify the behavior.
+
+## Open Issues/Questions
+
+N/A
+
+## Alternatives Considered
+
+One could imagine exposing settings with slightly different semantics - e.g. a simple
+boolean for turning the `ClusterRoleBinding` off, or a setting to change what service
+account the reconciler is running with. However, these both put a bigger burden on the
+user in order to utilize them even for the simple use cases, which is why changing which
+role to bind to is probably the most user-friendly knob to expose. With the inclusion of
+a sentinel value to create no binding, the first of these alternatives is effectively
+supported, if through a slightly worse API. 
--- a/docs/design-docs/02-custom-root-reconciler-clusterrole.md
+++ b/docs/design-docs/02-custom-root-reconciler-clusterrole.md
@@ -129,5 +129,5 @@ boolean for turning the `ClusterRoleBinding` off, or a setting to change what se
 account the reconciler is running with. However, these both put a bigger burden on the
 user in order to utilize them even for the simple use cases, which is why changing which
 role to bind to is probably the most user-friendly knob to expose. With the inclusion of
-a sentinel value to create no binding, the first of these alternatives is effectively
-supported, if through a slightly worse API. 
+a switch to turn off creating the binding entirely, the first of these alternatives is effectively
+supported as well, and the API supports extending for other use cases in the future. 
--- a/pkg/api/configsync/v1beta1/resource_override.go
+++ b/pkg/api/configsync/v1beta1/resource_override.go
@@ -92,6 +92,18 @@ type RootSyncOverrideSpec struct {
 	// +kubebuilder:validation:Enum=implicit;explicit
 	// +optional
 	NamespaceStrategy configsync.NamespaceStrategy `json:"namespaceStrategy,omitempty"`
+
+	// ClusterRole controls which role to bind the service account for this RootSync's
+	// reconciler to. Defaults to 'cluster-admin'.
+	//
+	// The specified ClusterRole needs to be created manually.
+	//
+	// You can use the special sentinel value '%none%' to disable binding to a ClusterRole
+	// entirely; it is then your responsibility to bind any roles to the relevant service
+	// account.
+	//
+	// +optional
+	ClusterRole string `json:"clusterRole,omitempty"`
 }
 
 // RepoSyncOverrideSpec allows to override the settings for a RepoSync reconciler pod
--- a/pkg/api/configsync/v1beta1/resource_override.go
+++ b/pkg/api/configsync/v1beta1/resource_override.go
@@ -93,17 +93,29 @@ type RootSyncOverrideSpec struct {
 	// +optional
 	NamespaceStrategy configsync.NamespaceStrategy `json:"namespaceStrategy,omitempty"`
 
-	// ClusterRole controls which role to bind the service account for this RootSync's
-	// reconciler to. Defaults to 'cluster-admin'.
+	// clusterRole controls which role to bind the service account for this RootSync's
+	// reconciler to.
 	//
-	// The specified ClusterRole needs to be created manually.
+	// +optional
+	ClusterRole ClusterRoleOverrideSpec `json:"clusterRole,omitempty"`
+}
+
+// ClusterRoleOverrideSpec allows to override settings for the root reconciler permissions
+type ClusterRoleOverrideSpec struct {
+	// name is the name of the ClusterRole to bind to; defaults to "cluster-admin".
 	//
-	// You can use the special sentinel value '%none%' to disable binding to a ClusterRole
-	// entirely; it is then your responsibility to bind any roles to the relevant service
+	// If you specify a different name, the ClusterRole needs to be created manually.
+	//
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// disabled turns off binding to a ClusterRole entirely.
+	//
+	// It is then your responsibility to bind any roles to the relevant service
 	// account.
 	//
 	// +optional
-	ClusterRole string `json:"clusterRole,omitempty"`
+	Disabled bool `json:"disabled,omitempty"`
 }
 
 // RepoSyncOverrideSpec allows to override the settings for a RepoSync reconciler pod
--- a/pkg/core/names.go
+++ b/pkg/core/names.go
@@ -27,11 +27,8 @@ const (
 	NsReconcilerPrefix = "ns-reconciler"
 	// RootReconcilerPrefix is the prefix usef for all Root reconcilers.
 	RootReconcilerPrefix = "root-reconciler"
-)
-
-var (
 	// RootSyncPermissionsPrefix is the prefix used for all ClusterRoleBindings granting access to Root Reconcilers
-	RootSyncPermissionsPrefix = fmt.Sprintf("%s:%s", configsync.RootSyncKind, RootReconcilerPrefix)
+	RootSyncPermissionsPrefix = configsync.RootSyncKind + ":" + RootReconcilerPrefix
 )
 
 // RootReconcilerName returns the root reconciler's name in the format root-reconciler-<name>.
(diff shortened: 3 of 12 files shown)


tomasaschan:
Started working on an implementation, but it turns out it was slightly less straightforward than I'd hoped 😄 The main source of complexity is that Config Sync currently creates a single CRB and adds all root reconciler KSAs as subjects.  We must split this into a CRB per RootSync, because

* they might not target the same role, so we cannot have just one CRB for all of them
* when deleting a RootSync we must be able to clean them up, and at that point [we only know the name of the `RootSync`, not the name of the role it used](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/garbage_collector.go#L235), so we cannot have one CRB per referenced role.

In practice, we can probably change [`RootSyncPermissionsName`](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/build_names.go#L31) to take the name of the `RootSync` as a parameter, fix all compiler errors, and have a working implementation (with different behavior than before wrt the number of bindings created). We'll then also have to change a few tests that assume that permissions will be managed by adding and removing subjects on a single CRB.

tomasaschan:
`gpg` timed out when it tried to sign those last two commits; not sure why. I hope GPG-signed commits isn't a requirement for contributing...

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -544,8 +556,8 @@ func (r *RootSyncReconciler) mapObjectToRootSync(obj client.Object) []reconcile.
 
 	// Ignore changes from resources without the root-reconciler prefix or configsync.gke.io:root-reconciler
 	// because all the generated resources have the prefix.
-	roleBindingName := RootSyncPermissionsName()
-	if !strings.HasPrefix(objRef.Name, core.RootReconcilerPrefix) && objRef.Name != roleBindingName {
+	if !strings.HasPrefix(objRef.Name, core.RootReconcilerPrefix) &&
+		!strings.HasPrefix(objRef.Name, fmt.Sprintf("%s:%s", configsync.GroupName, core.RootReconcilerPrefix)) {
```
tomasaschan, line 560: Maybe we shouldn't `fmt.Sprintf` here, but rather try to depend on something like `RootSyncPermissionsName(core.RootReconcilerPrefix)` to ensure that it stays consistent if we change the naming scheme?
```suggestion
	if !strings.HasPrefix(objRef.Name, core.RootReconcilerPrefix) &&
		!strings.HasPrefix(objRef.Name, RootSyncPermissionsName(core.RootReconcilerPrefix)) {
```

pkg/reconcilermanager/controllers/build_names.go
```diff
@@ -32,8 +32,11 @@ func RepoSyncPermissionsName() string {
 	return fmt.Sprintf("%s:%s", configsync.GroupName, core.NsReconcilerPrefix)
 }
 
-// RootSyncPermissionsName returns root reconciler permissions name.
-// e.g. configsync.gke.io:root-reconciler
-func RootSyncPermissionsName() string {
-	return fmt.Sprintf("%s:%s", configsync.GroupName, core.RootReconcilerPrefix)
+// RootSyncPermissionsName returns root reconciler ClusterRoleBinding name.
+// e.g. configsync.gke.io:root-reconciler:my-root-sync
+func RootSyncPermissionsName(reconcilerName string) string {
```
sdowell, line 37: I expect there may be a gap here during an upgrade. If the user upgrades from an older version, I expect that the old ClusterRoleBinding (the singleton `configsync.gke.io:root-reconciler`) will still include the Subjects for every `RootSync`. In this case, every pre-existing `RootSync` would still have a `ClusterRoleBinding` which grants `cluster-admin` in addition to the new `ClusterRoleBinding`  (`configsync.gke.io:root-reconciler:my-root-sync`).

pkg/reconcilermanager/controllers/rootsync_controller.go
tomasaschan: I think I'm gonna go with this, since I prefer it myself.

sdowell: Maybe we could add another const which stores this prefix value? e.g. `RootSyncClusterRoleBindingPrefix`

That way we can reuse the string and ensure consistency

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -0,0 +1,111 @@
...
+```yaml
+spec:
+  overrides:
+    clusterRole: my-custom-role
+```
+
+This role is defaulted to `cluster-admin` in order to stay backwards compatible.
+
+For the case where a single `ClusterRole` is not expressive enough to configure the
+permissions a user want, we also allow the special sentinel value `%none%`, which
```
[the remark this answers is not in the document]
tomasaschan: I don't know; I don't think so. I picked it for two reasons:
1. It is a value that can never be a valid name of a `ClusterRole` (because `%` is a disallowed character), so setting it to this can never end up binding to an unintended role created later.
2. I found it more descriptive than e.g. using a single invalid character or something else along those lines.
If you don't like it we can remove that part of this PR; although I do find it appealing to have _some_ way to completely turn off the binding...

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -0,0 +1,111 @@
...
+can always find the service account by its labels; in particular, the labels
+`configsync.gke.io/sync-kind` and `configsync.gke.io/sync-name` should be useful.
+
+## Risks and Mitigations
+
+This feature is entirely backwards-compatible, by means of being opt-in.
+
+## Test Plan
+
+Implementation of this feature can includes automated tests that verify the behavior.
```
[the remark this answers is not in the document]
tomasaschan: There is pretty good coverage of the new and changed features here, although not among the e2e tests. In particular, see the changes to [rootsync_controller_test.go](https://github.com/GoogleContainerTools/kpt-config-sync/pull/938/files#diff-96d9758339603eab1bf5deb9d64abf4acb751b8474cd85e2cd4d790a76514291).

pkg/reconcilermanager/controllers/build_names.go
tomasaschan: Yes, I'm aware; see the "Open Question" in the OP. In practice, this will be quite quickly remediated _if_ there is a `RootSync` called `root-sync`, since that one will then "take over" the previously-singleton role binding and remove all the other subjects from it. The hard part to solve is the case when no such `RootSync` exists. Not sure how you prefer to handle that though.

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -92,6 +92,18 @@ type RootSyncOverrideSpec struct {
 	// +kubebuilder:validation:Enum=implicit;explicit
 	// +optional
 	NamespaceStrategy configsync.NamespaceStrategy `json:"namespaceStrategy,omitempty"`
+
+	// ClusterRole controls which role to bind the service account for this RootSync's
```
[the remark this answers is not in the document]
tomasaschan: I followed the convention of the other doc comments, which are also `PascalCased` despite their JSON property names being `camelCased`.

Pushed a commit with the output of `make configsync-crds` too :)

pkg/reconcilermanager/controllers/rootsync_controller.go
tomasaschan: Yeah, that's not a bad idea.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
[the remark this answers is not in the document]
tomasaschan: Another option would be to make the property pointer-valued, and differentiate between "not set" and "set to an empty string", but that seems bad for other reasons...

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,14 +791,14 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
...
+func (r *RootSyncReconciler) upsertClusterRoleBinding(ctx context.Context, reconcilerRef types.NamespacedName, clusterRole string) (client.ObjectKey, error) {
 	crbRef := client.ObjectKey{Name: RootSyncPermissionsName(reconcilerRef.Name)}
 	childCRB := &rbacv1.ClusterRoleBinding{}
 	childCRB.Name = crbRef.Name
 
 	op, err := CreateOrUpdate(ctx, r.client, childCRB, func() error {
 		childCRB.OwnerReferences = nil
-		childCRB.RoleRef = rolereference("cluster-admin", "ClusterRole")
+		childCRB.RoleRef = rolereference(clusterRole, "ClusterRole")
 		childCRB.Subjects = addSubject(childCRB.Subjects, r.serviceAccountSubject(reconcilerRef))
```
tomasaschan, line 802: Or wait, are you saying we need to ignore existing bindings here, and set it to `addSubject(nil, ...)`? That's probably correct. That case should also have a test...

pkg/reconcilermanager/controllers/build_names.go
sdowell: Ah I missed that in the PR description and went straight to the design doc, sorry about that.

I'm thinking a potential approach could be:
- remove the special case for `root-reconciler` below, such that its `ClusterRoleBinding` will be named `configsync.gke.io:root-reconciler:root-sync`
- add a snippet to the Reconcile loop which deletes the legacy `ClusterRoleBinding` named `configsync.gke.io:root-reconciler`

This solution should handle the migration in all cases without user interference, even in the case that a `RootSync` named `root-sync` doesn't exist. WDYT?

pkg/reconcilermanager/controllers/rootsync_controller.go
sdowell: Yeah since there is a 1:1 relationship between `RootSync` and `ClusterRoleBinding`, I agree that it seems correct to set this to a single subject.

This would handle the upgrade path in the case that a `RootSync` named `root-sync` exists, but requires user intervention otherwise. See my other comment for a proposal on how to handle the upgrade in the general case.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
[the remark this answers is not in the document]
sdowell: Another option could be to make this a struct with a nested `name` and `enabled` field. This could create a clearer API for how to disable ClusterRoleBinding, although adds more API permutations (e.g. `name=my-cluster-role, enabled=false` probably doesn't make sense).

I don't feel too strongly about it and am okay with `%none%`, my main concern is whether this is unconventional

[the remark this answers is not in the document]
tomasaschan: I had the struct idea independently (sleeping on ideas is often effective! 😄), but I imagined it as
```
type ClusterRoleBindingOverrides struct {
  RoleName string
  Disabled   bool
}
```
This way the zero value of the struct is the current behavior (if an empty `RoleName` defaults it to `"cluster-admin"`); setting the name to something else does what this PR currently does for any value other than `%none%`, and setting `Disabled: true` disables the binding (and ignores `RoleName`, i.e. `Disabled` takes precedence).

I'll work on an update here address this (and other comments).

pkg/reconcilermanager/controllers/build_names.go
tomasaschan: That is cleaner, and makes the name `root-sync` less special. I like it!

tomasaschan:
@sdowell I think I've addressed all your feedback now. I updated the design doc to reflect the new schema (with separate properties for the name vs enabled/disabled status), and added a test that verifies that upgrading to a version with this change will cleanly remove the old binding and add the expected new ones.

I haven't added any e2e tests, but I hope the coverage in [`rootsync_controller_test.go`](https://github.com/GoogleContainerTools/kpt-config-sync/pull/938/files#diff-96d9758339603eab1bf5deb9d64abf4acb751b8474cd85e2cd4d790a76514291) suffices.

pkg/core/names.go
```diff
@@ -28,6 +29,11 @@ const (
 	RootReconcilerPrefix = "root-reconciler"
 )
 
+var (
+	// RootSyncPermissionsPrefix is the prefix used for all ClusterRoleBindings granting access to Root Reconcilers
+	RootSyncPermissionsPrefix = fmt.Sprintf("%s:%s", configsync.RootSyncKind, RootReconcilerPrefix)
```
sdowell, line 34: nit: this can be a `const` since it is comprised of two other consts

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,14 +779,37 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
...
+func (r *RootSyncReconciler) upsertClusterRoleBinding(ctx context.Context, reconcilerRef types.NamespacedName, clusterRole string) (client.ObjectKey, error) {
+	crbRef := client.ObjectKey{Name: RootSyncPermissionsName(reconcilerRef.Name)}
 	childCRB := &rbacv1.ClusterRoleBinding{}
 	childCRB.Name = crbRef.Name
 
 	op, err := CreateOrUpdate(ctx, r.client, childCRB, func() error {
 		childCRB.OwnerReferences = nil
-		childCRB.RoleRef = rolereference("cluster-admin", "ClusterRole")
+		childCRB.RoleRef = rolereference(clusterRole, "ClusterRole")
 		childCRB.Subjects = addSubject(childCRB.Subjects, r.serviceAccountSubject(reconcilerRef))
```
sdowell, line 813: Do we still need to support allowing multiple subjects on the CRB? It seems like we can set a single subject since the assumption is now that each CRB maps 1:1 with a RootSync. Is the idea to allow other clients to add additional subjects to the CRB?

If we leave this as-is, it seems like we might consider keeping the cleanup logic which first checks that the subjects field is empty. WDYT?

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -0,0 +1,133 @@
...
+```yaml
+apiVersion: configsync.gke.io/v1beta1
+kind: RootSync
+metadata:
+  name: root-sync
+  namespace: config-management-system
+spec:
+  overrides:
+    clusterRole:
+      name: my-cluster-role
```
sdowell, line 101: Was the intent here to set `disabled: true`?

sdowell (commented):
Looks good! A few more comments but this is close to approval

pkg/core/names.go
tomasaschan: It can't, at least not if we create it with `fmt.Sprintf` (my compiler complained that it wasn't a `const` expression).

sdowell: You should be able to do something like

```golang
const (
    RootSyncPermissionsPrefix = configsync.RootSyncKind + ":" + RootReconcilerPrefix
)
```

pkg/reconcilermanager/controllers/rootsync_controller.go
tomasaschan: Hm, yeah, you're right - we should probably make this assume a single subject, and use this here
```suggestion
		childCRB.Subjects = addSubject(nil, r.serviceAccountSubject(reconcilerRef))
```
Not sure if `addSubject` is used anywhere else - if not, we could probably simplify it to e.g. not take a parameter for the existing subjects.

pkg/core/names.go
tomasaschan: Ah, yeah, that might work.

tomasaschan:
@sdowell Hoping this is the one! :)

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -0,0 +1,133 @@
...
+N/A
+
+## Alternatives Considered
+
+One could imagine exposing settings with slightly different semantics - e.g. a simple
+boolean for turning the `ClusterRoleBinding` off, or a setting to change what service
+account the reconciler is running with. However, these both put a bigger burden on the
+user in order to utilize them even for the simple use cases, which is why changing which
+role to bind to is probably the most user-friendly knob to expose. With the inclusion of
+a sentinel value to create no binding, the first of these alternatives is effectively
```
sdowell, line 132: Please update this snippet to reflect that the sentinel value is no longer being used

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -92,6 +92,30 @@ type RootSyncOverrideSpec struct {
...
+	// clusterRole controls which role to bind the service account for this RootSync's
+	// reconciler to.
+	//
+	// +optional
+	ClusterRole ClusterRoleOverrideSpec `json:"clusterRole,omitempty"`
+}
+
+// ClusterRoleOverrideSpec allows to override settings for the root reconciler permissions
+type ClusterRoleOverrideSpec struct {
+	// name is the name of the ClusterRole to bind to; defaults to "cluster-admin".
```
sdowell, line 105: Sorry I got a new round of thoughts after not looking at this over the weekend 😅 

I'm thinking it may be better form to use the following:
```
// +kubebuilder:default:=cluster-admin
```
This would clearly document the default value on the field as well as provide the default behavior at the API level.

---

Additional consideration:

With this realization I'm also considering that we could go back to a single string field and treat it as "disabled" if the user sets the value to `""` (empty string). WDYT?

sdowell: Correction: it looks like the default value only applies if the nested struct (in this case `override`) is set. So that means if the user has an empty override field, the default value would not apply (and `namespaceStrategy` would be `""`). Given this, it looks like we would still need the defaulting logic in the controller and wouldn't be able to rely on an empty string check.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -0,0 +1,133 @@
...
+supported, if through a slightly worse API.
```
tomasaschan, line 133:
```suggestion
a switch to turn off creating the binding entirely, the first of these alternatives is effectively
supported as well, and the API supports extending for other use cases in the future. 
```

docs/design-docs/02-custom-root-reconciler-clusterrole.md
tomasaschan: Done.

sdowell (approved):

karlkfi:
Circling back to this, I think we may need to extend this to support RepoSyncs too, and also multiple bindings, not just one. And if we're quick about it, we can make reverse incompatible changes before the 1.17.0 release.

I'm thinking something like:

```
  override:
    roleRefs:
    - kind: ClusterRole
      name: crd-admin
    - kind: Role
      name: admin
      namespace: example
```

This ouwld hav emultiple advantages:
1. This would allow 0-many bindings.
2. Combining multiple roles means fewer use cases where you need to provision a custom role before using it.
3. You could grant permission to provision to multiple specific namespaces.
4. It would make managing RepoSyncs easier, because you don't need to specify the RoleBinding or the dependency, which is a common gotcha when deleting, especially with deletion-propagation.
5. It would make the two APIs more similar.

Would this satisfy your needs @tomasaschan?
If you can manage multiple, do you still need to be able to disable the cluster-admin on RootSyncs?

tomasaschan:
@karlkfi As long as adding this override means the reconciler isn't granted cluster-admin too, I think this would work for us.

We don't have an explicit use case for turning the binding off entirely, but I can imagine two:

1. The bindings you need are complicated enough that you cannot express them in a single cluster role. That is solved by your proposal, and arguably better since the user doesn't have to create the bindings manually.

2. You're using a different authz mechanism than k8s RBAC, eg WLI + GCP IAM, and don't want to create any RBAC permissions that interfere. I think this could be worked around by binding to a role with no permissions, and might be uncommon enough in practice that a workaround like that, hacky as it is, is OK.

In other words: go ahead!

tomasaschan:
@karlkfi @sdowell Just to clarify: will your team do this work? We still need this feature, and want to have it available in the shape it's going to take when released as soon as we can - so I'll do it if I have to, but I have other things I'd rather spend time on...

karlkfi:
@sdowell has been working on it: https://github.com/GoogleContainerTools/kpt-config-sync/pull/977

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -0,0 +1,133 @@
...
+This role is defaulted to `cluster-admin` in order to stay backwards compatible.
+
+For the case where a single `ClusterRole` is not expressive enough to configure the
+permissions a user want, you can instead set `disabled: true` on the override object:
+
+```yaml
+spec:
+  overrides:
+    clusterRole:
+      disabled: true
```
janetkuo, line 57: Whoever has access to RootSync has access to grant themselves any permission they want by specifying the right cluster role. This has the problem of privilege escalation. However, given that RootSync has cluster-admin access before this change, this change isn't that dangerous, but could be surprising to admins when they grant limited access but also grant RootSync access to someone else. @tomasaschan what do you think about this caveat? Ideally the toggle should live outside of RootSync, but it's less of a problem for RootSync compared to RepoSync API.

tomasaschan: In the setup we run which motivates this feature this won't be an issue - we do not intend to grant access to create new syncs or update existing ones. In fact, the whole point is to keep all syncs in a repo we control, so that we can review the setup and ensure permissions are what they should be, while allowing them to sync config from repos we don't. In other words, this feature is what allows us to _prevent_ privilege escalation.

So I don't think there's a privilege escalation vector here that wasn't here already. If you can create or edit syncs, you can sync in config from an arbitrary source, and grant yourself whatever permissions you want.

As for RepoSyncs, in addition to all of the above, they will not be able to create cluster scoped resources, so any escalation will be even less limited.

janetkuo: The privilege escalation problem is more on the RepoSync given that it has no permission to do anything by default. 

>  If you can create or edit syncs, you can sync in config from an arbitrary source, and grant yourself whatever permissions you want.

This isn't true for RepoSync. RepoSync editors don't have access to modify the RBAC rolebindings granted to the namespace reconciler SA.

PR #926, opened 2023-10-07
test: Add stress tests for OCI and Helm

karlkfi:
- Add TestStressManyDeployments to pre-load stress test clusters,
  so that the control plane auto-scales before using it for other
  stress tests.
- Add TestStressMemoryUsageHelm to ensure the helm-sync container
  works at scale.
- Add TestStressMemoryUsageOCI to ensure the oci-sync container
  works at scale.

Dependencies:
- https://github.com/kptdev/kpt/pull/4055
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/937
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/928
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/945
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/947
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/948

Related Profiling: go/config-sync-reconciler-profiling-2

PR #939, opened 2023-10-13
Add krane/helm/kustomize to gke-e2e & kind-e2e

karlkfi:
Increment image versions to v1.1.0-<hash>

Needed by https://github.com/GoogleContainerTools/kpt-config-sync/pull/926 which uses the krane & helm binaries in the stress tests. Kustomize stress test pending.


build/prow/e2e/Dockerfile
```diff
@@ -69,15 +85,9 @@ ENTRYPOINT ["/usr/bin/bash"]
 FROM gke-e2e as kind-e2e
 
 WORKDIR /repo
-# Steps after here can't be cached since they touch the local filesystem.
-# Just copy everything in the nomos repository so we have whatever we might need.
-COPY . .
-
-# install the nomos CLI, which is used in several e2e tests.
-RUN go install ./cmd/nomos
```
sdowell, line 77: The `nomos` cli is built in this layer because the `kind-e2e` image is special in that is rebuilt for every job execution (using docker-in-docker). As such, `kind-e2e` is always built from the current commit. We always want to use the current version of the nomos binary for testing, so it needs to be created during the job execution.

For gke-e2e periodics, it's pulled from postsubmit artifacts in the gke-e2e container as part of the job entrypoint (make target). The gke-e2e image itself shouldn't have the `nomos` binary pre-installed.

For kind-e2e presubmits, it's built as part of the kind-e2e image build on every job execution.

tools/tools.go
```diff
@@ -0,0 +1,29 @@
...
+// This file will never be built, but `go mod tidy` will see the packages
+// imported here as dependencies and not remove them from `go.mod`.
+
+package main
+
+import (
+	// krane is used in e2e tests to package and push OCI images.
+	// Since we are only using krane for testing, we don't need to fork it
+	// and set up an independent build pipeline, like we do for helm/kustomize.
+	// Instead, we can just vendor the source and build it locally.
```
sdowell, line 27: Do we need to vendor this? Why not just `go install github.com/google/go-containerregistry/cmd/krane`?

karlkfi: I thought it might be a good way to keep krane up to date. This way it will be covered by the CVE warnings, which will cause it to be updated more frequently, even tho we don't ship it for production use. It also means the version is only in one place, which makes it easy to find and update. Plus we already vendor the parent project, so this keeps the versions in sync. 

But if you think it's too confusion or complex, we don't need to vendor it.

build/prow/e2e/Dockerfile
karlkfi: Don't 3/4 layers get re-built every time, since the source is copied into the builder?

I removed the `COPY . .` from the kind-e2e layer because it seemed redundant and expensive. I replaced it with a mount of the repo in the docker run call in Makefile.e2e. Previously, the docker build was copying like 4GiB of context data every time it built, probably due to the .git & .output dirs, which was hella slow.

The way you're describing how it works makes me think we're conflating the target environment and whether it's presubmit/postsumbmit. We also run kind in the postsubmits now, and may want to use gke in presubmits in the future. So maybe we should move this nomos build elsewhere?

build/prow/e2e/Dockerfile
```diff
@@ -27,7 +27,12 @@ FROM ${GOLANG_IMAGE} as builder
 
 WORKDIR /workspace
 
-COPY . .
```
nan-yu, line 30: This seems to be missed before. I don't think we need to copy any source to the `builder` image.

build/prow/e2e/Dockerfile
nan-yu: Agreed that `gke-e2e` should remain separate from the source, and only gets changed when the golang base image needs to be updated. For periodic jobs, `nomos` CLI is installed as a runtime make target, `__install-nomos-local`.

Currently, the `builder` image has a copy of the source, which seems unnecessary. I left a comment above to remove that layer.

tools/tools.go
nan-yu: I would prefer to install a specific version of the binary,  rather than to vendor it as a dependency. `krane` is one option to push OCI images, and we don't want to bundle it with the production code.

build/prow/e2e/Dockerfile
karlkfi: The source is used to build both nomos and junit-report. Both might hypothetically change on any given commit.

karlkfi:
Replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/945

build/prow/e2e/Dockerfile
nan-yu: Oh, right, `junit-report` postprocesses the junit report: http://b/260274932, but I think that rarely changes.

We can still keep `nomos` outside of the builder image.

PR #944, opened 2023-10-16
[WIP] fix: Allow finalizers to be set from source

karlkfi:
Finalizers generally should NOT be set from source, because it will cause Config Sync to error when the finalizer is removed by a controller durring deletion: "no new finalizers can be added if the object is being deleted". BUT some users may want to manually manage finalizers simply to stop deletion of an object until manually removed.

NOTE: I'm not actually sure if this is safe or not. I explicitly removed this capability in https://github.com/GoogleContainerTools/kpt-config-sync/pull/242/files#diff-1f337ad11272c3e7579c8c68abebd2e872b60d378cb62634a1192d9599d46f7aR75 but I don't rembember why exactly. It's possible this will cause test failures when using composition. It might also have broken the kubevirt test, which has a lot of complex finalizer behaviors (e.g. to delete the VM when the VM object is deleted, which depended on some RBAC still existing, requiring deletion ordering).


nan-yu (approved):
/lgtm

PR #945, opened 2023-10-17
ci: add crane binary to e2e environments

sdowell:
This change adds the crane binary to the execution environments for e2e tests. The crane binary is intended to be used by the test framework to push OCI images during test execution.


karlkfi (approved):
/lgtm

PR #947, opened 2023-10-17
fix: Don't panic on empty helm charts

karlkfi:
Add TestHelmEmptyChart to validate


pkg/helm/helm.go
```diff
@@ -191,11 +192,11 @@ func (h *Hydrator) getChartVersion(ctx context.Context) error {
 func fetchNewToken(ctx context.Context) (*oauth2.Token, error) {
 	creds, err := google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")
 	if err != nil {
-		return nil, fmt.Errorf("failed to find default credentials: %w", err)
+		return nil, errors.Wrap(err, "failed to find default credentials")
```
nan-yu, line 195: Just curious why changing to `github.com/pkg/errors`? IIRC, you had a comment before for this archived and random library and would prefer to use `fmt.Errorf`.

e2e/testcases/helm_sync_test.go
```diff
@@ -809,8 +817,174 @@ func TestHelmARTokenAuth(t *testing.T) {
...
+	err = updateYAMLFile(localChartFileName, func(chartMap map[string]interface{}) error {
+		chartMap["name"] = chartName
+		chartMap["version"] = chartVersion
+		return nil
+	})
+	if err != nil {
+		nt.T.Fatalf("failed to update Chart.yaml: %v", err)
+	}
+
+	nt.T.Log("Packaging Helm Chart")
```
nan-yu, line 897: Do we need to pack and push the empty Helm Chart inside the test?

If the Helm Chart always exists, the test doesn't need the write permission. It also doesn't need to pack and push every time, especially when the test runs on multiple clusters.

karlkfi: Ideally the e2e test suite would be runnable in a new project without any manual setup. It's not today, because it assumes a ton of artifacts pre-exist, and we dont have them documented or scripted anywhere.

If we had some system to pre-populate test artifacts, I would use it, but since we don't, I opted to build & push it in the test and also add cleanup and unique versioning to avoid overlap between tests.

pkg/helm/helm.go
karlkfi: I don't like the github.com/pkg/errors library, but I do like consistency, and we're already using this library throughout Config Sync. So I added it here for consistency. If we want to stop using it, someone will need to do a very large refactor.

I didn't add it to all of the errors, because some of them are printing the error first and then the stdout/stderr, which isn't compatible. But that kind of error is also problematic when logged, due to the use to syslog, which interprets each line as its own log message. So all that needs to be fixed at some point later anyway.

e2e/testcases/helm_sync_test.go
karlkfi: fwiw, doing this in the test also helps validate that our docs work when explaining how to manage OCI images and Helm charts.

karlkfi: Another requirement we might consider in the future: 3rd parties should be able to run all the e2e tests. Today they can't, because they use things like Artifact Registry in our project, which requires setting up IAM to a ServiceAccount that you can use for Workload Identity.

nan-yu: We do have scripts and docs for these artifacts: https://github.com/GoogleContainerTools/kpt-config-sync/tree/main/e2e/testinfra#ci-setup.

I believe we need some manual setup anyway, e.g. set-up-kcc-configs.sh and set-up-workload-identity-configs.sh.

For the sake of consistency, I would prefer to create a permanent Helm chart.

It is also not heavy to pack and push the Helm Chart in the test, so I'm okay if you still want to include it in the test.

e2e/testcases/helm_sync_test.go
```diff
@@ -809,8 +817,174 @@ func TestHelmARTokenAuth(t *testing.T) {
...
+	// Create temp dir & delete on exit
+	tempDir, err := os.MkdirTemp("", fmt.Sprintf("%s-", nt.T.Name()))
+	if err != nil {
+		nt.T.Fatalf("failed to create temp directory: %v", err)
+	}
+	nt.T.Cleanup(func() {
+		if err := os.RemoveAll(tempDir); err != nil {
+			nt.T.Errorf("failed to delete temp dir: %v", err)
+		}
+	})
```
nan-yu, line 863: We don't need to create the tempDir. The local chart can be copied to a subfolder of `nt.TmpDir`, which is deleted in the end.

e2e/testcases/helm_sync_test.go
```diff
@@ -809,8 +817,174 @@ func TestHelmARTokenAuth(t *testing.T) {
...
+	}
+
+	// Use unique chart name:version so it doesn't overlap with parallel runs
+	// of this test in the same GCP project.
+	chartName := strcase.ToKebab(nt.T.Name())
+	chartVersion := time.Now().UTC().Format("20060102.150405.999999999")
+	nt.T.Logf("Creating Helm chart: %s:%s", chartName, chartVersion)
+
+	// Copy chart to temp dir
+	localChartPath := filepath.Join(tempDir, "empty")
```
nan-yu, line 882: We can reuse the `CopyChartFromLocal` function: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/helm/helm.go#L97.

karlkfi: Rewrote the test to use the RemoteHelmChart tool. Had to significantly modify RemoteHelmChart to make it work as intended tho.

e2e/testcases/helm_sync_test.go
karlkfi: Done. RemoteHelmChart uses TmpDir.

e2e/testcases/helm_sync_test.go
karlkfi: I switched to using RemoteHelmChart to be consistent with the other helm tests.

pkg/helm/helm.go
karlkfi: Reverted changes to the errors.

karlkfi:
/hold

manually testing on GKE

karlkfi:
/unhold

Tests finally pass

e2e/nomostest/helm/helm.go
```diff
@@ -18,174 +18,284 @@ import (
...
+		}
+		// repository does not exist, continue with creation
+	} else {
+		// repository already exists, skip creation
+		return nil
+	}
+
+	r.Logger.Info("Creating image repository")
+	_, err = r.Shell.ExecWithDebug("gcloud", "artifacts", "repositories",
+		"create", r.RepositoryName,
```
nan-yu, line 77: This requires the CI service account (`e2e-test-runner`) to have the `Artifact Registry Administrator` role to create repositories.
The SA is not managed by latchkey yet, so we need to update it in both the main project and release project.

nan-yu (approved):
/lgtm

e2e/nomostest/helm/helm.go
karlkfi: I have updated the IAM in stolos-dev and oss-prow-build-kpt-config-sync to replace Writer with Admin.

sdowell: Those two are the wrong projects, I can take a look

nan-yu: We need to update `kpt-config-sync-ci-main` and `kpt-config-sync-ci-release`.
`stolos-dev` only has monorepo CI jobs, which doesn't support OCI and Helm.
`oss-prow-build-kpt-config-sync` runs the prow jobs, but the job creates the registry in the main and release projects.

PR #950, opened 2023-10-18
Override the resource requests for helm-sync to address the flakiness of TestPublicHelm

haiyanmeng:
The override setting is borrowed from
https://github.com/GoogleContainerTools/kpt-config-sync/pull/665/files#diff-d79a2bfd62bc74a4236bb679fd53f68e6e7618a1e04fb3b7788add5e36a64243.

 This is to fix the test flakiness on the v1.16 branch.  The main branch  does not have the flakiness.

### DIFF
--- a/e2e/testdata/root-sync-helm-chart-cr.yaml
+++ b/e2e/testdata/root-sync-helm-chart-cr.yaml
@@ -18,6 +18,13 @@ metadata:
   name: root-sync
   namespace: config-management-system
 spec:
+  # This is to fix the test flakiness on the v1.16 branch.
+  # The main branch does not have the flakiness.
+  override:
+    resources:
+    - containerName: helm-sync
+      cpuRequest: "200m"
+      memoryRequest: "600Mi"
   sourceFormat: unstructured
   sourceType: helm
   helm:


e2e/testdata/root-sync-helm-chart-cr.yaml
```diff
@@ -18,6 +18,11 @@ metadata:
   name: root-sync
   namespace: config-management-system
 spec:
+  override:
+    resources:
+    - containerName: helm-sync
+      cpuRequest: "200m"
+      memoryRequest: "600Mi"
```
karlkfi, line 25: I was thinking maybe set the override for all tests in the test setup.
Is this artifact used in all the tests that are failing due to OOMKill?

haiyanmeng: Based on the CI dashboard, fixing this one should get the CI mostly green and avoid the buganizer tickets to be filed.

e2e/testdata/root-sync-helm-chart-cr.yaml
```diff
@@ -18,6 +18,11 @@ metadata:
   name: root-sync
   namespace: config-management-system
 spec:
+  override:
```
karlkfi, line 21: Add a comment explaining why this needs to be set and that it's fixed in 1.17

karlkfi (changes requested):

e2e/testdata/root-sync-helm-chart-cr.yaml
haiyanmeng: Done.

karlkfi (approved):
/lgtm

PR #312, opened 2022-11-18
use variable expansion for image targets

sdowell:
This is intended to make it easier to manage the various image related targets by defining the list of images in one place.

PR #951, opened 2023-10-19
cherry picks for 1.16.3

sdowell:
This PR contains a sequence of cherry picks targeted for 1.16.3, primarily focused on security fixes.

git-sync cherry-picks were omitted since the main branch has moved forward a major version and includes breaking changes. A separate PR will be created for updating git-sync.


sdowell:
> Do #312 and #807 need to be cherrypicked?

807 is needed to prevent conflicts. It looks like 312 can be omitted

PR #956, opened 2023-10-23
set namespaceStrategy default on CRD

sdowell:
This change sets the default value of the namespaceStrategy field explicitly in the CRD.

### DIFF
--- a/manifests/rootsync-crd.yaml
+++ b/manifests/rootsync-crd.yaml
@@ -1445,6 +1445,7 @@ spec:
                     - containerName
                     x-kubernetes-list-type: map
                   namespaceStrategy:
+                    default: implicit
                     description: 'namespaceStrategy controls how the reconciler handles
                       Namespaces which are used by resources in the source but not
                       declared. Only applies when using the unstructured sourceFormat.
--- a/pkg/api/configsync/v1beta1/resource_override.go
+++ b/pkg/api/configsync/v1beta1/resource_override.go
@@ -90,6 +90,7 @@ type RootSyncOverrideSpec struct {
 	// are not declared in the source.
 	//
 	// +kubebuilder:validation:Enum=implicit;explicit
+	// +kubebuilder:default:=implicit
 	// +optional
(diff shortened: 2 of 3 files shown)


sdowell:
The default only applies if the parent struct is set. In this case that struct is `override`, which is an optional field in the `RootSync` spec. This means if the `override` field is unset, the value of the `namespaceStrategy` field will be empty rather than the default `implicit` value. This means we still need the defaulting logic in the controller, because we want the default value to be `implicit` even when no overrides are set.

tomasaschan:
@sdowell This could be changed in a backward-compatible way, I believe, if the `override` field on the CRD was converted from pointer to a plain struct [here](https://github.com/sdowell/kpt-config-sync/blob/634af43e1ccff4025155edd0f2380e58a2ca7d86/pkg/api/configsync/v1beta1/rootsync_types.go#L79) and [here](https://github.com/sdowell/kpt-config-sync/blob/634af43e1ccff4025155edd0f2380e58a2ca7d86/pkg/api/configsync/v1beta1/reposync_types.go#L79). If you do that, you might also want to make corresponding changes on [all the fields in `OverrideSpec`](https://github.com/sdowell/kpt-config-sync/blob/634af43e1ccff4025155edd0f2380e58a2ca7d86/pkg/api/configsync/v1beta1/resource_override.go#L24-L77) but just from reading the fields and their docstrings - but not actually looking at how they're used in controller code - it seems to me that this could be possible without changing any observable behavior, although it depends a bit on how an unset [`GitSyncDepth`](https://github.com/sdowell/kpt-config-sync/blob/634af43e1ccff4025155edd0f2380e58a2ca7d86/pkg/api/configsync/v1beta1/resource_override.go#L37) is handled in practice:

* If not setting `GitSyncDepth` means it defaults to a shallow clone, i.e. equivalent with `GitSyncDepth: 0`, all of this will work out.
* If not setting `GitSyncDepth` defaults to some value other than `0` in the controller, then making this change here would mean that we can no longer distinguish between "I want a shallow clone of depth 0" vs "I haven't set this override". Whether that makes sense depends on how the controller handles a value of `0` today.

PR #957, opened 2023-10-23
ci: publish buildenv in postsubmit

sdowell:
This change publishes the buildenv image in the postsubmit target if the image tag does not already exist. This removes any need for manual steps to publish a new buildenv image.


nan-yu (approved):
/lgtm

PR #958, opened 2023-10-23
ci: run all of test-presubmit in a single container

sdowell:
This refactors the test-presubmit entrypoint to wrap the entrypoint and all nested targets in a single container. This ensures that all of the presubmit targets run inside the same nested container. Before this change, some targets such as tidy/vendor ran on the host machine. This also has the benefit on minimizing the number of extra containers that need to be pulled and/or created.

### DIFF
--- a/scripts/license-headers.sh
+++ b/scripts/license-headers.sh
@@ -25,6 +25,8 @@ ignores=(
   "-ignore=e2e/testdata/helm-charts/**"
   "-ignore=.output/**"
   "-ignore=e2e/testdata/*.xml"
+  "-ignore=.idea/**"
+  "-ignore=.vscode/**"
 )
 
 case "$1" in
--- a/scripts/lint-bash.sh
+++ b/scripts/lint-bash.sh
@@ -26,28 +26,6 @@ mapfile -t check_files < <(
     uniq -u
 )
 
-# Handle bats tests
-bats_tmp="$(mktemp -d lint-bash-XXXXXX)"
-function cleanup() {
-  rm -rf "${bats_tmp}"
-}
-trap cleanup EXIT
-
-readonly linter=koalaman/shellcheck:v0.6.0
-
-if ! docker image inspect "$linter" &>/dev/null; then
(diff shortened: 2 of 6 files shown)


scripts/license-headers.sh
```diff
@@ -25,6 +25,7 @@ ignores=(
   "-ignore=e2e/testdata/helm-charts/**"
   "-ignore=.output/**"
   "-ignore=e2e/testdata/*.xml"
+  "-ignore=.idea/**"
```
karlkfi, line 28: Add .vscode while you're at it.
Maybe also to the .dockerignore.
Both are in .gitignore

scripts/lint-bash.sh
```diff
@@ -26,28 +26,6 @@ mapfile -t check_files < <(
...
-  cmd+=(--tty)
-fi
-cmd+=(
-  --rm
-  "$linter" "${check_files[@]}"
-)
-
 echo "Linting scripts..."
-"${cmd[@]}"
+shellcheck "${check_files[@]}"
```
karlkfi, line 30: shellcheck is installed in the buildenv?

scripts/license-headers.sh
sdowell: done

scripts/lint-bash.sh
karlkfi: oh, i see you added it. but you didn't set the version. Should we be worried about the version changing randomly?

sdowell: it is with this change

karlkfi: has anything changed since 0.6.0? what's the new version?

sdowell: The current version this is installing is 0.7.1. shellcheck complained about the wordsplitting in generate-clientset which I fixed in this change, but that was the only issue.

https://packages.debian.org/bullseye/shellcheck

karlkfi (approved):
/lgtm

PR #959, opened 2023-10-23
refactor: build resource-group image in repo

sdowell:
This changes the resource-group-controller image to be built alongside the other first party images rather than built separately and imported as a third party image. The ResourceGroup controller is a core component of Config Sync, and this is intended to simplify the maintenance of this component.


cmd/resource-group/main.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2022 Google LLC
```
nan-yu, line 1: nit: 2023

cmd/resource-group/main.go
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"os"
+
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
nan-yu, line 20: I think I missed that in the resource-group repo, but the import is no longer needed.

cmd/resource-group/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"kpt.dev/resourcegroup/controllers/runner"
+	// +kubebuilder:scaffold:imports
```
nan-yu, line 22: same here

cmd/resource-group/main.go
sdowell: done

cmd/resource-group/main.go
sdowell: done

cmd/resource-group/main.go
sdowell: done

nan-yu (approved):
/lgtm

PR #960, opened 2023-10-24
ci: check analysis status in vuln scan script

sdowell:
Some images may fail to be scanned for various reasons. At the moment this leads to a silent failure where the fixable_total is reported as zero. This change checks the analysisStatus and treats any non-success status as a failure.

### DIFF
--- a/scripts/vulnerabilities.sh
+++ b/scripts/vulnerabilities.sh
@@ -37,6 +37,7 @@ source "${scripts_dir}/lib/manifests.sh"
 read -r -a images <<< "$(config_sync_images)"
 [[ ${#images[@]} -eq 0 ]] && exit 1
 fixable_total=0
+scan_failure_total=0
 declare -A vuln_map
 
 echo -n "Scanning" >&2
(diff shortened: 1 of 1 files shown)


scripts/vulnerabilities.sh
```diff
@@ -37,19 +37,26 @@ source "${scripts_dir}/lib/manifests.sh"
...
   echo -n "."
-  vulnerabilities=$(gcloud beta container images describe --show-package-vulnerability --format json --verbosity error "${image}" |
-    jq -r '.package_vulnerability_summary.vulnerabilities')
-  fixable=$(echo "${vulnerabilities}" |
-    jq -r 'select(.vulnerability != {}) | map(map(select(.vulnerability.packageIssue[].fixAvailable == true)) | length) + [0] | add')
-  vuln_map[${image}]=${fixable}
-  fixable_total=$((fixable_total + fixable))
+  results=$(gcloud beta container images describe --show-package-vulnerability --format json --verbosity error "${image}")
+  status=$(echo "${results}" | jq -r '.discovery_summary.discovery[].discovery.analysisStatus')
+  if [[ "${status}" != "FINISHED_SUCCESS" ]]; then
```
nan-yu, line 50: Are all failed statuses unsupported failures?
I'm wondering whether we should distinguish between unsupported and other failures.
Or at least rename `unsupported_total` to something like `scan_failure_total`.

sdowell: I feel it's safest to treat anything that isn't `FINISHED_SUCCESS` as something unexpected and treat it as a failure. This will alert us that something went wrong so that we can take a closer look. This may include statuses other than  `FINISHED_UNSUPPORTED`. I can rename the variable

sdowell: done

PR #961, opened 2023-10-24
Validate all Config Sync controller Namespaces

tiffanny29631:
Config Sync checks and errors out if config-management-system namespace is declared in SoT. 

This PR extends the check to config-management-monitoring namespace and resource-group-controller to avoid accidental modification or removal to the managed namespaces.

### DIFF
--- a/pkg/api/configmanagement/register.go
+++ b/pkg/api/configmanagement/register.go
@@ -57,12 +57,16 @@ const (
 	// RGControllerNamespace is the namespace used for the resource-group controller
 	RGControllerNamespace = "resource-group-system"
 
+	// MonitoringNamespace is the namespace used for Config Sync monitoring
+	MonitoringNamespace = "config-management-monitoring"
+
 	// RGControllerName is the name used for the resource-group controller
 	RGControllerName = "resource-group-controller-manager"
 )
 
-// IsControllerNamespace returns true if the namespace is the ACM Controller Namespace.
+// IsControllerNamespace returns true if the namespace is one of the Config Sync controller Namespace.
 func IsControllerNamespace(name string) bool {
-	// For now we only forbid syncing the Namespace containing the ACM controllers.
-	return name == ControllerNamespace
+	return name == ControllerNamespace ||
+		name == RGControllerNamespace ||
+		name == MonitoringNamespace
 }
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -110,7 +110,7 @@ func IsReconcilerManagerConfigMap(obj client.Object) bool {
 // otel-collector Deployment in the config-management-monitoring namespace.
 func isOtelCollectorDeployment(obj client.Object) bool {
 	return obj.GetName() == ocmetrics.OtelCollectorName &&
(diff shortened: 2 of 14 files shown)


pkg/validate/raw/validate/namespace_validator.go
```diff
@@ -38,7 +38,7 @@ func validateNamespace(obj ast.FileObject) status.Error {
 	if !isValidNamespace(obj.GetName()) {
 		return nonhierarchical.InvalidNamespaceError(obj)
 	}
-	if configmanagement.IsControllerNamespace(obj.GetName()) {
+	if configmanagement.IsConfigSyncManagedNamespace(obj.GetName()) {
 		return nonhierarchical.IllegalNamespace(obj)
```
nan-yu, line 42: `IllegalNamespace` also needs to be updated to reflect the namespace.

pkg/api/configmanagement/register.go
```diff
@@ -57,12 +57,16 @@ const (
...
+
 	// RGControllerName is the name used for the resource-group controller
 	RGControllerName = "resource-group-controller-manager"
 )
 
-// IsControllerNamespace returns true if the namespace is the ACM Controller Namespace.
-func IsControllerNamespace(name string) bool {
-	// For now we only forbid syncing the Namespace containing the ACM controllers.
-	return name == ControllerNamespace
+// IsConfigSyncManagedNamespace returns true if the namespace is the Config Sync managed Namespace.
```
nan-yu, line 67: `Config Sync managed Namespace` sounds confusing. Namespaces declared in the source repo and applied by Config Sync are all managed Namespaces.

`IsControllerNamespace` seems to be okay as RG and otel-collector are also controllers required by CS.

pkg/api/configmanagement/register.go
```diff
@@ -57,12 +57,16 @@ const (
 	// RGControllerNamespace is the namespace used for the resource-group controller
 	RGControllerNamespace = "resource-group-system"
 
+	// MonitoringNamespace is the namespace used for Config Sync monitoring
+	MonitoringNamespace = "config-management-monitoring"
```
nan-yu, line 61: The same variable is already declared in `pkg/metrics/otel.go`. We can consolidate them.

nan-yu (changes requested):

pkg/validate/raw/validate/namespace_validator.go
tiffanny29631: Done in the function.

pkg/api/configmanagement/register.go
tiffanny29631: Makes sense, changed back to IsControllerNamespace(). Also aligned the PR title.

pkg/api/configmanagement/register.go
tiffanny29631: Using metrics.MonitoringNamespace would cause an import cycle.

tiffanny29631: That would work, changed.

nan-yu (approved):
/lgtm

PR #966, opened 2023-10-24
TestHelmLatestVersion: Remove helm chart version suffix

victorpras:
Helm is still unable to retrieve from AR using empty version if the version has suffixes, so we're removing version suffix addition. Chart version that exceed 20 characters will throw an error

### DIFF
--- a/e2e/nomostest/artifactregistry/image.go
+++ b/e2e/nomostest/artifactregistry/image.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 
 	"github.com/ettle/strcase"
-	"github.com/google/uuid"
 	"kpt.dev/configsync/e2e"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/testlogger"
@@ -51,6 +50,11 @@ func SetupImage(nt *nomostest.NT, name, version string) (*Image, error) {
 	if version == "" {
 		return nil, fmt.Errorf("image version must not be empty")
 	}
+
+	// Version will error out if it's longer than 20 characters
+	if err := validateImageVersion(version); err != nil {
+		return nil, err
+	}
 	chart := &Image{
 		Shell:   nt.Shell,
 		Logger:  nt.Logger,
@@ -62,9 +66,8 @@ func SetupImage(nt *nomostest.NT, name, version string) (*Image, error) {
 		// Use chart name to avoid overlap.
 		BuildPath: filepath.Join(nt.TmpDir, name),
 		// Use test name to avoid overlap. Truncate to 40 characters.
-		Name: generateImageName(nt, name),
-		// Use random suffix to avoid overlap. Truncate to 20 characters.
-		Version: generateImageVersion(version),
+		Name:    generateImageName(nt, name),
+		Version: version,
 	}
 	nt.T.Cleanup(func() {
 		if err := chart.Delete(); err != nil {
@@ -216,10 +219,12 @@ func (r *Image) SetName(nt *nomostest.NT, name string) {
 
 // SetVersion updates the local version of the image to the specified tag with a
 // random suffix
-func (r *Image) SetVersion(version string) {
-	version = generateImageVersion(version)
-	r.Logger.Infof("Updating image version to %q", version)
+func (r *Image) SetVersion(version string) error {
+	if err := validateImageVersion(version); err != nil {
+		return err
+	}
 	r.Version = version
+	return nil
 }
 
 // Delete the package from the remote registry, including all versions and tags.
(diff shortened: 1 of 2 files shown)


e2e/nomostest/artifactregistry/image.go
```diff
@@ -64,7 +63,7 @@ func SetupImage(nt *nomostest.NT, name, version string) (*Image, error) {
 		// Use test name to avoid overlap. Truncate to 40 characters.
 		Name: generateImageName(nt, name),
 		// Use random suffix to avoid overlap. Truncate to 20 characters.
```
karlkfi, line 65: We either need to truncate to 20 characters or just error if the name is too long. Error might be better if it needs to be semver for helm. And this comment need updating.

victorpras: version length check has been added and will error out when it's longer than 20 chars. Comments have been updated

e2e/nomostest/artifactregistry/image.go
```diff
@@ -51,6 +50,12 @@ func SetupImage(nt *nomostest.NT, name, version string) (*Image, error) {
 	if version == "" {
 		return nil, fmt.Errorf("image version must not be empty")
 	}
+
+	// Version will error out if longer than 20 characters
+	v, err := validateImageVersion(version)
```
karlkfi, line 55: I don't think this needs to return the version if it's not being changed. Just validate and return the error.

e2e/nomostest/artifactregistry/image.go
```diff
@@ -248,14 +255,12 @@ func generateImageName(nt *nomostest.NT, chartName string) string {
...
+// validateImageVersion will validate if chart version string
+// is 20 characters or less and can function as a k8s metadata.name.
 // Chart name and version must be less than 63 characters combined.
-func generateImageVersion(chartVersion string) string {
-	if len(chartVersion) > 12 {
-		chartVersion = chartVersion[:12]
-		chartVersion = strings.Trim(strings.Trim(chartVersion, "-"), ".")
+func validateImageVersion(chartVersion string) (string, error) {
+	if len(chartVersion) > 20 {
+		return chartVersion, fmt.Errorf("chart version string %s should not exceed 20 characters", chartVersion)
```
karlkfi, line 263: Use %q for the value so it's easy to tell if it's empty or has spaces.

karlkfi (changes requested):

e2e/nomostest/artifactregistry/image.go
```diff
@@ -63,8 +68,8 @@ func SetupImage(nt *nomostest.NT, name, version string) (*Image, error) {
 		BuildPath: filepath.Join(nt.TmpDir, name),
 		// Use test name to avoid overlap. Truncate to 40 characters.
 		Name: generateImageName(nt, name),
-		// Use random suffix to avoid overlap. Truncate to 20 characters.
-		Version: generateImageVersion(version),
+		// Version will error out if longer than 20 characters
```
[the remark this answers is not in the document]
victorpras: removed

e2e/nomostest/artifactregistry/image.go
victorpras: removed

e2e/nomostest/artifactregistry/image.go
victorpras: replaced with %q

e2e/nomostest/artifactregistry/image.go
```diff
@@ -248,14 +254,12 @@ func generateImageName(nt *nomostest.NT, chartName string) string {
...
-// generateImageVersion returns the version with a random 8 character suffix.
-// Result will be no more than 20 characters and can function as a k8s metadata.name.
+// validateImageVersion will validate if chart version string
+// is 20 characters or less and can function as a k8s metadata.name.
 // Chart name and version must be less than 63 characters combined.
-func generateImageVersion(chartVersion string) string {
-	if len(chartVersion) > 12 {
-		chartVersion = chartVersion[:12]
-		chartVersion = strings.Trim(strings.Trim(chartVersion, "-"), ".")
+func validateImageVersion(chartVersion string) (string, error) {
```
karlkfi, line 260: The version return value isn't used. You can remove it.

karlkfi (changes requested):

e2e/nomostest/artifactregistry/image.go
victorpras: removed string return

e2e/nomostest/artifactregistry/image.go
```diff
@@ -51,6 +50,11 @@ func SetupImage(nt *nomostest.NT, name, version string) (*Image, error) {
 	if version == "" {
```
karlkfi, line 50: Can you move this empty check into `validateImageVersion` so it is also checked when SetVersion is called?

karlkfi (approved):
/lgtm

e2e/nomostest/artifactregistry/image.go
victorpras: moved

PR #974, opened 2023-10-26
Refactor reconciler container to use controller-runtime framework

nan-yu:
The reconciler container has a parser running in the main thread, a remeidator running in a separate goroutine, and a Kubernetes controller manager for the finalizer running in another goroutine.

This commit refactors it to use the controller-runtime framework. After the change, the main process just starts the Kubernetes controller manager. Three controllers are registered to the manager:
- Syncer: read, parse, validate, apply source configs to the cluster, and start a watcher for the GKNN of each declared resources.
- Remediator: watch for the remediate event and correct dirft.
- Finalizer: manage the lifycycle of R*Sync's finalizer and finalizing condition, and garbage-collect the managed resources.


cmd/reconciler/main.go
```diff
@@ -131,6 +159,122 @@ func main() {
...
+		SyncScope:          reconcilerScope,
+		SyncName:           *syncName,
+		Applier:            baseApplier,
+		Resources:          decls,
+		RemediateResources: remResources,
+		Remediator:         remInterface,
+	}
+
+	// Register the Remediator controller
+	if err = rem.SetupWithManager(mgr); err != nil {
```
sdowell, line 274: When registering multiple controllers with the same manager, is there any guarantee that different controllers will not run concurrently?

The reason I ask is that we probably want to ensure the Syncer and Remediator controllers are never running at the same time

nan-yu: Probably not. Kubernetes controllers operate independently, and their execution is not coordinated by default.

To ensure the Syncer and Remdiator controllers are not running at the same time, I used a shared flag `canRemediate`: https://github.com/GoogleContainerTools/kpt-config-sync/pull/974/files#diff-55e9dd0d45f70809718f2bafb46a7e8b0328caa9a2a018e442c3032ec35ba27cR46. Before the Syncer applies the declared resources, it pauses the remediator, and resumes it after the apply succeeds. The Remediator starts enqueueing and processing requests only when it the flag is true.

pkg/remediator/remediator.go
```diff
@@ -175,8 +104,7 @@ func (r *Remediator) Pause() {
 	defer r.lifecycleMux.Unlock()
 
 	klog.V(1).Info("Remediator pausing...")
-	r.stopFn() // tell the workers to stop
-	<-r.doneCh // wait until workers are done (if running)
+	r.canRemediate = false
```
sdowell, line 107: This will prevent new remediator executions from starting after `Pause`, but does not guarantee that a prior execution is not still running. This means that if a remediator Reconcile starts before a syncer Reconcile, they could be running concurrently. The syncer still needs to block and wait for remediator executions to complete

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,335 @@
...
+			case <-r.ControllerCtx.Done():
+				// The controller is shutting down, close the timerEvents channel and exit
+				close(timerEvents)
+				close(r.DoneCh) // inform finalizer the remediator is done
+				return
+
+			case <-remediateTimer.C:
+				if r.CanRemediate() {
+					// Skip enqueuing remediator requests if the Applier is running.
+					r.enqueueEvents(timerEvents)
```
sdowell, line 183: Since we are enqueueing objects for remediation, the following scenario seems possible:
- sync to commit A
- enqueue remediation requests for commit A
- new commit B is created, pausing remediator
- commit B is synced, and the inventory differs substantially from commit B
- remediator continues and beings processing remediation requests for commit A

In this hypothetical, would the remediator try to correct for drift based on the previous commit? If so, what would that look like? I'm thinking we may want to reset the queue after the syncer runs the applier

nan-yu: If the above case happens, the remediator is basically a no-op for queued objects from commit A.
Here are the possible scenarios:
1. new commit B doesn't have the same GVK, so the watch is stopped. No new events will be added to the queue. Remaining items in the queue will be drained as a DELETE operation (the object is already deleted by the syncer for commit B, so no-op). 
2. new commit has the same GVK, but the object is updated. When commit B is synced, the watch receives the update event, updates the queue, and marks canRemediate as true. The remediator processes the request. The object in the queue is the same as the actual object, so no op.
3. new commit has the same GVK, and the object is the same. The watch won't receive any new event, so no remediation.
4. new commit has new GVK, new watch will be set up to trigger new remediator.

If we reset the queue after the syncer runs the applier, it helps avoid draining the queue in the first scenario only, which is still beneficial, but it requires all Remediate requests to be in the queue, which also means we can't use the controller-runtime to requeue the requests.

pkg/remediator/remediator.go
nan-yu: The remediator reconciles a single object at a time. When a remediator reconcile starts before a syncer reconcile, there are at most one remediator reconciliation. When it finishes, remaining remediations will be skipped and requeued, so there won't have continuous fights.

If we block the Syncer while Remediator is processing each object, there will be very frequent pause and resume signals for the Syncer, which might be an issue.

sdowell: This is fundamentally a race condition which could lead to surprising/unexpected behavior, so should be prevented IMO. The remediator generally should only be running between syncs, and never concurrently with syncer.

It seems essential for the syncer to send a pause signal to the remediator and receive a done signal before proceeding.

nan-yu: If the syncer needs to block and wait for remediator to complete, it is possible that the syncer is blocked by remediator errors because those errors will be retried.

sdowell: In this example I think it would be worse to let the syncer run while there's a remediator thread retry-looping, as this just increases the risk of a race condition (remediator thread continuing to run concurrently during syncer execution).

To minimize the impact of this we can wrap the remediator thread in a context with a reasonable timeout for the remediator (I expect this should probably be relatively short for remediator). We could also potentially cancel the context in the Pause function

nan-yu: Syncer will pause the remediator. With the current logic, if an object fails to be remediated, it keeps retrying until the next syncer kicks in. The syncer pauses the remediator, so no new remediation will run. After the syncer finishes applying, the remediate queue will be updated, and the remediator is resumed.
When dequeuing the object, it may already get updated, so becomes a no-op. If not, it continues with retry.

sdowell: I was responding to your previous comment, which implies the remediator may continue retry-looping after being paused

> If the syncer needs to block and wait for remediator to complete, it is possible that the syncer is blocked by remediator errors because those errors will be retried.

nan-yu: Another caveat inherited from the previous architecture is that the Syncer applies a list of declared objects, while the Remediator reconciles a single object. I'm thinking if we should change the Remediator to process a list of resources. If so, then it becomes a mini-apply. Do we want to use the kpt applier to remediate them, and how do we handle the fight?

nan-yu: > I was responding to your previous comment, which implies the remediator may continue retry-looping after being paused
> 
> > If the syncer needs to block and wait for remediator to complete, it is possible that the syncer is blocked by remediator errors because those errors will be retried.

Nope, it won't. After being paused, the Remediator's request will be skipped (but requeued) if `canRemediate` is false.

sdowell: In that case, it sounds like the syncer just needs to wait for the remediator to complete any prior requests after pausing

nan-yu: I see. So basically when the syncer applies declared objects, it:
1. pause the Remediator
2. wait for the Remediator to be paused
3. start applying
4. resume the Remediator if apply succeeds.

sdowell: Yes, that's how I see it. If we don't perform step 2 (`wait for the Remediator to be paused`) then we allow the possibility for unpredictable race conditions

nan-yu: That seems to go hand-in-hand with your other comment. After waiting for Remediator to be paused, we can clear the remediate queue. The only caveat is to not requeue the request with the controller-runtime library, but only use the remediate queue for requeueing and dequeueing.

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	if !r.CanRemediate() {
+		// Mark the object as done so that it can be requeued later when the Applier is not running.
+		r.DoneObj(id)
+		return ctrl.Result{}, nil
+	}
+
+	// Mark the Remediator is working to block the Syncer.
+	r.Remediating()
```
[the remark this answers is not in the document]
nan-yu: They are not atomic because `Remediating` doesn't always run after `CanRemediate`.
`CanRemediate` is invoked in two places, while `Remediating` is invoked only once:
- When registering the Remediator in `SetupWithManager`, we started a goroutine to enqueue the request when `CanRemediate`, but `Remediating` isn't invoked here.
- In the `reconcile` function, it calls `Remediating` when `CanRemediate` is true.

If we make them atomic, we need to allow enqueuing when the applier is running. I think that is okay as one object only gets updated once in the apply, and will be enqueued after the apply completes. Even though it is enqueued earlier, it won't get processed until the apply completes.

pkg/remediator/remediator.go
```diff
@@ -92,102 +95,82 @@ var _ Interface = &Remediator{}
...
-	klog.V(3).Info("Remediator paused")
+	klog.V(1).Info("Remediating...")
+	r.wg.Add(1)
 }
 
-// Resume the Remediator by starting the workers.
-func (r *Remediator) Resume() {
+// WaitRemediating blocks until the Remediator finishes remediating the requests.
+func (r *Remediator) WaitRemediating() {
 	r.lifecycleMux.Lock()
```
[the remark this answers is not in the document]
nan-yu: Yes, you're right. The `sync.WaitGroup` is indeed thread-safe.

pkg/reconciler/controllers/remediator.go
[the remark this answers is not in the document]
nan-yu: After a second thought, we still need to check `CanRemediate` when enqueueing requests. Otherwise, it will keep attempting to enqueue while the applier is running.
The enqueue attempts are unnecessary for long-running apply.

We can still have two functions, one check if it can remediate only, and the other does both check and remediate atomically.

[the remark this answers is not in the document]
sdowell: What's important here is that the `canRemediate` check and the `waitgroup.Add` call happen in the same critical section.

If we need a separate function signature which doesn't call `waitgroup.Add` to support the enqueue logic that's fine.

For example, change the `Remediating` function signature to:
```golang
// Remediating indicates the Remediator starts processing the remediate requests.
func (r *Remediator) Remediating() bool {
	r.lifecycleMux.Lock()
	defer r.lifecycleMux.Unlock()

	klog.V(1).Info("Remediating...")
    if r.canRemediate {
        r.wg.Add(1)
    }
    return r.canRemediate
}
```

pkg/remediator/remediator.go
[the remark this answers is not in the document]
nan-yu: done

pkg/reconciler/controllers/remediator.go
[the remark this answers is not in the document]
nan-yu: done

pkg/reconciler/controllers/remediator.go
nan-yu: done

pkg/remediator/remediator.go
nan-yu: done

e2e/testcases/remediator_test.go
```diff
@@ -77,4 +87,129 @@ func TestSurfaceFightError(t *testing.T) {
...
+	nt.T.Log("Simulate delete fight")
+	doneCh := make(chan struct{})
+	go func() {
+		close(doneCh)
+		// Make the # of updates exceed the fightThreshold defined in pkg/syncer/reconcile/fight/detector.go
+		for i := 0; i <= 5; i++ {
+			if err := nt.KubeClient.Delete(rb, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil && !apierrors.IsNotFound(err) {
+				nt.T.Error(err)
+			}
+			time.Sleep(time.Second)
```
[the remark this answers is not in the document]
nan-yu: Updated to use the wait condition

e2e/testcases/remediator_test.go
```diff
@@ -77,4 +87,129 @@ func TestSurfaceFightError(t *testing.T) {
...
+		fmt.Sprintf("acme/namespaces/%s/rb.yaml", ns.Name), rb))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Add Namespace and RoleBinding"))
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Simulate delete fight")
+	doneCh := make(chan struct{})
+	go func() {
+		close(doneCh)
```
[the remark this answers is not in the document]
nan-yu: Removed the goroutine

e2e/testcases/remediator_test.go
```diff
@@ -77,4 +87,129 @@ func TestSurfaceFightError(t *testing.T) {
...
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add(
+		fmt.Sprintf("acme/namespaces/%s/rb.yaml", ns.Name), rb))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Add Namespace and RoleBinding"))
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Simulate delete fight")
+	doneCh := make(chan struct{})
+	go func() {
```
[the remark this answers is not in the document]
nan-yu: The fight will occur after the goroutine is done, so it is not necessary. Removed.

pkg/remediator/remediator.go
```diff
@@ -71,6 +62,16 @@ type Interface interface {
 	Pause()
 	// Resume the Remediator by starting the workers.
 	Resume()
+	// CanRemediate returns whether the Remediator can start remediating.
+	CanRemediate() bool
+	// Remediating indicates the Remediator starts processing the remediate requests if it can remediate.
+	Remediating() bool
```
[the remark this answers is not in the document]
nan-yu: Renamed to `StartRemediating` to match `DoneRemediating`.

pkg/reconciler/controllers/syncer.go
```diff
@@ -0,0 +1,298 @@
...
+		for {
+			select {
+			case <-r.controllerCtx.Done():
+				// The controller is shutting down, close the timerEvents channel and exit
+				close(timerEvents)
+				close(r.doneCh) // inform finalizer the syncer is done
+				return
+
+			case <-r.syncing:
+				// Skip enqueueing any events when the reconcile function is running.
```
sdowell, line 214: What is the intent with this `case`? It doesn't seem like the current implementation would prevent enqueueing new events while the reconcile function is running. Only the first iteration of the for loop would receive `true` on this channel

pkg/reconciler/controllers/syncer.go
```diff
@@ -0,0 +1,298 @@
...
+			r.statusUpdateTimer.Stop()
+		}()
+
+		r.state.InitRetryTimerAndPeriod(r.retryTimer, r.retryPeriod)
+		for {
+			select {
+			case <-r.controllerCtx.Done():
+				// The controller is shutting down, close the timerEvents channel and exit
+				close(timerEvents)
+				close(r.doneCh) // inform finalizer the syncer is done
```
sdowell, line 210: Same comment here, do we need to wait for reconcile loop to exit?

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,342 @@
...
+	go func() {
+		remediateTimer := time.NewTimer(time.Second)
+		defer remediateTimer.Stop()
+
+		for {
+			select {
+			case <-r.ControllerCtx.Done():
+				// The controller is shutting down, close the timerEvents channel and exit
+				close(timerEvents)
+				close(r.DoneCh) // inform finalizer the remediator is done
```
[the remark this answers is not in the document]
nan-yu: The context cancel doesn't kill those threads, but those threads will fail because the context is canceled, and won't be requeued because the controller is shutting down. The downside is that there will be remediate errors in the logs because of the context cancelation.

We can pause the remediator (prevents new threads from being executed) and wait for the currently running thread (should be a single one) to complete before closing the doneCh (prevents errors from spamming the remediator log).

pkg/reconciler/controllers/syncer.go
nan-yu: The intent is to avoid enqueueing any retry/statusUpdate events, and also avoid repeatedly resetting the retry timer while the reconcile function is still running.

You're right. The channel doesn't block to wait for r.syncing to become true.

The retry/statusUpdate events will only be enqueued once during the long-running reconciliation because the retry timer is set after the retry execution.
We can move  `r.retryTimer.Reset(r.retryPeriod)` to the reconcile function and reset the retry timer before the function returns to prevent it from being reset repeatedly.

pkg/reconciler/controllers/syncer.go
nan-yu: done. Added a signal to wait for the current reconcile loop to complete.

pkg/reconciler/controllers/syncer.go
nan-yu: done

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	r.RemoveFightError(id)
+	return remediatedRV, nil
+}
+
+// remediateDiff takes diff (declared & actual) and ensures the server matches the
+// declared state.
+// It returns the ResourceVersion of the new remediated object.
+func (r *Remediator) remediateDiff(ctx context.Context, id core.ID, objDiff diff.Diff) (string, status.Error) {
```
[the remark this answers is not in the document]
nan-yu: unit tests added

pkg/reconciler/controllers/remediator.go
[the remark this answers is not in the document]
nan-yu: done

cmd/reconciler/main.go
```diff
@@ -148,9 +292,103 @@ func main() {
...
+//   - Supervisor is used by the Finalizer to finalize managed resources. It is also
+//     used by the Syncer to apply/prune declared resources.
+func configureClients(scope declared.Scope) (client.Client, discovery.DiscoveryInterface, reconcile.Applier, applier.Supervisor, error) {
+	reconcileTimeoutDuration, err := timeStringToDuration(*reconcileTimeout)
+	if err != nil {
+		return nil, nil, nil, nil, fmt.Errorf("reconcileTimeout format error: %v", err)
+	}
+
+	// Get a config to talk to the apiserver.
+	apiServerTimeoutDuration, err := time.ParseDuration(*apiServerTimeout)
```
janetkuo, line 330: Why isn't `timeStringToDuration` used here?

cmd/reconciler/main.go
```diff
@@ -122,6 +139,17 @@ var flags = struct {
 	namespaceStrategy: "namespace-strategy",
 }
 
+func timeStringToDuration(t string) (time.Duration, error) {
```
janetkuo, line 142: nit: name it something more specific given that it expects Duration to be non-negative.

janetkuo, line 142: nit: name it something more specific given that it expects Duration to be non-negative.

cmd/reconciler/main.go
janetkuo, line 330: Why isn't `timeStringToDuration` used here?

Do we need to check whether it's non-negative? I see this being used for parsing other timeout strings but the check wasn't there

cmd/reconciler/main.go
```diff
@@ -75,10 +94,8 @@ var (
 		"The rate of updates per minute to an API Resource at which the Syncer logs warnings about too many updates to the resource.")
 	resyncPeriod = flag.Duration("resync-period", configsync.DefaultReconcilerResyncPeriod,
 		"Period of time between forced re-syncs from source (even without a new commit).")
-	workers = flag.Int("workers", 1,
-		"Number of concurrent remediator workers to run at once.")
 	pollingPeriod = flag.Duration("filesystem-polling-period",
-		controllers.PollingPeriod(reconcilermanager.ReconcilerPollingPeriod, configsync.DefaultReconcilerPollingPeriod),
+		rmcontrollers.PollingPeriod(reconcilermanager.ReconcilerPollingPeriod, configsync.DefaultReconcilerPollingPeriod),
```
karlkfi, line 98: Why are we using a function from the reconciler-manager in the reconciler? 
Is it just re-use because they have the same flag and behavior?

cmd/reconciler/main.go
karlkfi: non-negative seems like validation, which is different form the parsing behavior. perhaps they should be seperate.

cmd/reconciler/main.go
```diff
@@ -131,6 +159,122 @@ func main() {
...
+	fight.SetFightThreshold(*fightDetectionThreshold)
+
+	cl, discoveryClient, baseApplier, supervisor, err := configureClients(reconcilerScope)
+	if err != nil {
+		klog.Error(err)
+	}
+
+	// Start listening to signals
+	signalCtx := signals.SetupSignalHandler()
+	mgrOptions := ctrl.Options{
```
karlkfi, line 172: Do we want to set the MapperProvider here? Seems like we already have a discovery client and usually the mapper wraps that. But I'm not sure what the default mapper is.

karlkfi: We might also want to set the Logger

karlkfi: NewClient maybe too, if we already have a client.

cmd/reconciler/main.go
```diff
@@ -131,6 +159,122 @@ func main() {
...
+	// Otherwise, all namespaced informers will watch at the cluster-scope.
+	// This prevents Namespaced Reconcilers from needing cluster-scoped read
+	// permissions.
+	if reconcilerScope != declared.RootReconciler {
+		mgrOptions.Namespace = string(reconcilerScope)
+	}
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), mgrOptions)
+	if err != nil {
+		klog.Fatalf("Error starting manager: %v", err)
```
karlkfi, line 188: suggest: "Error starting controller manager"

cmd/reconciler/main.go
```diff
@@ -131,6 +159,122 @@ func main() {
...
+	if err = finalizerController.SetupWithManager(mgr); err != nil {
+		klog.Fatalf("Instantiating Finalizer: %v", err)
+	}
+
+	// Configure Syncer and Remediator.
+
+	// Get a separate config for the remediator to talk to the apiserver since
+	// we want a longer REST config timeout for the remediator to avoid restarting
+	// idle watches too frequently.
+	cfgForWatch, err := restconfig.NewRestConfig(watch.RESTConfigTimeout)
```
karlkfi, line 227: Maybe not relevant to this PR, but I think we might need to send multiple clients to the remediator at some point to separate the watch timeout and list timeout from all the other operations. Or maybe we needs to be more explicit about those timeouts when we make those requests. Otherwise they all get the 1hr timeout.

cmd/reconciler/main.go
karlkfi: Do you mean de-queuing and processing requests?

The current behavior is that watch events are still added to the queue, they just don't get handled until the applier exits successfully.

nan-yu: The Remediator controller won't enqueue and dequeue+process requests if `canRemedite` is false.

cmd/reconciler/main.go
karlkfi: Might want to avoid using package variables in this helper and instead pass them into the function pre-parsed and pre-validated.

cmd/reconciler/main.go
```diff
@@ -161,68 +399,97 @@ func main() {
...
+		HydratedLink: *hydratedLinkDir,
+		SyncDir:      relSyncDir,
+		SourceType:   v1beta1.SourceType(*sourceType),
+		SourceRepo:   *sourceRepo,
+		SourceBranch: *sourceBranch,
+		SourceRev:    *sourceRev,
+	}
+	ds := declared.Scope(*scope)
+	if ds == declared.RootReconciler {
+		parser, err = parse.NewRootRunner(
```
karlkfi, line 450: these constructors are crazy long. perhaps we should replace them with a builder so we can just populate one builder and use it to build both runners.

e2e/testcases/custom_resources_test.go
```diff
@@ -218,21 +217,46 @@ func TestCRDDeleteBeforeRemoveCustomResourceV1(t *testing.T) {
...
-	// and record it as a conflict.
-	// TODO: distinguish between management conflict (spec/generation drift) and concurrent status update conflict (resource version change)
-	err = nomostest.ValidateMetrics(nt,
-		nomostest.ReconcilerErrorMetrics(nt, rootSyncLabels, firstCommitHash, metrics.ErrorSummary{
-			Conflicts: 1,
-		}))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+	// No conflicts anymore. There was one before the reconciler refactor, because
```
karlkfi, line 220: perhaps pull this long comment out to a github issue as bug report with details, mark it solved, and link to it from the code comment?

karlkfi: or a buganizer link.

pkg/core/id.go
```diff
@@ -42,6 +42,27 @@ func (i ID) String() string {
 	return fmt.Sprintf("%s, %s/%s", i.GroupKind.String(), i.Namespace, i.Name)
 }
 
+// FromID returns a GKNN representing the provided ID string.
+func FromID(id string) (ID, error) {
```
karlkfi, line 46: core.FromID is a confusing name, since it doesn't return a "core". How about ParseID?

The func comment should also explain the format.

karlkfi: The comment also says it returns a GKNN, but it doesn't... it returns an ID.

pkg/parse/opts.go
```diff
@@ -95,6 +95,8 @@ type Parser interface {
 	K8sClient() client.Client
 	// setRequiresRendering sets the requires-rendering annotation on the RSync
 	setRequiresRendering(ctx context.Context, renderingRequired bool) error
+	HasManagementConflict() bool
```
karlkfi, line 98: public interface methods need comments

pkg/parse/root_test.go
```diff
@@ -93,6 +97,29 @@ func (r *noOpRemediator) Errors() status.MultiError {
 	return nil
 }
 
+func (r *noOpRemediator) CanRemediate() bool {
```
karlkfi, line 100: public methods need comments

pkg/parse/state.go
```diff
@@ -58,9 +58,10 @@ func (gs syncStatus) equal(other syncStatus) bool {
 	return gs.syncing == other.syncing && gs.commit == other.commit && status.DeepEqual(gs.errs, other.errs)
 }
 
-type reconcilerState struct {
-	// lastApplied keeps the state for the last successful-applied syncDir.
-	lastApplied string
+// ReconcilerState stores the shared state across multiple Syncer reconciliations.
+type ReconcilerState struct {
+	// lastAppliedDir keeps the state for the last successful-applied syncDir.
+	lastAppliedDir string
```
karlkfi, line 64: was confused when i first read this name. maybe lastAppliedSyncDir?

pkg/parse/state.go
```diff
@@ -99,70 +109,70 @@ const retryLimit = 12
 //	4m24.712222056s, 9m18.97652295s, 17m15.344384599s, 35m15.603237976s.
 func defaultBackoff() wait.Backoff {
```
karlkfi, line 110: Might be cleaner to move this default code into ResetRetryBackoffAndTimer and then always call ResetRetryBackoffAndTimer when you want to reset the retry backoff.

Then while you're at it, also move the NewTimer initialization into ResetRetryBackoffAndTimer so it doesn't panic when nil.

pkg/parse/state.go
```diff
@@ -192,3 +202,51 @@ func (s *reconcilerState) needToSetSyncStatus(newStatus syncStatus) bool {
...
+	return s.retryBackoff.Steps
+}
+
+// NeedRetry returns whether a retry is needed.
+func (s *ReconcilerState) NeedRetry() bool {
+	return s.cache.needToRetry
+}
+
+// SourceCommitInStatus returns the source commit in the status field.
+func (s *ReconcilerState) SourceCommitInStatus() string {
```
karlkfi, line 240: suggest: 
- StatusSourceCommit
- StatusRenderingCommit
- StatusSyncCommit

Then it's closer to how it looks like the RSync api

pkg/parse/updater.go
```diff
@@ -157,13 +157,19 @@ func (u *updater) Update(ctx context.Context, cache *cacheForCommit) status.Mult
...
-func (u *updater) update(ctx context.Context, cache *cacheForCommit) status.MultiError {
+func (u *updater) update(ctx context.Context, cache *CacheForCommit) status.MultiError {
 	// Stop remediator workers.
 	// This prevents objects been updated in the wrong order (dependencies).
 	// Continue watching previously declared objects and updating the queue.
 	// Queued objects will be remediated when the workers are started again.
 	u.remediator.Pause()
 
+	// block and wait until the current remediation is done.
+	u.remediator.WaitRemediating()
```
karlkfi, line 168: Could Pause do the waiting or does this need to be a separate call?

karlkfi: Looks like this set of 3 calls is made in the same order in two places. Could it be reduced to one method?

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// Remediator is a controller that watches for timer-based events.
+// When an object is added to the remediate resource list, it enqueues the GKNN
+// of the object as a remediate request. The reconciler dequeues the request
+// and remediate the object to match the declared resource.
+type Remediator struct {
+	// ControllerCtx is set in reconciler/main.go and is canceled by the Finalizer.
+	ControllerCtx context.Context
```
karlkfi, line 50: If Remediator IS a controller, then the controller prefix here is redundant. You could just call this field Context.

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+	// DoneCh indicates the Remediator stops processing new requests.
```
karlkfi, line 51: stops processing -> has stopped and is no longer processing

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+	*cache.RemediateResources
+	// Remediator is the actor to check whether it is ready to start reconciliation.
+	*remediator.Remediator
+}
+
+// Reconcile processes the remediate request to correct drift.
+func (r *Remediator) Reconcile(_ context.Context, req ctrl.Request) (ctrl.Result, error) {
+	if errors.Is(r.ControllerCtx.Err(), context.Canceled) {
+		// Drain the request queue
+		klog.V(3).Infof("Stop remediating new objects as the context is canceled..")
```
karlkfi, line 74: Suggest: 
`klog.V(3).Infof("Remediator already stopped - Ignoring reconcile request: %s", req)`

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	klog.V(1).Infof("New remediator reconciliation for %q", req.Name)
+
+	id, err := core.FromID(req.Name)
```
karlkfi, line 80: should this not include the name and namespace?

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		klog.Errorf("invalid request: %v", err)
```
karlkfi, line 82: suggest: failed to parse reconcile request

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+		klog.V(1).Infof("Skip remediating object %s because the Remediator is paused", id)
+		r.DoneObj(id)
+		return ctrl.Result{}, nil
+	}
+	// Mark the Remediator as done before it exits, so the Syncer can start the apply work.
+	defer r.DoneRemediating()
+
+	obj, found := r.GetObjs()[id]
+	if !found {
+		klog.V(4).Infof("Skip remediating object %q because it is no longer on the remediate list", id)
```
karlkfi, line 98: What does "no longer on the remediate list" mean in this case? The watch saw an event but we dont care about that object for some reason?

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+				// block and wait until the current remediation is done.
+				r.Remediator.WaitRemediating()
+				// clear remediate cache to abandon obsolete remediate requests.
+				r.Remediator.ClearCache()
+				close(r.DoneCh) // inform finalizer the remediator is done
+				return
+
+			case <-remediateTimer.C:
+				if r.CanRemediate() {
+					// Skip enqueuing remediator requests if the Applier is running.
```
karlkfi, line 195: I'm confused by this. My understanding when I added the pausing the first time was that the watch events were still handled and it still added objects to the queue to be processed, it just paused the dequeuing and handling of those object events.

With this new code, it seems like we might miss events that happen while the applier is running, which wasn't my original intent. We still need to handle all those events from the applier and make sure none of them added drift. Otherwise someone can make changes while the applier is running and the remediator won't handle or revert them until the NEXT change happens on that object.

cmd/reconciler/main.go
sdowell: I made this comment a few patchsets ago, before synchronization was implemented for Syncer/Remediator. In the earlier patchset, it was possible for both the Remediator and Syncer to be running simultaneously.

The current patchset has synchronization which should prevent concurrent execution. The `CanRemediate` call doesn't prevent concurrent enqueueing given that's it's an unsafe boolean, but it is an optimization that should prevent some enqueueing of watch events. The actual synchronization is in the `StartRemediating` call in the reconcile loop

pkg/reconciler/controllers/remediator.go
sdowell: In that case I think we can remove this, as it just seems like an optimization to me. The `StartRemediating` function handles formal synchronization between the syncer/remediator

pkg/reconciler/controllers/constants.go
```diff
@@ -12,19 +12,14 @@
...
-
-	"k8s.io/klog/v2"
+// reconciler controller names
+const (
+	// SyncerController is the name of the syncer controller running in the reconciler container.
+	// It is named syncer instead of parser because it includes parsing, validating and applying.
+	// Also, there is already a Parser struct declared in pkg/parse/opts.go.
+	SyncerController = "Syncer"
+	// RemediatorController is the name of the remediator controller running in the reconciler container.
+	RemediatorController = "Remediator"
```
janetkuo, line 24: Suggest putting `const` where it's used instead of creating a const file for all constants.

pkg/reconciler/controllers/remediator.go
```diff
@@ -0,0 +1,348 @@
...
+		// is for an Object that was deleted.
+		toRemediate = nil
+		// If the object was deleted, it has no ResourceVersion, use the constant.
+		currentRV = cache.DeletedResourceVersion
+	} else {
+		toRemediate = obj
+	}
+
+	// If the current ResourceVersion was last remediated, skip remediation because
+	// it should match the declared resource.
```
janetkuo, line 119: Should we also skip remediation when the only change on the obj is its status field? ResourceVersion changes on status updates, too.

janetkuo: This might not be easily done given that comparing `generation` won't capture metadata updates, so can be done later as an enhancement.

cmd/reconciler/main.go
karlkfi: Problem is, it needs to keep enqueuing requests, adding them to a queue and then handling them all when the applier succeeds. Otherwise it might miss drift that happened while the applier was running.

pkg/reconciler/controllers/remediator_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2023 Google LLC
```
karlkfi, line 1: google policy is to leave the copyright at the original year the file was added.

pkg/reconciler/controllers/syncer.go
```diff
@@ -0,0 +1,343 @@
...
+		doneCh:             doneCh,
+		canSync:            true, // The syncer should be able to sync at any time unless the controller context is canceled.
+	}
+}
+
+// Reconcile processes the Syncer request to apply the source configs to the cluster.
+func (s *Syncer) Reconcile(_ context.Context, req ctrl.Request) (ctrl.Result, error) {
+	if errors.Is(s.controllerCtx.Err(), context.Canceled) {
+		// Drain the request queue
+		klog.V(3).Infof("Stop syncing new events as the context is canceled..")
```
karlkfi, line 112: Suggest:
`klog.V(3).Info("Skip syncing because the Syncer is stopped")`

pkg/reconciler/controllers/syncer.go
```diff
@@ -0,0 +1,343 @@
...
+	"k8s.io/klog/v2"
+	"kpt.dev/configsync/pkg/declared"
+	"kpt.dev/configsync/pkg/parse"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// Syncer is a controller that watches for timer-based Syncer requests, and also
```
karlkfi, line 33: Confused by this description. 
How did you get teconciler-manager to trigger reconcile based on a timer?
And why would a namespace watch event trigger a sync?

pkg/reconciler/controllers/syncer.go
karlkfi: Not sure how to modify that to include the request, but it seems like it should. I'm not really clear what requests trigger reconciles.

pkg/reconciler/controllers/syncer.go
```diff
@@ -0,0 +1,343 @@
...
+				s.WaitSyncing()
+				close(s.doneCh) // inform finalizer the syncer is done
+				return
+
+			// Re-apply even if no changes have been detected.
+			// This case should be checked first since it resets the cache.
+			// If the reconciler is in the process of reconciling a given commit, the resync won't
+			// happen until the ongoing reconciliation is done.
+			case <-s.resyncTimer.C:
+				timerEvents <- timerEvent(parse.Resync)
```
karlkfi, line 238: The old code reset the other timers after each event, but I don't see that code here. How is is delaying other events when a timer triggers?

pkg/reconciler/controllers/syncer.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// StartSyncing marks the Syncer is working.
+func (s *Syncer) StartSyncing() bool {
+	s.lifecycleMux.Lock()
+	defer s.lifecycleMux.Unlock()
+
+	if s.canSync {
+		klog.V(1).Infof("Start syncing...")
+		s.wg.Add(1)
```
karlkfi, line 318: Are wait groups re-usable? can you re-add and re-wait after previously waiting until done?

If so, maybe using a waitgroup makes sense. Otherwise i would have suggested a channel, since there's only ever one task being waited on.

pkg/reconciler/controllers/syncer.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// DoneSyncing marks the Syncer as done when it finishes syncing.
+func (s *Syncer) DoneSyncing() {
+	klog.V(1).Info("Syncer completes...")
+	s.wg.Done()
+}
+
+// PauseSyncing stops the Syncer to process new requests.
+func (s *Syncer) PauseSyncing() {
```
karlkfi, line 336: It doesn't looks like we ever really need to resume syncing. I think this could be replaced with a Context and a DoneCh, like you have in the remediator.

pkg/reconciler/finalizer/reposync_finalizer.go
```diff
@@ -29,18 +29,18 @@ import (
...
-	// StopControllers will be called by the Finalize() method to stop the Parser and Remediator.
+	// StopControllers will be called by the Finalize() method to stop the Syncer and Remediator.
 	StopControllers context.CancelFunc
 
-	// ControllersStopped will be closed by the caller when the parser and
+	// ControllersStopped will be closed by the caller when the syncer and
 	// remediator have fully stopped. This unblocks Finalize() to destroy
 	// managed resource objects.
-	ControllersStopped <-chan struct{}
+	ControllersStopped []<-chan struct{}
```
karlkfi, line 43: Dont love this interface. Slices are mutable. I wonder if it would make sense to use a waitgroup or a waitgroup blocking a goroutine that closes a single channel.

nan-yu:
We decided to punt on the refactor after 1.17.0.

PR #977, opened 2023-10-28
feat: add roleRefs to RSync override API

sdowell:
The roleRefs field is intended to allow more flexible and simple management of bindings for both RootSync and RepoSync objects. The field allows a list of references to Role and ClusterRole objects. For RepoSyncs, this will always create a RoleBinding in the same namespace as the RepoSync. For RootSyncs, the user can either choose to create a RoleBinding or ClusterRoleBinding via the namespace field.

This is an iteration upon the previous design which added a single ClusterRoleBinding configuration for RootSyncs. This design allows more flexible configuration for RootSyncs, as well as a consistent API for RepoSyncs which simplifies the management of RepoSync RoleBindings.

Issue: https://github.com/GoogleContainerTools/kpt-config-sync/issues/935
Iterates on this PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/938


sdowell:
/assign @karlkfi 
/assign @mortent 

FYI @tomasaschan

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -1,15 +1,21 @@
...
 When creating a reconciler deployment for a `RootSync`, Config Sync currently [creates a
 `ClusterRoleBinding`] for the generated `ServiceAccount` granting it `cluster-admin`.
 
-This proposal introduces a mechanism to customize this behavior.
+On the other hand, when creating a reconciler deployment for a `RepoSync`, Config Sync
+currently creates `RoleBinding` for the generated `ServiceAccount` granting
+it `configsync.gke.io:ns-reconciler`. This default `ClusterRole` grants basic permissions required for the `RepoSync`
+reconciler to function, but it's up to users to manage additional `RoleBinding`s.
+
+This proposal introduces a manage the lifecycle of RBAC bindings for both `RootSync`
```
tomasaschan, line 17:
```suggestion
This proposal introduces a way to manage the lifecycle of RBAC bindings for both `RootSync`
```

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+This essentially requires for the `reconciler-manager` to be able to track an
+inventory of bindings that were previously created for a given `RootSync` or `RepoSync`.
+This can be accomplished by applying a label whenever a new binding is created,
+and then querying using a label selector on subsequent reconciliation loops.
+
+The following label will be applied to binding objects:
+```yaml
+metadata:
+  labels:
+    configsync.gke.io-reconciler: <RECONCILER_NAME>
```
tomasaschan, line 117: Did you mean for this to be slash-separated?
```suggestion
    configsync.gke.io/reconciler: <RECONCILER_NAME>
```

pkg/reconcilermanager/controllers/build_names.go
```diff
@@ -21,19 +21,26 @@ import (
 	"kpt.dev/configsync/pkg/core"
 )
 
+const (
+	// ReconcilerBindingLabelKey is the name of the label applied to RoleBinding
+	// and ClusterRoleBinding objects created for a reconciler. The value of
+	// the label is set to the name of the reconciler.
+	ReconcilerBindingLabelKey = configsync.GroupName + "-reconciler"
```
tomasaschan, line 28:
```suggestion
	ReconcilerBindingLabelKey = configsync.GroupName + "/reconciler"
```
?

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,52 +794,112 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
...
-	if op != controllerutil.OperationResultNone {
-		r.logger(ctx).Info("Managed object upsert successful",
-			logFieldObjectRef, crbRef.String(),
-			logFieldObjectKind, "ClusterRoleBinding",
-			logFieldOperation, op)
+	declaredRoleRefs := roleRefs
+	// If no RoleRefs are declared, default to cluster-admin for backwards compatibility
+	if len(declaredRoleRefs) == 0 {
+		declaredRoleRefs = append(declaredRoleRefs, v1beta1.RootSyncRoleRef{
+			RoleRefBase: v1beta1.RoleRefBase{
```
sdowell, line 854: One thing to note is that while this backwards-compatible default does keep the `cluster-admin` default ClusterRoleBinding, the name/structure of the ClusterRoleBinding does change. Instead of a many RootSync -> one ClusterRoleBinding named `configsync.gke.io:root-reconciler`, it's now a many->many ClusterRoleBinding where each RootSync will get it's own ClusterRoleBinding with a generated name.

This assumes that the name/structure of the ClusterRoleBinding is an implementation detail and therefore changes are not breaking. We could alternatively maintain a separate code path for the legacy behavior of a singleton `configsync.gke.io:root-reconciler` ClusterRoleBinding with many subjects, which IMO has the downside of more code variance/complexity.

tomasaschan (commented):
Very nice, thanks for putting this together so quickly!

A few small comments on things I assume are either typos or just stylistic disagreements :)

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: Our general policy is simply not to require customers to need to do any work when a service is auto-upgraded, such as Config Sync on Config Controller. But we don't have any way to determine if any customers depend on the exact result of the generated resources. So historically we have not considered the generated resources to be a customer facing interface.

We don't want to leak any resources, tho. So if we change the naming scheme, we need to clean up the legacy resources. This can be done in the config-management-operator on upgrade with a version check or in the reconciler-manager teardown code.

karlkfi: You might also consider de-duping similar ClusterRoleBindings and RoleBindings, to avoid sprawl.

pkg/reconcilermanager/controllers/build_names.go
sdowell: Accepted change

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: Accepted change

pkg/reconcilermanager/controllers/rootsync_controller.go
sdowell: The legacy resources created by older versions reconciler-manager are already being cleaned up by the current reconciler-manager, see below.

---

De-duping may work for the ClusterRoleBindings/RoleBindings with known names, such as the binding to the "base" ClusterRole. However it's less clear to me how we can handle de-duping for user-provided roleRefs - given we need to maintain an inventory of previously applied bindings.

In the current approach, each binding is attached to a specific reconciler and can be listed using a label selector. If we consolidate bindings and attach multiple reconcilers as subjects, it becomes less clear to me how we can track which bindings were attached to a given reconciler on previous iterations.

Solutions that come to mind but don't scale:
- list all config sync bindings and filter the subjects in code
- attach a label for each reconciler subject to the binding

karlkfi: I think the only de-duping we might want would be a single admin ClusterRoleBinding, and just add a new service account to it for every RootSync using the default config. Then if a custom config is specified, use reconciler-specific ClusterRoleBindings/RoleBindings.

tomasaschan: For reference, the migration cleanup implementation in #938 is [here](https://github.com/GoogleContainerTools/kpt-config-sync/blob/3a526f88a0a978a84c6b559ba506d47e582e1c35/pkg/reconcilermanager/controllers/rootsync_controller.go#L783-L788).

sdowell: I kept that snippet but moved it to the end of the function, the idea being that we should add new bindings before removing old ones

sdowell: @karlkfi updated to use a consolidated binding for both the base bindings and for the cluster-admin binding

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+    - kind: ClusterRole # Creates a RoleBinding in my-tenant-namespace bound to ClusterRole tenant-cluster-role
+      name: tenant-cluster-role
+      namespace: my-tenant-namespace
+    - kind: Role # Creates a RoleBinding in my-tenant-namespace bound to Role my-tenant-role
+      name: my-tenant-role
+      namespace: my-tenant-namespace
 ```
 
-This role is defaulted to `cluster-admin` in order to stay backwards compatible.
+For `RootSync` objects, the default behavior will create a ClusterRoleBinding
```
mortent, line 71: Exactly how will the defaulting work here? Is the assumption that is no RoleRefs are specified for a RootSync, the binding to cluster-admin will automatically be set up, but if the RoleRefs list is not empty, only the specified bindings will be created?

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+
+Given this API provides configuration for a list of RoleRefs, it requires some form
+of lifecycle management to clean up stale bindings. For example if a user removes
+one roleRef from the list of roleRefs, they would reasonably expect that the
+binding will be garbage collected.
+
+This essentially requires for the `reconciler-manager` to be able to track an
+inventory of bindings that were previously created for a given `RootSync` or `RepoSync`.
+This can be accomplished by applying a label whenever a new binding is created,
+and then querying using a label selector on subsequent reconciliation loops.
```
mortent, line 111: Can we end up in a situation where the `reconciler-manager` will not be able to delete the bindings (maybe because it no longer has permissions)? If so, how is the issue surfaced to users?

manifests/rootsync-crd.yaml
```diff
@@ -444,6 +444,33 @@ spec:
                           x-kubernetes-int-or-string: true
                       type: object
                     type: array
+                  roleRefs:
+                    description: roleRefs is a list of Roles or ClusterRoles to create
+                      bindings. If unset, a binding to cluster-admin will be created.
```
mortent, line 449: Presumably then the cluster-admin binding will also be removed if the user adds a roleRefs through an update?

sdowell: Yes that is correct, if a user adds roleRefs then the `cluster-admin` default will be removed. Conversely if a user removes all roleRefs then the custom bindings will be deleted and the `cluster-admin` default will be re-created.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: Yeah that sums it up pretty accurately. Keep in mind this is necessary for backwards compatibility - I may not have chosen this behavior if designing from scratch.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: I don't think that should ever happen, unless a user has a standalone Config Sync installation (no managing operator) and manually modifies the reconciler-manager ClusterRole/ClusterRoleBinding.

If such a error did happen, I expect the behavior would be the same regardless of which managed resource it's erroring on (e.g. ServiceAccount, Deployment, RoleBinding, etc). I haven't tested this to verify, but my expectation is that the reconciler-manager would stall and surface the error on the RSync.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,52 +794,147 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
...
+	// Add the base ClusterRole for basic root reconciler functionality
+	if err := r.upsertSharedClusterRoleBinding(ctx, RootSyncBaseClusterRoleBindingName, RootSyncBaseClusterRoleName, reconcilerRef); err != nil {
+		return err
+	}
+	declaredRoleRefs := roleRefs
+	// Create RoleBindings which are declared but do not exist on the cluster
+	for _, roleRef := range declaredRoleRefs {
+		if _, ok := currentRefMap[roleRef]; !ok {
+			// we need to call a separate create method here for generateName
+			if _, err := r.createRBACBinding(ctx, reconcilerRef, roleRef.RoleRefBase, roleRef.Namespace); err != nil {
```
nan-yu, line 881: Do we need to verify whether the referenced Role/ClusterRole exists, and fail earlier if it doesn't exist?

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,52 +794,147 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
...
+	if err := r.deleteSharedClusterRoleBinding(ctx, RootSyncBaseClusterRoleBindingName, reconcilerRef); err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrap(err, "deleting base binding")
+	}
+	return nil
+}
 
-	_, err := r.upsertClusterRoleBinding(ctx, reconcilerRef, roleConfig.Name)
-	return errors.Wrap(err, "upserting cluster role binding")
+func (r *RootSyncReconciler) configureClusterRoleBinding(ctx context.Context, reconcilerRef types.NamespacedName, roleRefs []v1beta1.RootSyncRoleRef) error {
+	currentRefMap, err := r.listCurrentRoleRefs(ctx, reconcilerRef)
```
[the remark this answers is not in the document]
sdowell: My initial thoughts is that it's fair game to assume that a resource key that is prefixed with our `GroupName` (`configsync.gke.io`) is a resource that we can manage. The same argument could be made for other resources as well, say if a user tried to manage the `configsync.gke.io/ns-reconciler` base RoleBinding in the source repo.

If this is a concern, we could potentially add another label to make it more explicit - say something like `configsync.gke.io/managed-by: reconciler-manager`.


As for tracking the inventory of bindings in RSync annotations - I'm not sure how this would scale to larger numbers of bindings. We would need to track `Kind`, `Name`, and `Namespace` for an arbitrary number of bindings. This is unlikely to fit in a single annotation, and it's not clear to me how we would shard this across multiple annotations.

I could potentially see adding new Status fields on RootSync/RepoSync, however I'm not sure how compelling the case for this is over using a label selector.

pkg/reconcilermanager/controllers/rootsync_controller.go
sdowell: I'm not sure it would add much value to verify that earlier on, as such condition would cause creation of the binding to error and surface on the RSync. Even if we did check for the existence of the Role/ClusterRole early on, it's always possible for race condition where another client deletes the Role/ClusterRole after our existence check.

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1171,3 +1171,51 @@ func RootSyncSpecOverrideEquals(expected *v1beta1.RootSyncOverrideSpec) Predicat
 		return nil
 	}
 }
+
+func subjectsEqual(want []string, got []rbacv1.Subject) error {
```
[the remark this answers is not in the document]
sdowell: Not exactly, since this function is only comparing the name field of each `Subject`. We would still need to convert the data structures to look the same in order to use `cmp.Equal`

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -1,15 +1,21 @@
...
+* Approver: @mortent, @karlkfi
 * Status: approved
 
 ## Summary
 
 When creating a reconciler deployment for a `RootSync`, Config Sync currently [creates a
 `ClusterRoleBinding`] for the generated `ServiceAccount` granting it `cluster-admin`.
 
-This proposal introduces a mechanism to customize this behavior.
+On the other hand, when creating a reconciler deployment for a `RepoSync`, Config Sync
```
karlkfi, line 12: Is this right? I thought this was just how the e2e tests work. In normal use, the user has to create a role & rolebinding, right?

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -1,15 +1,21 @@
...
+currently creates `RoleBinding` for the generated `ServiceAccount` granting
+it `configsync.gke.io:ns-reconciler`. This default `ClusterRole` grants basic permissions required for the `RepoSync`
+reconciler to function, but it's up to users to manage additional `RoleBinding`s.
+
+This proposal introduces a way to manage the lifecycle of RBAC bindings for both
```
karlkfi, line 17: I would move this to the top. It's the actual summary. The other stuff above is more like historical context, not summary.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
 If one wants to allow deploying resources from a source where one doesn't have full
```
karlkfi, line 24: I know you didn't write this, but this sentence is pretty hard to read. I would get rid of the "one" and rephrase with "you" and maybe reorganize this section as a list of use cases. This use case is for a platform admin wanting to allow tenants to use RootSyncs with reduced permissions. 

The use case for RepoSync is to follow least privilege to only give the reconciler the permissions it needs, to limit the blast area if the source of truth is compromised or a developer mistake is made. For example, to disallow making service accounts and RBAC.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+kind: RootSync
+metadata:
+  name: my-root-sync
+  namespace: config-management-system
 spec:
   overrides:
-    clusterRole:
-      name: my-custom-role
+    roleRefs:
+    - kind: ClusterRole # Creates a ClusterRoleBinding bound to ClusterRole my-cluster-role
```
karlkfi, line 61: nit: the CRB isn't the object being bound, it's the subject doing the binding. The CRB binds the CR to the KSA.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
karlkfi: Suggest: 
> For `RootSync` objects, a default `cluster-admin` ClusterRoleBinding will be applied when `spec.override.roleRefs` is empty or nil, for reverse compatibility.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+    - kind: Role # Creates a RoleBinding in my-tenant-namespace bound to Role my-tenant-role
+      name: my-tenant-role
+      namespace: my-tenant-namespace
 ```
 
-This role is defaulted to `cluster-admin` in order to stay backwards compatible.
+For `RootSync` objects, the default behavior will create a ClusterRoleBinding
+to `cluster-admin` in order to stay backwards compatible.
+
+For convenience, `RootSync` reconcilers will also be bound to a base `ClusterRole`
```
karlkfi, line 74: "also"? Isn't this the same as the above? It's the `cluster-admin` ClusterRole, right?

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+which gives the reconciler the permissions for basic functionality. This will
+be comparable to the pre-existing base ClusterRole for `RepoSync` reconcilers,
```
karlkfi, line 76: I'm confused. Are these RootSync and RepoSync swapped?

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+which includes permissions such as status writing on `RepoSync` objects. Leaving
+this permission to the user to manage would create unneeded toil.
+
+### `RepoSync`s
 
-For the case where a single `ClusterRole` is not expressive enough to configure the
-permissions a user want, you can instead set `disabled: true` on the override object:
+For `RepoSync` objects, the `namespace` field will not be present on `roleRefs`.
```
karlkfi, line 82: Suggest:

> For `RepoSync` objects, `roleRefs` entries will not need or allow specifying a `namespace`.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
karlkfi: In general, when the reconciler-manager or reconciler cannot delete things that it manages, it errors and that error is surfaced in the RSync status.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -18,65 +24,130 @@ This proposal introduces a mechanism to customize this behavior.
...
+This essentially requires for the `reconciler-manager` to be able to track an
+inventory of bindings that were previously created for a given `RootSync` or `RepoSync`.
+This can be accomplished by applying a label whenever a new binding is created,
+and then querying using a label selector on subsequent reconciliation loops.
+
+The following label will be applied to binding objects:
+```yaml
+metadata:
+  labels:
+    configsync.gke.io/reconciler: <RECONCILER_NAME>
```
karlkfi, line 117: I get that this works as a single unique identifier, but these objects are managed by the reconciler-manager, not the reconciler. They're dependencies for the reconciler. I wonder if it might make more sense to use the sync-name, sync-namespace, and sync-kind labels instead. We already have precedent for setting those in `upsertManagedObjects`.

karlkfi: In fact, a 3rd party made a draft PR to add these labels to the RoleBinding already https://github.com/GoogleContainerTools/kpt-config-sync/pull/981

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: What you're thinking of is something the e2e tests used to do - they always created a RoleBinding which granted cluster-admin so that RepoSyncs could manage any resource in the namespace. This has since been changed, and each e2e test needs to actually declare which permissions are granted to the RepoSync reconciler. This enables more realistic testing of RepoSync reconcilers and their typical RoleBindings.

The `configsync.gke.io:ns-reconciler` RoleBinding here is separate, and this RoleBinding is always created. The RoleBinding binds to the ClusterRole [here](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/ns-reconciler-cluster-role.yaml), which just grants basic permissions such as RepoSync status updates etc. My understanding is that this is for convenience, as this is the bare minimum that is necessary for a RepoSync reconciler to run and would just add burden for the user to create.

I realized that the similar concept applies to RootSync if we strip the cluster-admin binding, so I added a similar base ClusterRole for RootSync that will be automatically created.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: No, this is different. This is a binding to the `root-reconciler-base-cluster-role` which is added by this change, and is modeled after the existing ns-reconciler-cluster-role`. This contains just basic permission for RootSync reconciler functionality, such as RootSync status updates.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: no, see my explanation above

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: That seems reasonable.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: done

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: Done

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: rephrased

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: Done

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: Done

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: I'm working on a change to the labels for the generated custom bindings based on roleRefs. Note that the kind/name/namespace labels don't really work for the shared bindings, because a single binding is shared by many RSyncs. I think those will continue to not have any of the labels.

sdowell: Done

docs/design-docs/02-custom-root-reconciler-clusterrole.md
karlkfi: Ahh. ok. That wasn't clear from this doc.

I think perhaps it could be clarified with a before and after table or list or something.

docs/design-docs/02-custom-root-reconciler-clusterrole.md
karlkfi: Do the shared bindings have fixed names so we can always delete them when all the RSyncs that use them are deleted?

docs/design-docs/02-custom-reconciler-bindings.md
```diff
@@ -0,0 +1,247 @@
...
+
+When creating a reconciler deployment for a `RootSync`, Config Sync currently [creates a
+`ClusterRoleBinding`] for the generated `ServiceAccount` granting it `cluster-admin`.
+
+On the other hand, when creating a reconciler deployment for a `RepoSync`, Config Sync
+currently creates `RoleBinding` for the generated `ServiceAccount` granting
+it `configsync.gke.io:ns-reconciler`. This default `ClusterRole` grants basic permissions required for the `RepoSync`
+reconciler to function, but it's up to users to manage additional `RoleBinding`s.
+
+[creates a `ClusterRoleBinding`]: https://github.com/GoogleContainerTools/kpt-config-sync/blob/199db6dbfa0b9eb1824925c8c687574de095a294/pkg/reconcilermanager/controllers/rootsync_controller.go#L789
```
karlkfi, line 20: suggest linking to a specific release tag, like https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.16.0/pkg/reconcilermanager/controllers/rootsync_controller.go#L808

docs/design-docs/02-custom-root-reconciler-clusterrole.md
sdowell: Yes they do, and this change currently implements the behavior your describe. The logic for the shared bindings should be relatively unchanged, it's just the new custom bindings that will be tracked using labels

docs/design-docs/02-custom-reconciler-bindings.md
```diff
@@ -0,0 +1,247 @@
...
+* Author(s): @tomasaschan, @sdowell
+* Approver: @mortent, @karlkfi
+* Status: approved
+
+## Summary
+
+This proposal introduces a way to manage the lifecycle of RBAC bindings for both
+`RootSync` and `RepoSync` objects.
+
+When creating a reconciler deployment for a `RootSync`, Config Sync currently [creates a
```
karlkfi, line 12: I think these 2 paragraphs might make more sense as a `Historical Behavior` section, followed by a `Proposed Behavior` section which briefly explains the new behavior.

e2e/nomostest/nt.go
```diff
@@ -966,3 +968,39 @@ func cloneCloudSourceRepo(nt *NT, repo string) (string, error) {
 	}
 	return cloneDir, nil
 }
+
+// ListReconcilerRoleBindings is a convenience method for listing all RoleBindings
+// associated with a reconciler.
+func (nt *NT) ListReconcilerRoleBindings(syncKind string, rsRef types.NamespacedName) ([]rbacv1.RoleBinding, error) {
+	opts := &client.ListOptions{}
+	opts.LabelSelector = client.MatchingLabelsSelector{
```
karlkfi, line 976: slightly redundant to do this twice. You could have a function that returns the label selector. in fact, I'm surprised there isn't one under `pkg/reconcilermanager/controllers` already. Don't we need it there too for cleanup?

sdowell: There is one but it's a method of `reconcilerBase`. I suppose I can refactor that to a function

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1171,3 +1171,51 @@ func RootSyncSpecOverrideEquals(expected *v1beta1.RootSyncOverrideSpec) Predicat
...
+	if len(got) != len(want) {
+		return errors.Wrapf(ErrFailedPredicate, "want %v subjects; got %v", want, got)
```
karlkfi, line 1177: This ErrFailedPredicate pattern seems weird. I would expect the caller of the predicates to add it as a wrapper, rather that every individual predicate to need to return it, especially as the root cause...

e2e/testcases/override_role_refs_test.go
```diff
@@ -0,0 +1,346 @@
+// Copyright 2022 Google LLC
```
karlkfi, line 1: new files should use the current year copyright

e2e/testcases/override_role_refs_test.go
```diff
@@ -0,0 +1,346 @@
...
+		Namespace: rootSyncA.Namespace,
+	}
+	if err := nt.Validate(controllers.RootSyncLegacyClusterRoleBindingName, "", &rbacv1.ClusterRoleBinding{},
+		testpredicates.ClusterRoleBindingHasSubjects(nomostest.DefaultRootReconcilerName, syncAReconcilerName)); err != nil {
+		nt.T.Fatal(err)
+	}
+	// Set some custom roleRef overrides
+	rootSyncA.Spec.SafeOverride().RoleRefs = []v1beta1.RootSyncRoleRef{
+		{
+			RoleRefBase: v1beta1.RoleRefBase{
```
karlkfi, line 53: does this nesting of the base type help make something easier? 
seems like there could just be a namespaced ref and a non-namespaced ref.

manifests/reposync-crd.yaml
```diff
@@ -421,6 +421,26 @@ spec:
                           x-kubernetes-int-or-string: true
                       type: object
                     type: array
+                  roleRefs:
```
karlkfi, line 424: i see the `roleRefs` addition here, but shouldn't the `roleRef` have been removed? I can't find it...

karlkfi: oh, nvm. this is reposync

sdowell: The `ClusterRole` api from the recent change was only added for RootSync.

e2e/testcases/override_role_refs_test.go
sdowell: The main goal here is that code paths can be shared for reposync/rootsync controllers, given they share a common struct. If the structs were not nested, we would either need separate implementations or some glue logic to translate between struct types

pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -233,8 +232,40 @@ func (r *reconcilerBase) deleteDeployment(ctx context.Context, reconcilerRef typ
...
+func (r *reconcilerBase) deleteRoleBinding(ctx context.Context, name, namespace string) error {
+	rb := &rbacv1.RoleBinding{}
+	rb.Name = name
+	rb.Namespace = namespace
+
+	return r.cleanup(ctx, rb)
+}
 
-	return r.cleanup(ctx, crb)
+func (r *RootSyncReconciler) deleteSharedClusterRoleBinding(ctx context.Context, name string, reconcilerRef types.NamespacedName) error {
```
karlkfi, line 243: where is this used?

karlkfi: nvm. collapsed by default

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,52 +793,149 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
...
+			},
+		}
+		currentRoleMap[roleRef] = &crbList.Items[idx]
 	}
+	return currentRoleMap, nil
+}
 
-	if roleConfig.Name == "" {
-		roleConfig.Name = "cluster-admin"
+func (r *RootSyncReconciler) cleanRBACBindings(ctx context.Context, reconcilerRef, rsRef types.NamespacedName) error {
```
karlkfi, line 833: cleanRBACBindings -> cleanBindings ? This cleans Cluster and Namespaced RoleBindings, right?

sdowell: Yes this cleans both ClusterRoleBindings and RoleBindings, so I figured cleanRBACBindings would be a precise yet generic name

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,52 +793,149 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
...
 	}
+	if err := r.deleteSharedClusterRoleBinding(ctx, RootSyncBaseClusterRoleBindingName, reconcilerRef); err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrap(err, "deleting base binding")
+	}
+	return nil
+}
 
-	_, err := r.upsertClusterRoleBinding(ctx, reconcilerRef, roleConfig.Name)
-	return errors.Wrap(err, "upserting cluster role binding")
+func (r *RootSyncReconciler) configureClusterRoleBinding(ctx context.Context, reconcilerRef, rsRef types.NamespacedName, roleRefs []v1beta1.RootSyncRoleRef) error {
```
karlkfi, line 852: kindof a weird name. might need a func comment to explain. This is upserting and deleting? Maybe it should just say upsert? or apply?

karlkfi:
Lotta code for one PR. 
Would it be possible to pull out the labels added for GC into a separate PR to merge first?
Or maybe do the RootSync changes and RepoSync changed as separate PRs?
Anything else that might come out cleanly?

sdowell:
> Lotta code for one PR. Would it be possible to pull out the labels added for GC into a separate PR to merge first? Or maybe do the RootSync changes and RepoSync changed as separate PRs? Anything else that might come out cleanly?

@karlkfi Let's start with approving the design doc so I don't spend too much time spinning my wheels https://github.com/GoogleContainerTools/kpt-config-sync/pull/990

sdowell:
Ended up breaking this up into several smaller PRs. Pending review on https://github.com/GoogleContainerTools/kpt-config-sync/pull/991 for the final change to implement the RootSync API

PR #978, opened 2023-10-30
Add container args validation and e2e test for log verbosity override

victorpras:
Add validation for container args in reconciler and add e2e test scenario for overriding log level API

b/308638701

### DIFF
--- a/e2e/testcases/override_log_level_test.go
+++ b/e2e/testcases/override_log_level_test.go
@@ -1,0 +1,108 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"strings"
+	"testing"
+
+	"kpt.dev/configsync/e2e/nomostest"
+	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/e2e/nomostest/testpredicates"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/metrics"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+	"kpt.dev/configsync/pkg/testing/fake"
+)
+
+func TestOverrideLogLevel(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.OverrideAPI)
+
+	rootSyncName := nomostest.RootSyncNN(configsync.RootSyncName)
+	rootReconcilerName := core.RootReconcilerObjectKey(rootSyncName.Name)
+	rootSyncV1 := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
+
+	// validate initial container log level value
+	err := nt.Watcher.WatchObject(kinds.Deployment(),
+		rootReconcilerName.Name, rootReconcilerName.Namespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.Reconciler, "-v=0"),
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.GitSync, "-v=5"),
+			testpredicates.DeploymentContainerArgsContains(metrics.OtelAgentName, "-v=0"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// apply override to one container and validate the others are unaffected
+	nt.MustMergePatch(rootSyncV1, `{"spec": {"override": {"logLevels": [{"containerName": "reconciler", "logLevel": 3}]}}}`)
+	err = nt.Watcher.WatchObject(kinds.Deployment(),
+		rootReconcilerName.Name, rootReconcilerName.Namespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.Reconciler, "-v=3"),
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.GitSync, "-v=5"),
+			testpredicates.DeploymentContainerArgsContains(metrics.OtelAgentName, "-v=0"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// apply override to all containers and validate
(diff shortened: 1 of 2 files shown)


e2e/testcases/override_log_level_test.go
```diff
@@ -0,0 +1,51 @@
...
+	rootSyncName := nomostest.RootSyncNN(configsync.RootSyncName)
+	rootReconcilerName := core.RootReconcilerObjectKey(rootSyncName.Name)
+
+	rootSyncV1 := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
+	nt.MustMergePatch(rootSyncV1, `{"spec": {"override": {"logLevels": [{"containerName": "reconciler", "logLevel": 5}]}}}`)
+
+	err := nt.Watcher.WatchObject(kinds.Deployment(),
+		rootReconcilerName.Name, rootReconcilerName.Namespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.Reconciler, fmt.Sprintf("-v=%d", 5)),
```
sdowell, line 45: Perhaps we should add coverage for other containers as well, WDYT?

e2e/testcases/override_log_level_test.go
```diff
@@ -0,0 +1,51 @@
...
+		},
+		testwatcher.WatchTimeout(30*time.Second))
+	if err != nil {
+		nt.T.Fatal(err)
+	}
```
sdowell, line 50: I think it would be worth having some followup steps/assertions as well, for example:
- change the override to a different logLevel value and assert that the flag changes
- remove the override and assert that the flag returns to the default value

victorpras:
holding this PR since there's an issue with setting log level. Currently being addressed in https://github.com/GoogleContainerTools/kpt-config-sync/pull/987

e2e/testcases/override_log_level_test.go
victorpras: `git-sync` and `otel-agent` containers have been included in the test case

e2e/testcases/override_log_level_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// apply override to all containers and validate
+	nt.MustMergePatch(rootSyncV1, `{"spec": {"override": {"logLevels": [{"containerName": "reconciler", "logLevel": 5}, {"containerName": "git-sync", "logLevel": 7}, {"containerName": "otel-agent", "logLevel": 9}]}}}`)
+
+	err = nt.Watcher.WatchObject(kinds.Deployment(),
+		rootReconcilerName.Name, rootReconcilerName.Namespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.Reconciler, fmt.Sprintf("-v=%d", 5)),
```
sdowell, line 74: nit: just write `"-v=5"` instead of using `Sprintf` for readability, since the value is hardcoded anyways

e2e/testcases/override_log_level_test.go
```diff
@@ -0,0 +1,105 @@
...
+	// apply override to one container and validate the others are unaffected
+	nt.MustMergePatch(rootSyncV1, `{"spec": {"override": {"logLevels": [{"containerName": "reconciler", "logLevel": 3}]}}}`)
+	err = nt.Watcher.WatchObject(kinds.Deployment(),
+		rootReconcilerName.Name, rootReconcilerName.Namespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.Reconciler, fmt.Sprintf("-v=%d", 3)),
+			testpredicates.DeploymentContainerArgsContains(reconcilermanager.GitSync, fmt.Sprintf("-v=%d", 5)),
+			testpredicates.DeploymentContainerArgsContains(metrics.OtelAgentName, fmt.Sprintf("-v=%d", 0)),
+		},
+		testwatcher.WatchTimeout(30*time.Second))
```
sdowell, line 63: I think a `30s` timeout may be too short in some cases, for example on autopilot the `reconciler-manager` Pod may get rescheduled and take a while to come back up. I'd suggest just leaving the timeout unspecified so that it uses the default timeout

e2e/testcases/override_log_level_test.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+}
+
+func TestInvalidOverrideLogLevel(t *testing.T) {
```
sdowell, line 98: I'd suggest just rolling this into the other test case. It's usually good to separate out testcases like this but the e2e test scaffolding involves non-trivial cost to set up a test case. For a small example like this is probably better to consolidate it.

e2e/testcases/override_log_level_test.go
```diff
@@ -0,0 +1,105 @@
...
+	nt := nomostest.New(t, nomostesting.OverrideAPI)
+
+	rootSyncV1 := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
+	if err := nt.KubeClient.MergePatch(rootSyncV1, `{"spec": {"override": {"logLevels": [{"containerName": "reconciler", "logLevel": 13}]}}}`); err == nil {
+		nt.T.Fatal("Expecting invalid value error, got none")
```
sdowell, line 103: Can this check be a bit more explicit than just not nil? There's other unexpected errors that could happen here which would silently succeed.

e2e/testcases/override_log_level_test.go
victorpras: done

e2e/testcases/override_log_level_test.go
victorpras: good point about autopilot, removed the timeout

e2e/testcases/override_log_level_test.go
victorpras: sounds good, merged this into the first test

e2e/testcases/override_log_level_test.go
victorpras: the validation has been changed to check for a more specific error message

e2e/testcases/override_log_level_test.go
```diff
@@ -0,0 +1,108 @@
...
+		nt.T.Fatal(err)
+	}
+
+	// try invalid log level value
+	maxError := "RootSync.configsync.gke.io \"root-sync\" is invalid: spec.override.logLevels[0].logLevel: Invalid value: 13: spec.override.logLevels[0].logLevel in body should be less than or equal to 10"
+	minError := "RootSync.configsync.gke.io \"root-sync\" is invalid: spec.override.logLevels[0].logLevel: Invalid value: -3: spec.override.logLevels[0].logLevel in body should be greater than or equal to 0"
+
+	err = nt.KubeClient.MergePatch(rootSyncV1, `{"spec": {"override": {"logLevels": [{"containerName": "reconciler", "logLevel": 13}]}}}`)
+	if !strings.Contains(err.Error(), maxError) {
+		nt.T.Fatalf("Expecting invalid value error, got %s", err.Error())
```
sdowell, line 101: Suggest logging the `want` value (`maxError`/`minError`) here as well just so that the logs are clearer when this fails

sdowell (commented):
looks good, one final nit

sdowell (approved):

PR #981, opened 2023-10-30
Add same labels as other config sync resources

gbmeuk:
When role bindings are created they done so without any labels. Unlike other resources like service accounts.

There than the obvious audit and selector reasons for this, we've found a specific requirement for this where when using hierarchical namespaces, where there will be RepoSyncs in descendant namespaces, the current implementation conflicts because the RoleBinding created by Config Sync has the same name every time.

The solution to this is to use a label selector to tel HNC to not propagate RoleBindings with specific labels. Unfortunately, the the current implementation has no labels.

More details on https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/341


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -944,6 +944,7 @@ func (r *RepoSyncReconciler) upsertRoleBinding(ctx context.Context, reconcilerRe
 	childRB.Namespace = rbRef.Namespace
 
 	op, err := CreateOrUpdate(ctx, r.client, childRB, func() error {
+		core.AddLabels(childRB, labelMap)
```
sdowell, line 947: The same binding is used for many RepoSyncs, so it doesn't fit to apply the typical sync-kind/sync-name/sync-namespace label.

gbmeuk: deleted the sync name label

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -934,16 +934,18 @@ func (r *RepoSyncReconciler) validateNamespaceSecret(ctx context.Context, repoSy
...
-func (r *RepoSyncReconciler) upsertRoleBinding(ctx context.Context, reconcilerRef, rsRef types.NamespacedName) (client.ObjectKey, error) {
+func (r *RepoSyncReconciler) upsertRoleBinding(ctx context.Context, reconcilerRef, rsRef types.NamespacedName, labelMap map[string]string) (client.ObjectKey, error) {
 	rbRef := client.ObjectKey{
 		Namespace: rsRef.Namespace,
 		Name:      RepoSyncPermissionsName(),
 	}
 	childRB := &rbacv1.RoleBinding{}
 	childRB.Name = rbRef.Name
 	childRB.Namespace = rbRef.Namespace
+	delete(labelMap, metadata.SyncNameLabel)
```
sdowell, line 945: This modifies the map that is being passed in, which will have a side effect on the subsequent lines of code that use this map.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -934,16 +934,24 @@ func (r *RepoSyncReconciler) validateNamespaceSecret(ctx context.Context, repoSy
...
-func (r *RepoSyncReconciler) upsertRoleBinding(ctx context.Context, reconcilerRef, rsRef types.NamespacedName) (client.ObjectKey, error) {
+func (r *RepoSyncReconciler) upsertRoleBinding(ctx context.Context, reconcilerRef, rsRef types.NamespacedName, labelMap map[string]string) (client.ObjectKey, error) {
 	rbRef := client.ObjectKey{
 		Namespace: rsRef.Namespace,
 		Name:      RepoSyncPermissionsName(),
 	}
 	childRB := &rbacv1.RoleBinding{}
 	childRB.Name = rbRef.Name
 	childRB.Namespace = rbRef.Namespace
+	// copy labelMap
```
sdowell, line 945: nit: if you rebase you can use `ManagedObjectLabelMap(r.syncKind, rsRef)` to create a new map with the standard labels, which may simplify this snippet

This could also use a comment explaining why we do not include the `sync-name` label here. For example:
```
// Omit the sync-name label, because the RoleBinding may be shared by multiple RepoSyncs
```

sdowell (commented):
Needs a rebase to resolve conflicts.

Are you having issues with signing the CLA?

sdowell:
Replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1052

PR #983, opened 2023-10-31
fix: Stop using the legacy gcp auth plugin

karlkfi:
The gcp auth plugin was removed form client-go and replaced with a stub that panics if the legacy auth plugin is not installed. This was causing nomos not to work in environments without the legacy plugin installed, like containers and prow.

Now, in order to auth to gcp, use gcloud:
gcloud container clusters get-credentials CLUSTER_NAME

BREAKING CHANGE: The legacy auth plugin will no longer be supported by the nomos CLI. GCP and Azure users will need to configure their kubeconfig to use the appropriate credential helper. This fixes nomos to no longer require the legacy auth plugin, migrating use to the new credential helpers, the same way kubectl has done.
- https://kubernetes.io/blog/2022/12/22/kubelet-credential-providers/
- https://cloud.google.com/kubernetes-engine/docs/deprecations/auth-plugin
- https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#install_plugin

To use nomos with GCP:
1. Install nomos:
    ```
    gcloud components install nomos
    ```
2. Install the gke-gcloud-auth-plugin:
    ```
    gcloud components install gke-gcloud-auth-plugin
    ```
3. Configure kubectl to use the gcloud auth plugin:
    ```
    gcloud container clusters get-credentials CLUSTER_NAME --region=COMPUTE_REGION --project=PROJECT_ID
    ```
4. (Optional) List configured contexts:
    ```
    kubectl config get-contexts
    ```
6. List the status of all syncs, in all configured contexts:
    ```
    nomos status
    ```

For Debian/Ubuntu install of gcloud components with apt, see: https://cloud.google.com/sdk/docs/install#deb


karlkfi:
/retest

known flaky unit test

karlkfi:
/retest

known flaky unit test

janetkuo (approved):
/approve 

This breaking change and steps to auth should be documented.

karlkfi:
Doc change request: b/308697610

PR #985, opened 2023-10-31
POC for trying out ksa support as 3rd party idP

droot:
This PR adds support for new auth type `gcpksa` (BYOID) in config-sync.


nan-yu:
replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1121

PR #987, opened 2023-11-01
Set default log level value and configure at runtime

victorpras:
After the refactor in https://github.com/GoogleContainerTools/kpt-config-sync/pull/876, reconciler container log level will only be configured after override is specified. To mitigate this, the default log level value will be set earlier alongside the container resource.

Also in this PR, removed some hardcoded log level value and updated unit test data.

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_container_log_level.go
+++ b/pkg/reconcilermanager/controllers/reconciler_container_log_level.go
@@ -1,0 +1,101 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controllers
+
+import (
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/metrics"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+)
+
+// ReconcilerContainerLogLevelDefaults are the default log level to use for the
+// reconciler deployment containers.
+// All containers default value are 0 except git-sync which default value is 5
+func ReconcilerContainerLogLevelDefaults() map[string]v1beta1.ContainerLogLevelOverride {
+	return map[string]v1beta1.ContainerLogLevelOverride{
+		reconcilermanager.Reconciler: {
+			ContainerName: reconcilermanager.Reconciler,
+			LogLevel:      0,
+		},
+		reconcilermanager.HydrationController: {
+			ContainerName: reconcilermanager.HydrationController,
+			LogLevel:      0,
+		},
+		reconcilermanager.OciSync: {
+			ContainerName: reconcilermanager.OciSync,
+			LogLevel:      0,
+		},
+		reconcilermanager.HelmSync: {
+			ContainerName: reconcilermanager.HelmSync,
+			LogLevel:      0,
+		},
+		reconcilermanager.GitSync: {
+			ContainerName: reconcilermanager.GitSync,
+			LogLevel:      5, // git-sync default is 5 so the logs will store git commands ran
+		},
+		reconcilermanager.GCENodeAskpassSidecar: {
+			ContainerName: reconcilermanager.GCENodeAskpassSidecar,
+			LogLevel:      0,
+		},
+		metrics.OtelAgentName: {
+			ContainerName: metrics.OtelAgentName,
+			LogLevel:      0,
+		},
+	}
+}
+
+// setContainerLogLevelDefaults will compile the default and override value of container log level
+func setContainerLogLevelDefaults(overrides []v1beta1.ContainerLogLevelOverride, defaultsMap map[string]v1beta1.ContainerLogLevelOverride) []v1beta1.ContainerLogLevelOverride {
+
+	// replace defaultsMap value with values from overrides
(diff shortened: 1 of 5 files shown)


pkg/reconcilermanager/controllers/reconciler_container_log_level.go
```diff
@@ -0,0 +1,112 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/metrics"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+)
+
+// ReconcilerContainerLogLevelDefaults are the default log level to use for the
+// reconciler deployment containers.
+// All containers default value are 0 excep git-sync which default value is 5
```
karlkfi, line 29: typo: except

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// setContainerLogLevelDefaults will compile the default and override value of container log level
+func setContainerLogLevelDefaults(overrides []v1beta1.ContainerLogLevelOverride, defaultsMap map[string]v1beta1.ContainerLogLevelOverride) []v1beta1.ContainerLogLevelOverride {
+	// create a map to store container name and log level data
+	overrideMap := make(map[string]v1beta1.ContainerLogLevelOverride)
+
+	// copy override data to map
+	for _, override := range overrides {
+		if override.LogLevel >= 0 && override.LogLevel <= 10 {
```
karlkfi, line 70: why check this here? shouldn't this have already been checked in validation func previously?

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
victorpras: fixed!

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
```diff
@@ -0,0 +1,112 @@
...
+			ContainerName: reconcilermanager.OciSync,
+			LogLevel:      0,
+		},
+		reconcilermanager.HelmSync: {
+			ContainerName: reconcilermanager.HelmSync,
+			LogLevel:      0,
+		},
+		reconcilermanager.GitSync: {
+			ContainerName: reconcilermanager.GitSync,
+			LogLevel:      5, // unlike other containers, git-sync default is 5
```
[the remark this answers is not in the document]
victorpras: from my understanding this is so that we will log the git commands that are ran in git-sync? LMK if there's any other reason that should be mentioned

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
victorpras: this validation has been removed

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+		metrics.OtelAgentName: {
+			ContainerName: metrics.OtelAgentName,
+			LogLevel:      0,
+		},
+	}
+}
+
+// setContainerLogLevelDefaults will compile the default and override value of container log level
+func setContainerLogLevelDefaults(overrides []v1beta1.ContainerLogLevelOverride, defaultsMap map[string]v1beta1.ContainerLogLevelOverride) []v1beta1.ContainerLogLevelOverride {
```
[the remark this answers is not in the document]
victorpras: refactored the func to copy the override into defaults map and convert it back to list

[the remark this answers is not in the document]
karlkfi: This assumes that it's ok to mutate the caller supplied defaultsMap. Usually that's not a great idea, in case the same defaultsMap is used multiple times, in tests, for example.

Is this safe in all uses?

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, override := range overrides {
+		defaultsMap[override.ContainerName] = v1beta1.ContainerLogLevelOverride{
+			ContainerName: override.ContainerName,
+			LogLevel:      override.LogLevel,
+		}
+	}
+
+	// convert defaultsMap back to list
+	overrideList := make([]v1beta1.ContainerLogLevelOverride, 0, len(defaultsMap))
+	for _, override := range defaultsMap {
```
karlkfi, line 76: for loops over maps have semi-random order, depending on size... do we need to sort these to keep the ordering consistent?

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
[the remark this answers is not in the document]
victorpras: i think it's safe to mutate defaultsMap in this specific scenario, but I agree it won't be if we start using it multiple times. I've changed so that the map is copied locally for mutation usage.

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
victorpras: from how the list is used in the `func mutateContainerLogLevel`, i  don't think the ordering matters since the function perform a lookup. What do you think?

karlkfi (approved):
/lgtm

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
karlkfi: actually... i dont see any earlier validation... please add some.

karlkfi:
/hold

for validation

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
karlkfi: Validation should happen before setup, in validateRootSync and validateRepoSync

karlkfi: validation performed by crd

karlkfi:
/unhold

Validation should be performed by CRD. Tests pending: https://github.com/GoogleContainerTools/kpt-config-sync/pull/978

PR #988, opened 2023-11-01
fix: use GroupName for RootSyncPermissionPrefix

sdowell:
This prefix was unintentionally changed to the wrong value in https://github.com/GoogleContainerTools/kpt-config-sync/pull/938


karlkfi (approved):
/lgtm

PR #989, opened 2023-11-01
fix: map RepoSync RoleBinding to RepoSync namespace

sdowell:
The RoleBinding created for RepoSyncs is actually in the namespace of the RepoSync, rather than the controller namespace. This change fixes the watch configuration and mapping logic to correctly map the RoleBinding.

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -512,7 +512,9 @@ func (r *RepoSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
 		Watches(&source.Kind{Type: withNamespace(&corev1.ServiceAccount{}, configsync.ControllerNamespace)},
 			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRepoSync),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
-		Watches(&source.Kind{Type: withNamespace(&rbacv1.RoleBinding{}, configsync.ControllerNamespace)},
+		// Watch RoleBindings in all namespaces, because RoleBindings are created
+		// in the namespace of the RepoSync. Only maps to existing RepoSyncs.
+		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}},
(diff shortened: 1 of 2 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -512,7 +512,7 @@ func (r *RepoSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
 		Watches(&source.Kind{Type: withNamespace(&corev1.ServiceAccount{}, configsync.ControllerNamespace)},
 			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRepoSync),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
-		Watches(&source.Kind{Type: withNamespace(&rbacv1.RoleBinding{}, configsync.ControllerNamespace)},
+		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}},
```
karlkfi, line 515: Maybe add a comment here to explain:
 `Watch RoleBindings in all namespaces, but only trigger Reconcile for RoleBindings that map to an existing RepoSync.`

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -759,12 +759,6 @@ func (r *RepoSyncReconciler) mapConfigMapToRepoSyncs(obj client.Object) []reconc
 func (r *RepoSyncReconciler) mapObjectToRepoSync(obj client.Object) []reconcile.Request {
 	objRef := client.ObjectKeyFromObject(obj)
 
-	// Ignore changes from other namespaces because all the generated resources
-	// exist in the config-management-system namespace.
-	if objRef.Namespace != configsync.ControllerNamespace {
```
karlkfi, line 764: Should we perhaps keep this early exit for non-RoleBindings (Deployment and ServiceAccount)?

sdowell: done

pkg/reconcilermanager/controllers/reposync_controller.go
sdowell: done

karlkfi (approved):
/lgtm

PR #990, opened 2023-11-02
docs: add design doc for RSync roleRefs

sdowell:
The roleRefs field is intended to allow more flexible and simple management of bindings for both RootSync and RepoSync objects. The field allows a list of references to Role and ClusterRole objects. For RepoSyncs, this will always create a RoleBinding in the same namespace as the RepoSync. For RootSyncs, the user can either choose to create a RoleBinding or ClusterRoleBinding via the namespace field.

This is an iteration upon the previous design which added a single ClusterRoleBinding configuration for RootSyncs. This design allows more flexible configuration for RootSyncs, as well as a consistent API for RepoSyncs which simplifies the management of RepoSync RoleBindings.

Issue: https://github.com/GoogleContainerTools/kpt-config-sync/issues/935
Iterates on this PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/938

The implementation will come with followup PRs

### DIFF
--- a/docs/design-docs/02-custom-root-reconciler-clusterrole.md
+++ b/docs/design-docs/02-custom-root-reconciler-clusterrole.md
@@ -1,31 +1,51 @@
-# Custom `ClusterRole` assignments for `RootSync`s
+# Custom RBAC assignments for `RootSync` and `RepoSync`
 
-* Author(s): @tomasaschan
-* Approver: @sdowell, @janetkuo
+* Author(s): @tomasaschan, @sdowell
+* Approver: @mortent, @karlkfi
 * Status: approved
 
 ## Summary
 
-When creating a reconciler deployment for a `RootSync`, Config Sync currently [creates a
+This proposal introduces a way to manage the lifecycle of RBAC bindings for both
+`RootSync` and `RepoSync` objects.
+
+### Historical Behavior
+
+When creating a reconciler deployment for a `RootSync`, Config Sync historically [creates a
 `ClusterRoleBinding`] for the generated `ServiceAccount` granting it `cluster-admin`.
 
-This proposal introduces a mechanism to customize this behavior.
+On the other hand, when creating a reconciler deployment for a `RepoSync`, Config Sync
+currently creates `RoleBinding` for the generated `ServiceAccount` granting
+it `configsync.gke.io:ns-reconciler`. This default `ClusterRole` grants basic permissions required for the `RepoSync`
+reconciler to function, but it's up to users to manage additional `RoleBinding`s.
 
-[creates a `ClusterRoleBinding`]: https://github.com/GoogleContainerTools/kpt-config-sync/blob/199db6dbfa0b9eb1824925c8c687574de095a294/pkg/reconcilermanager/controllers/rootsync_controller.go#L789
+[creates a `ClusterRoleBinding`]: https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.16.0/pkg/reconcilermanager/controllers/rootsync_controller.go#L808
+
+### Proposed Behavior
+
+The proposed behavior will allow full configuration of which bindings are
+created for a `RootSync` or `RepoSync` reconciler. Config Sync will manage the
+lifecycle of all `RoleBinding` and `ClusterRoleBinding` objects declared using
+this new API.
 
 ## Motivation
 
-If one wants to allow deploying resources from a source where one doesn't have full
-control of the contents of the upstream config source, and want to limit what another
-team or third party can accomplish in the cluster, one can imagine limiting what
-permissions are granted to the `ServiceAccount` running the reconciler.
+There are several common use cases where a platform admin may want to limit the
+scope of what a `RootSync` or `RepoSync` can manage in the cluster. For example,
+a platform admin may want to allow a tenant to use a `RootSync` to manage a
+subset of cluster-scoped resources.
 
-However, as Config Sync is currently granting `cluster-admin`, any custom role bindings
-are effectively ignored; the reconciler will, due to the binding added by Config Sync,
-have access to do _anything_ regardless.
+However, as Config Sync is currently granting `cluster-admin` to all `RootSync`
+reconcilers, any custom role bindings are effectively ignored; the `RootSync`
+reconciler will, due to the binding added by Config Sync, have access to do
+_anything_ regardless.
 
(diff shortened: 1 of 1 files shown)


docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -1,82 +1,165 @@
...
-This proposal introduces a mechanism to customize this behavior.
+On the other hand, when creating a reconciler deployment for a `RepoSync`, Config Sync
+currently creates `RoleBinding` for the generated `ServiceAccount` granting
+it `configsync.gke.io:ns-reconciler`. This default `ClusterRole` grants basic permissions required for the `RepoSync`
+reconciler to function, but it's up to users to manage additional `RoleBinding`s.
+
+[creates a `ClusterRoleBinding`]: https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.16.0/pkg/reconcilermanager/controllers/rootsync_controller.go#L808
 
-[creates a `ClusterRoleBinding`]: https://github.com/GoogleContainerTools/kpt-config-sync/blob/199db6dbfa0b9eb1824925c8c687574de095a294/pkg/reconcilermanager/controllers/rootsync_controller.go#L789
+### Proposed Behavior
```
karlkfi, line 24: these new sections are much easier to understand high level, thank you.

karlkfi (approved):
/lgtm

docs/design-docs/02-custom-root-reconciler-clusterrole.md
```diff
@@ -1,82 +1,165 @@
...
+  namespace: example-ns
 spec:
   overrides:
-    clusterRole:
-      disabled: true
+    roleRefs:
+    - kind: ClusterRole # Creates a RoleBinding in example-ns bound to ClusterRole my-cluster-role
+      name: my-cluster-role
+    - kind: Role # Creates a RoleBinding in example-ns bound to Role my-tenant-role
+      name: my-tenant-role
```
janetkuo, line 108: This gives privilege escalation, i.e. someone who only has access to RepoSync to have access to modify Rolebinding. It is not safe to add this feature in.

nan-yu: It only grants the Role to the root reconciler's service account, which is reconciled by the reconciler-manager, so it seems not an issue.

karlkfi: This does allow escalation to namespace admin IF you have write access to the RepoSync. That's different than how it has worked in the past. But there's another way to look at this.

Historically, we've considered two personas: Platform Admin, and Tenant. Platform Admin controlled the RootSyncs, and Tenants controlled the RepoSyncs, but there's another way to frame the personas by considering who has control over the Sources of Truth.

In practice, we've seen many customers use a single RootSync to provision namespaces, RBAC, and RepoSyncs. In this scenario, the Platform Admin controls both the RootSyncs and the RepoSyncs. With this setup, Tenants need to make PRs to the Platform Admin's git repo in order to make changes to those RepoSyncs. That way the Platform Admin can configure Git to require them to approve any changes to the RepoSync and their permissions.

This setup has the advantage that the reconciler-manager owns the RoleBindings and can ensure that they are deleted in the right order, without requiring the Platform Admin to add any depends-on to block deletion of the RoleBindings.

But you're right, with this change, if the Tenant is given RBAC to write RepoSyncs, they effectively are given admin permission in that namespace, and that's different than previous behavior where write access to RoleBindings needed to be managed separately.

sdowell: I suppose the question here is whether the tenant is allowed to create/update `RepoSync` objects themselves, or if the `RepoSync` is created on their behalf (for example, by a platform admin). The premise of https://github.com/GoogleContainerTools/kpt-config-sync/issues/935 makes the implicit assumption that tenants do not have write access to the RSync object itself. I realize I may be incorrectly concluding that assumption aligns with the design of RSync tenancy.

Given that each RepoSync reconciler has a unique ServiceAccount; if a tenant has permission to create RepoSyncs but not RoleBindings, the tenant would need to ask a privileged user to create RoleBindings on their behalf any time they create a new RepoSync. Is this the intended tenancy UX?

In other words, is the tenancy design intended to be that tenants have permission to manage their own RepoSyncs? Or is it assumed that a privileged user (e.g. platform admin) will be creating RepoSyncs on their behalf? 

---
Also, does the same concern apply to `RootSync`? Or is that not the case because `RootSync` reconcilers are granted `cluster-admin` by default?

janetkuo: It's intentional that Tenant only get RBAC access from the admin and can't grant themselves (or the RepoSyncs they own) more permission then what they already have. This is access model of multi-repo. It's also documented in https://cloud.google.com/anthos-config-management/docs/how-to/multiple-repositories. 

This technically applies to RootSync as well, but then RootSync historically already have cluster-admin access.  

> It only grants the Role to the root reconciler's service account, which is reconciled by the reconciler-manager, so it seems not an issue.

Isn't RepoSync is controlled by namespace reconciler? Whatever access the namespace reconciler has, the owner of the RepoSync can get the same access to modify resources via RepoSync.

karlkfi: Actually, server-side apply might allow for a hybrid setup. 

What if the Platform Admin managed RepoSyncs centrally in their SoT, and the source just doesn't include the `branch`/`revision` field? Then Tenants and their CI could be given RBAC to manage the RepoSync, but their changes would be reverted for fields specified in the Platform Admin SoT. That would allow the Tenant to update the `branch`/`revision` whenever they need. 

Still, shared control of the spec of a resource is problematic UX we should try to address.

tomasaschan: FWIW, our preferred model would have been to have Platform Admins own the definitions of both types of RSync. Tenants can change the synced ref by having the RepoSync reference a movable target that is not mainline if they want more controlled deployments - eg, target an OCI image tag `prod` which is manually updated at the source of truth to trigger a deploy. (We abandoned that because config sync could not handle the scale of our number of RepoSyncs, and are now using other, push-based, means to get tenant config into our clusters...).

The model where tenants create RepoSyncs themselves has never been attractive to us.

PR #991, opened 2023-11-02
feat: implement roleRefs API for RootSyncs

sdowell:
This change adds the roleRefs API field to the CRD for both RootSync objects. This API is intended to streamline the management of RBAC bindings for RootSync reconcilers.

Design doc: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/docs/design-docs/02-custom-root-reconciler-clusterrole.md


e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1171,3 +1171,36 @@ func RootSyncSpecOverrideEquals(expected *v1beta1.RootSyncOverrideSpec) Predicat
...
+			return errors.Errorf("missing subject %q", name)
+		}
+	}
+
+	return nil
+}
+
+// ClusterRoleBindingHasSubjects checks that the ClusterRoleBinding has a list
+// of subjects whose names match the specified list of names.
+func ClusterRoleBindingHasSubjects(subjects ...string) func(o client.Object) error {
```
karlkfi, line 1195: Might make sense to name this SubjectsEqual, instead of HasSubjects, since it's validating an exact match, not a subset.

e2e/testcases/override_role_refs_test.go
```diff
@@ -0,0 +1,226 @@
+// Copyright 2022 Google LLC
```
karlkfi, line 1: current year copyright for new files

e2e/testcases/override_role_refs_test.go
```diff
@@ -0,0 +1,226 @@
...
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
+	"kpt.dev/configsync/pkg/testing/fake"
+)
+
+func TestRootSyncRoleRefs(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.OverrideAPI, ntopts.Unstructured,
+		ntopts.RootRepo("sync-a"), // will declare namespace-a explicitly
```
karlkfi, line 38: what does this comment mean? I don't see a namespace-a in use.

e2e/testcases/override_role_refs_test.go
```diff
@@ -0,0 +1,226 @@
...
+			RoleRefBase: v1beta1.RoleRefBase{
+				Kind: "ClusterRole",
+				Name: "foo-role",
+			},
+			Namespace: "foo",
+		},
+	}
+	roleObject := fake.RoleObject(core.Name("foo-role"), core.Namespace("foo"))
+	clusterRoleObject := fake.ClusterRoleObject(core.Name("foo-role"))
+	clusterRoleObject.Rules = []rbacv1.PolicyRule{
```
karlkfi, line 81: kinda weird having this in a specific test when these objects are defined in the test setup. Will we need this for other tests later? Or only for RootSync tests validating custom roleRefs?

e2e/testcases/override_role_refs_test.go
```diff
@@ -0,0 +1,226 @@
...
+	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/e2e/nomostest/testpredicates"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
+	"kpt.dev/configsync/pkg/testing/fake"
+)
+
+func TestRootSyncRoleRefs(t *testing.T) {
```
karlkfi, line 36: nit: For readability and debugging, I'd suggest adding some line breaks and log messages between sections. I tend to think of the flow of e2e tests as:
1. setup
2. make changes
3. validate expected results
4. repeat steps 2 & 3

So I like to add line breaks between those sections and use log messages and/or comments to summarize the changes made and the results expected. And then maybe add a function comment to summarize and list the steps.

These are all just nice to haves, tho, not blockers.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -235,7 +235,7 @@ func validateRepoSyncDependencies(nt *nomostest.NT, ns, rsName string) []client.
 
 	repoSyncRB := &rbacv1.RoleBinding{}
 	setNN(repoSyncRB, client.ObjectKey{
```
karlkfi, line 237: nit: this setNN func seems a little silly when it takes 4 lines, when the alternative setters only take 2 lines. I think it only really makes sense to use if the NN variable is pre-made and re-used.

e2e/testcases/override_role_refs_test.go
sdowell: copied the boilerplate from another test and forgot to remove the comment

e2e/testcases/override_role_refs_test.go
sdowell: The reason this exists is because of the "safety" ClusterRole/Namespace that get added to all RootSync repositories by the test framework. Normally the RootSync gets cluster-admin by default and can manage ClusterRoles/Namespaces, but here the RootSync would error because it doesn't have permission to manage these resources.

For now this is the only test that adjusts the `roleRefs` field for `RootSyncs`, but yes if we added more tests that set `roleRefs` then we may need to reuse this `ClusterRole`.

e2e/testcases/reconciler_manager_test.go
sdowell: Done

e2e/nomostest/testpredicates/predicates.go
sdowell: Done

e2e/testcases/override_role_refs_test.go
sdowell: Done

e2e/testcases/override_role_refs_test.go
sdowell: Done

e2e/testcases/override_role_refs_test.go
```diff
@@ -0,0 +1,226 @@
...
+	})
+	tg.Go(func() error {
+		return nt.ValidateNotFound(controllers.RootSyncBaseClusterRoleBindingName, "", &rbacv1.ClusterRoleBinding{})
+	})
+	if err := tg.Wait(); err != nil {
+		nt.T.Fatal(err)
+	}
+}
+
+func validateRoleRefs(nt *nomostest.NT, syncKind string, rsRef types.NamespacedName, expected []v1beta1.RootSyncRoleRef) error {
```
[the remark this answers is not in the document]
sdowell: Done

docs/design-docs/02-custom-reconciler-bindings.md
janetkuo, line 1: Why's this file renamed?

pkg/api/configsync/v1alpha1/resource_override.go
```diff
@@ -93,34 +93,51 @@ type RootSyncOverrideSpec struct {
...
+	Namespace string `json:"namespace,omitempty"`
+}
 
-	// disabled turns off binding to a ClusterRole entirely.
-	//
-	// It is then your responsibility to bind any roles to the relevant service
-	// account.
+// RoleRefBase is the base struct for references to Roles and ClusterRoles. These
+// references are used to create bindings for both RootSyncs and RepoSyncs.
+type RoleRefBase struct {
```
karlkfi, line 120: If we're not going to add this to RepoSync, you can probably simplify and remove the RoleRefBase

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1240,3 +1240,36 @@ func ResourceGroupStatusEquals(expected v1alpha1.ResourceGroupStatus) Predicate
 		return nil
 	}
 }
+
+func subjectsEqual(want []string, got []rbacv1.Subject) error {
```
janetkuo, line 1244: This only compares subjects name but not other fields. Suggest renaming to make it more clear.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -779,52 +801,151 @@ func (r *RootSyncReconciler) validateRootSecret(ctx context.Context, rootSync *v
 	return validateSecretData(rootSync.Spec.Auth, secret)
 }
 
-func (r *RootSyncReconciler) configureClusterRoleBinding(ctx context.Context, reconcilerRef types.NamespacedName, roleConfig v1beta1.ClusterRoleOverrideSpec) error {
-	// in order to be backwards compatible with behavior before https://github.com/GoogleContainerTools/kpt-config-sync/pull/938
-	// we delete any ClusterRoleBinding that uses the old name.
-	// This ensures smooth migrations for users upgrading past that version boundary.
-	if err := r.deleteClusterRoleBinding(ctx, fmt.Sprintf("%s:%s", core.RootReconcilerPrefix, configsync.RootSyncName)); err != nil && !apierrors.IsNotFound(err) {
-		return errors.Wrap(err, "deleting old binding")
+func (r *RootSyncReconciler) listCurrentRoleRefs(ctx context.Context, rsRef types.NamespacedName) (map[v1beta1.RootSyncRoleRef]client.Object, error) {
```
karlkfi, line 804: probably complex enough to need a comment explaining what's happening

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1240,3 +1240,36 @@ func ResourceGroupStatusEquals(expected v1alpha1.ResourceGroupStatus) Predicate
...
+			return errors.Errorf("missing subject %q", name)
+		}
+	}
+
+	return nil
+}
+
+// ClusterRoleBindingSubjectsEqual checks that the ClusterRoleBinding has a list
+// of subjects whose names match the specified list of names.
+func ClusterRoleBindingSubjectsEqual(subjects ...string) func(o client.Object) error {
```
janetkuo, line 1264: Again this only compares subject names, so it needs to be more clear.

pkg/reconcilermanager/controllers/build_names.go
```diff
@@ -21,19 +21,27 @@ import (
...
+	// for RepoSync objects. This contains basic permissions for RepoSync reconcilers
+	//(e.g. RepoSync status update).
+	RepoSyncBaseRoleBindingName = RepoSyncBaseClusterRoleName
+	// RootSyncLegacyClusterRoleBindingName is the name of the legacy ClusterRoleBinding created
+	// for RootSync objects. It is always bound to cluster-admin.
+	RootSyncLegacyClusterRoleBindingName = RootSyncBaseClusterRoleName
+	// RootSyncBaseClusterRoleBindingName is the name of the default ClusterRoleBinding created
+	// for RootSync objects. This contains basic permissions for RootSync reconcilers
+	// (e.g. RootSync status update).
+	RootSyncBaseClusterRoleBindingName = RootSyncBaseClusterRoleName + "-base"
```
janetkuo, line 41: Is it a breaking change by adding a trailing `-base`?

sdowell: This name is for a new binding, so this is not breaking. The old cluster-admin binding name is unchanged

e2e/nomostest/testpredicates/predicates.go
sdowell: Renamed

e2e/nomostest/testpredicates/predicates.go
sdowell: Renamed

docs/design-docs/02-custom-reconciler-bindings.md
sdowell: This was from before when we were considering the API for RepoSync. Reverted

pkg/reconcilermanager/controllers/rootsync_controller.go
sdowell: Added a comment

pkg/api/configsync/v1alpha1/resource_override.go
sdowell: Removed.

manifests/root-reconciler-base-cluster-role.yaml
```diff
@@ -0,0 +1,34 @@
+# Copyright 2022 Google LLC
```
janetkuo, line 1: nit: 2023

manifests/root-reconciler-base-cluster-role.yaml
```diff
@@ -0,0 +1,34 @@
...
+  labels:
+    configmanagement.gke.io/system: "true"
+    configmanagement.gke.io/arch: "csmr"
+rules:
+- apiGroups: ["configsync.gke.io"]
+  resources: ["rootsyncs"]
+  verbs: ["get","list","watch","update","patch"]
+- apiGroups: ["configsync.gke.io"]
+  resources: ["rootsyncs/status"]
+  verbs: ["get","list","watch","update","patch"]
```
janetkuo, line 28: It needs to be able to access RepoSyncs (RSyncs of Rsyncs), Secrets, and ConfigMaps as well?

pkg/reconcilermanager/controllers/garbage_collector.go
```diff
@@ -233,8 +232,32 @@ func (r *reconcilerBase) deleteDeployment(ctx context.Context, reconcilerRef typ
...
+		}
+		return NewObjectOperationErrorWithKey(err, crb, OperationGet, crbKey)
+	}
+	count := len(crb.Subjects)
+	crb.Subjects = removeSubject(crb.Subjects, r.serviceAccountSubject(reconcilerRef))
+	if count == len(crb.Subjects) {
+		// No change
+		return nil
+	}
+	if len(crb.Subjects) == 0 {
```
janetkuo, line 255: If `count` was zero when it's defined (L249), this line won't be executed.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -437,6 +436,10 @@ func (r *RootSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
 			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRootSync),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
 		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}},
+			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRootSync),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
+		// TODO: is it possible to watch with a label filter?
```
janetkuo, line 441: This is supported https://github.com/kubernetes-sigs/controller-runtime/issues/244

pkg/reconcilermanager/controllers/garbage_collector.go
sdowell: To be clear the behavior is actually unchanged from before, although the diff is a bit misleading. The original function for reference: https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.16/pkg/reconcilermanager/controllers/garbage_collector.go#L235

I agree with you though that we should switch the order so that the CRB always gets garbage collected when the subjects are empty. I'll update this

manifests/root-reconciler-base-cluster-role.yaml
sdowell: It's not clear to me that we should grant those permissions as a baseline. It's possible that a platform admin doesn't want to grant permission for a given RootSync to manage RepoSyncs

manifests/root-reconciler-base-cluster-role.yaml
sdowell: fixed

manifests/rootsync-crd.yaml
```diff
@@ -459,6 +444,33 @@ spec:
                           x-kubernetes-int-or-string: true
                       type: object
                     type: array
+                  roleRefs:
+                    description: roleRefs is a list of Roles or ClusterRoles to create
+                      bindings. If unset, a binding to cluster-admin will be created.
+                    items:
+                      description: RootSyncRoleRef indicates references for Roles
```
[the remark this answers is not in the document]
sdowell: Done. Note that other fields have this problem, may want to address in a followup.

pkg/reconcilermanager/controllers/rootsync_controller.go
sdowell: Dug into this a bit and it looks like the implementation you linked is a setting at the level of the controller manager (`SelectorsByObject`). Since we use a shared controller manager for both RootSync and RepoSync reconcilers, I don't think we could filter all RoleBindings by a single label selector (based on the current labels).

This could potentially work if we added a new label that was applied to all managed objects (RoleBindings in this case, but could also apply to other objects) that is the same for both RootSync and RepoSync

karlkfi: Generally when using the controller-manager you have to assume your watches are being shared, so you can't filter them without filtering them for all controllers (e.g. [SelectorsByObject](https://github.com/kubernetes-sigs/controller-runtime/pull/1435)). There was discussion of making these dynamic based on watch selectors here, but it was never implemented due to caching complexities.

The general workaround is to use WithPredicates instead, which allows skipping reconciling of events for a specific controller unless they match the predicate.

karlkfi: There's a builder.LabelSelectorPredicate, if you want.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -571,13 +574,19 @@ func (r *RootSyncReconciler) mapObjectToRootSync(obj client.Object) []reconcile.
 		reconcilerName := core.RootReconcilerName(rs.GetName())
```
janetkuo, line 574: Comments above need updating

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -437,6 +436,10 @@ func (r *RootSyncReconciler) SetupWithManager(mgr controllerruntime.Manager, wat
 			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRootSync),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
 		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}},
+			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRootSync),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
+		// TODO: is it possible to watch with a label filter?
+		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}},
```
janetkuo, line 442: Can you only start watches when users opt-in to this feature?

sdowell: My understanding is that the watches are instantiated at controller startup, so I don't think this would be possible given that the feature is enabled on the RootSync.

As a side note I'm looking into adding a label selector for the shared cache, which should optimize all of these watches.

pkg/reconcilermanager/controllers/rootsync_controller.go
sdowell: Added a comment

janetkuo (commented):
/approve

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: Actually, I think we have code already that adds watches at runtime. It's not as pretty as this interface but it's doable. This could be a future optimization.

karlkfi: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/reposync_controller.go#L550

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: This could be added as a future optimization.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -207,7 +207,7 @@ func validateRootSyncDependencies(nt *nomostest.NT, rsName string) []client.Obje
 	// only happens when upgrading from a very old unsupported version.
 
 	rootSyncCRB := &rbacv1.ClusterRoleBinding{}
-	setNN(rootSyncCRB, client.ObjectKey{Name: controllers.RootSyncPermissionsName(rootSyncReconciler.Name)})
+	setNN(rootSyncCRB, client.ObjectKey{Name: controllers.RootSyncLegacyClusterRoleBindingName})
```
karlkfi, line 210: another case where setNN is unnecessary abstraction, since it's just setting the name.

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -93,29 +93,36 @@ type RootSyncOverrideSpec struct {
...
+//
+//nolint:revive
+type RootSyncRoleRef struct {
+
+	// kind refers to the Kind of the RBAC resource.
+	// Accepted values are Role and ClusterRole.
 	//
-	// +optional
+	// +kubebuilder:validation:Pattern=^(Role|ClusterRole)$
+	Kind string `json:"kind,omitempty"`
```
karlkfi, line 114: Are these all required when you don't specify +optional?
If it's required, should it still have omitempty?

karlkfi: Same for Name

pkg/reconcilermanager/controllers/rootsync_controller_test.go
```diff
@@ -70,6 +72,15 @@ const (
 
 var rootReconcilerName = core.RootReconcilerName(rootsyncName)
 
+func role(t *testing.T, name string, opts ...core.MetaMutator) *rbacv1.Role {
+	t.Helper()
+
+	result := fake.RoleObject(opts...)
+	result.Name = name
```
karlkfi, line 79: nit: i think there's a name mutator. that would allow this to be inlined.

pkg/reconcilermanager/controllers/rootsync_controller_test.go
```diff
@@ -70,6 +72,15 @@ const (
 
 var rootReconcilerName = core.RootReconcilerName(rootsyncName)
 
+func role(t *testing.T, name string, opts ...core.MetaMutator) *rbacv1.Role {
+	t.Helper()
```
karlkfi, line 76: i dont think we need the t.Helper() on functions that can't call t.Fail

sdowell: removed

pkg/reconcilermanager/controllers/rootsync_controller_test.go
sdowell: There is, but I'm not sure of a cleaner way to combine that into a single variadic argument

pkg/api/configsync/v1beta1/resource_override.go
sdowell: You're right, the `omitempty` parameter effectively made these optional. I removed `omitempty`, and they should be properly validated as required by the CRD now.

e2e/testcases/reconciler_manager_test.go
sdowell: Removed

pkg/reconcilermanager/controllers/rootsync_controller_test.go
karlkfi: `append(Name(name), opts...)...`

sdowell: I feel that's less readable than the alternative here, but I don't feel strongly

karlkfi (approved):
/lgtm

Thanks for cleaning this up!

pkg/reconcilermanager/controllers/rootsync_controller_test.go
karlkfi: i don't disagree. just a cute way to make it short. no biggie.

PR #992, opened 2023-11-02
consolidate apt-get calls to a single layer

sdowell:
Several images using the same debian-nonroot base were performing similar apt-get update/install calls. Consolidating these to a single layer should improve runtime.

### DIFF
--- a/build/all/Dockerfile
+++ b/build/all/Dockerfile
@@ -54,6 +54,7 @@ ARG USER_UID=65532
 ARG USER_GID=$USER_UID
 RUN groupadd --gid $USER_GID $USERNAME && \
   useradd --uid $USER_UID --gid $USER_GID -m $USERNAME
+RUN apt-get update && apt-get install -y bash git
 USER nonroot:nonroot
 
 # Hydration controller image
@@ -104,7 +105,6 @@ COPY --from=bins /workspace/.output/third_party/kustomize/kustomize /usr/local/b
 COPY --from=bins /workspace/.output/third_party/kustomize/NOTICES /third_party/kustomize/NOTICES
 COPY --from=bins /workspace/LICENSE LICENSE
 COPY --from=bins /workspace/LICENSES.txt LICENSES.txt
-RUN apt-get update && apt-get install -y git
 USER nonroot:nonroot
 ENTRYPOINT ["/hydration-controller"]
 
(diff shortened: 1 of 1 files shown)


build/all/Dockerfile
```diff
@@ -55,6 +55,7 @@ ARG USER_GID=$USER_UID
 RUN groupadd --gid $USER_GID $USERNAME && \
   useradd --uid $USER_UID --gid $USER_GID -m $USERNAME
 USER nonroot:nonroot
+RUN apt-get update && apt-get install -y bash git
```
karlkfi, line 58: should this be done before changing the user?

karlkfi, line 58: This puts bash in hydration-controller-with-shell. Is that desirable?

karlkfi, line 58: Does doing the update here actually avoid CVEs? Or would using the latest DEBIAN_BASE_IMAGE already solve that problem for packages that aren't installed by default?

sdowell: You are correct!

sdowell: It wasn't there before but I feel like it should be fine. The debian base image already comes with a shell, this just adds bash as another shell. The benefit is that we only need to perform the install in one layer

sdowell: I don't think the base image is guaranteed to have an up to date package index. It seems plausible that packages which aren't already installed may be indexed to a stale version which has since been patched. I don't think the upstream base image would be updated for packages which are not installed, thus I don't think we can rely on just updating to the latest base image

sdowell: If it's a big deal, we could cache everything except for the bash install and just `apt-get install bash` in the layer for the `nomos` image

karlkfi: any idea what the size diff is with or without bash?

sdowell: As in the diff in the size of the resulting image?

karlkfi: right. like, is it going to significantly increase costs for our users in download & storage? is it like 1MiB, or 100MiB?

sdowell: Based on local A/B testing:

Size without bash: 325279874 Bytes
Size with bash: 327555914 Bytes

Difference: 2276040 Bytes (Approximately 2 MB)

Diving into the container itself, it looks like the bash binary is a little over 1 MB:
```
-rwxr-xr-x 1 root root   1265648 Apr 23  2023  bash
```

karlkfi: Thanks for checking. That seems fine.

karlkfi (approved):
/lgtm

PR #997, opened 2023-11-03
ci: remove in-repo artifacts from gke-e2e image

sdowell:
This removes any remaining in-repo artifacts from the gke-e2e image, which simplifies the build by virtue of not needing to copy local files into the build context. Furthermore, the CI jobs will now always use the latest code that is defined in the repo checkout rather than what was published in the original image.


nan-yu (approved):
/lgtm

PR #998, opened 2023-11-06
test: migrate ResourceGroup e2e tests to this repo

sdowell:
This implements the e2e test cases as they are defined in the upstream ResourceGroup repository. Implementing these tests here will enable migrating the remaining controller logic and API definitions to this repository.

### DIFF
--- a/e2e/nomostest/testpredicates/predicates.go
+++ b/e2e/nomostest/testpredicates/predicates.go
@@ -27,6 +27,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"kpt.dev/configsync/e2e/nomostest/retry"
@@ -40,6 +41,7 @@ import (
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/rootsync"
 	"kpt.dev/configsync/pkg/util/log"
+	"kpt.dev/resourcegroup/apis/kpt.dev/v1alpha1"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -1198,4 +1200,43 @@ func containerArgsContains(container *corev1.Container, expectedArg string) erro
 		}
 	}
 	return fmt.Errorf("expected arg not found: %s", expectedArg)
+}
+
+// ResourceGroupStatusEquals checks that the RootSync's spec.override matches
+// the specified RootSyncOverrideSpec.
+func ResourceGroupStatusEquals(expected v1alpha1.ResourceGroupStatus) Predicate {
+	return func(obj client.Object) error {
+		rg, ok := obj.(*v1alpha1.ResourceGroup)
+		if !ok {
+			return WrongTypeErr(obj, &v1alpha1.ResourceGroup{})
+		}
+		found := rg.DeepCopy().Status
+		// Set LastTransitionTime to nil value for comparisons
+		for x := range found.Conditions {
+			found.Conditions[x].LastTransitionTime = metav1.Time{}
+		}
+		for x := range found.ResourceStatuses {
+			for y := range found.ResourceStatuses[x].Conditions {
+				found.ResourceStatuses[x].Conditions[y].LastTransitionTime = metav1.Time{}
+				// Allow substring matching for ResourceStatus messages
+				if x < len(expected.ResourceStatuses) && y < len(expected.ResourceStatuses[x].Conditions) {
+					foundMsg := found.ResourceStatuses[x].Conditions[y].Message
+					expectedMsg := expected.ResourceStatuses[x].Conditions[y].Message
+					if strings.Contains(foundMsg, expectedMsg) {
+						expected.ResourceStatuses[x].Conditions[y].Message = foundMsg
+					}
+				}
+			}
+		}
+		for x := range found.SubgroupStatuses {
+			for y := range found.SubgroupStatuses[x].Conditions {
+				found.SubgroupStatuses[x].Conditions[y].LastTransitionTime = metav1.Time{}
+			}
+		}
+		if !equality.Semantic.DeepEqual(found, expected) {
+			return errors.Errorf("expected %s to have status: %s, but got %s",
+				kinds.ObjectSummary(obj), log.AsJSON(expected), log.AsJSON(found))
+		}
+		return nil
+	}
 }
--- a/e2e/nomostest/testresourcegroup/resourcegroup.go
+++ b/e2e/nomostest/testresourcegroup/resourcegroup.go
@@ -1,0 +1,150 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testresourcegroup
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
+	"kpt.dev/configsync/pkg/api/configmanagement"
+	"kpt.dev/resourcegroup/apis/kpt.dev/v1alpha1"
+	"kpt.dev/resourcegroup/controllers/resourcegroup"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// NotOwnedMessage is the status message for a resource which does not have
+	// an inventory id.
+	NotOwnedMessage = "This object is not owned by any inventory object. The status for the current object may not reflect the specification for it in current ResourceGroup."
+)
+
+// New creates a new ResourceGroup object
+func New(nn types.NamespacedName, id string) *v1alpha1.ResourceGroup {
+	rg := &v1alpha1.ResourceGroup{}
+	rg.SetNamespace(nn.Namespace)
+	rg.SetName(nn.Name)
+	if id != "" {
+		rg.SetLabels(map[string]string{
+			common.InventoryLabel: id,
+		})
+	}
+	return rg
+}
+
+// GenerateResourceStatus is a helper function to generate an array of ResourceStatus
+// from ObjMetadata
+func GenerateResourceStatus(resources []v1alpha1.ObjMetadata, s v1alpha1.Status, sourceHash string) []v1alpha1.ResourceStatus {
+	resourceStatus := []v1alpha1.ResourceStatus{}
+	for _, resource := range resources {
+		resourceStatus = append(resourceStatus, v1alpha1.ResourceStatus{
+			ObjMetadata: resource,
+			Status:      s,
+			SourceHash:  sourceHash,
+		})
+	}
+	return resourceStatus
+}
+
+// EmptyStatus returns the initial/default ResourceStatus of a
+// ResourceGroup object. This is the status set by the controller upon a successful
+// reconcile.
+func EmptyStatus() v1alpha1.ResourceGroupStatus {
+	return v1alpha1.ResourceGroupStatus{
+		Conditions: []v1alpha1.Condition{
+			{
+				Type:    v1alpha1.Reconciling,
+				Status:  v1alpha1.FalseConditionStatus,
+				Reason:  resourcegroup.FinishReconciling,
+				Message: "finish reconciling",
+			},
+			{
+				Type:    v1alpha1.Stalled,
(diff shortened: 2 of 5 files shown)


e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1199,3 +1201,42 @@ func containerArgsContains(container *corev1.Container, expectedArg string) erro
 	}
 	return fmt.Errorf("expected arg not found: %s", expectedArg)
 }
+
+// ResourceGroupHasStatus checks that the RootSync's spec.override matches
+// the specified RootSyncOverrideSpec.
+func ResourceGroupHasStatus(expected v1alpha1.ResourceGroupStatus) Predicate {
```
karlkfi, line 1207: rename: ResourceGroupStatusEquals

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1199,3 +1201,42 @@ func containerArgsContains(container *corev1.Container, expectedArg string) erro
...
+	return func(obj client.Object) error {
+		rg, ok := obj.(*v1alpha1.ResourceGroup)
+		if !ok {
+			return WrongTypeErr(obj, &v1alpha1.ResourceGroup{})
+		}
+		found := rg.DeepCopy().Status
+		// Set LastTransitionTime to nil value for comparisons
+		for x := range found.Conditions {
+			found.Conditions[x].LastTransitionTime = metav1.Time{}
```
karlkfi, line 1216: Should this nil LastUpdateTime too? Or does it not have one?

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1199,3 +1201,42 @@ func containerArgsContains(container *corev1.Container, expectedArg string) erro
...
+		}
+		for x := range found.ResourceStatuses {
+			for y := range found.ResourceStatuses[x].Conditions {
+				found.ResourceStatuses[x].Conditions[y].LastTransitionTime = metav1.Time{}
+				// Allow substring matching for ResourceStatus messages
+				if x < len(expected.ResourceStatuses) && y < len(expected.ResourceStatuses[x].Conditions) {
```
karlkfi, line 1222: should there be an error if the indexes are out of range?

e2e/nomostest/testresourcegroup/resourcegroup.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+const (
+	// NotOwnedMessage is the status message for a resource which does not have
+	// an inventory id.
+	NotOwnedMessage = "This object is not owned by any inventory object. The status for the current object may not reflect the specification for it in current ResourceGroup."
+)
+
+// NewKptResourceGroup creates a new ResourceGroup object
+func NewKptResourceGroup(nn types.NamespacedName, id string) *v1alpha1.ResourceGroup {
```
karlkfi, line 40: rename: New (result is named by package)

e2e/nomostest/testresourcegroup/resourcegroup.go
```diff
@@ -0,0 +1,150 @@
...
+		rg.SetLabels(map[string]string{
+			common.InventoryLabel: id,
+		})
+	}
+	return rg
+}
+
+// AggregateResourceStatus is a helper function to generate an array of ResourceStatus
+// from ObjMetadata
+func AggregateResourceStatus(resources []v1alpha1.ObjMetadata, s v1alpha1.Status, sourceHash string) []v1alpha1.ResourceStatus {
```
karlkfi, line 54: "aggregate" seems like not the right verb for this. 
Perhaps "map" would be more accurate? It's mapping the specified status onto all the specified objects.

e2e/nomostest/testresourcegroup/resourcegroup.go
```diff
@@ -0,0 +1,150 @@
...
+			Status:      s,
+			SourceHash:  sourceHash,
+		})
+	}
+	return resourceStatus
+}
+
+// InitialResourceStatus returns the initial/default ResourceStatus of a
+// ResourceGroup object.
+func InitialResourceStatus() v1alpha1.ResourceGroupStatus {
```
karlkfi, line 68: I'm confused. Why is the initial status FinishReconciling?  I would have expected the initial status to just not includes these conditions at all.

e2e/nomostest/testresourcegroup/resourcegroup.go
```diff
@@ -0,0 +1,150 @@
...
+				Reason:  resourcegroup.FinishReconciling,
+				Message: "finish reconciling",
+			},
+		},
+	}
+}
+
+// PatchResourceGroup modifies a ResourceGroup object on the cluster using
+// the provided mutateFn.
+func PatchResourceGroup(kubeClient *testkubeclient.KubeClient, nn types.NamespacedName, mutateFn func(group *v1alpha1.ResourceGroup)) error {
```
karlkfi, line 89: Rename: Update

e2e/nomostest/testresourcegroup/resourcegroup.go
```diff
@@ -0,0 +1,150 @@
...
+		return err
+	}
+	obj2 := obj.DeepCopy()
+	mutateFn(obj2)
+	return kubeClient.Update(obj2)
+}
+
+// ApplyResources applies the list of resources to the cluster with the provided
+// ResourceGroup inventory ID.
+func ApplyResources(kubeClient *testkubeclient.KubeClient, resources []v1alpha1.ObjMetadata, id string) error {
```
karlkfi, line 101: Rename: CreateOrUpdate

e2e/nomostest/testresourcegroup/resourcegroup.go
```diff
@@ -0,0 +1,150 @@
...
+		} else {
+			return err
+		}
+	}
+	return nil
+}
+
+// ControllerPodsNoRestart checks that no pods have restarted in the ResourceGroup
+// controller namespace.
+func ControllerPodsNoRestart(kubeClient *testkubeclient.KubeClient) error {
```
karlkfi, line 134: This is only used in one place. I think you can inline it or make it a private local func near where it's used. Or maybe call it ListControllerPods and do the restart check in the test that calls it.

e2e/testcases/kcc_test.go
```diff
@@ -129,3 +135,93 @@ func TestKCCResourcesOnCSR(t *testing.T) {
...
+
+	ns := v1.Namespace{}
+	ns.Name = "resourcegroup-e2e"
+	nt.T.Cleanup(func() {
+		// all test resources are created in this namespace
+		if err := nt.KubeClient.Delete(ns.DeepCopy()); err != nil {
+			nt.T.Error(err)
+		}
+	})
+	if err := nt.KubeClient.Create(ns.DeepCopy()); err != nil {
```
karlkfi, line 150: why deepcopy?

e2e/testcases/kcc_test.go
```diff
@@ -129,3 +135,93 @@ func TestKCCResourcesOnCSR(t *testing.T) {
...
 }
+
+func TestResourceGroupKCC(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.ACMController, ntopts.KCCTest)
+
+	ns := v1.Namespace{}
+	ns.Name = "resourcegroup-e2e"
+	nt.T.Cleanup(func() {
+		// all test resources are created in this namespace
+		if err := nt.KubeClient.Delete(ns.DeepCopy()); err != nil {
```
karlkfi, line 146: DeepCopy is expensive. Maybe just make a var for the namespace name and recreate the Namespace for the create and delete.

e2e/testcases/kcc_test.go
```diff
@@ -129,3 +135,93 @@ func TestKCCResourcesOnCSR(t *testing.T) {
...
+		},
+	}
+	err = nt.Watcher.WatchObject(kinds.ResourceGroup(), rgNN.Name, rgNN.Namespace, []testpredicates.Predicate{
+		testpredicates.ResourceGroupHasStatus(expectedStatus),
+	})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.T.Log("Delete the ResourceGroup")
+	if err := nt.KubeClient.Delete(rg.DeepCopy()); err != nil {
```
karlkfi, line 224: should delete be in Cleanup or do we only want to delete on success?

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1199,3 +1201,42 @@ func containerArgsContains(container *corev1.Container, expectedArg string) erro
...
+		rg, ok := obj.(*v1alpha1.ResourceGroup)
+		if !ok {
+			return WrongTypeErr(obj, &v1alpha1.ResourceGroup{})
+		}
+		found := rg.DeepCopy().Status
+		// Set LastTransitionTime to nil value for comparisons
+		for x := range found.Conditions {
+			found.Conditions[x].LastTransitionTime = metav1.Time{}
+		}
+		for x := range found.ResourceStatuses {
```
karlkfi, line 1218: should we sort these or do we expect them to be consistently and predictably ordered?

e2e/nomostest/testpredicates/predicates.go
sdowell: Doesn't look like it has one

e2e/nomostest/testpredicates/predicates.go
sdowell: It would error below in the deep equal check

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: Seems to be a quirk of the resource-group controller. Looks like it comes from https://github.com/GoogleContainerTools/kpt-resource-group/blob/main/controllers/resourcegroup/resourcegroup_controller.go#L216

e2e/testcases/kcc_test.go
sdowell: Not sure what the intent is, the upstream test deletes the ResourceGroup at the end of the test. Cleanup will manage deletion regardless by deleting the namespace.

e2e/nomostest/testpredicates/predicates.go
sdowell: They seem to be predictably ordered with the test cases thus far

e2e/testcases/kcc_test.go
sdowell: Done.

e2e/testcases/kcc_test.go
sdowell: I found it an easy abstraction to work with when writing the tests. Removed.

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: I realized I missed a couple spots in the translated tests, it's used by three different test cases now

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: Done

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: Done

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: Renamed to `GenerateResourceStatus`

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: Done

e2e/nomostest/testpredicates/predicates.go
sdowell: Done

e2e/nomostest/testresourcegroup/resourcegroup.go
karlkfi: Oh, I see. It's not the initial status so much as the success status set after all resource statuses have been computed without stalling. But if there's no resources, then it reconciles immediately.

How about `EmptyStatus()` with a bit more comment?

e2e/nomostest/testresourcegroup/resourcegroup.go
karlkfi: If you want to keep it, it needs a verb in the name, like ValidateNoControllerPodRestarts maybe.

e2e/testcases/kcc_test.go
karlkfi: Still got deepcopy used in a dozen other places in this file. It's probably not a big deal in terms of test time, but it's not very efficient. I'd rather get in the habit of avoiding deepcopy wherever possible.

e2e/testcases/kcc_test.go
karlkfi: can you make these tests consistent? either always delete at the end or delete in cleanup.

sdowell: Removed the delete from the test(s) which aren't doing any validation after deleting the ResourceGroup.

FWIW I'm trying to stay as close as possible to the spirit of the tests in our upstream repo, so that with this initial commit we can easily map the test coverage to upstream.

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: Done.

e2e/nomostest/testresourcegroup/resourcegroup.go
sdowell: Done.

e2e/testcases/kcc_test.go
sdowell: IMO this is such a low cost thing, particularly for a test case, that it's not worth the mental bandwidth optimizing for. But I removed the DeepCopy calls.

karlkfi (approved):
/lgtm

e2e/testcases/kcc_test.go
karlkfi: Engineering is full of conflicting motivations, opinions, and priorities.

If you think something I asked for is a waste of time feel free to push back or ask me to do it in a follow up. Just because I asked for it doesn't necessarily mean I feel strongly about it. 

I'm trying to keep project quality high but if it annoys everyone in the process then it isn't worth the cost.

e2e/testcases/kcc_test.go
karlkfi: Agree that it's not really worth optimizing for the speed alone. The other reason I should have mentioned was that it confused me when I read it and took me a while to understand. It's effectively a workaround to get around a gotcha that you have to know about in order to understand why it was done this way. So in this case not using deepcopy also helps improve readability. 

So thanks for cleaning it up.

PR #999, opened 2023-11-06
docs: move testing to its own doc page

sdowell:
The testing instructions are fairly verbose/complex, so moving them to a dedicated page should help make the development doc faster to digest.


droot:
Thanks @sdowell

PR #1000, opened 2023-11-06
docs: update RSync RBAC design doc for RootSync

sdowell:
After further discussion, extending this API for RepoSync would be a breaking change to the delegated control pattern. This change updates the design doc to reduce the scope of the API to RootSync objects.


janetkuo (approved):
/lgtm

PR #1001, opened 2023-11-07
Implement known_host ssh support

victorpras:
Currently config sync is not supporting known_host configuration for git connection over ssh. With this PR, we are updating the controller logic to check for `known_host` key in the secret. Also added e2e test for Root and Repo sync that demonstrates the end to end process of this feature.

go/cs-known-hosts

b/309005641

### DIFF
--- a/e2e/testcases/private_cert_secret_test.go
+++ b/e2e/testcases/private_cert_secret_test.go
@@ -57,6 +57,10 @@ func secretDataPatch(key, value string) string {
 	return fmt.Sprintf(`{"data": {"%s": "%s"}}`, key, value64)
 }
 
+func secretDataDeletePatch(key string) string {
+	return fmt.Sprintf(`{"data": {"%s": null}}`, key)
+}
+
 func TestCACertSecretRefV1Alpha1(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipNonLocalGitProvider,
 		ntopts.NamespaceRepo(backendNamespace, configsync.RepoSyncName))
--- a/e2e/testcases/ssh_known_hosts_test.go
+++ b/e2e/testcases/ssh_known_hosts_test.go
@@ -1,0 +1,221 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/policy"
+	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/e2e/nomostest/testpredicates"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
+	"kpt.dev/configsync/pkg/status"
+	"kpt.dev/configsync/pkg/testing/fake"
+)
+
+func TestRootSyncSSHKnownHost(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipNonLocalGitProvider, ntopts.Unstructured)
+	var err error
+	rootSecret := &corev1.Secret{}
+	nt.T.Cleanup(func() {
+		nt.MustMergePatch(rootSecret, secretDataDeletePatch(controllers.KnownHostsKey))
+		err = nt.Watcher.WatchObject(kinds.Deployment(),
+			core.RootReconcilerPrefix, configsync.ControllerNamespace,
+			[]testpredicates.Predicate{
+				testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "false"),
+			},
+		)
+		if err != nil {
+			nt.T.Error(err)
+		}
+	})
+
+	// validate secret and deployment initial stage
+	err = nt.Validate(nomostest.RootAuthSecretName, configsync.ControllerNamespace, rootSecret)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	err = nt.Watcher.WatchObject(kinds.Deployment(),
+		core.RootReconcilerPrefix, configsync.ControllerNamespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "false"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// get known host key value
+	knownHostValue, err := nomostest.GetKnownHosts(nt)
+	if err != nil {
+		nt.T.Fatalf("error: %s", err)
+	}
+
+	// apply known host key and validate
+	nt.MustMergePatch(rootSecret, secretDataPatch(controllers.KnownHostsKey, knownHostValue))
+	err = nt.Watcher.WatchObject(kinds.Deployment(),
+		core.RootReconcilerPrefix, configsync.ControllerNamespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "true"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// try syncing resource and validate
+	cmName := "configmap-test"
+	cmPath := "acme/configmap.yaml"
+	cm := fake.ConfigMapObject(core.Name(cmName))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add(cmPath, cm))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Adding test ConfigMap"))
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+	err = nt.Validate(cmName, "default", &corev1.ConfigMap{})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Remove(cmPath))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Removing test ConfigMap"))
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+	if err := nt.ValidateNotFound(cmName, "default", &corev1.ConfigMap{}); err != nil {
+		nt.T.Fatalf("error: %s", err)
+	}
+
+	// validate root sync error using invalid known host value
+	knownHostValue = "invalid value"
+	nt.MustMergePatch(rootSecret, secretDataPatch(controllers.KnownHostsKey, knownHostValue))
+	err = nt.Watcher.WatchObject(kinds.Deployment(),
+		core.RootReconcilerPrefix, configsync.ControllerNamespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "true"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	err = nt.Watcher.WatchObject(kinds.RootSyncV1Beta1(), configsync.RootSyncName, configsync.ControllerNamespace, []testpredicates.Predicate{
+		testpredicates.RootSyncHasSourceError(status.SourceErrorCode, "No ED25519 host key is known"),
+	})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+}
+
+func TestRepoSyncSSHKnownHost(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipNonLocalGitProvider, ntopts.Unstructured,
+		ntopts.NamespaceRepo(backendNamespace, configsync.RepoSyncName), ntopts.WithDelegatedControl,
+		ntopts.RepoSyncPermissions(policy.AppsAdmin(), policy.CoreAdmin()))
+	var err error
+	repoSecret := &corev1.Secret{}
+	repoReconcilerName := core.NsReconcilerName(backendNamespace, configsync.RepoSyncName)
+	repoSyncNN := nomostest.RepoSyncNN(backendNamespace, configsync.RepoSyncName)
+	nt.T.Cleanup(func() {
+		nt.MustMergePatch(repoSecret, secretDataDeletePatch(controllers.KnownHostsKey))
+		err = nt.Watcher.WatchObject(kinds.Deployment(),
+			repoReconcilerName, configsync.ControllerNamespace,
+			[]testpredicates.Predicate{
+				testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "false"),
+			},
+		)
+		if err != nil {
+			nt.T.Error(err)
+		}
+	})
+
+	// validate secret and deployment initial stage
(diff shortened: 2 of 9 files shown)


e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
+)
+
+func TestSSHKnownHost(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipNonLocalGitProvider, ntopts.NamespaceRepo(backendNamespace, configsync.RepoSyncName))
+	knownHostValue := "dir"
```
sdowell, line 36: I'd like to see a real value being used here based on the ssh keys generated for the in-cluster git server. That way we can actually validate that syncing is working as expected. The current test assertions validate that environment variables are being set but don't actually test the full integration and ensure syncing is behaving as expected.

There's also value in including an incorrect value like this one, with an assertion that git-sync errors and an error status is surfaced on the RSync

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.MustMergePatch(rootSecret, secretDataPatch(controllers.KnownHostsKey, knownHostValue))
+
+	// validate known_hosts key in secret
+	require.NoError(nt.T,
```
sdowell, line 59: This seems a bit overkill, I think we can safely assume that if the patch was successful that the Secret was updated. There's no reconciliation happening here that we should need to verify, since the test it patching the Secret directly.

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,106 @@
...
+		core.RootReconcilerPrefix, v1.NSConfigManagementSystem,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "false"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.MustMergePatch(rootSecret, secretDataPatch(controllers.KnownHostsKey, knownHostValue))
```
sdowell, line 56: The test should remove the key/value in a Cleanup block

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -1169,3 +1170,20 @@ func enableAskpassSidecar(sourceType string, auth configsync.AuthType) bool {
 	}
 	return false
 }
+
+func (r *RepoSyncReconciler) getSecretKnownHost(ctx context.Context, repoSync *v1beta1.RepoSync) bool {
```
sdowell, line 1174: A couple suggestions here:
- The logic is the same for RootSync and RepoSync, so consolidate this to a single method of `reconcilerBase`
- The Secret is already fetched earlier in the reconcile loop during the Secret validation step. I suggest refactoring this so that we only fetch the Secret once

e2e/testcases/ssh_known_hosts_test.go
sdowell: You can actually see there error in the presubmit came from this:
```
            - code: "2004"
              errorMessage: |-
                KNV2004: error in the git-sync container: {"Msg":"error syncing repo, will retry","Err":"Run(git ls-remote -q git@test-git-server.config-management-system-test:/git-server/repos/config-management-system/root-sync main main^{}): exit status 128: { stdout: \"\", stderr: \"No ED25519 host key is known for test-git-server.config-management-system-test and you have requested strict checking.\\r\\nHost key verification failed.\\r\\nfatal: Could not read from remote repository.\\n\\nPlease make sure you have the correct access rights\\nand the repository exists.\" }","Args":{"failCount":24}}
        
                For more information, see https://g.co/cloud/acm-errors#knv2004
```
Because the cluster is shared, and the subsequent test that reuses this cluster still has the Secret set with the fake known_hosts value.

Side note: this is an error that we should have an assertion for in this test case to verify that git-sync will reject a host whose key does not match the specified known_hosts value

e2e/testcases/ssh_known_hosts_test.go
victorpras: sounds good, this part of validation has been removed

pkg/reconcilermanager/controllers/reposync_controller.go
victorpras: Refactored to:
- combined both r*sync to use reconciler base
- only one getSecret call with the temp data stored in reconcilerBase

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,119 @@
...
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "true"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	err = nt.Watcher.WatchObject(kinds.RepoSyncV1Beta1(), configsync.RepoSyncName, backendNamespace, []testpredicates.Predicate{
+		testpredicates.RepoSyncHasSourceError("2004", "No ED25519 host key is known"),
```
tiffanny29631, line 109: status.SourceErrorCode

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Cleanup(func() {
```
tiffanny29631, line 115: nit: optional: Cleanup is usually at the beginning of test

e2e/testcases/private_cert_secret_test.go
```diff
@@ -57,6 +57,10 @@ func secretDataPatch(key, value string) string {
 	return fmt.Sprintf(`{"data": {"%s": "%s"}}`, key, value64)
 }
 
+func secretDataDelete(key string) string {
```
tiffanny29631, line 60: Can add comment or `patch` to the name to reduce confusion of the function.

e2e/testcases/ssh_known_hosts_test.go
sdowell: not optional 😄 

the cleanup should be registered at the same time that the resources is created/updated

e2e/testcases/ssh_known_hosts_test.go
victorpras: as we discussed offline, `GetKnownHost` func is now used to construct a real value being used for known host key. 

Invalid value scenario have been added and properly validated the error message

e2e/testcases/ssh_known_hosts_test.go
victorpras: done!

e2e/testcases/ssh_known_hosts_test.go
victorpras: moved cleanup to the beginning

e2e/testcases/private_cert_secret_test.go
victorpras: changed name to `secretDataDeletePatch`

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,119 @@
...
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// repo sync test
```
sdowell, line 83: This test currently covers the important assertions, but spreads them out across RootSync and RepoSync. I think it would be ideal to get full assertion coverage for both RootSync and RepoSync. For this it may be preferable to split this out into two tests - one for RootSync and one for RepoSync.

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	err = nt.Watcher.WatchObject(kinds.RootSyncV1Beta1(), configsync.RootSyncName, configsync.ControllerNamespace, []testpredicates.Predicate{
+		testpredicates.RootSyncHasSourceError(status.SourceErrorCode, "No ED25519 host key is known"),
+	})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+}
+
+func TestRepoReconcilerSSHKnownHost(t *testing.T) {
```
sdowell, line 117: nit: we typically refer to RepoSync reconcilers as "namespace reconcilers" rather than "repo reconcilers"

Perhaps rename to `TestNsReconcilerSSHKnownHost`, although I think I prefer `TestRepoSyncSSHKnownHost`. If you opt for the latter, also rename the other test to `TestRootSyncSSHKnownHost` for consistency

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"kpt.dev/configsync/pkg/status"
+)
+
+func TestRootReconcilerSSHKnownHost(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.SkipNonLocalGitProvider)
+	var err error
+	rootSecret := &corev1.Secret{}
+	nt.T.Cleanup(func() {
+		nt.MustMergePatch(rootSecret, secretDataDeletePatch(controllers.KnownHostsKey))
+		nomostest.DeletePodByLabel(nt, "app", reconcilermanager.ManagerName, true)
```
sdowell, line 42: Why does the reconciler-manager Pod need to be deleted?

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,204 @@
...
+		repoReconcilerName, configsync.ControllerNamespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "false"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// validate repo sync error using invalid known host value
```
sdowell, line 145: Suggest ordering the steps the same way for both tests, so that it is easier to follow when reading each test case

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,204 @@
...
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "true"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// try syncing resource and validate
+	deploymentPath := nomostest.StructuredNSPath(backendNamespace, "deployment-helloworld")
+	deployment := loadDeployment(nt, "../testdata/deployment-helloworld.yaml")
```
sdowell, line 184: nits:
- use unstructured
- just use a simple inline object, such as a basic ConfigMap. We don't really care what sort of object is synced here, so it's preferable to use something simple/inline. It's harder to read if you need to reference another file to see the object

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,119 @@
...
+		core.RootReconcilerPrefix, configsync.ControllerNamespace,
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "true"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	if err := nt.WatchForAllSyncs(); err != nil {
```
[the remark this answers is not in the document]
victorpras: the test has been changed to commit ConfigMap resource and validate it.

e2e/testcases/ssh_known_hosts_test.go
victorpras: the test has been split for rootsync and reposync and both will cover the full assertion

e2e/testcases/ssh_known_hosts_test.go
```diff
@@ -0,0 +1,204 @@
...
+		[]testpredicates.Predicate{
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "true"),
+		},
+	)
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	// try syncing resource and validate
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Copy(fmt.Sprintf("%s/accounting-namespace.yaml", yamlDir), "acme/namespaces/accounting/namespace.yaml"))
```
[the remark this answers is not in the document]
victorpras: added `ntopts.Unstructured`

e2e/testcases/ssh_known_hosts_test.go
victorpras: renamed to `TestRootSyncSSHKnownHost` and `TestRepoSyncSSHKnownHost`

e2e/testcases/ssh_known_hosts_test.go
victorpras: from my observation, the root deployment isn't always updated after running `nt.MustMergePatch` in cleanup, which cause flakiness to the next running test.

e2e/testcases/ssh_known_hosts_test.go
victorpras: both tests have been streamlined to follow the same steps and ordering

e2e/testcases/ssh_known_hosts_test.go
victorpras: changed both test use and validate config map instead

e2e/testcases/ssh_known_hosts_test.go
sdowell: Hmm, I'm curious as to why that is. I expected the following events to happen:
- An update to the Secret triggers a watch event for reconciler-manager
- The Secret is mapped to the RootSync, triggering a reconcile
- During the reconcile loop, the known_hosts environment variable for the Deployment is set to false
- Changing the environment variable mutates the Deployment, which causes the Pod to be restarted

I wonder if one of these assumptions are incorrect.

Could you try replacing this with a wait to make ensure the Deployment is updated before proceeding?
```golang
	err = nt.Watcher.WatchObject(kinds.Deployment(),
		core.RootReconcilerPrefix, configsync.ControllerNamespace,
		[]testpredicates.Predicate{
			testpredicates.DeploymentHasEnvVar(reconcilermanager.GitSync, controllers.GitSyncKnownHosts, "false"),
		},
	)
	if err != nil {
		nt.T.Error(err)
	}
```

victorpras: sorry looks like this was an oversight on my part. The controller in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1001/files#diff-87f46982b7989d2ed226f9f1007d2327aae710a78e09c68bb4575ba18f0e340cR809 was previously only setting to true without any way of returning back to false (except by restarting the reconciler manager). Should be fixed now

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -934,6 +936,13 @@ func (r *RepoSyncReconciler) validateNamespaceSecret(ctx context.Context, repoSy
 		}
 		return errors.Wrapf(err, "Secret %s get failed", namespaceSecretName)
 	}
+
+	if _, ok := secret.Data[KnownHostsKey]; ok {
```
sdowell, line 940: nit:
```golang
_, r.knownHostExist = secret.Data[KnownHostsKey]
```

sdowell (commented):
LGTM, please just consolidate the commits and add some more context to the commit message

pkg/reconcilermanager/controllers/gitsync_env.go
```diff
@@ -163,11 +167,18 @@ func gitSyncEnvs(_ context.Context, opts options) []corev1.EnvVar {
...
-		Name:  gitSyncKnownHosts,
-		Value: "false",
-	})
+
+	if opts.knownHost {
+		result = append(result, corev1.EnvVar{
+			Name:  GitSyncKnownHosts,
+			Value: "true",
+		})
+	} else {
```
sdowell, line 176: Perhaps add a comment here for future readers:
```
// This must default to false for backwards compatibility
```

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -671,3 +672,10 @@ func (r *reconcilerBase) updateRBACBinding(ctx context.Context, reconcilerRef ty
 		logFieldObjectKind, binding.GetObjectKind().GroupVersionKind().Kind)
 	return nil
 }
+
+func (r *reconcilerBase) getSecretKnownHost(auth configsync.AuthType) bool {
```
sdowell, line 676: nit: `isKnownHostsEnabled`

pkg/reconcilermanager/controllers/gitsync_env.go
victorpras: added

pkg/reconcilermanager/controllers/reconciler_base.go
victorpras: changed func name

sdowell (approved):

PR #1009, opened 2023-11-08
test: fix name pattern of kcc and stress test

sdowell:
These recently added test cases for ResourceGroup do not match the expected pattern for KCC and stress tests. This change will make it so that the periodic jobs properly run these tests.

### DIFF
--- a/e2e/testcases/kcc_test.go
+++ b/e2e/testcases/kcc_test.go
@@ -135,7 +135,7 @@ func TestKCCResourcesOnCSR(t *testing.T) {
 	}
 }
 
-func TestResourceGroupKCC(t *testing.T) {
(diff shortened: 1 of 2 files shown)


karlkfi (approved):
/lgtm

I didn't know these has naming requirements in addition to the nt option.

sdowell:
> /lgtm
> 
> I didn't know these has naming requirements in addition to the nt option.

I wasn't either, it's a bit of a gotcha. Seems a bit redundant IMO, I think we can probably remove the naming requirement https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml#L329

PR #1011, opened 2023-11-08
use go install for golangci-lint

sdowell:
This simplifies the tooling for the go linter and makes it more consistent with our similar tooling. This also removes the reliance on wget to fetch the binary.


nan-yu (approved):
/lgtm

PR #1012, opened 2023-11-09
Optimize git/oci/helm container resources

karlkfi:
- Reduce Memory and increase CPU to match usage under load, as observed durring stress tests.
- Adjust standard requests to half of new autopilot limits. Boost is allowed on non-autopilot clusters, because no limit is set.

Profiling:
- git-sync: go/config-sync-reconciler-profiling
- helm-sync & oci-sync: go/config-sync-reconciler-profiling-2


pkg/reconcilermanager/controllers/reconciler_container_resources.go
```diff
@@ -91,17 +91,17 @@ func ReconcilerContainerResourceDefaultsForAutopilot() map[string]v1beta1.Contai
...
 		reconcilermanager.HelmSync: {
 			ContainerName: reconcilermanager.HelmSync,
-			CPURequest:    resource.MustParse("100m"),
-			CPULimit:      resource.MustParse("100m"),
-			MemoryRequest: resource.MustParse("256Mi"),
-			MemoryLimit:   resource.MustParse("256Mi"),
+			CPURequest:    resource.MustParse("50m"),
+			CPULimit:      resource.MustParse("50m"),
+			MemoryRequest: resource.MustParse("64Mi"),
+			MemoryLimit:   resource.MustParse("64Mi"),
```
sdowell, line 104: I'm fairly confident this is going to cause OOMKill on autopilot, given that we were getting sporadic OOMKill in our e2e tests when the request/limit was set to `200Mi`. As evidence you can see the current request/limit for [`v1.16` is 200Mi](https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.16/manifests/templates/reconciler-manager-configmap.yaml#L182), and observe the [autopilot flakiness for that branch](https://testgrid.k8s.io/googleoss-kpt-config-sync-release#autopilot-regular)

karlkfi: Thanks. Yeah, the called that out as a risk in the profiling doc. The stress test I used for profiling used for loops to generate 5k objects, but otherwise doesn't necessarily represent all the possible ways to consume a lot of memory with helm.

It looks like TestHelmVersionRange is failing because `helm show chart` is being killed. And TestHelmConfigMapOverride is failing from `helm template` being killed.

I'll run those with profiling and see what the metrics say.

karlkfi: I have added profiling for these tests to the proposal and modified the proposed values. You are right, they are much higher for helm. [go/config-sync-reconciler-profiling-2](http://goto.google.com/config-sync-reconciler-profiling-2)

janetkuo: `helm show chart` is known to take up more resources (especially) and there should be override in the test for that. It's also documented here that version range users need to add more cpu via override: https://cloud.google.com/anthos-config-management/docs/reference/rootsync-reposync-fields#configuration_for_the_helm_repository

karlkfi: Do you not want me to increase the default to handle that case?

janetkuo: No, I just want to make sure that we don't bump the default just for version range because that one is spiky and not used by all users of helm source type

janetkuo (approved):
/lgtm

PR #1013, opened 2023-11-09
Add the scope marker to ConfigManagement CRDs

nan-yu:
This commit configures the scope and the apiVersion of those CRDs. If the scope is not set, the scope of those CRD manifests will be set to Namespaced after regeneration.
Setting the apiVersion is for preparation of moving some types into a separate package.

### DIFF
--- a/pkg/api/configmanagement/v1/types.go
+++ b/pkg/api/configmanagement/v1/types.go
@@ -28,6 +28,7 @@ import (
 // +genclient
 // +genclient:nonNamespaced
 // +kubebuilder:object:root=true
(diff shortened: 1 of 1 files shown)


karlkfi:
So when you generated them after this config, there was no diff?

nan-yu:
> So when you generated them after this config, there was no diff?

There are some diff, but only reorder and expansion: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1014#issuecomment-1804678500

sdowell (approved):

PR #1010, opened 2023-11-08
import ResourceGroup api and controller code

sdowell:
This change migrates the ResourceGroup API and controller code from the
upstream repository into this repository. The CRD is now built alongside
the ConfigSync CRDs and the controller image is built alongside Config
Sync images. This is intended to simplify the management and release
lifecycle of Config Sync and its subcomponents.

PR #1014, opened 2023-11-09
Update the configmanagement v1 CRDs to match the generated manifests

nan-yu:
The configmanagement v1 CRDs haven't been updated quite a while. There is no make target to automatically update them.

This commit updates them to match the generated version. The change includes field reordering and expansion.

It also removes the namespace-config-crd.yaml because it is not used by the RootSync/RepoSync API.


karlkfi:
Are there any breaking changes here?

manifests/cluster-selector-crd.yaml
sdowell, line 1: Are these built from the [apis in this repo](https://github.com/GoogleContainerTools/kpt-config-sync/tree/main/pkg/api/configmanagement/v1)? If so, it seems like it might be worth adding these to the presubmit (for example, `make configsync-crds`) to make sure that the CRDs stay up-to-date.

nan-yu: Yes, I'm working on another PR to add them to the presubmit, but it requires significant changes. I'll push up a new PR later.

nan-yu:
> Are there any breaking changes here?

No breaking changes. It only reorders some fields, and expand the `metav1.TypeMeta` and `metav1.ObjectMeta`. There will be a new field added later, that's why I want to make a separate PR for non-breaking changes.

manifests/cluster-selector-crd.yaml
sdowell: I feel it would make sense to start with presubmits, so that it becomes easier to follow/verify subsequent changes.

I expect it would look similar to the changes which add the [ResourceGroup CRD to the build](https://github.com/GoogleContainerTools/kpt-config-sync/pull/1010/files#diff-c67d460364b26335c0851c49105ac4b3150869f0f234059580dd875d8e12d0a1). What significant changes are required for adding these to presubmit?

nan-yu: The controller-gen failed because of the `ResourceGVK schema.GroupVersionKind` field (https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.17.0-rc.1/pkg/api/configmanagement/v1/types.go#L525) used in some legacy APIs:
```
/go/src/kpt.dev/configsync/vendor/k8s.io/apimachinery/pkg/runtime/schema/group_version.go:143:2: encountered struct field "Group" without JSON tag in type "GroupVersionKind"
```

To fix that, I splitted those configmanagement/v1 types into 2 separate directories: `monorepo` and `common`. In the presubmit, we only generate CRDs for the `common` package. This way we can bypass the error.

Another option is to update the legacy field to be `inline`, but that sounds like a breaking change.

nan-yu: I also thought about moving the common types into a separate go file, but `controller-gen` only takes a package as the input path, not a go file.
It also doesn't support skipping types in the package.

sdowell: Ack. Would you prefer to submit this first in that case?

nan-yu: If I submit the first, the package split change will be mixed with the CRD manifests change. I thought it would be hard to review.

nan-yu: I'm still trying to fix the client-gen package path. Let me upload the PR when it is done. Then we can decide which PR should go first.

The intent to split into 3 PRs is to make the review easier.

manifests/cluster-selector-crd.yaml
```diff
@@ -18,36 +18,94 @@
...
-          - spec
-          properties:
-            spec:
-              # ClusterSelectorSpec
-              type: object
-              required:
-              - selector
-              properties:
-                selector:
-                  type: object # metav1.LabelSelector
```
sdowell, line 51: I'm a bit confused by the previous CRD. What is the structure of the previous object? It doesn't look like it defines the same fields as the current API

manifests/cluster-selector-crd.yaml
nan-yu: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1015 
^ is the large PR that makes the CRD manifests generation work.

#1015 includes changes in #1014 and #1013.

manifests/cluster-selector-crd.yaml
nan-yu: The previous CRD declares `x-kubernetes-preserve-unknown-fields: true` for this field, which doesn't expand the nested fields.

I don't know how was it initially generated. My best guess would be the CRD used to be v1beta1, and is now changed to v1/

janetkuo (approved):
/approve

PR #1015, opened 2023-11-09
Separate the monorepo-only API from the common ConfigManagement API

nan-yu:
The NamespaceSelector CRD needs to be updated to support a new `mode` field, which allows selecting resources matching on-cluster Namespaces.

To update the manifest, we need to regenerate the CRDs under pkg/api/configmanagement/v1 package. However, the controller-gen failed because the `ResourceGVK schema.GroupVersionKind` field (https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.17.0-rc.1/pkg/api/configmanagement/v1/types.go#L525) used in some legacy APIs:
```
/go/src/kpt.dev/configsync/vendor/k8s.io/apimachinery/pkg/runtime/schema/group_version.go:143:2: encountered struct field "Group" without JSON tag in type "GroupVersionKind"
```

There is no way to skip those failing types, and the path has to be a package.

Some legacy CRD manifests are not used by the multirepo mode, so this commit moves the monorepo apis into a separate package.

The `configsync-crds` is updated to only generate CRD manifests for the configmanagement/v1 package, but not monorepo/v1 package.


pkg/hydrate/flatten.go
```diff
@@ -16,7 +16,7 @@ package hydrate
 
 import (
 	corev1 "k8s.io/api/core/v1"
-	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
```
sdowell, line 19: The function in this file is unused (aside from unit test) and can be removed

pkg/reconcilermanager/controllers/secret_test.go
```diff
@@ -48,7 +47,7 @@ const (
 )
 
 var nsReconcilerKey = types.NamespacedName{
-	Namespace: v1.NSConfigManagementSystem,
+	Namespace: configsync.ControllerNamespace,
```
sdowell, line 50: I think all of these changes which switch out this constant can be moved to another PR for easy review

clientgen/informer/monorepo/v1/sync.go
sdowell, line 1: What's the general use for the monorepo types in this repository now? Is it for nomos CLI?

nan-yu: Yes, only for nomos CLI.

pkg/reconcilermanager/controllers/secret_test.go
nan-yu: Thought about that. Sure. Let me do another PR for this change.

pkg/hydrate/flatten.go
nan-yu: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1016

pkg/reconcilermanager/controllers/secret_test.go
nan-yu: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1017

nan-yu:
/hold
This PR should be rebased on #1013 and #1014

After #1014 is merged, this should not include any CRD manifests change.

pkg/syncer/differ/namespace.go
```diff
@@ -15,18 +15,18 @@
 package differ
 
 import (
-	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
-	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
+	v12 "kpt.dev/configsync/pkg/api/monorepo/v1"
```
sdowell, line 20: nit: monorepov1

pkg/testing/fake/objects.go
```diff
@@ -28,16 +28,17 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	clusterregistry "k8s.io/cluster-registry/pkg/apis/clusterregistry/v1alpha1"
-	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
+	v12 "kpt.dev/configsync/pkg/api/configmanagement/v1"
```
sdowell, line 31: nit: configmanagementv1 and monorepov1

karlkfi:
I don't think there's any great need to preserve reverse compatibility in the CRD struct itself, just the CRD yaml. Could we instead just replace the ResourceGVK type with a different type that has json tags and convert it to a schema.GroupVersionKind as needed?

nan-yu:
> I don't think there's any great need to preserve reverse compatibility in the CRD struct itself, just the CRD yaml. Could we instead just replace the ResourceGVK type with a different type that has json tags and convert it to a schema.GroupVersionKind as needed?
> 
> A quick search shows ResourceGVK is only used in one place that isn't generated. So it should be a tiny change.

If backward compatibility is not a concern, sure, we can replace with a custom GVK with json tags.

nan-yu:
replaced by #1027

PR #1018, opened 2023-11-10
test: wait for test dependencies asyncronously

karlkfi:
Apply config sync, prometheus, and config connector async. This should speed up test setup, since some of these take a minutes to reconcile.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -28,6 +28,7 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	testmetrics "kpt.dev/configsync/e2e/nomostest/metrics"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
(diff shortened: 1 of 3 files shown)


sdowell (approved):

e2e/nomostest/new.go
```diff
@@ -369,11 +366,19 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		}
 	})
 
-	if err := installConfigSync(nt, opts.Nomos); err != nil {
-		nt.T.Fatal(err)
+	tg := taskgroup.New()
```
sdowell, line 369: Might be able to install the git server in parallel as well

sdowell: /hold

I think this might be worth adding

karlkfi: turns out it's a PITA. I need to rewrite all the git setup func to return error instead of fatal.

sdowell: That's fine, this is a good improvement so we can add git server as a followup

/unhold

karlkfi: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1020

PR #1019, opened 2023-11-10
ci: create junit report for unit tests

sdowell:
This generalizes the junit wrapper so that it can be ran for both the
e2e tests and the unit tests in a similar way. This is intended to make
it easier to digest test results in the prow UI.

Based on https://github.com/GoogleContainerTools/kpt-config-sync/pull/997 for the common installation path of the local junit-report binary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ TEST_INFRA_REGISTRY ?= $(LOCATION)-docker.pkg.dev/$(TEST_INFRA_PROJECT)/test-inf
 # When upgrading this tag, the image will be rebuilt locally during presubmits.
 # After the change is submitted, a postsubmit job will publish the new tag.
 # There is no need to manually publish this image.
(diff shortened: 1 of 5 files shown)


sdowell:
Doesn't look like the report is getting rendered quite right, need to do a bit more digging

/hold

sdowell:
Looks like the issue was that you need to pass the -v flag to `go test` to get the right output format.

/retest
/unhold

scripts/test-unit.sh
```diff
@@ -1,12 +1,11 @@
 #!/bin/bash
-#
-# Copyright 2016 The Kubernetes Authors.
```
karlkfi, line 3: Do we actually own this or did we copy it?
If we copied it, we need to retain the copyright header and add our copyright line underneath.

sdowell: There's not much left of this file other than calling other local scripts, so I don't think there's much to be copied. I haven't dug into the history of this file

PR #1021, opened 2023-11-10
fix: add ClusterRoleOverrideSpec to v1alpha

sdowell:
This API was added to v1beta but not v1alpha, and did not have any conversion code added. This API is intended to be replaced by the roleRefs API, but this will fix clientgen in the interim.

Will be replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/991, but this fixes clientgen in the interim


karlkfi (approved):
/lgtm

PR #1022, opened 2023-11-10
ci: enable spellchecking linter

sdowell:


e2e/testcases/main_test.go
```diff
@@ -118,7 +116,6 @@ func main(m *testing.M) int {
 		fmt.Printf("Error setting test.parallel: %v\n", err)
 		return 1
 	}
-	rand.Seed(time.Now().UnixNano())
```
sdowell, line 121: `staticcheck` flagged this as being unnecessary after being bumped to Go `1.21`. See https://pkg.go.dev/math/rand

karlkfi: huh. TIL

karlkfi (approved):
/lgtm

cool cool. most of these were probably mine. O:-)

PR #1023, opened 2023-11-10
chore: enable linting for resourcegroup API

sdowell:


manifests/resourcegroup-crd.yaml
```diff
@@ -58,7 +58,7 @@ spec:
           metadata:
             type: object
           spec:
-            description: spec defines the desired state of ResourceGroup
+            description: ResourceGroupSpec defines the desired state of ResourceGroup
```
sdowell, line 61: I'm not sure whether this is how we want descriptions to read on the CRD. I think the previous descriptions may read better. I'm hoping this will spark some discussion

/assign @karlkfi

karlkfi: pretty sure we have nolint set on other api structs to avoid this.

sdowell: We generally don't - we have other API fields which surface the name of the go struct on the CRD.

But I added some nolint comments here to prevent any updates to the CRD with this change.

PR #1024, opened 2023-11-13
Add a new Namespace Controller

nan-yu:
The Namespace controller watches for Namespace events. If an event is detected, it checks whether the corresponding Namespace was previously selected or will be selected. If it does, it updates a flag to indicate a reconciliation needs to be executed by the reconciler.

### DIFF
--- a/cmd/reconciler/main.go
+++ b/cmd/reconciler/main.go
@@ -100,6 +100,8 @@ var (
 	namespaceStrategy = flag.String(flags.namespaceStrategy, util.EnvString(reconcilermanager.NamespaceStrategy, ""),
 		fmt.Sprintf("Set the namespace strategy for the reconciler. Must be %s or %s. Default: %s.",
 			configsync.NamespaceStrategyImplicit, configsync.NamespaceStrategyExplicit, configsync.NamespaceStrategyImplicit))
+
+	nsControllerEnabled = flag.Bool("ns-controller-enabled", util.EnvBool(reconcilermanager.NSControllerEnabled, false), "")
 )
 
 var flags = struct {
@@ -193,6 +195,7 @@ func main() {
 		ReconcileTimeout:        *reconcileTimeout,
 		APIServerTimeout:        *apiServerTimeout,
 		RenderingEnabled:        *renderingEnabled,
+		NSControllerEnabled:     *nsControllerEnabled,
 	}
 
 	if declared.Scope(*scope) == declared.RootReconciler {
--- a/pkg/parse/run.go
+++ b/pkg/parse/run.go
@@ -29,6 +29,7 @@ import (
 	"kpt.dev/configsync/pkg/importer/filesystem/cmpath"
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/metrics"
+	"kpt.dev/configsync/pkg/reconciler/namespacecontroller"
 	"kpt.dev/configsync/pkg/status"
 	"kpt.dev/configsync/pkg/util"
 	webhookconfiguration "kpt.dev/configsync/pkg/webhook/configuration"
@@ -41,6 +42,7 @@ const (
 	triggerRetry              = "retry"
 	triggerManagementConflict = "managementConflict"
 	triggerWatchUpdate        = "watchUpdate"
+	namespaceEvent            = "namespaceEvent"
 )
 
 const (
@@ -65,7 +67,7 @@ const (
 )
 
 // Run keeps checking whether a parse-apply-watch loop is necessary and starts a loop if needed.
-func Run(ctx context.Context, p Parser) {
+func Run(ctx context.Context, p Parser, nsControllerState *namespacecontroller.State) {
 	opts := p.options()
 	// Use timers, not tickers.
 	// Tickers can cause memory leaks and continuous execution, when execution
@@ -81,6 +83,10 @@ func Run(ctx context.Context, p Parser) {
 
 	statusUpdateTimer := time.NewTimer(opts.statusUpdatePeriod)
 	defer statusUpdateTimer.Stop()
+
+	nsEventPeriod := time.Second
+	nsEventTimer := time.NewTimer(nsEventPeriod)
+	defer nsEventTimer.Stop()
 
 	state := &reconcilerState{
 		backoff:     defaultBackoff(),
@@ -179,6 +185,26 @@ func Run(ctx context.Context, p Parser) {
 			statusUpdateTimer.Reset(opts.statusUpdatePeriod) // Schedule status update attempt
 			// we should not reset retryTimer under this `case` since it is not aware of the
 			// state of backoff retry.
+
+		// Execute the entire parse-apply-watch loop for a namespace event.
+		case <-nsEventTimer.C:
+			if nsControllerState == nil {
+				// If the Namespace Controller is not running, stop the timer without
+				// closing the channel.
+				nsEventTimer.Stop()
+				continue
+			}
+			if nsControllerState.ScheduleSync() {
+				klog.Infof("A new sync is triggered by a Namespace event")
+				// Reset the cache partially to make sure all the steps of a parse-apply-watch loop will run.
+				// The cached sourceState will not be reset to avoid reading all the source files unnecessarily.
+				// The cached needToRetry will not be reset to avoid resetting the backoff retries.
+				state.resetPartialCache()
+				run(ctx, p, namespaceEvent, state)
+			}
+			// we should not reset retryTimer under this `case` since it is not aware of the
+			// state of backoff retry.
+			nsEventTimer.Reset(nsEventPeriod) // Schedule namespace event check attempt
 		}
 	}
 }
--- a/pkg/reconciler/namespacecontroller/namespace_controller.go
+++ b/pkg/reconciler/namespacecontroller/namespace_controller.go
@@ -1,0 +1,82 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package namespacecontroller
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
+	"kpt.dev/configsync/pkg/status"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// NamespaceController is a controller that watches for Namespace events and
+// sets a flag to indicate whether the Reconciler thread needs to reconcile for
+// this namespace event.
+type NamespaceController struct {
+	client client.Client
+	state  *State
+}
+
+// New instantiates the namespace controller.
+func New(cl client.Client, state *State) *NamespaceController {
+	return &NamespaceController{
+		client: cl,
+		state:  state,
+	}
+}
+
+// Reconcile processes the Namespace request to set the flag to indicate whether
+// a new reconciliation needs to be executed by the Reconciler thread.
+func (nc *NamespaceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ns := &corev1.Namespace{}
+	if err := nc.client.Get(ctx, req.NamespacedName, ns); err != nil {
+		if apierrors.IsNotFound(err) {
+			if nc.state.isSelectedNamespace(req.Name) {
+				klog.Infof("A previously selected Namespace %s is deleted", req.Name)
+				nc.state.setSyncPending()
+			}
+			// Ignore the event of the non-selected namespaces
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, status.APIServerError(err, "failed to get Namespace "+req.Name)
+	}
+
+	if nc.state.matchChanged(ns) {
+		// The Namespace is either previously selected but no longer selected, or it
+		// is not selected but its labels match at least one NamespaceSelector's labels.
+		klog.Infof("The Namespace %s is either selected or unselected", req.Name)
+		nc.state.setSyncPending()
+	}
+	// Ignore the event of the non-selected and non-matching namespaces
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager registers the Namespace Controller.
+func (nc *NamespaceController) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("NamespaceController").
+		Watches(&source.Kind{Type: &corev1.Namespace{}},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(UnmanagedNamespacePredicate{})).
+		Complete(nc)
+
+}
--- a/pkg/reconciler/namespacecontroller/namespace_predicate.go
+++ b/pkg/reconciler/namespacecontroller/namespace_predicate.go
@@ -1,0 +1,76 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package namespacecontroller
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"kpt.dev/configsync/pkg/syncer/differ"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+var _ predicate.Predicate = UnmanagedNamespacePredicate{}
+
+// UnmanagedNamespacePredicate triggers a new reconciliation of the namespace
+// controller when receiving an unmanaged Namespace event.
+type UnmanagedNamespacePredicate struct {
+	predicate.Funcs
+}
+
+// isUnmanagedNamespace returns whether the Namespace is not managed by Config Sync.
+// If it is a managed Namespace, the event should be reconciled by the remediator.
+func isUnmanagedNamespace(obj client.Object) bool {
+	ns, ok := obj.(*corev1.Namespace)
+	if !ok {
+		// Skip the event if it is not a Namespace object.
+		return false
+	}
+	// Skip the event if the Namespace is already managed by CS
+	return !differ.ManagedByConfigSync(ns)
+}
+
+// Create implements default CreateEvent filter for validating namespace with additional labels.
+func (n UnmanagedNamespacePredicate) Create(e event.CreateEvent) bool {
+	return isUnmanagedNamespace(e.Object)
+}
+
+// Delete implements default DeleteEvent filter for validating namespace with additional labels.
+func (n UnmanagedNamespacePredicate) Delete(_ event.DeleteEvent) bool {
+	// It is not reliable to check if the deleted event is managed by Config Sync
+	// based on the last known labels because it may not match the actual last state.
+	// So it always return true to enqueue the request.
+	return true
+}
+
+// Update implements default UpdateEvent filter for validating namespace label change.
+func (n UnmanagedNamespacePredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil {
+		klog.Errorf("update event has no old object to update, event: %v", e)
+		return false
+	}
+	if e.ObjectNew == nil {
+		klog.Errorf("update event has no new object to update, event: %v", e)
+		return false
+	}
+
+	return isUnmanagedNamespace(e.ObjectOld) || isUnmanagedNamespace(e.ObjectNew)
+}
+
+// Generic implements Predicate.
+func (n UnmanagedNamespacePredicate) Generic(e event.GenericEvent) bool {
+	return isUnmanagedNamespace(e.Object)
+}
--- a/pkg/reconciler/namespacecontroller/state.go
+++ b/pkg/reconciler/namespacecontroller/state.go
@@ -1,0 +1,129 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package namespacecontroller
+
+import (
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// State stores the state of the Namespace Controller.
+type State struct {
+	// eventMux is a RMMutex to protect read/write to 'isSyncingPending'.
+	// It is RWMutex because the cache is read more often than written.
+	eventMux sync.RWMutex
+
+	// isSyncPending indicates whether the namespace controller fires up an event that requires a new sync.
+	isSyncPending bool
+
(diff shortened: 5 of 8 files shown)


pkg/parse/run.go
```diff
@@ -120,6 +126,24 @@ func Run(ctx context.Context, p Parser) {
 			// state of backoff retry.
 			statusUpdateTimer.Reset(opts.statusUpdatePeriod) // Schedule status update attempt
 
+		// Execute the entire parse-apply-watch loop for a namespace event.
+		case <-nsEventTimer.C:
+			if nsController == nil {
```
karlkfi, line 131: This is one of those situations where using reflect.Select might be a good idea. If the nsController isn't running/enabled then we don't need the nsEventTimer or this select case at all.

pkg/parse/run.go
```diff
@@ -120,6 +126,24 @@ func Run(ctx context.Context, p Parser) {
...
+				// The Namespace Controller is not running, so no need to reset nsEventTimer.
+				continue
+			} else if nsController.HasNewNSEvent() {
```
karlkfi, line 134: don't need the else if when the previous case continues. you can just make this a new if statement.

pkg/parse/run.go
```diff
@@ -120,6 +126,24 @@ func Run(ctx context.Context, p Parser) {
...
+				// Reset the cache partially to make sure all the steps of a parse-apply-watch loop will run.
+				// The cached sourceState will not be reset to avoid reading all the source files unnecessarily.
+				// The cached needToRetry will not be reset to avoid resetting the backoff retries.
+				state.resetPartialCache()
+				run(ctx, p, namespaceEvent, state)
+			}
+			nsEventTimer.Reset(nsEventPeriod) // Schedule namespace event check attempt
+			// we should not reset retryTimer under this `case` since it is not aware of the
+			// state of backoff retry.
+			statusUpdateTimer.Reset(opts.statusUpdatePeriod) // Schedule status update attempt
```
karlkfi, line 145: If statusUpdatePeriod > nsEventPeriod, this will prevent the status update from ever happening.

pkg/parse/run.go
```diff
@@ -120,6 +126,24 @@ func Run(ctx context.Context, p Parser) {
 			// state of backoff retry.
 			statusUpdateTimer.Reset(opts.statusUpdatePeriod) // Schedule status update attempt
 
+		// Execute the entire parse-apply-watch loop for a namespace event.
+		case <-nsEventTimer.C:
```
karlkfi, line 130: This case should probably be moved to the bottom of the select because it's the shortest timer. Technically, selects don't work top down, but people read them like switch statements, expecting the higher cases to take precedence. So since this is the shortest timer period, it should go last and let the others take precedence, at least for readability.

karlkfi: FWIW, when multiple select channels have messages available, one of them in randomly selected to execute.

karlkfi: ^^ This is another reason this kind of timer based select is not optimal. Ideally these would all be re-written into a reconcile loop with an explicit precedence top to bottom, checking now against timestamps to see if enough time has elapsed to execute one of the cases.

cmd/reconciler/main.go
```diff
@@ -100,6 +100,8 @@ var (
 	namespaceStrategy = flag.String(flags.namespaceStrategy, util.EnvString(reconcilermanager.NamespaceStrategy, ""),
 		fmt.Sprintf("Set the namespace strategy for the reconciler. Must be %s or %s. Default: %s.",
 			configsync.NamespaceStrategyImplicit, configsync.NamespaceStrategyExplicit, configsync.NamespaceStrategyImplicit))
+
+	nsControllerEnabled = flag.Bool("ns-controller-enabled", util.EnvBool(reconcilermanager.NSControllerEnabled, false), "")
```
karlkfi, line 104: did you implement how this envvar/flag gets set? I can't find that code.

pkg/parse/run.go
nan-yu: It sounds like we don't need to reschedule the statusUpdate attempt when a nsEventTimer kicks in.

There may already have the issue that prevents the status update from happening in the retry case: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L163

cmd/reconciler/main.go
nan-yu: It will be in a separate PR.
The envvar flag will be read from the NamespaceSelector mode, and set as RSync annotation. Then reconciler-manager sets the envvar by reading the annotation.

pkg/reconciler/namespacecontroller/namespace_controller.go
```diff
@@ -0,0 +1,144 @@
...
+// sets a flag to indicate whether the Reconciler thread needs to reconcile for
+// this namespace event.
+type NamespaceController struct {
+	client client.Client
+
+	// eventMux is a RMMutex because the cache is read more often than writen.
+	eventMux sync.RWMutex
+
+	// newNSEvent indicates whether the namespace controller fires up a new namespace event.
+	newNSEvent bool
```
karlkfi, line 42: Suggest naming this field and the setters/getters after what it's intended to do rather than what caused it.

Like, maybe `namespacesUpdated` or `resyncRequired`.

pkg/reconciler/namespacecontroller/namespace_controller.go
```diff
@@ -0,0 +1,144 @@
...
+
+// HasNewNSEvent returns whether a new namespace event is detected.
+func (nc *NamespaceController) HasNewNSEvent() bool {
+	nc.eventMux.Lock()
+	defer nc.eventMux.Unlock()
+	return nc.newNSEvent
+}
+
+// SetSelectorCache is invoked by the reconciler to set the cache after parsing the NamespaceSelectors.
+func (nc *NamespaceController) SetSelectorCache(nsSelectorMap map[string]labels.Selector, selectedNamespaces map[string]struct{}) {
```
karlkfi, line 78: never used?

usually you dont want to set a map like this because maps are mutable. probably want to have individual Clear, Add, Remove methods.

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: The field is in the namespace controller, but it is the Reconciler that takes the action. If the field is in the reconciler's cache, yes, it should indicate what it is intended to do.
Looking at how it is invoked, it is `nsController.HasNewNSEvent()` in the Reconciler, which makes sense. The reconciler resyncs when the namespace controller has a new event.
If we rename to resyncRequired, then it looks like `nsController.ResyncRequired`, it sounds like the namespace controller requires a new reconciliation, which seems more confusing.

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: It will be used by the follow-up PR when parsing the NamespaceSelector.

pkg/reconciler/namespacecontroller/namespace_controller.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+// NamespaceController is a controller that watches for Namespace events and
+// sets a flag to indicate whether the Reconciler thread needs to reconcile for
+// this namespace event.
+type NamespaceController struct {
+	client client.Client
+
+	// eventMux is a RMMutex because the cache is read more often than writen.
+	eventMux sync.RWMutex
```
karlkfi, line 39: I'm not loving how messy this looks with the mutexes and protected values on the controller struct. You also shouldn't have to pass the whole controller to the Run method when it never needs to call Reconcile. That way you dont need to build the whole controller to test the Run func. We probably need some intermediate struct  or something.

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: We don't want to have  ADD/Remove/Clear methods because we update the whole map after parsing all NamespaceSelectors. They should be handled atomically.

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: extracted them into a state structure.

pkg/parse/run.go
nan-yu: done

pkg/parse/run.go
nan-yu: If the nsConroller isn't running, we stop the timer.

pkg/parse/run.go
nan-yu: Moved to the bottom

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: Just changes that require resync.

pkg/parse/run.go
```diff
@@ -179,6 +185,26 @@ func Run(ctx context.Context, p Parser) {
...
+			if nsControllerState == nil {
+				// If the Namespace Controller is not running, stop the timer without
+				// closing the channel.
+				nsEventTimer.Stop()
+				continue
+			}
+			if nsControllerState.HasNewNSEvent() {
+				klog.Infof("A new reconciliation is triggered by a Namespace event")
```
karlkfi, line 198: s/reconciliation/sync

yes, a sync is technically a reconciliation, but we've been calling it a sync in this context.

pkg/reconciler/namespacecontroller/namespace_controller.go
```diff
@@ -0,0 +1,80 @@
...
+			// Ignore the event of the non-selected namespaces
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, status.APIServerError(err, "failed to get Namespace "+req.Name)
+	}
+
+	if nc.state.selectedOrMatchSelector(ns) {
+		// The created/updated namespace is either previously selected, or its
+		// labels match at least one NamespaceSelector's labels.
+		nc.state.setNewNSEvent()
```
karlkfi, line 65: It looks like we only check that the labels match the selectors and don't check for any kind of diff we might care about. So ANY change to a matching namespace will trigger sync. We might want to avoid unnecessary syncs by checking if the set of matching namespaces has changed.

The deletion detection has the same problem, but we can probably reduce those with the same method.

These are the use cases, afaik:
1. if it used to be selected and is no longer selected, we need to sync
2. if it was not selected and now is selected, we need to sync
3. is it was selected and is now deleted, we need to sync.

So I think we need to cache the set of namespace that were selected as of the last successful sync. That way we can check if any change to that list has been made with the current state and only then trigger a new sync.

pkg/reconciler/namespacecontroller/namespace_controller.go
karlkfi: how about copying the maps then, just to ensure they can't be changed by the caller?

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: The deletion detection seems to be okay. `apierrors.IsNotFound(err)` and `nc.state.isSelectedNamespace(req.Name)` matches use case 3.
The `selectedNamespaces` caches the set of namespaces that were selected, but may not always get to a successful sync.
We can change the value from an empty struct into the label selector, and detects whether it still matches or not.

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: SG. Done.

pkg/parse/run.go
nan-yu: done

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: updated `selectedOrMatchSelector` to `matchChanged`

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: done

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: It is intended to be updated when parsing the source, and the namespace controller is running.

pkg/reconciler/namespacecontroller/state.go
```diff
@@ -0,0 +1,110 @@
...
+	defer s.selectorMux.Unlock()
+
+	selector, found := s.selectedNamespaces[ns.Name]
+	// The Namespace is previously selected, but not selected anymore.
+	if found && !selector.Matches(labels.Set(ns.GetLabels())) {
+		return true
+	}
+
+	// The Namespace is not selected before, but is now selected.
+	for _, selector = range s.nsSelectors {
```
tiffanny29631, line 104: Does the  !found need to be a criteria for the `not selected before`?

pkg/reconciler/namespacecontroller/state.go
```diff
@@ -0,0 +1,110 @@
...
+
+// IsSyncPending returns whether the state triggers a sync.
+func (s *State) IsSyncPending() bool {
+	s.eventMux.Lock()
+	defer s.eventMux.Unlock()
+	return s.isSyncPending
+}
+
+// SetSelectorCache is invoked by the reconciler to set the cache after parsing the NamespaceSelectors.
+func (s *State) SetSelectorCache(nsSelectors, selectedNamespaces map[string]labels.Selector) {
```
janetkuo, line 66: Where is this used?

pkg/reconciler/namespacecontroller/namespace_controller.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Reconcile processes the Namespace request to set the flag to indicate whether
+// a new reconciliation needs to be executed by the Reconciler thread.
+func (nc *NamespaceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ns := &corev1.Namespace{}
+	if err := nc.client.Get(ctx, req.NamespacedName, ns); err != nil {
+		if apierrors.IsNotFound(err) {
+			if nc.state.isSelectedNamespace(req.Name) {
+				// A previously selected namespace is deleted, so set the namespace event.
+				nc.state.setSyncPending()
```
janetkuo, line 54: Is the cache (`selectedNamespaces`) updated to remove this namespace?

janetkuo (commented):
Would you add some tests?

pkg/parse/run.go
```diff
@@ -179,6 +185,26 @@ func Run(ctx context.Context, p Parser) {
...
+				// If the Namespace Controller is not running, stop the timer without
+				// closing the channel.
+				nsEventTimer.Stop()
+				continue
+			}
+			if nsControllerState.IsSyncPending() {
+				klog.Infof("A new sync is triggered by a Namespace event")
+				// Reset the cache partially to make sure all the steps of a parse-apply-watch loop will run.
+				// The cached sourceState will not be reset to avoid reading all the source files unnecessarily.
+				// The cached needToRetry will not be reset to avoid resetting the backoff retries.
```
karlkfi, line 201: Does this mean that the `run` here will just return and skip applying if we're in backoff?
I can't remember how backoff works.
I think I once expected it to be implemented here with these timers but it wasn't.

pkg/reconciler/namespacecontroller/namespace_predicate.go
```diff
@@ -0,0 +1,90 @@
...
+func canTriggerReconcile(obj client.Object) bool {
+	if configmanagement.IsControllerNamespace(obj.GetName()) {
+		return false // Don't allow managing resources in CS controller namespaces
+	}
+	if differ.ManagedByConfigSync(obj) {
+		return false // Skip the event because the namespace is already managed by CS.
+	}
+	labels := obj.GetLabels()
+	delete(labels, metadataNameLabel)
+	// Skip the event because the namespace only has the `kubernetes.io/metadata.name` label.
```
karlkfi, line 59: Do we not allow using this label for selecting by label?

pkg/reconciler/namespacecontroller/namespace_predicate.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Create implements default CreateEvent filter for validating namespace with additional labels.
+func (n NamespaceLabelChangedPredicate) Create(e event.CreateEvent) bool {
+	return canTriggerReconcile(e.Object)
+}
+
+// Delete implements default DeleteEvent filter for validating namespace with additional labels.
+func (n NamespaceLabelChangedPredicate) Delete(e event.DeleteEvent) bool {
+	return canTriggerReconcile(e.Object)
```
karlkfi, line 70: Delete events are triggered with Object set to the last known state, from the cache, which may or may not be the actual last state, if events were missed or the watch just started. 

It's possible that the Namespace was in the selected namespaces and then the label was removed and then the object was deleted before the reconciler had a chance to process the update event. So we might not want to filter Delete events based on the last known labels.

pkg/reconciler/namespacecontroller/namespace_predicate.go
nan-yu: Using this label is the same as setting the namespace directly because it will only select a single namespace.

I need to change the predicate as we also want to enqueue the namespace with this single label because existing labels might be removed, which should trigger a resync.

pkg/reconciler/namespacecontroller/namespace_predicate.go
```diff
@@ -0,0 +1,90 @@
...
+
+// canTriggerReconcile returns whether the Namespace has labels that can trigger
+// a reconciliation.
+// It can't trigger a reconciliation for the following scenarios:
+//   - If it only contains the `kubernetes.io/metadata.name` label. The Kubernetes
+//     control plane sets the immutable label `kubernetes.io/metadata.name` on all
+//     namespaces, the value is the same as Namespace name. If a Namespace only
+//     has that label, there is no need to use NamespaceSelector. It reduces the
+//     number of reconciliations by eliminating this scenario.
+//   - If it is a CS controller Namespace, it should not trigger a reconciliation
```
karlkfi, line 46: Didn't we recently have a customer issue where they had added config-management-system to their source of truth and it was managing the labels or something?

Is this a hard-coded constraint somewhere?

pkg/reconciler/namespacecontroller/state.go
karlkfi: It's not yet, which means isSelectedNamespace will always be false.

I think this part will need to be implemented to make tests works.

pkg/reconciler/namespacecontroller/state.go
nan-yu: updated

pkg/reconciler/namespacecontroller/state.go
```diff
@@ -0,0 +1,110 @@
...
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// State stores the state of the Namespace Controller.
+type State struct {
+	// eventMux is a RMMutex to protect read/write to 'isSyncingPending'.
+	// It is RMMutex because the cache is read more often than writen.
```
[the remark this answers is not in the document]
nan-yu: done

pkg/reconciler/namespacecontroller/state.go
nan-yu: It will be used after NamespaceSelector selects resources, and will be in a followup PR.

pkg/reconciler/namespacecontroller/namespace_controller.go
nan-yu: Nope. `selectedNamespaces` is only updated by the NamespaceSelector.

pkg/reconciler/namespacecontroller/namespace_predicate.go
nan-yu: Removed the check for the label.

pkg/reconciler/namespacecontroller/namespace_predicate.go
nan-yu: We never allow managing `config-management-system`: https://cloud.google.com/anthos-config-management/docs/reference/errors#knv1034. 

https://github.com/GoogleContainerTools/kpt-config-sync/pull/961 recently extends the constraints to config-management-monitoring and resource-group-controller namespaces. IIRC, the customer issue was adding config-management-monitoring to the SoT.

pkg/reconciler/namespacecontroller/namespace_predicate.go
```diff
@@ -0,0 +1,90 @@
...
+//     because we don't allow managing resources in CS controller namespaces.
+//   - If it is managed by CS, it should not trigger a reconciliation because
```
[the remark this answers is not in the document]
nan-yu: If the namespace is already managed by Config Sync, regardless whether it is managed by this reconciler, or other reconcilers, any update to the managed namespace should be fixed by the remediator. 
We should not trigger a resync event for the drifted state of the managed namespace.

pkg/reconciler/namespacecontroller/namespace_predicate.go
nan-yu: updated to not rely on labels.

pkg/parse/run.go
nan-yu: The backoff is only for retry events. The run function doesn't check  the backoff.

nan-yu:
> Would you add some tests?

Unit tests for the namespace controller state is added, more unit tests and e2e tests will be added when updating NamespaceSelector.

pkg/reconciler/namespacecontroller/namespace_predicate.go
```diff
@@ -0,0 +1,81 @@
...
+// labels are changed, so NamespaceSelector can select a different set of resources.
+type NamespaceLabelChangedPredicate struct {
+	predicate.Funcs
+}
+
+// canTriggerReconcile returns whether the Namespace has labels that can trigger
+// a reconciliation.
+// It can't trigger a reconciliation for the following scenarios:
+//   - If it is a CS controller Namespace, it should not trigger a reconciliation
+//     because we don't allow managing resources in CS controller namespaces.
```
janetkuo, line 38: We still allow managing some resources in the controller ns, such as RootSyncs, Secrets and VPA objects.

pkg/reconciler/namespacecontroller/namespace_predicate.go
```diff
@@ -0,0 +1,81 @@
...
+
+var _ predicate.Predicate = NamespaceLabelChangedPredicate{}
+
+// NamespaceLabelChangedPredicate triggers a new reconciliation when Namespace
+// labels are changed, so NamespaceSelector can select a different set of resources.
+type NamespaceLabelChangedPredicate struct {
+	predicate.Funcs
+}
+
+// canTriggerReconcile returns whether the Namespace has labels that can trigger
```
janetkuo, line 34: Is this function generic for any objects or just for namespaces? Suggest making it more clear, e.g. change the input type, do a type check

pkg/reconciler/namespacecontroller/state.go
```diff
@@ -0,0 +1,123 @@
...
+
+// IsSyncPending returns whether the state triggers a sync.
+func (s *State) IsSyncPending() bool {
+	s.eventMux.Lock()
+	defer s.eventMux.Unlock()
+	isSyncingPending := s.isSyncPending
+	if s.isSyncPending {
+		// Reset the Sync pending state after a sync regardless if succeeds or not.
+		// If it fails, a retry should be triggered.
+		s.isSyncPending = false
```
sdowell, line 66: nit: the function signature `IsSyncPending` gives the impression this is a simple getter style method, when this in reality may mutate state.

Suggest a change in the name to include a verb, for example `StartResync`

pkg/reconciler/namespacecontroller/state.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return isSyncingPending
+}
+
+// SetSelectorCache is invoked by the reconciler to set the cache after parsing the NamespaceSelectors.
+func (s *State) SetSelectorCache(nsSelectors map[string]labels.Selector, selectedNamespaces map[string][]labels.Selector) {
```
sdowell, line 72: This method appears to be unused, or am I missing something? What is the intent?

pkg/reconciler/namespacecontroller/state.go
```diff
@@ -0,0 +1,123 @@
...
+	isSyncPending bool
+
+	// selectorMux is the mutex to protect read/write to 'nsSelectors' and 'selectedNamespaces'.
+	selectorMux sync.Mutex
+
+	// nsSelectors is a map of NamespaceSelector names to the LabelSelectors.
+	nsSelectors map[string]labels.Selector
+
+	// selectedNamespaces records the dynamic/on-cluster Namespaces selected by NamespaceSelectors
+	selectedNamespaces map[string][]labels.Selector
```
sdowell, line 40: Could you explicitly document the values for each map? I assume the keys for this map are `Namespace` names, but it's not totally clear from the comment

pkg/reconciler/namespacecontroller/namespace_predicate.go
nan-yu: Oh, right. The controller namespaces can't be managed, but resources under it can be managed. Removed the check for controller namespaces.

pkg/reconciler/namespacecontroller/namespace_predicate.go
nan-yu: The predicate is only used for Namespace watch, but it doesn't hurt to have a type check.
done

pkg/reconciler/namespacecontroller/state.go
nan-yu: Renamed to `ScheduleSync`. 

`StartResync` seems to belong to a method of the reconciler thread, not the state.

sdowell: Sounds good, thank you

pkg/reconciler/namespacecontroller/state.go
nan-yu: Added more comments on the key/value.
It is not clear because the it is initialized and updated in a followup PR.

pkg/reconciler/namespacecontroller/state.go
nan-yu: The intent is to keep this PR only for the namespace controller. 

The method will be invoked after parsing NamespaceSelectors, which is a large change and I want to keep it separate and easy for review.

janetkuo (approved):
/approve

PR #1027, opened 2023-11-14
Auto-generate manifests for ConfigManagement CRDs

nan-yu:
The following manifests haven't been updated and are not auto-generated:
- cluster-selector-crd.yaml
- hierachyconfig-crd.yaml
- namespace-selector-crd.yaml

This commit updates the configsync-crd target to include them in the auto-generated manifests.

### DIFF
--- a/Makefile.reconcilermanager
+++ b/Makefile.reconcilermanager
@@ -18,18 +18,32 @@ configsync-crds: install-controller-gen "$(KUSTOMIZE)" "$(GOBIN)/addlicense"
 	$(CONTROLLER_GEN) crd \
 		paths="./pkg/api/configsync/v1alpha1" \
 		paths="./pkg/api/configsync/v1beta1" \
+		paths="./pkg/api/configmanagement/v1" \
 		paths="./pkg/api/kpt.dev/v1alpha1" \
 		output:artifacts:config=./manifests \
 	&& mv ./manifests/configsync.gke.io_reposyncs.yaml ./manifests/patch/reposync-crd.yaml \
 	&& mv ./manifests/configsync.gke.io_rootsyncs.yaml ./manifests/patch/rootsync-crd.yaml \
+	&& mv ./manifests/configmanagement.gke.io_clusterselectors.yaml ./manifests/patch/cluster-selector-crd.yaml \
+	&& mv ./manifests/configmanagement.gke.io_hierarchyconfigs.yaml ./manifests/patch/hierarchyconfig-crd.yaml \
+	&& mv ./manifests/configmanagement.gke.io_namespaceselectors.yaml ./manifests/patch/namespace-selector-crd.yaml \
 	&& mv ./manifests/kpt.dev_resourcegroups.yaml ./manifests/patch/resourcegroup-crd.yaml \
 	&& "$(KUSTOMIZE)" build ./manifests/patch -o ./manifests \
 	&& mv ./manifests/*customresourcedefinition_rootsyncs* ./manifests/rootsync-crd.yaml \
 	&& mv ./manifests/*customresourcedefinition_reposyncs* ./manifests/reposync-crd.yaml \
+	&& mv ./manifests/*customresourcedefinition_clusterselectors* ./manifests/cluster-selector-crd.yaml \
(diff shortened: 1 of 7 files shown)


nan-yu:
Replace #1015 and rebase on #1014

pkg/api/configmanagement/v1/types.go
```diff
@@ -528,7 +528,23 @@ type ErrorResource struct {
 	// ResourceGVK is the GroupVersionKind of the affected K8S resource. This field may be empty for
 	// errors that are not associated with a specific resource.
 	// +optional
-	ResourceGVK schema.GroupVersionKind `json:"resourceGVK"`
+	ResourceGVK GroupVersionKind `json:"resourceGVK"`
```
karlkfi, line 531: I think we may need to convert this to a pointer and add omitempty so it's blank if not set. Otherwise the optional does nothing.

Even if we just add omitempty, it doesn't work on structs that are not pointers, due to a yaml formatter quirk (long standing bug).

karlkfi: If we do make this optional, we need to make sure the frontend parses it correctly when missing.

I don't think we have nay other clients that read the status on the ConfigManagement object, unless nomos cli does.

Makefile.reconcilermanager
```diff
@@ -18,18 +18,32 @@ configsync-crds: install-controller-gen "$(KUSTOMIZE)" "$(GOBIN)/addlicense"
 	$(CONTROLLER_GEN) crd \
 		paths="./pkg/api/configsync/v1alpha1" \
 		paths="./pkg/api/configsync/v1beta1" \
+    paths="./pkg/api/configmanagement/v1" \
 		paths="./pkg/api/kpt.dev/v1alpha1" \
 		output:artifacts:config=./manifests \
 	&& mv ./manifests/configsync.gke.io_reposyncs.yaml ./manifests/patch/reposync-crd.yaml \
 	&& mv ./manifests/configsync.gke.io_rootsyncs.yaml ./manifests/patch/rootsync-crd.yaml \
+  && mv ./manifests/configmanagement.gke.io_clusterselectors.yaml ./manifests/patch/cluster-selector-crd.yaml \
```
karlkfi, line 26: use tabs in makefiles, same as the other lines.

manifests/cluster-selector-crd.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2023 Google LLC
```
karlkfi, line 1: did the generator update this? it shoudn't update the year if the header already exists, but maybe the header doesn't already exist if the code that runs the generator deleted all the generated files first.

nan-yu: Yes, same as other files. The generate recreates these manifests.

pkg/api/configmanagement/v1/types.go
nan-yu: Nomos CLI doesn't parse ConfigManagement status. It only parses RSync status.

nan-yu: Checking the usage of `ErrorResource`, it is not even used on the ConfigManagement object, but only on legacy mono-repo CRDs: ClusterConfig, NamespaceConfig, Repo, Sync, and etc.

Nomos CLI does parse the Repo's status field, but not `ErrorResource` field specifically. 

It should be safe to convert the field to a pointer, but do we have to make the change for monorepo code?

janetkuo: > It should be safe to convert the field to a pointer, but do we have to make the change for monorepo code?

No, given that's deprecated

janetkuo (commented):
/approve

pkg/api/configmanagement/v1/types.go
karlkfi: Alright cool. I'm fine with this being a known bug in a deprecated API.

karlkfi (approved):
/lgtm

Thanks for cleaning this up!

sdowell (approved):

PR #1029, opened 2023-11-15
API: add a new field to the NamespaceSelector API

nan-yu:
The new field is `mode`. It supports static mode and dynamic mode. `static` mode means it can only select resources matching the Namespaces statically declared in the source of truth.
`dynamic` mode means it includes both statically declared Namespaces and Namespaces present on the cluster.


pkg/api/configmanagement/v1/types.go
```diff
@@ -272,6 +272,15 @@ type NamespaceSelectorSpec struct {
 	// This field is NOT optional and follows standard label selector semantics. An empty selector
 	// matches all namespaces.
 	Selector metav1.LabelSelector `json:"selector"`
+
+	// Mode specifies the selection mode of the NamespaceSelector.
```
sdowell, line 276: use lowercase `mode` for the CRD description

nan-yu: It is the beginning of a sentence. We use uppercase for consistency.

sdowell: My understanding is that we use camelCase for most of the CRD fields (to match the field name), at least for the Config Sync CRDS.

nan-yu: Alright, sounds like inconsistency between configsync CRD and configmanagement CRD descriptions.
Given that this field is to be used by Config Sync users, changed to lowercase.

sdowell: If there's an inconsistency then I don't feel strongly about it. It may be preferable to stay consistent within the CRD, if this CRD is using uppercase.

nan-yu: For this particular CRD, it is okay as the description of the other field doesn't start with the field name.
It looks different from other CRDs in the package, but other CRDs are already inconsistency with configsync CRDs, so I'm okay either way.

This PR changes to lowercase.

PR #1030, opened 2023-11-15
Update the reconciler-manager to set the ns-controller-requires flag

nan-yu:
The reconciler-manager checks the
`configsync.gke.io/requries-ns-controller` annotation in the RSync object and sets the `ns-controller-requires` flag in the reconciler container if the annoation is true.

### DIFF
--- a/pkg/metadata/annotations.go
+++ b/pkg/metadata/annotations.go
@@ -145,6 +145,14 @@ const (
 	// reconciler-manager will create the reconciler with the hydration-controller
 	// sidecar container.
 	RequiresRenderingAnnotationKey = configsync.ConfigSyncPrefix + "requires-rendering"
+
+	// DynamicNSSelectorEnabledAnnotationKey is the annotation key set on R*Sync
+	// object to indicate whether the source of truth contains at least one
+	// NamespaceSelector using the dynamic mode, which requires the Namespace
+	// controller. The reconciler writes the value of this annotation and the
+	// reconciler-manager reads it. If set to true, the reconciler-manager will
+	// create the reconciler with the Namespace controller in the reconciler container.
+	DynamicNSSelectorEnabledAnnotationKey = configsync.ConfigSyncPrefix + "dynamic-ns-selector-enabled"
 )
 
 // Lifecycle annotations
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -851,7 +851,9 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 			statusMode:        rs.Spec.SafeOverride().StatusMode,
 			reconcileTimeout:  v1beta1.GetReconcileTimeout(rs.Spec.SafeOverride().ReconcileTimeout),
 			apiServerTimeout:  v1beta1.GetAPIServerTimeout(rs.Spec.SafeOverride().APIServerTimeout),
-			requiresRendering: enableRendering(rs.GetAnnotations()),
+			requiresRendering: annotationEnabled(metadata.RequiresRenderingAnnotationKey, rs.GetAnnotations()),
+			// Namespace reconciler doesn't support NamespaceSelector at all.
+			dynamicNSSelectorEnabled: false,
 		}),
 	}
 	switch v1beta1.SourceType(rs.Spec.SourceType) {
@@ -1125,7 +1127,7 @@ func (r *RepoSyncReconciler) mutationsFor(ctx context.Context, rs *v1beta1.RepoS
 			case reconcilermanager.Reconciler:
 				container.Env = append(container.Env, containerEnvs[container.Name]...)
 			case reconcilermanager.HydrationController:
-				if !enableRendering(rs.GetAnnotations()) {
+				if !annotationEnabled(metadata.RequiresRenderingAnnotationKey, rs.GetAnnotations()) {
 					// if the sync source does not require rendering, omit the hydration controller
 					// this minimizes the resource footprint of the reconciler
 					addContainer = false
--- a/pkg/reconcilermanager/controllers/rootsync_controller.go
+++ b/pkg/reconcilermanager/controllers/rootsync_controller.go
@@ -687,20 +687,21 @@ func (r *RootSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 		}),
 		reconcilermanager.Reconciler: append(
 			reconcilerEnvs(reconcilerOptions{
-				clusterName:       r.clusterName,
-				syncName:          rs.Name,
-				syncGeneration:    rs.Generation,
-				reconcilerName:    reconcilerName,
-				reconcilerScope:   declared.RootReconciler,
-				sourceType:        rs.Spec.SourceType,
-				gitConfig:         rs.Spec.Git,
-				ociConfig:         rs.Spec.Oci,
-				helmConfig:        rootsync.GetHelmBase(rs.Spec.Helm),
-				pollPeriod:        r.reconcilerPollingPeriod.String(),
-				statusMode:        rs.Spec.SafeOverride().StatusMode,
-				reconcileTimeout:  v1beta1.GetReconcileTimeout(rs.Spec.SafeOverride().ReconcileTimeout),
-				apiServerTimeout:  v1beta1.GetAPIServerTimeout(rs.Spec.SafeOverride().APIServerTimeout),
-				requiresRendering: enableRendering(rs.GetAnnotations()),
+				clusterName:              r.clusterName,
+				syncName:                 rs.Name,
+				syncGeneration:           rs.Generation,
+				reconcilerName:           reconcilerName,
+				reconcilerScope:          declared.RootReconciler,
+				sourceType:               rs.Spec.SourceType,
+				gitConfig:                rs.Spec.Git,
+				ociConfig:                rs.Spec.Oci,
+				helmConfig:               rootsync.GetHelmBase(rs.Spec.Helm),
+				pollPeriod:               r.reconcilerPollingPeriod.String(),
+				statusMode:               rs.Spec.SafeOverride().StatusMode,
+				reconcileTimeout:         v1beta1.GetReconcileTimeout(rs.Spec.SafeOverride().ReconcileTimeout),
+				apiServerTimeout:         v1beta1.GetAPIServerTimeout(rs.Spec.SafeOverride().APIServerTimeout),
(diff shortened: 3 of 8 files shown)


pkg/reconcilermanager/controllers/util.go
```diff
@@ -85,7 +85,11 @@ func hydrationEnvs(sourceType string, gitConfig *v1beta1.Git, ociConfig *v1beta1
 }
 
 // reconcilerEnvs returns environment variables for namespace reconciler.
-func reconcilerEnvs(clusterName, syncName string, syncGeneration int64, reconcilerName string, reconcilerScope declared.Scope, sourceType string, gitConfig *v1beta1.Git, ociConfig *v1beta1.Oci, helmConfig *v1beta1.HelmBase, pollPeriod, statusMode string, reconcileTimeout string, apiServerTimeout string, requiresRendering bool) []corev1.EnvVar {
+func reconcilerEnvs(clusterName, syncName string, syncGeneration int64,
```
sdowell, line 88: I think it's time to refactor these parameters, perhaps into an options struct. We can do that as a separate change if your prefer

pkg/metadata/annotations.go
```diff
@@ -145,6 +145,13 @@ const (
 	// reconciler-manager will create the reconciler with the hydration-controller
 	// sidecar container.
 	RequiresRenderingAnnotationKey = configsync.ConfigSyncPrefix + "requires-rendering"
+
+	// RequiresNSControllerAnnotationKey is the annotation key set on R*Sync
+	// object to indicate whether the source of truth requires the Namespace
+	// controller. The reconciler writes the value of this annotation and the
```
sdowell, line 151: What are the conditions in which the reconciler will set this to true? In other words, what in the source of truth indicates that the namespace controller must be running?

pkg/reconcilermanager/controllers/util.go
nan-yu: will do in a separate change.

pkg/metadata/annotations.go
nan-yu: Updated the comment to clarify the condition.

pkg/reconcilermanager/controllers/util.go
```diff
@@ -181,6 +185,15 @@ func reconcilerEnvs(clusterName, syncName string, syncGeneration int64, reconcil
 		},
 	)
 
+	if requiresNSController {
```
sdowell, line 188: Is the conditional necessary? For `RenderingEnabled` we always set the env var, including for a value of `False`.

I assume it doesn't make much functional difference since the reconciler defaults to false, but WDYT?

pkg/metadata/annotations.go
sdowell: Thank you

pkg/reconcilermanager/controllers/util.go
nan-yu: Yes, I was also thinking about whether to add the condition for `requiresRendering` when implementing this.
In the flags, we explicitly set the default to false when env var doesn't exist. I feel like we don't need to set the default in two places.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -684,7 +684,8 @@ func (r *RootSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 				r.reconcilerPollingPeriod.String(), rs.Spec.SafeOverride().StatusMode,
 				v1beta1.GetReconcileTimeout(rs.Spec.SafeOverride().ReconcileTimeout),
 				v1beta1.GetAPIServerTimeout(rs.Spec.SafeOverride().APIServerTimeout),
-				enableRendering(rs.GetAnnotations()),
+				enableAnnotation(metadata.RequiresRenderingAnnotationKey, rs.GetAnnotations()),
+				enableAnnotation(metadata.RequiresNSControllerAnnotationKey, rs.GetAnnotations()),
```
sdowell, line 688: Thinking about this a bit more, is it necessary to communicate between `reconciler` and `reconciler-manager` to set this variable? In the case of `RequiresRendering`, this communication is necessary to enable/disable the hydration-controller sidecar.

In this case we are toggling the namespace controller, which is an additional controller in the existing reconciler container. I'm wondering if we can enable/disable this controller without restarting the reconciler Pod, and by extension not require the annotation to communicate with `reconciler-manager`. WDYT?

nan-yu: The namespace controller is registered in the controller-manager. When the controller-manager starts, it starts.
I we want to enable/disable it within the reconciler, we will need to restart the controller-manager, which seems to be feasible by listening to a channel. I'll give it a try.

sdowell: For the record I'm not altogether opposed to this abstraction. If this ends up being a cleaner abstraction than trying to restart the controller-manager while the reconciler is running, then I'm on board with this implementation.

Feel free to time box your investigation and please let me know what you find

nan-yu: Thanks, will keep you posted.

nan-yu: The controller-manager is shared by both the Finalizer controller and the namespace controller.
If a namespace controller is enabled/disabled, and the controller-manager is restarted, the finalizer will also be restarted, which might be okay because the reconciliation loop should be idempotent, although it still feels hacky to restart a controller-manager shared by multiple controllers.

That might be another reason why we would like to move the controller-manager to the top layer, so that its lifecycle go with the Pod container's lifecycle.

nan-yu: @karlkfi  WDYT?

pkg/reconcilermanager/controllers/util.go
nan-yu: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1037

pkg/metadata/annotations.go
```diff
@@ -145,6 +145,14 @@ const (
...
 	// sidecar container.
 	RequiresRenderingAnnotationKey = configsync.ConfigSyncPrefix + "requires-rendering"
+
+	// RequiresNSControllerAnnotationKey is the annotation key set on R*Sync
+	// object to indicate whether the source of truth contains at least one
+	// NamespaceSelector using the dynamic mode, which requires the Namespace
+	// controller. The reconciler writes the value of this annotation and the
+	// reconciler-manager reads it. If set to true, the reconciler-manager will
+	// create the reconciler with the Namespace controller in the reconciler container.
+	RequiresNSControllerAnnotationKey = configsync.ConfigSyncPrefix + "requires-ns-controller"
```
janetkuo, line 155: I just found that this page needs updating https://cloud.google.com/anthos-config-management/docs/reference/labels-and-annotations

Also, ns controller is an implementation detail. Are we planning to keep that architecture? If that will be gone with the controller-runtime refactor, this annotation should talk about the feature (user facing) instead of the controller (impl detail).

nan-yu: Yes, the doc update for that page is on the list.

What about `requires-dynamic-ns-selector`?

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -1029,22 +1030,22 @@ func (r *RootSyncReconciler) updateSyncStatus(ctx context.Context, rs *v1beta1.R
...
-// enableRendering returns whether rendering should be enabled for the reconciler
-// of this RSync. This is determined by an annotation that is set on the RSync
-// by the reconciler.
-func enableRendering(annotations map[string]string) bool {
-	val, ok := annotations[metadata.RequiresRenderingAnnotationKey]
-	if !ok { // default to disabling rendering
+// enableAnnotation returns whether the annotation should be enabled for the
+// reconciler of this RSync. This is determined by an annotation that is set on
+// the RSync by the reconciler.
+func enableAnnotation(key string, annotations map[string]string) bool {
```
janetkuo, line 1036: This reads like a "set" instead of a "get" function. Suggest `checkAnnotationEnabled`.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -828,8 +828,20 @@ func requeueRepoSyncRequest(obj client.Object, rsRef types.NamespacedName) []rec
...
+			rs.Generation, reconcilerName, declared.Scope(rs.Namespace),
+			rs.Spec.SourceType, rs.Spec.Git, rs.Spec.Oci,
+			reposync.GetHelmBase(rs.Spec.Helm), r.reconcilerPollingPeriod.String(),
+			rs.Spec.SafeOverride().StatusMode,
+			v1beta1.GetReconcileTimeout(rs.Spec.SafeOverride().ReconcileTimeout),
+			v1beta1.GetAPIServerTimeout(rs.Spec.SafeOverride().APIServerTimeout),
+			enableAnnotation(metadata.RequiresRenderingAnnotationKey, rs.GetAnnotations()),
+			// Namespace reconciler never needs a Namespace Controller because the
+			// Namespace repository doesn't support NamespaceSelector at all.
+			false),
```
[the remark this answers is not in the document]
nan-yu: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1037 changes to use options structs to replace the long list of paramenters.

pkg/reconcilermanager/controllers/rootsync_controller.go
janetkuo: Or make it shorter, `isAnnotationEnabled` or just `annotationEnabled `

pkg/metadata/annotations.go
janetkuo: Minor tweak: `has-dynamic-ns-selector` (the RootSync has dynamic ns selector, instead of requires)

janetkuo:
Just a few nits and lgtm otherwise.

pkg/metadata/annotations.go
nan-yu: `has-dynamic-ns-selector` indicates the current mode of NamespaceSelector in the RootSync's source repo, which may not match the actual mode.
What about `dynamic-ns-selector-enabled`, which implies the state of a former event. The annotation might be enabled by a previous state. If it doesn't match the current state, it should be updated.

pkg/reconcilermanager/controllers/rootsync_controller.go
nan-yu: done

nan-yu:
rebased and comments addressed

janetkuo (approved):

PR #1031, opened 2023-11-15
Implement the select mode in the NamespaceSelector API

nan-yu:
Previously, NamespaceSelector only supports selecting namespace-scoped
resources whose labels match those labels of the Namespaces that are
statically declared in the source of truth.

This commit updates the NamespaceSelector to support selecting
dynmaic/on-cluster Namespaces as well.
When it applies the NamespaceSelector, it sends out a LIST request to
get all Namespaces and select resources match the on-cluster Namespaces.

### DIFF
--- a/cmd/nomos/hydrate/hydrate.go
+++ b/cmd/nomos/hydrate/hydrate.go
@@ -22,6 +22,7 @@ import (
 	"kpt.dev/configsync/cmd/nomos/flags"
 	nomosparse "kpt.dev/configsync/cmd/nomos/parse"
 	"kpt.dev/configsync/cmd/nomos/util"
+	"kpt.dev/configsync/pkg/declared"
 	"kpt.dev/configsync/pkg/hydrate"
 	"kpt.dev/configsync/pkg/importer/analyzer/ast"
 	"kpt.dev/configsync/pkg/importer/filesystem"
@@ -105,6 +106,9 @@ which you could kubectl apply -fR to the cluster, or have Config Sync sync to th
 
 		if sourceFormat == filesystem.SourceFormatHierarchy {
 			files = filesystem.FilterHierarchyFiles(rootDir, files)
+		} else {
+			// hydrate as a root repository to preview all the hydrated configs
+			validateOpts.Scope = declared.RootReconciler
 		}
 
 		filePaths := reader.FilePaths{
@@ -149,7 +153,7 @@ which you could kubectl apply -fR to the cluster, or have Config Sync sync to th
 
 			allObjects = append(allObjects, fileObjects...)
 		}
-		hydrate.ForEachCluster(parseOpts, validateOpts, clusterFilterFunc)
+		hydrate.ForEachCluster(cmd.Context(), parseOpts, validateOpts, clusterFilterFunc)
 
 		multiCluster := numClusters > 1
 		fileObjects := hydrate.GenerateFileObjects(multiCluster, allObjects...)
--- a/cmd/nomos/vet/vet_impl.go
+++ b/cmd/nomos/vet/vet_impl.go
@@ -152,7 +152,7 @@ func runVet(ctx context.Context, namespace string, sourceFormat filesystem.Sourc
 			allObjects = append(allObjects, fileObjects...)
 		}
 	}
-	hydrate.ForEachCluster(parseOpts, validateOpts, clusterFilterFunc)
+	hydrate.ForEachCluster(ctx, parseOpts, validateOpts, clusterFilterFunc)
 	if keepOutput {
 		multiCluster := numClusters > 1
 		fileObjects := hydrate.GenerateFileObjects(multiCluster, allObjects...)
--- a/e2e/nomostest/testpredicates/predicates.go
+++ b/e2e/nomostest/testpredicates/predicates.go
@@ -30,6 +30,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"kpt.dev/configsync/e2e/nomostest/retry"
 	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
 	"kpt.dev/configsync/e2e/nomostest/testlogger"
@@ -1236,6 +1237,59 @@ func ResourceGroupStatusEquals(expected v1alpha1.ResourceGroupStatus) Predicate 
 		if !equality.Semantic.DeepEqual(found, expected) {
 			return errors.Errorf("expected %s to have status: %s, but got %s",
 				kinds.ObjectSummary(obj), log.AsJSON(expected), log.AsJSON(found))
+		}
+		return nil
+	}
+}
+
+// ResourceGroupHasObjects checks whether the objects are in the ResourceGroup's inventory.
+func ResourceGroupHasObjects(objects []client.Object) Predicate {
+	return func(obj client.Object) error {
+		rg, ok := obj.(*v1alpha1.ResourceGroup)
+		if !ok {
+			return WrongTypeErr(obj, &v1alpha1.ResourceGroup{})
+		}
+
+		rgResources := toManagedResourceIDs(rg)
+		for _, o := range objects {
+			if _, found := rgResources[core.IDOf(o)]; !found {
+				return fmt.Errorf("resource %s missing from the inventory of ResourceGroup %s",
+					core.IDOf(o), core.IDOf(rg))
+			}
+		}
+		return nil
+	}
+}
+
+func toManagedResourceIDs(rg *v1alpha1.ResourceGroup) map[core.ID]struct{} {
+	rgResources := make(map[core.ID]struct{}, len(rg.Spec.Resources))
+	for _, res := range rg.Spec.Resources {
+		id := core.ID{
+			GroupKind: schema.GroupKind{
+				Group: res.Group,
+				Kind:  res.Kind,
+			},
+			ObjectKey: client.ObjectKey{Name: res.Name, Namespace: res.Namespace},
+		}
+		rgResources[id] = struct{}{}
+	}
+	return rgResources
+}
+
+// ResourceGroupMissingObjects checks whether the objects are NOT in the ResourceGroup's inventory.
+func ResourceGroupMissingObjects(objects []client.Object) Predicate {
+	return func(obj client.Object) error {
+		rg, ok := obj.(*v1alpha1.ResourceGroup)
+		if !ok {
+			return WrongTypeErr(obj, &v1alpha1.ResourceGroup{})
+		}
+
+		rgResources := toManagedResourceIDs(rg)
+		for _, o := range objects {
+			if _, found := rgResources[core.IDOf(o)]; found {
+				return fmt.Errorf("resource %s found from the inventory of ResourceGroup %s",
+					core.IDOf(o), core.IDOf(rg))
+			}
 		}
 		return nil
 	}
--- a/e2e/testcases/namespace_selectors_test.go
+++ b/e2e/testcases/namespace_selectors_test.go
@@ -15,37 +15,78 @@
 package e2e
 
 import (
+	"fmt"
 	"testing"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/e2e/nomostest/testpredicates"
+	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
+	"kpt.dev/configsync/pkg/reconcilermanager"
 	fake "kpt.dev/configsync/pkg/testing/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	bookstoreNS = "bookstore"
+	shoestoreNS = "shoestore"
+)
+
+var (
+	bookstoreNSS = fake.NamespaceSelectorObject(core.Name(bookstoreNS))
+	bookstoreCM  = fake.ConfigMapObject(core.Name("cm-bookstore"),
+		core.Annotation(metadata.NamespaceSelectorAnnotationKey, bookstoreNSS.Name))
+	bookstoreRQ = fake.ResourceQuotaObject(core.Name("rq-bookstore"),
+		core.Annotation(metadata.NamespaceSelectorAnnotationKey, bookstoreNSS.Name))
+
+	shoestoreNSS = fake.NamespaceSelectorObject(core.Name(shoestoreNS))
+	shoestoreCM  = fake.ConfigMapObject(core.Name("cm-shoestore"),
+		core.Annotation(metadata.NamespaceSelectorAnnotationKey, shoestoreNSS.Name))
+
+	selectedResourcesWithBookstoreNSSAndShoestoreNSS = []client.Object{
+		fake.ConfigMapObject(core.Namespace(bookstoreNS), core.Name(bookstoreCM.Name)),
+		fake.ResourceQuotaObject(core.Namespace(bookstoreNS), core.Name(bookstoreRQ.Name)),
+		fake.ConfigMapObject(core.Namespace(shoestoreNS), core.Name(shoestoreCM.Name)),
+	}
+
+	unselectedResourcesWithBookstoreNSSAndShoestoreNSS = []client.Object{
+		fake.ConfigMapObject(core.Namespace(bookstoreNS), core.Name(shoestoreCM.Name)),
+		fake.ConfigMapObject(core.Namespace(shoestoreNS), core.Name(bookstoreCM.Name)),
+		fake.ResourceQuotaObject(core.Namespace(shoestoreNS), core.Name(bookstoreRQ.Name)),
+	}
+
+	selectedResourcesWithShoestoreNSSOnly = []client.Object{
+		fake.ConfigMapObject(core.Namespace(shoestoreNS), core.Name(shoestoreCM.Name)),
+	}
+
+	unselectedResourcesWithShoestoreNSSOnly = []client.Object{
+		fake.ConfigMapObject(core.Namespace(bookstoreNS), core.Name(bookstoreCM.Name)),
+		fake.ResourceQuotaObject(core.Namespace(bookstoreNS), core.Name(bookstoreRQ.Name)),
+		fake.ConfigMapObject(core.Namespace(bookstoreNS), core.Name(shoestoreCM.Name)),
+		fake.ConfigMapObject(core.Namespace(shoestoreNS), core.Name(bookstoreCM.Name)),
+		fake.ResourceQuotaObject(core.Namespace(shoestoreNS), core.Name(bookstoreRQ.Name)),
+	}
 )
 
 func TestNamespaceSelectorHierarchicalFormat(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.Selector)
 
 	nt.T.Log("Add Namespaces, NamespaceSelectors and Namespace-scoped resources")
-	bookstoreNS := "bookstore"
-	bookstoreNSS := fake.NamespaceSelectorObject(core.Name(bookstoreNS))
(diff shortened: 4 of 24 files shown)


pkg/validate/validate.go
```diff
@@ -155,7 +156,7 @@ func Hierarchical(objs []ast.FileObject, opts Options) ([]ast.FileObject, status
 
 // Unstructured validates and hydrates the given FileObjects from an
 // unstructured repo.
-func Unstructured(objs []ast.FileObject, opts Options) ([]ast.FileObject, status.MultiError) {
+func Unstructured(ctx context.Context, c client.Client, objs []ast.FileObject, opts Options, skipAPICall, nsControllerEnabled bool, nsControllerState *namespacecontroller.State) ([]ast.FileObject, status.MultiError) {
```
janetkuo, line 159: nit: too many parameters

cmd/nomos/hydrate/hydrate.go
```diff
@@ -105,6 +106,9 @@ which you could kubectl apply -fR to the cluster, or have Config Sync sync to th
 
 		if sourceFormat == filesystem.SourceFormatHierarchy {
 			files = filesystem.FilterHierarchyFiles(rootDir, files)
+		} else {
+			// hydrate as a root repository to preview all the hydrated configs
+			options.Scope = declared.RootReconciler
```
janetkuo, line 111: Why is the scope related to source format?

pkg/validate/scoped/scoped.go
```diff
@@ -48,13 +53,13 @@ func Hierarchical(objs *objects.Scoped) status.MultiError {
 // Unstructured performs the second round of validation and hydration for an
 // unstructured repo against the given Scoped objects. Note that this will
 // modify the Scoped objects in-place.
-func Unstructured(objs *objects.Scoped) status.MultiError {
+func Unstructured(ctx context.Context, c client.Client, objs *objects.Scoped, skipAPICall, nsControllerEnabled bool, nsControllerState *namespacecontroller.State) status.MultiError {
```
janetkuo, line 56: nit: too many parameters

cmd/nomos/hydrate/hydrate.go
nan-yu: The scope is used by namespace selectors and other validators for unstructured source format.

The hierarchy mode doesn't rely on the scope because the folder structure has already defined the scope.

nan-yu: Before the change, it uses a `IsNamespaceReconciler` flag, which is semantically the same as scope.
It wasn't set explicitly here because the flag is default to false, which indicates a root scope.

cmd/nomos/hydrate/hydrate.go
```diff
@@ -116,7 +116,7 @@ which you could kubectl apply -fR to the cluster, or have Config Sync sync to th
 		var allObjects []ast.FileObject
 		encounteredError := false
 		numClusters := 0
-		hydrate.ForEachCluster(parser, options, sourceFormat, filePaths, func(clusterName string, fileObjects []ast.FileObject, err status.MultiError) {
+		hydrate.ForEachCluster(cmd.Context(), parser, options, sourceFormat, filePaths, func(clusterName string, fileObjects []ast.FileObject, err status.MultiError) {
```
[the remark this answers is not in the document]
nan-yu: refactored

pkg/validate/scoped/scoped.go
nan-yu: refactored

pkg/validate/validate.go
nan-yu: refactored

pkg/validate/scoped/hydrate/namespace_hydrator.go
```diff
@@ -15,85 +15,148 @@
...
+		// Select dynamic/on-cluster Namespaces that match the NamespaceSelector's labels.
+		for _, ns := range nsList.Items {
+			if selector.Matches(labels.Set(ns.GetLabels())) {
+				selectedNamespaces = append(selectedNamespaces, ns.GetName())
+				// selectedNamespaceMap only stores dynamic Namespaces.
+				selectedNamespaceMap[ns.GetName()] = append(selectedNamespaceMap[ns.GetName()], selector)
+			}
+		}
+
+		// Select statically declared Namespaces that match the NamespaceSelector's labels.
```
sdowell, line 148: Does `static` refer to a namespace that is declared in the sync source?

pkg/validate/scoped/hydrate/namespace_hydrator.go
```diff
@@ -15,85 +15,148 @@
...
-			continue
+			return nil, errs
 		}
+	}
 
+	nsList := &corev1.NamespaceList{}
+	// Fetch on-cluster Namespaces if allowing API calls and hasDynamicNSSelector,
+	// so that the NamespaceSelector can select resources matching on-cluster Namespaces.
+	if objs.AllowAPICall && hasDynamicNSSelector {
+		if err := c.List(ctx, nsList); err != nil {
```
sdowell, line 130: Instead of listing all namespaces and then filtering by labels in code, can we use label selectors on the list call?

nan-yu: There might be multiple NamespaceSelectors. If we use label selector, that means multiple API calls or each label selector.
Without using label selector, there is only one single API call.

pkg/validate/scoped/hydrate/namespace_hydrator.go
nan-yu: Correct
- `static` only selects resources matching Namespaces statically declared in the source of truth.
- `dynamic` selects both statically declared Namespaces and on-cluster Namespaces.

That is explained in the CRD.

e2e/testcases/namespace_selectors_test.go
```diff
@@ -83,49 +82,167 @@ func TestNamespaceSelectorUnstructuredFormat(t *testing.T) {
...
+		nt.T.Fatal(err)
+	}
+
+	nt.Logger.Info("Creating a Namespace with labels, matching resources should be selected")
+	bookstoreNamespace := fake.NamespaceObject(bookstoreNS, core.Label("app", bookstoreNS))
+	if err := nt.KubeClient.Create(bookstoreNamespace); err != nil {
+		nt.T.Fatal(err)
+	}
+	nomostest.Wait(nt.T, "matching resources should be applied", nt.DefaultWaitTimeout, func() error {
+		return allResourcesSelectedAndApplied(nt)
```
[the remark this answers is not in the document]
nan-yu: In this version, the underlying functions don't use watch, but anyway, I switched to use watch instead of wait.

e2e/testcases/namespace_selectors_test.go
```diff
@@ -83,49 +82,167 @@ func TestNamespaceSelectorUnstructuredFormat(t *testing.T) {
...
+
+	nt.Logger.Info("Stop the admission webhook so the Namespace can be deleted")
+	// If the webhook is running, the namespace DELETE call will fail because the
+	// underlying resources cannot be deleted due to the admission-webhook.
+	nomostest.StopWebhook(nt)
+	nt.Logger.Info("Delete Namespace, resources should NOT be selected")
+	if err := nt.KubeClient.Delete(bookstoreNamespace); err != nil {
+		t.Fatal(err)
+	}
+	nomostest.Wait(nt.T, "previously matched resources should be pruned", nt.DefaultWaitTimeout, func() error {
```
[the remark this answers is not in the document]
nan-yu: sounds good. Refactored the assertions.

e2e/testcases/namespace_selectors_test.go
```diff
@@ -83,49 +82,167 @@ func TestNamespaceSelectorUnstructuredFormat(t *testing.T) {
...
+	); err != nil {
 		nt.T.Fatal(err)
 	}
-	if err := nt.ValidateNotFound(bookstoreCM.Name, shoestoreNS, &corev1.ConfigMap{}); err != nil {
+
+	nt.Logger.Info("Update NamespaceSelector to use dynamic mode")
+	bookstoreNSS.Spec.Mode = v1.NSSelectorDynamicMode
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespace-selector-bookstore.yaml", bookstoreNSS))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update NamespaceSelector to use dynamic mode"))
+	nomostest.Wait(nt.T, "Namespace controller should start", nt.DefaultWaitTimeout, func() error {
```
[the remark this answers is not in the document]
nan-yu: done

e2e/testcases/namespace_selectors_test.go
```diff
@@ -83,49 +82,167 @@ func TestNamespaceSelectorUnstructuredFormat(t *testing.T) {
...
+	if err := nt.WatchForAllSyncs(); err != nil {
 		nt.T.Fatal(err)
 	}
-	if err := nt.ValidateNotFound(shoestoreCM.Name, bookstoreNS, &corev1.ConfigMap{}); err != nil {
+	if err := onlyShoestoreResourcesSelectedAndApplied(nt); err != nil {
 		nt.T.Fatal(err)
 	}
 }
+
+func allResourcesSelectedAndApplied(nt *nomostest.NT) error {
```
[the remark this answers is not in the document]
nan-yu: refactored

e2e/testcases/namespace_selectors_test.go
```diff
@@ -59,73 +99,226 @@ func TestNamespaceSelectorHierarchicalFormat(t *testing.T) {
...
+		nt.T.Fatal(err)
+	}
+	validateSelectedAndUnselectedResources(nt,
+		append(selectedResourcesWithShoestoreNSSOnly),
+		append(unselectedResourcesWithShoestoreNSSOnly),
+	)
+
+	nt.Logger.Info("Creating a Namespace with labels, matching resources should be selected")
+	bookstoreNamespace := fake.NamespaceObject(bookstoreNS, core.Label("app", bookstoreNS))
+	if err := nt.KubeClient.Create(bookstoreNamespace); err != nil {
```
sdowell, line 186: This needs an accompanying `Delete` wrapped in a `Cleanup`, in case the test fails before reaching line 254 where the namespace is deleted

e2e/testcases/namespace_selectors_test.go
```diff
@@ -59,73 +99,226 @@ func TestNamespaceSelectorHierarchicalFormat(t *testing.T) {
...
-	if err := nt.Validate(shoestoreCM.Name, shoestoreNS, &corev1.ConfigMap{}); err != nil {
+	validateSelectedAndUnselectedResources(nt,
+		selectedResourcesWithBookstoreNSSAndShoestoreNSS,
+		unselectedResourcesWithBookstoreNSSAndShoestoreNSS,
+	)
+
+	nt.Logger.Info("Stop the admission webhook so the Namespace can be deleted")
+	// If the webhook is running, the namespace DELETE call will fail because the
+	// underlying resources cannot be deleted due to the admission-webhook.
+	nomostest.StopWebhook(nt)
```
sdowell, line 252: Why does the webhook block deletion? It looks like the namespace is created out-of-band, so presumably should not have Config Sync management labels. Is it because the namespace contains managed resources?

e2e/testcases/namespace_selectors_test.go
```diff
@@ -59,73 +99,226 @@ func TestNamespaceSelectorHierarchicalFormat(t *testing.T) {
...
 		nt.T.Fatal(err)
 	}
-	if err := nt.ValidateNotFound(bookstoreRQ.Name, shoestoreNS, &corev1.ResourceQuota{}); err != nil {
+	validateSelectedAndUnselectedResources(nt,
+		append(selectedResourcesWithShoestoreNSSOnly),
+		append(unselectedResourcesWithShoestoreNSSOnly),
+	)
+
+	nt.Logger.Info("Update NamespaceSelector to use static mode")
+	bookstoreNSS.Spec.Mode = v1.NSSelectorStaticMode
```
sdowell, line 277: What happens if the mode is switched from `dynamic` -> `static` while there are currently resources in a "dynamic" namespace?

That seems like an interesting scenario, but I don't think the current test case covers that because the dynamic namespace is deleted before switching from `dynamic` -> `static`.

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1241,6 +1241,64 @@ func ResourceGroupStatusEquals(expected v1alpha1.ResourceGroupStatus) Predicate
...
+		if len(unselected) == 0 {
+			return nil
+		}
+		return fmt.Errorf("the following resources are not in the inventory of ResourceGroup %s: %v",
+			core.IDOf(obj), unselected)
+	}
+}
+
+// ResourceGroupHasNoObjects checks whether the objects are NOT in the ResourceGroup's inventory.
+func ResourceGroupHasNoObjects(objects []client.Object) Predicate {
```
sdowell, line 1276: nit: `ResourceGroupMissingObjects`

e2e/testcases/namespace_selectors_test.go
nan-yu: Correct. When deleting a namespace, resources in the namespace will also be deleted. But they're managed by Config Sync, the webhook will reject the deletion.

I have the `underlying resources cannot be deleted` comment. It seems not clear? Clarified.

e2e/testcases/namespace_selectors_test.go
nan-yu: Moves the case right after the resources are dynamically selected, and then change it back to `dynamic` mode to cover more scenarios.

e2e/nomostest/testpredicates/predicates.go
nan-yu: done

e2e/testcases/namespace_selectors_test.go
nan-yu: good catch. Done.

e2e/testcases/namespace_selectors_test.go
```diff
@@ -59,73 +100,264 @@ func TestNamespaceSelectorHierarchicalFormat(t *testing.T) {
...
+		append(unselectedResourcesWithShoestoreNSSOnly),
+	)
+
+	nt.Logger.Info("Creating a Namespace with labels, matching resources should be selected")
+	bookstoreNamespace := fake.NamespaceObject(bookstoreNS, core.Label("app", bookstoreNS))
+	if err := nt.KubeClient.Create(bookstoreNamespace); err != nil {
+		nt.T.Fatal(err)
+	}
+	t.Cleanup(func() {
+		if err := nt.KubeClient.Delete(bookstoreNamespace); err != nil && !apierrors.IsNotFound(err) {
```
sdowell, line 191: I think you may need to disable the webhook here in case it's still running

sdowell:
One outstanding comment, but lgtm.
/lgtm
/assign @janetkuo

e2e/testcases/namespace_selectors_test.go
nan-yu: done. Also added a check for the webhook state to avoid re-stop.

pkg/validate/scoped/hydrate/namespace_hydrator.go
janetkuo: Is it talking to the API server? With namespace watches can it read from cache instead of calling the API server?

nan-yu: Yes, it talks to the API server. The namespace watch is in the  namespace controller within the controller-manager running in a separate goroutine.  This namespace_hydrator is in the reconciler thread, so it won't be able to access the cache.

cmd/nomos/vet/vet_impl.go
```diff
@@ -102,9 +102,9 @@ func runVet(ctx context.Context, namespace string, sourceFormat filesystem.Sourc
 		files = filesystem.FilterHierarchyFiles(rootDir, files)
 	case filesystem.SourceFormatUnstructured:
 		if namespace == "" {
-			validateOpts = parse.OptionsForScope(validateOpts, declared.RootReconciler)
+			validateOpts = parse.OptionsForScope(validateOpts, declared.RootReconciler, "")
 		} else {
-			validateOpts = parse.OptionsForScope(validateOpts, declared.Scope(namespace))
+			validateOpts = parse.OptionsForScope(validateOpts, declared.Scope(namespace), "")
```
janetkuo, line 107: Seems a bit weird passing empty syncName here?

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1241,6 +1241,64 @@ func ResourceGroupStatusEquals(expected v1alpha1.ResourceGroupStatus) Predicate
...
+				}
+			}
+			if !selected {
+				unselected = append(unselected, core.IDOf(o))
+			}
+		}
+
+		if len(unselected) == 0 {
+			return nil
+		}
```
janetkuo, line 1269: The code will be cleaner if the for loop becomes a function that returns a `bool` (found or not), which can be shared by the function below as well.

pkg/validate/scoped/hydrate/namespace_hydrator.go
```diff
@@ -15,85 +15,148 @@
...
+			if hasSelector {
+				copies, err := makeNamespaceCopies(obj, nsSelectors)
+				if err != nil {
+					errs = status.Append(errs, err)
+				} else {
+					result = append(result, copies...)
+				}
 			} else {
-				result = append(result, copies...)
+				if obj.GetNamespace() == "" {
```
janetkuo, line 62: `else if {`

pkg/validate/scoped/hydrate/namespace_hydrator.go
```diff
@@ -15,85 +15,148 @@
...
+	"kpt.dev/configsync/pkg/rootsync"
 	"kpt.dev/configsync/pkg/status"
 	"kpt.dev/configsync/pkg/validate/objects"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // NamespaceSelectors hydrates the given Scoped objects by performing namespace
 // selection to copy objects into namespaces which match their selector. It also
 // sets a default namespace on any namespace-scoped object that does not already
 // have a namespace or namespace selector set.
```
janetkuo, line 44: comment should be updated

pkg/validate/scoped/hydrate/namespace_hydrator.go
janetkuo: Ah, good point. I'm a bit concerned about directly calling the API server. Each reconciler could be listing all namespaces from the API server, and that's pretty expensive.

e2e/nomostest/testpredicates/predicates.go
nan-yu: If we use a shared function, then we need to iterate through all items. 
Currently, we can break the loop when found or not found without iterating through all items.

cmd/nomos/vet/vet_impl.go
nan-yu: moved syncName setter out of `parse.OptionsForScope`.

pkg/validate/scoped/hydrate/namespace_hydrator.go
nan-yu: done

pkg/validate/scoped/hydrate/namespace_hydrator.go
nan-yu: updated

pkg/validate/scoped/hydrate/namespace_hydrator.go
nan-yu: Also considered storing all namespaces in the memory (namespace controller's state) with the namespace controller's cache, and let the reconciler thread read from the namespaces from the state. However, found https://medium.com/@timebertt/kubernetes-controllers-at-scale-clients-caches-conflicts-patches-explained-aa0f7a8b4332#:~:text=Controllers%20must%20not,should%20be%20improved.

> Controllers must not use the same in-memory object concurrently in multiple goroutines. For example, decoding results from the API server in multiple goroutines into the same maps (e.g., labels, annotations) will cause panics because of concurrent map writes. Also, reading from an in-memory API object in one goroutine while decoding into it in another goroutine will yield non-atomic reads, meaning data might be corrupt and represent an invalid/non-existing API object.
Therefore, if you need to use the same in-memory object in multiple goroutines concurrently (e.g., shared state), remember to leverage proper synchronization techniques like channels, mutexes, atomic.Value and/or copy the object prior to use. The average controller however, will not need to share in-memory API objects between goroutines, and it's typically an indicator that the controller's design should be improved.

It feels like using the cache requires to refactor the reconciler thread to use the controller-runtime library, which sets up the watch and uses the cache in the same thread.

pkg/validate/scoped/hydrate/namespace_hydrator.go
nan-yu: reverted, because it appends the obj in the end o the else branch even if `obj.GetNamespace()` is not empty.

e2e/nomostest/testpredicates/predicates.go
janetkuo: The function can use a set/map to and then 1 loop to check whether the item can be found. This would avoid the need of `unselected` and make the code look cleaner.

pkg/validate/scoped/hydrate/namespace_hydrator.go
janetkuo: Let's enhance that later as a follow-up

e2e/nomostest/testpredicates/predicates.go
nan-yu: converted to use map.

The `unselected` is used to capture all objects missing from the ResourceGroup spec, which may not be necessary. After the change, it returns error as long as one object is missing.

janetkuo (approved):
/lgtm
/approve

PR #1032, opened 2023-11-15
test: fix go linter install

karlkfi:
Install with "go install" into .output/go/bin, instead of the local user's $GOPATH/bin.

The latest versions of go install respect GOPATH, but not GOBIN. But we can't set GOPATH as a global in the Makefile, otherwise we wouldn't be able to build, cache, and vendor using the system Go.


mikebz:
exciting change!

sdowell (approved):

PR #1033, opened 2023-11-15
test: fix addlicense & controller-gen install

karlkfi:
Install with "go install" into .output/go/bin, instead of the local user's $GOPATH/bin.

The latest versions of go install respect GOPATH, but not GOBIN. But we can't set GOPATH as a global in the Makefile, otherwise we wouldn't be able to build, cache, and vendor using the system Go.

### DIFF
--- a/Makefile.reconcilermanager
+++ b/Makefile.reconcilermanager
@@ -1,11 +1,12 @@
 
 
-CONTROLLER_GEN := $(GOBIN)/controller-gen
+CONTROLLER_GEN_VERSION := v0.13.0
+CONTROLLER_GEN := $(BIN_DIR)/controller-gen
 
 .PHONY: generate
 # Generate DeepCopy and runtime.Object implementation methods.
-generate: install-controller-gen
-	$(CONTROLLER_GEN) \
+generate: "$(CONTROLLER_GEN)"
+	"$(CONTROLLER_GEN)" \
 		object:headerFile="hack/boilerplate.txt" \
 		paths="./pkg/api/configsync/v1alpha1" \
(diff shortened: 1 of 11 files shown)


Makefile.reconcilermanager
```diff
@@ -44,21 +45,27 @@ configsync-crds: install-controller-gen "$(KUSTOMIZE)" "$(GOBIN)/addlicense"
...
 # Run make generate-in-docker in the docker buildenv container
-generate-in-docker: buildenv-dirs
+generate-in-docker: "$(CONTROLLER_GEN)"
 	@docker run $(DOCKER_RUN_ARGS) \
 		make generate
 
 .PHONY: configsync-crds-in-docker
 # Run make configsync-crds in the docker buildenv container
-configsync-crds-in-docker: buildenv-dirs
+configsync-crds-in-docker: "$(CONTROLLER_GEN)" "$(KUSTOMIZE)" "$(ADDLICENSE)"
```
sdowell, line 69: I don't think this needs `"$(CONTROLLER_GEN)" "$(KUSTOMIZE)" "$(ADDLICENSE)"` as dependencies, since those are covered in `configsync-crds`

ditto for `generate-in-docker` above

I think for these commands, we explicitly want to use the go binary inside the container to install each tool for portability

karlkfi: I added them explicitly to make sure they were installed by the local user. But I guess DOCKER_RUN_ARGS does set the UID, so it shouldn't matter.

karlkfi: reverted

sdowell (approved):

PR #1035, opened 2023-11-15
ci: add clientgen/generate to presubmit

sdowell:
This adds the clientgen and generate calls to the presubmit check. This ensures that the generated code is always up to date.

### DIFF
--- a/scripts/generate-clientset.sh
+++ b/scripts/generate-clientset.sh
@@ -32,7 +32,7 @@
 # Alternatively, make a symlink from the GOPATH package source directory to the
 # repo directory.
 
-set -euox pipefail
+set -euo pipefail
 
 GOPATH="${GOPATH:-$(go env GOPATH)}"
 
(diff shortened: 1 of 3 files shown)


scripts/generate-clientset.sh
```diff
@@ -32,7 +32,7 @@
 # Alternatively, make a symlink from the GOPATH package source directory to the
 # repo directory.
 
-set -euox pipefail
+set -euo pipefail
```
tiffanny29631, line 35: Why removing -x?

sdowell: it makes the output a lot more verbose than it needs to be IMO, and the extra verbosity doesn't seem to add much value

tiffanny29631: Fair, thanks!

karlkfi:
Looks like some sorting differences in the generate output. Is it not consistent or was it generated with a different permutation of tool versions?

PR #1036, opened 2023-11-15
Add validations for unknown and unsupported NamespaceSelector mode

nan-yu:


cmd/nomoserrors/examples/examples.go
```diff
@@ -321,6 +321,12 @@ func Generate() AllExamples {
 	// 1069
 	result.add(validate.SelfReconcileError(fake.RootSyncV1Beta1(configsync.RootSyncName)))
 
+	// 1070
+	result.add(selectors.UnsupportedNamespaceSelectorModeError(fake.NamespaceSelector()))
+
+	// 1071
+	result.add(selectors.UnknownNamespaceSelectorModeError(fake.NamespaceSelector()))
```
janetkuo, line 328: Is it necessary to separate unsupported vs. unknown? The error can just be "unsupported / invalid" to let users know that the specification is not valid.

nan-yu: `unknown` applies to both hierarchy mode and unstructured mode, while `unsupported` only applies the hierarchy mode.

We have KNV1014 for invalidSelector, which applies to both ClusterSelector and NamespaceSelector: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/importer/analyzer/transform/selectors/errors.go#L29. There is also a TODO to refactor it properly. 

What's the best practice to reuse error code? Do we need to group them for a particular feature?

pkg/validate/tree/hydrate/namespace_selector_hydrator.go
```diff
@@ -59,6 +59,13 @@ func labelSelector(obj ast.FileObject) (labels.Selector, status.Error) {
 	}
 	nss := s.(*v1.NamespaceSelector)
 
+	if nss.Spec.Mode == v1.NSSelectorDynamicMode {
+		return nil, selectors.UnsupportedNamespaceSelectorModeError(nss)
+	}
+	if nss.Spec.Mode != "" && nss.Spec.Mode != v1.NSSelectorStaticMode {
+		return nil, selectors.UnknownNamespaceSelectorModeError(nss)
+	}
```
janetkuo, line 67: Can we just return `InvalidSelectorError(obj, err)` here (combined with below), and `err` tells users the cause, e.g. "dynamic mode isn't supported in hierachical repositories" etc.

cmd/nomoserrors/examples/examples.go
janetkuo: Please see my other comment

pkg/validate/tree/hydrate/namespace_selector_hydrator.go
nan-yu: Updated to reuse KNV1014, but created separate functions to wrap the error message, which seems the common way to build error message.

cmd/nomoserrors/examples/examples.go
nan-yu: changed to reuse KNV1014

janetkuo (commented):
/lgtm 

#1031 should be merged first

janetkuo (approved):
/lgtm

PR #1037, opened 2023-11-16
refactor: use option structs to replace the long list of function parameters

nan-yu:


nan-yu:
/hold

Needs to rebase on top of #1030

janetkuo (commented):
/lgtm

janetkuo (approved):

nan-yu:
/unhold

Updated #1030 to base on top of this PR

pkg/reconcilermanager/controllers/util.go
```diff
@@ -79,43 +88,61 @@ func hydrationEnvs(sourceType string, gitConfig *v1beta1.Git, ociConfig *v1beta1
...
 }
 
+type reconcilerOptions struct {
+	clusterName       string
+	syncName          string
+	syncGeneration    int64
+	reconcilerName    string
+	reconcilerScope   declared.Scope
+	sourceType        string
+	gitConfig         *v1beta1.Git
```
sdowell, line 103: nit: Do we want to pass in pointers here? Granted there are some tradeoffs, but since we have no intention of modifying the variable in this function I feel it may be clearer to pass by value. WDYT?

nan-yu: Ideally we should pass by value, but there is some trick here.  The spec.git field is a pointer. If we pass by value, we need to check for nil first. If it is nil, then build an empty struct. Only one of the spec is not nil, so we need to build empty struct for the other two.
We can make it better, but seems another refactor.

sdowell: Ack, SGTM

PR #1040, opened 2023-11-16
chore: Update otel-collector to v0.89.0

karlkfi:
- Should fix the CVE in v0.87

### DIFF
--- a/manifests/templates/otel-collector.yaml
+++ b/manifests/templates/otel-collector.yaml
@@ -93,7 +93,7 @@ spec:
     spec:
       containers:
       - name: otel-collector
-        image: gcr.io/config-management-release/otelcontribcol:v0.87.0-gke.2
+        image: gcr.io/config-management-release/otelcontribcol:v0.89.0
         command:
         - /otelcontribcol
         args:
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -161,7 +161,7 @@ data:
                - NET_RAW
              runAsUser: 65533
          - name: otel-agent
-           image: gcr.io/config-management-release/otelcontribcol:v0.87.0-gke.2
+           image: gcr.io/config-management-release/otelcontribcol:v0.89.0
            command:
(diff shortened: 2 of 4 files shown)


manifests/templates/otel-collector.yaml
```diff
@@ -93,7 +93,7 @@ spec:
     spec:
       containers:
       - name: otel-collector
-        image: gcr.io/config-management-release/otelcontribcol:v0.87.0-gke.2
```
sdowell, line 96: Where is the `-gke.N` suffix?

nan-yu: +1. The image should be pushed by the Louhi flow, which appends `-gke.N`.

karlkfi: Does it need it? This is just the regular tag. https://gke-internal.googlesource.com/config-sync/open-telemetry/opentelemetry-collector-contrib/+/refs/tags/v0.89.0

It wasn't clear to me if there all get built or if i need to make a release branch to trigger a gke build.

karlkfi: hmm. maybe these don't exist

sdowell: It looks like that tag hasn't been imported yet. You don't need to make a release branch, it should be automated. It's possible something went wrong, I can take a closer look

sdowell: Either way, do we want to upgrade minor versions? I think we should be deliberate about upgrading minor versions with a review of potential breaking changes. If it's just for CVE fixes, it's lower risk to create a new patch. Not to say we shouldn't have a cadence for upgrading minor versions

karlkfi: the branch didn't exist. so i just made one... should i not have?

karlkfi: the louhi build isn't triggered yet. should i hit it manually or is it just delayed for polling?

sdowell:
replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1046

PR #1049, opened 2023-11-17
fix: setup/teardown keep finalizer on reconcile err

sdowell:
Prior to this change, if one of the managed objects failed to reconcile during setup or teardown the error would get swallowed and the reconciler-manager would proceed. This resulted in the finalizer exiting prematurely if one of the deletions timed out while waiting to reconcile.

For example, on autopilot it has been observed that the Deployment deletion may timeout during reconcile. When this happens, the error is swallowed and the finalizer exits even if the other managed objects have not been deleted (e.g. ServiceAccount, ClusterRoleBinding, etc).

Example logs:
```
        I1117 14:10:13.788408       1 rootsync_controller.go:376] controllers/RootSync "msg"="Deleting managed objects" "reconciler"="config-management-system/root-reconciler" "sync"="config-management-system/root-sync" "syncKind"="RootSync"
        I1117 14:10:13.788456       1 garbage_collector.go:65] controllers/RootSync "msg"="Deleting managed object" "object"="config-management-system/root-reconciler" "objectKind"="Deployment" "reconciler"="config-management-system/root-reconciler" "sync"="config-management-system/root-sync" "syncKind"="RootSync"
        I1117 14:10:13.889507       1 delete.go:155] Deleting object "object"="config-management-system/root-reconciler", "objectKind"="Deployment"
        I1117 14:10:13.902522       1 rootsync_controller.go:609] Changes to Deployment (config-management-system/root-reconciler) triggers a reconciliation for the RootSync (config-management-system/root-sync)
        I1117 14:10:13.902581       1 rootsync_controller.go:609] Changes to Deployment (config-management-system/root-reconciler) triggers a reconciliation for the RootSync (config-management-system/root-sync)
        I1117 14:10:13.985079       1 rootsync_controller.go:609] Changes to Deployment (config-management-system/root-reconciler) triggers a reconciliation for the RootSync (config-management-system/root-sync)
        I1117 14:10:13.985168       1 rootsync_controller.go:609] Changes to Deployment (config-management-system/root-reconciler) triggers a reconciliation for the RootSync (config-management-system/root-sync)
        I1117 14:10:14.170274       1 rootsync_controller.go:609] Changes to Deployment (config-management-system/root-reconciler) triggers a reconciliation for the RootSync (config-management-system/root-sync)
        I1117 14:10:14.170320       1 rootsync_controller.go:609] Changes to Deployment (config-management-system/root-reconciler) triggers a reconciliation for the RootSync (config-management-system/root-sync)
        W1117 14:10:43.904405       1 reflector.go:347] k8s.io/client-go/tools/watch/informerwatcher.go:146: watch of apps/v1, Kind=Deployment ended with: an error on the server ("unable to decode an event from the watch stream: context canceled") has prevented the request from succeeding
        E1117 14:10:43.904650       1 rootsync_controller.go:346] controllers/RootSync "msg"="Teardown waiting for event" "error"="deleting reconciler deployment: Deployment (config-management-system/root-reconciler) Terminating: failed to watch for deletion of Deployment: config-management-system/root-reconciler: timed out waiting for the condition" "object"="config-management-system/root-reconciler" "objectKind"="Deployment" "objectStatus"="Terminating" "reconciler"="config-management-system/root-reconciler" "sync"="config-management-system/root-sync" "syncKind"="RootSync"
        I1117 14:10:43.919739       1 rootsync_controller.go:1054] controllers/RootSync "msg"="Sync status update successful" "reconciler"="config-management-system/root-reconciler" "sync"="config-management-system/root-sync" "syncKind"="RootSync"
        I1117 14:10:43.919791       1 rootsync_controller.go:317] controllers/RootSync "msg"="Teardown successful" "reconciler"="config-management-system/root-reconciler" "sync"="config-management-system/root-sync" "syncKind"="RootSync"
        I1117 14:10:43.939844       1 reconciler_base.go:636] controllers/RootSync "msg"="Finalizer removal successful" "reconciler"="config-management-system/root-reconciler" "sync"="config-management-system/root-sync" "syncKind"="RootSync"
```

### DIFF
--- a/e2e/nomostest/testpredicates/predicates.go
+++ b/e2e/nomostest/testpredicates/predicates.go
@@ -41,6 +41,7 @@ import (
 	"kpt.dev/configsync/pkg/declared"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
+	"kpt.dev/configsync/pkg/reposync"
 	"kpt.dev/configsync/pkg/rootsync"
 	"kpt.dev/configsync/pkg/util/log"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
@@ -1146,6 +1147,35 @@ func RootSyncHasCondition(expected *v1beta1.RootSyncCondition) Predicate {
 }
 
 func validateRootSyncCondition(actual *v1beta1.RootSyncCondition, expected *v1beta1.RootSyncCondition) error {
+	e := expected.DeepCopy()
+	e.LastUpdateTime = actual.LastUpdateTime
+	e.LastTransitionTime = actual.LastTransitionTime
+	if diff := cmp.Diff(e, actual); diff != "" {
+		return fmt.Errorf("unexpected diff: %s", diff)
+	}
+	return nil
+}
+
+// RepoSyncHasCondition returns a Predicate that errors if the RepoSync does not
+// have the specified RepoSyncCondition. Fields such as timestamps are ignored.
+func RepoSyncHasCondition(expected *v1beta1.RepoSyncCondition) Predicate {
+	return func(o client.Object) error {
+		if o == nil {
+			return ErrObjectNotFound
+		}
+		rs, ok := o.(*v1beta1.RepoSync)
+		if !ok {
+			return WrongTypeErr(rs, &v1beta1.RepoSync{})
+		}
+		condition := reposync.GetCondition(rs.Status.Conditions, expected.Type)
+		if condition == nil {
+			return fmt.Errorf("RepoSyncCondition with type %s not found", expected.Type)
+		}
+		return validateRepoSyncCondition(condition, expected)
+	}
+}
+
+func validateRepoSyncCondition(actual *v1beta1.RepoSyncCondition, expected *v1beta1.RepoSyncCondition) error {
 	e := expected.DeepCopy()
 	e.LastUpdateTime = actual.LastUpdateTime
 	e.LastTransitionTime = actual.LastTransitionTime
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -36,6 +36,7 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
+	"kpt.dev/configsync/e2e/nomostest/testutils"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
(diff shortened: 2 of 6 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -393,17 +393,9 @@ func (r *RepoSyncReconciler) handleReconcileError(ctx context.Context, err error
 			logFieldObjectRef, statusErr.ID.ObjectKey.String(),
 			logFieldObjectKind, statusErr.ID.Kind,
 			logFieldObjectStatus, statusErr.Status)
-		switch statusErr.Status {
-		case kstatus.InProgressStatus, kstatus.TerminatingStatus:
```
sdowell, line 397: @karlkfi PTAL, as a I may be missing something here.

It seems to be the case, at least for finalizer teardown, that we should not swallow this error - as that results in the finalizer exiting prematurely even if not all managed objects have been deleted.

karlkfi: Returning the error causes the controller-runtime to immediately retry the Reconcile (with backoff). That's why it was returning nil. But there's nothing to do, so retrying immediately won't change anything.

Once the status is Stalled, anything waiting for completion should just keep waiting until the next even that removes the Stalled condition, right?

sdowell: We synced offline - TL;DR we decided to retain the Reconciling condition but return a `NoRetryError` so that the finalizer can be retained whilst also not immediately retriggering a Reconcile

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1155,6 +1156,35 @@ func validateRootSyncCondition(actual *v1beta1.RootSyncCondition, expected *v1be
...
+		}
+		return validateRepoSyncCondition(condition, expected)
+	}
+}
+
+func validateRepoSyncCondition(actual *v1beta1.RepoSyncCondition, expected *v1beta1.RepoSyncCondition) error {
+	e := expected.DeepCopy()
+	e.LastUpdateTime = actual.LastUpdateTime
+	e.LastTransitionTime = actual.LastTransitionTime
+	if diff := cmp.Diff(e, actual); diff != "" {
```
karlkfi, line 1182: `cmp.Equal(a, b, compare.IgnoreTimestampUpdates)` would work without needing to DeepCopy or change times. then diff on err.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -195,6 +196,162 @@ func TestReconcilerManagerTeardownInvalidRSyncs(t *testing.T) {
...
+			if apierrors.IsNotFound(err) { // Happy path - exit
+				return
+			}
+			nt.T.Fatal(err) // unexpected error
+		}
+		if testutils.RemoveFinalizer(rootSyncReconciler, nomostest.ConfigSyncE2EFinalizer) {
+			// The test failed to remove the finalizer. Remove to enable deletion.
+			if err := nt.KubeClient.Update(rootSyncReconciler); err != nil {
+				nt.T.Fatal(err)
+			} else {
```
karlkfi, line 227: dont need the else clause, since fatal panics.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -195,6 +196,162 @@ func TestReconcilerManagerTeardownInvalidRSyncs(t *testing.T) {
...
+	if err := nt.KubeClient.Update(repoSyncReconciler); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Deleting the RepoSync should result in a Reconciling error")
+	if err := nt.KubeClient.Delete(repoSync); err != nil {
+		nt.T.Fatal(err)
+	}
+	expectedCondition := v1beta1.RepoSyncCondition{
+		Message: "deleting reconciler deployment: Deployment (config-management-system/ns-reconciler-reconcile-timeout) Terminating: failed to watch for deletion of Deployment: config-management-system/ns-reconciler-reconcile-timeout: timed out waiting for the condition",
```
karlkfi, line 325: oof. verbose!
Would be nice to make this more user readable and reduce duplicate details, but not for this PR.

e2e/nomostest/testpredicates/predicates.go
sdowell: Done, although you do still need to change times due to the implementation of `compare.IgnoreTimestampUpdates`. I changed the function to pass by value instead of pointer

e2e/testcases/reconciler_manager_test.go
sdowell: Done

e2e/nomostest/testpredicates/predicates.go
karlkfi: Passing by value is problematic, because it's only a shallow copy and the k8s meta timestamps are themselves structs that contain a Time struct as a field. So without the DeepCopy here you've pointed the actual Times at the expected Times, meaning they will change in sync, which might be a gotcha if used elsewhere.

Not sure what you mean by the impl of IgnoreTimestampUpdates not working tho. Does it need to be run with cmpopts.EquateEmpty()? or does it have a bug?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -621,7 +627,12 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
 
 	if controllerutil.ContainsFinalizer(syncObj, metadata.ReconcilerManagerFinalizer) {
 		// Our finalizer is present, so delete managed resource objects.
-		if err := teardownFn(ctx); err != nil {
+		var nre *NoRetryError
+		if err := teardownFn(ctx); errors.As(err, &nre) {
```
karlkfi, line 631: nit: might be slightly easier to read as:
```
		if err := teardownFn(ctx); err != nil {
			if errors.As(err, &nre) {
				r.logger(ctx).Error(err, "Teardown failed (waiting for resource status to change)")
				// Return nil to avoid triggering immediate retry.
				return nil
			}
			return errors.Wrap(err, "teardown failed")
		}
```

e2e/nomostest/testpredicates/predicates.go
sdowell: Both values must either be empty or not-empty. So since the actual value is non-empty and the expected value is empty it returns not-equal.

We could update the `expected` struct to contain an arbitrary non-nil value, but that seems less clear IMO. This is a pretty small struct so a DeepCopy doesn't seem that problematic IMO

pkg/reconcilermanager/controllers/reconciler_base.go
sdowell: Agreed, done

karlkfi (approved):
/lgtm

PR #1052, opened 2023-11-20
fix: apply managed resource labels consistently

sdowell:
This change updates the label logic for the resources managed by reconciler-manager to apply a consistent set of rules.

- Apply the sync-kind and sync-namespace label for shared RBAC bindings, leaving out the sync-name field given that multiple RSyncs may be mapped to the same binding.

### DIFF
--- a/pkg/reconcilermanager/controllers/rootsync_controller.go
+++ b/pkg/reconcilermanager/controllers/rootsync_controller.go
@@ -899,7 +899,7 @@ func (r *RootSyncReconciler) manageRBACBindings(ctx context.Context, reconcilerR
 	}
 	if len(roleRefs) == 0 {
 		// Backwards compatible behavior to default to cluster-admin
-		if err := r.upsertSharedClusterRoleBinding(ctx, RootSyncLegacyClusterRoleBindingName, "cluster-admin", reconcilerRef); err != nil {
+		if err := r.upsertSharedClusterRoleBinding(ctx, RootSyncLegacyClusterRoleBindingName, "cluster-admin", reconcilerRef, rsRef); err != nil {
 			return err
 		}
 		// Clean up any RoleRefs created previously that are no longer declared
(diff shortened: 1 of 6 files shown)


pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -956,12 +956,17 @@ func (r *RootSyncReconciler) manageRBACBindings(ctx context.Context, reconcilerR
 	return nil
 }
 
-func (r *RootSyncReconciler) upsertSharedClusterRoleBinding(ctx context.Context, name, clusterRole string, reconcilerRef types.NamespacedName) error {
+func (r *RootSyncReconciler) upsertSharedClusterRoleBinding(ctx context.Context, name, clusterRole string, reconcilerRef, rsRef types.NamespacedName) error {
 	crbRef := client.ObjectKey{Name: name}
 	childCRB := &rbacv1.ClusterRoleBinding{}
 	childCRB.Name = crbRef.Name
 
+	labelMap := ManagedObjectLabelMap(r.syncKind, rsRef)
```
nan-yu, line 964: In the 1-pager, sync-namespace can be omitted for RootSync resources. Is it still the case?

sdowell: In my first pass I did that, but discovered that this seems to break many of the metrics assertions which apparently rely on the sync-namespace label. I'm planning to update the 1-pager accordingly. I'm leaving the sync-namespace out of this change as it appears it may be a breaking change.

nan-yu (commented):
Do we want to add labels to `upsertHelmConfigMaps`?

pkg/reconcilermanager/controllers/rootsync_controller.go
nan-yu: ACK

sdowell:
> Do we want to add labels to `upsertHelmConfigMaps`?

@nan-yu It looks like that already adds the standard labels. Or am I missing something?

nan-yu:
> > Do we want to add labels to `upsertHelmConfigMaps`?
> 
> Nan Yu It looks like that already adds the standard labels. Or am I missing something?

I missed that. Right, the `labelMap` is already passed into `upsertHelmConfigMaps`.

nan-yu (approved):
/lgtm

janetkuo (approved):

sdowell:
FYI @gbmeuk 

This can replace https://github.com/GoogleContainerTools/kpt-config-sync/pull/981 and should resolve your issue

PR #1057, opened 2023-11-23
fix: Delay RSync controller registration

karlkfi:
Add a CRD controller that watches for CRD updates and registers the RSync controllers when the RSync CRDs become Established, meaning the resources are registered and availible for use.

This allows the RootSync and RepoSync controllers to come up independently, which can unblock uninstall when the CRDs are being deleted simultaniously, in case the reconciler-manager is restarted or rescheduled.

### DIFF
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -33,7 +33,9 @@ import (
 	"k8s.io/client-go/util/retry"
 	"kpt.dev/configsync/e2e"
 	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/policy"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/e2e/nomostest/testutils"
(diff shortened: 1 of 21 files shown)


pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -651,21 +652,21 @@ func (r *RootSyncReconciler) mapSecretToRootSyncs(secret client.Object) []reconc
 	}
 
 	attachedRootSyncs := &v1beta1.RootSyncList{}
-	listOps := &client.ListOptions{
-		FieldSelector: fields.OneTermEqualSelector(gitSecretRefField, sRef.Name),
-		Namespace:     sRef.Namespace,
-	}
-	if err := r.client.List(context.Background(), attachedRootSyncs, listOps); err != nil {
+	if err := r.client.List(ctx, attachedRootSyncs, client.InNamespace(sRef.Namespace)); err != nil {
```
sdowell, line 655: Could you pull out these snippets for RootSync/RepoSync secret mapping to a separate change? They seem unrelated to the controller registration.

Also, why is this preferred to the field selector?

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -401,23 +400,13 @@ func (r *RootSyncReconciler) deleteManagedObjects(ctx context.Context, reconcile
...
-		if rs.Spec.Git == nil || v1beta1.GetSecretName(rs.Spec.Git.SecretRef) == "" {
-			return nil
-		}
-		return []string{rs.Spec.Git.SecretRef.Name}
-	}); err != nil {
-		return err
-	}
+// Setup RootSync controller with reconciler-manager.
+func (r *RootSyncReconciler) Setup(_ controllerruntime.Manager) error {
+	return nil
```
sdowell, line 405: Do you anticipate that `Setup` will need to do something in the future? I'm curious why we need `Setup` as part of the interface if it's not used by RepoSyncReconciler and RootSyncReconciler

karlkfi: I pulled it out to put the indexing there, but then discovered indexing didn't work without CRDs either. So I had to remove the indexes completely.

I could change this back to the old SetupWithManager tho, to make the PR a little smaller. I kinda like the explicitness of the interface saying which one happens before and which happens after c-m Start, but it's not required.

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: I suppose I could remove the indexes as a separate PR, but it's only really required because of this PR.

Would that make it easier to review?

pkg/reconcilermanager/controllers/controller.go
```diff
@@ -23,8 +23,12 @@ import (
 // SetupWithManager
 type Controller interface {
 	reconcile.Reconciler
-	// SetupWithManager registers the controller with the controller-manager
-	SetupWithManager(mgr controllerruntime.Manager, watchFleetMembership bool) error
+	// Setup error handlers, indexers, and transformers with controller-manager.
```
nan-yu, line 26: This doesn't seem to be the case anymore. We can delete the comment.
Also, if `Setup` doesn't do real work, do we still need the interface?

e2e/testcases/reconciler_manager_test.go
```diff
@@ -1116,3 +1120,91 @@ func filterResourceMap(resourceMap map[string]v1beta1.ContainerResourcesSpec, co
...
+	reconcilerManagerKey := client.ObjectKey{
+		Name:      reconcilermanager.ManagerName,
+		Namespace: configsync.ControllerNamespace,
+	}
+
+	t.Log("Validate the reconciler-manager deployment is healthy")
+	nt.Must(nt.Validate(reconcilerManagerKey.Name, reconcilerManagerKey.Namespace, &appsv1.Deployment{},
+		testpredicates.StatusEquals(nt.Scheme, kstatus.CurrentStatus)))
+
+	nt.Must(nomostest.UninstallRootSyncCRD(nt))
```
sdowell, line 1139: This should also register a cleanup to call `InstallRootSyncCRD` in case it errors before reaching the end of the test

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -0,0 +1,185 @@
...
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	if err := r.client.Get(ctx, req.NamespacedName, crd); err != nil {
+		if apierrors.IsNotFound(err) {
+			// CRD was deleted before event could be handled
+			return reconcile.Result{}, nil
+		}
+		// Retry with backoff
+		return reconcile.Result{}, fmt.Errorf("getting %s CRD: %w", req.Name, err)
+	}
+	if !CRDIsEstablished(crd) {
```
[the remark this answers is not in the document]
karlkfi: I could probably make a precondition for it, yeah.

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -0,0 +1,185 @@
...
+				return false
+			}
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			switch e.ObjectNew.GetName() {
+			case RootSyncCRDName, RepoSyncCRDName:
+				return true
+			default:
+				return false
+			}
```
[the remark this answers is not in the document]
karlkfi: update has a different field name for the object

e2e/testcases/reconciler_manager_test.go
```diff
@@ -1116,3 +1120,91 @@ func filterResourceMap(resourceMap map[string]v1beta1.ContainerResourcesSpec, co
...
+		return validateReconcilerManagerPodStatusCurrent(nt)
+	})
+
+	t.Log("Waiting for the reconciler-manager deployment to become healthy")
+	nt.Must(nt.Watcher.WatchForCurrentStatus(kinds.Deployment(),
+		reconcilerManagerKey.Name, reconcilerManagerKey.Namespace))
+
+	nt.Must(nomostest.InstallRootSyncCRD(nt))
+
+	t.Log("Validate the reconciler-manager deployment stays healthy for 30s")
```
sdowell, line 1176: As we discussed offline, this assertion could be improved by instead creating a `RootSync` and ensuring that it reconciles

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			// TODO: Unregister controllers, if possible
+			return false
+		},
+	}
+	return controllerruntime.NewControllerManagedBy(mgr).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: 1,
```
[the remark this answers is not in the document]
karlkfi: nah, but we are on the others so i just left it for explicitness. remember in the past we used a mutex to do the same, which was redundant, just out of ignorance.

e2e/testcases/reconciler_manager_test.go
karlkfi: i think test setup installed configsync from scratch every time

sdowell: That happens at the beginning and end of the test suite, but for shared test environments it doesn't get installed between each test case. So if this test failed then the RootSync CRD may be missing for subsequent tests which use the same cluster

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -0,0 +1,185 @@
...
+type CRDReconciler struct {
+	loggingController
+
+	client               client.Client
+	manager              controllerruntime.Manager
+	rootSyncController   Controller
+	repoSyncController   Controller
+	watchFleetMembership bool
+
+	registerLock       sync.Mutex
```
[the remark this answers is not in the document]
nan-yu: controller-runtime uses a for-loop to iterate through `MaxConcurrentReconciles`. Given that `MaxConcurrentReconciles` is set to 1, so there is only one goroutine to process work items from the queue: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/internal/controller/controller.go#L222.

The goroutine should process one object at a time, so the `Reconcile` function should never run concurrently.

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: Removed the Setup method.

pkg/reconcilermanager/controllers/crd_controller.go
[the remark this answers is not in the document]
karlkfi: Replaced with a simpler predicate.

pkg/reconcilermanager/controllers/crd_controller.go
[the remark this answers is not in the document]
karlkfi: Removed

pkg/reconcilermanager/controllers/controller.go
karlkfi: Removed Setup. I initially moved the indexers there, but later realized they required the CRDs. So I had to remove the indexes all together.

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -0,0 +1,185 @@
...
+
+var _ reconcile.Reconciler = &CRDReconciler{}
+
+// CRDReconciler reconciles OpenTelemetry ConfigMaps.
+type CRDReconciler struct {
+	loggingController
+
+	client               client.Client
+	manager              controllerruntime.Manager
+	rootSyncController   Controller
```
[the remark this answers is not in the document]
karlkfi: Replaced this with a more generic map of handler funcs. This makes for less complexity in the CRDController.

pkg/reconcilermanager/controllers/crd_controller.go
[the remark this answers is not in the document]
karlkfi: After my latest refactor, to a more generic pattern, I left the lock in, just in case we want to later add new handlers after the manager is started.

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -0,0 +1,185 @@
...
+	repoSyncController   Controller
+	watchFleetMembership bool
+
+	registerLock       sync.Mutex
+	rootSyncRegistered bool
```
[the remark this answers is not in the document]
karlkfi: See latest refactor.

I don't know if it's a great idea to move the complexity down into the `reconcilerBase` if the other controllers don't need it.

e2e/testcases/reconciler_manager_test.go
karlkfi: re-wrote the test here to validate RootSync and RepoSync spec updates and syncing as proof of reconciler-manager health.

e2e/testcases/reconciler_manager_test.go
```diff
@@ -1116,3 +1118,139 @@ func filterResourceMap(resourceMap map[string]v1beta1.ContainerResourcesSpec, co
...
+//
+// This behavior is made possible by two features:
+// - The CRD controller allows delaying RSync controller registration
+// - The controller-manager retries failing watches when the resource is deleted
+//
+// Since the reconciler-manager may be rescheduled at any time, especially on
+// autopilot, we can't directly validate that it stays healthy continuously,
+// only that it eventually performs its duties as expected.
+//
+// Test Overview:
```
sdowell, line 1137: Thanks for updating this, the test scenarios and assertions are greatly improved

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -752,9 +740,28 @@ func (r *RepoSyncReconciler) mapConfigMapToRepoSyncs(obj client.Object) []reconc
 	return requests
 }
 
+func repoSyncHelmValuesFileNames(rs *v1beta1.RepoSync) []string {
```
sdowell, line 743: nit: this is only used to be wrapped with `slices.Contains`, so this function could probably be written as something like `helmValuesFileRefsContains`

sdowell (approved):
A couple nits but lgtm. I'll let @nan-yu take another look and grant LGTM since she was reviewing this as well

nan-yu (approved):
/lgtm

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: yeah, I considered that, it would have been slightly more efficient, but I thought this looked a slight bit more readable/reusable. I don't think the list will ever get big enough to actually impact performance significantly.

PR #1059, opened 2023-11-29
fix: remove a false condition that sets the dynamic mode

nan-yu:
`hasDynamicNSSelector` should only be set to true when the mode is set to `dynamic` explicitly. It was probably a merging issue that mistakenly sets the flag to true for unknown mode.

### DIFF
--- a/pkg/validate/scoped/hydrate/namespace_hydrator.go
+++ b/pkg/validate/scoped/hydrate/namespace_hydrator.go
@@ -112,7 +112,6 @@ func buildSelectorMap(ctx context.Context, c client.Client, objs *objects.Scoped
 				hasDynamicNSSelector = true
 			} else if nsSelector.Spec.Mode != "" && nsSelector.Spec.Mode != v1.NSSelectorStaticMode {
 				errs = status.Append(errs, selectors.UnknownNamespaceSelectorModeError(nsSelector))
-				hasDynamicNSSelector = true
 			}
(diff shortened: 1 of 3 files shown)


pkg/validate/scoped/hydrate/namespace_hydrator.go
```diff
@@ -112,7 +112,6 @@ func buildSelectorMap(ctx context.Context, c client.Client, objs *objects.Scoped
 				hasDynamicNSSelector = true
 			} else if nsSelector.Spec.Mode != "" && nsSelector.Spec.Mode != v1.NSSelectorStaticMode {
```
sdowell, line 113: It does seem right that we shouldn't set `hasDynamicNSSelector = true` here, but I don't think this is why the test is flaking.

After taking a closer look, I see that the test is using a shared variable with other test cases in the file. Since these tests run in parallel, it's not safe to use a shared variable in this manner. Each test should get its own copy of the object.

https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/namespace_selectors_test.go#L48

sdowell (approved):
/lgtm

karlkfi:
Does this need to be cherry-picked for 1.17.1?

nan-yu:
> Does this need to be cherry-picked for 1.17.1?

It was already cherry-picked in #1076.

PR #1067, opened 2023-12-04
fix: pass correct argument for otel-agent logging

sdowell:
The opentelemetry image does not accept -v as the argument for logging verbosity as the other containers do. This change updates it to set the appropriate argument using the zap logging library.


pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -162,8 +162,8 @@ type ContainerLogLevelOverride struct {
 
 	// logLevel specifies the log level override value for the container.
 	// The default value for git-sync container is 5, while all other containers will default to 0.
```
nan-yu, line 164: Is the default log level for otel-agent -1 or 0?

sdowell: The default is 0, which maps to `INFO`

nan-yu (approved):
/lgtm

PR #1066, opened 2023-12-04
fix: Prevent RSync from premature Current status

karlkfi:
Setting the default observedGeneration to 0 should ensure that it is initialized even when using Unstructured resources. This should prevent kstatus.Compute from considering the object to be Current before the controller has updated the status. This should prevent the Config Sync reconciler from considering the object to be reconciled and returning prematurely. This should prevent any dependencies from being applied or deleted prematurely, avoiding out of order problems.

PR #1068, opened 2023-12-06
chore: Validate RSync reconciliation status

karlkfi:
Add unit tests for changes from #1066

Now that observedGeneration has a default, the status should always be non-Current until reconciliation is complete.

Validate this in setup by running a deployment controller simulator and validator in parallel.

Validate this in teardown by blocking teardown with a ServiceAccount finalizer.

b/315199932

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
@@ -25,13 +25,19 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/util/retry"
+	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
+	"kpt.dev/configsync/pkg/reposync"
+	syncerFake "kpt.dev/configsync/pkg/syncer/syncertest/fake"
 	watchutil "kpt.dev/configsync/pkg/util/watch"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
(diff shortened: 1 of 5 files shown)


pkg/reconcilermanager/controllers/rootsync_controller_manager_test.go
```diff
@@ -78,46 +84,89 @@ func TestRootSyncReconcilerDeploymentLifecycle(t *testing.T) {
...
+	err = fakeClient.Get(ctx, crbKey, saObj)
+	require.NoError(t, err)
+
+	// Validate RootSync still exists (means reconciler-manager finalizer blocks waiting for ServiceAccount not found)
+	err = fakeClient.Get(ctx, rsKey, rs)
+	require.NoError(t, err)
+
+	t.Log("Simulating teardown unblocked")
+	require.True(t, controllerutil.RemoveFinalizer(saObj, testFinalizer))
+	err = fakeClient.Update(ctx, saObj)
```
tiffanny29631, line 165: Why updating the service account again after finalizer removal?

karlkfi: RemoveFinalizer only modifies the local object. Gotta send the update to the server.

pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
```diff
@@ -494,3 +543,61 @@ func testRepoSyncDriftProtection(t *testing.T, exampleObj client.Object, objKeyF
...
+
+			// Wait for RepoSync to be reconciled and Current.
+			// This ensures that blocking with kstatus works on RepoSyncs,
+			// at least for the reconciler-manager Reconciling condition.
+			t.Log("verifying the RepoSync is ready")
+			result, err := status.Compute(uObj)
+			if err != nil {
+				return fmt.Errorf("computing RepoSync status: %w", err)
+			}
+			if result.Status != status.CurrentStatus {
```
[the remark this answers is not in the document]
karlkfi: The way the WatchUntil condition works is that this condition function is called for each new event received and will keep watching until `return nil`. It also caches the last error and returns if there's a timeout/cancel. So returning an error means to keep watching.

pkg/reconcilermanager/controllers/rootsync_controller_manager_test.go
tiffanny29631: Ack.

pkg/reconcilermanager/controllers/reposync_controller_manager_test.go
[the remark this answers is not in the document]
tiffanny29631: Ack

tiffanny29631 (approved):
/lgtm

PR #1069, opened 2023-12-07
chore: Make fake client more like the real client

karlkfi:
- In the future, we should change the tests to avoid needing GVK to be set on typed objects, but for now, ensure the fake clients consistently set the GVK on all objects.
- Update the fake clients to copy the new object state back into the input object.
- Update the fake clients to consistently skip updates where nothing was changed except the resourceVersion or generation.
- Update the fake clients to ignore spec updates when updating status.
- Update the fake clients to always set defaults from the scheme on objects before persisting them.
- Add some helper functions to the DynamicClient similar to the other Client.
- Remove minimizing from the fake client storage. This is no longer needed now that defaults are being set.
- Update unit tests to Get the object and then modify it with expected field changes, rather than needing to specify all the default values in the expectations.

Hopefully these changes make the unit tests a little less flakey, since it makes the behavior more predictable.


sdowell (approved):
/lgtm

PR #1070, opened 2023-12-08
fix: prevent remediator watch race with applier

sdowell:
This change starts the remediator watches before the applier applies objects to the cluster. This prevents a race condition which can occur if an update is made after the applier runs but before watches are started. If this race condition occurs, it is possible that the remediator will miss the watch event and drift is not corrected. By starting the watches before the applier runs, applier updates should result in a remediator no-op but actual drift will be remediated.

### DIFF
--- a/pkg/parse/updater.go
+++ b/pkg/parse/updater.go
@@ -186,6 +186,24 @@ func (u *Updater) update(ctx context.Context, cache *cacheForCommit) status.Mult
 		}
 	}
 
+	// Update the resource watches (triggers for the Remediator).
+	// Start watches before the applier runs to detect any drift that may occur
+	// immediately after an apply.
+	var watchErrs status.MultiError
+	if !cache.watchesUpdated {
+		declaredGVKs, _ := u.Resources.DeclaredGVKs()
+		// Delay returning watchErrs until after resources are applied.
+		// This enables applying a CRD and CR in the same commit, because watches
+		// for a CRD cannot be started before the CRD exists. In this case,
+		// the updater will need to run a second time to start watches before apply.
+		watchErrs = u.watch(ctx, declaredGVKs)
+		// Only mark the watches as updated if there were no (non-blocking) parse errors.
+		// This ensures the update will be retried until parsing fully succeeds.
+		if watchErrs == nil && cache.parserErrs == nil {
+			cache.watchesUpdated = true
+		}
+	}
+
 	// Apply the declared resources
 	if !cache.applied {
 		declaredObjs, _ := u.Resources.DeclaredObjects()
@@ -195,23 +213,13 @@ func (u *Updater) update(ctx context.Context, cache *cacheForCommit) status.Mult
 		}
 		// Only mark the commit as applied if there were no (non-blocking) parse errors.
 		// This ensures the apply will be retried until parsing fully succeeds.
-		if cache.parserErrs == nil {
+		if watchErrs == nil && cache.parserErrs == nil {
 			cache.applied = true
 		}
 	}
 
-	// Update the resource watches (triggers for the Remediator).
-	if !cache.watchesUpdated {
-		declaredGVKs, _ := u.Resources.DeclaredGVKs()
-		err := u.watch(ctx, declaredGVKs)
-		if err != nil {
-			return err
-		}
-		// Only mark the watches as updated if there were no (non-blocking) parse errors.
-		// This ensures the update will be retried until parsing fully succeeds.
-		if cache.parserErrs == nil {
-			cache.watchesUpdated = true
-		}
+	if watchErrs != nil {
+		return watchErrs
 	}
 
 	// Restart remediator workers.
--- a/pkg/remediator/watch/filteredwatcher.go
+++ b/pkg/remediator/watch/filteredwatcher.go
@@ -72,7 +72,7 @@ const maxWatchRetryFactor = 18
 // Runnable defines the custom watch interface.
 type Runnable interface {
 	Stop()
-	Run(ctx context.Context) status.Error
+	Run(ctx context.Context, startCh chan error) status.Error
 	ManagementConflict() bool
 	SetManagementConflict(object client.Object, commit string)
 	ClearManagementConflict()
@@ -109,6 +109,7 @@ type filteredWatcher struct {
 	stopped            bool
 	managementConflict bool
 	conflictHandler    conflict.Handler
+	cancel             func()
 }
 
 // filteredWatcher implements the Runnable interface.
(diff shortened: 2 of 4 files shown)


pkg/parse/updater.go
```diff
@@ -214,6 +202,20 @@ func (u *Updater) update(ctx context.Context, cache *cacheForCommit) status.Mult
 		}
 	}
 
+	// Apply the declared resources
+	if !cache.applied {
+		declaredObjs, _ := u.Resources.DeclaredObjects()
+		_, err := u.apply(ctx, declaredObjs, cache.source.commit)
```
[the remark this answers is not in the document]
sdowell: The watches are by GVK, so we can't necessarily stop watches for individual objects that fail to be applied (unless _all_ objects of a given GVK failed to apply). We could potentially ignore watch events for a object which failed to apply, although I'm not sure what the value add would be for that new behavior.

What is the motivation for the behavior you propose?

pkg/parse/updater.go
```diff
@@ -186,21 +186,9 @@ func (u *Updater) update(ctx context.Context, cache *cacheForCommit) status.Mult
...
-		}
-		// Only mark the commit as applied if there were no (non-blocking) parse errors.
-		// This ensures the apply will be retried until parsing fully succeeds.
-		if cache.parserErrs == nil {
-			cache.applied = true
-		}
-	}
-
 	// Update the resource watches (triggers for the Remediator).
+	// Start watches before the applier runs to detect any drift that may occur
```
sdowell, line 190: Ah looks like this fails for CRDs if a CRD and CR are added in the same commit, because the watch can't be started for a CRD that does not exist yet. @nan-yu do you have any thoughts for a workaround?

pkg/parse/updater.go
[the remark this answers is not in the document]
nan-yu: I am concerned that we may miss some events.

- For actuation, it works alright. If an object failed to be applied, while the watch is still running. Any drift will be corrected by the remediator.
- For pruning, when the apply (prune) failed, the watch was not stopped before, but it is stopped with the change. Any drift before the next retry won't get remediated.

pkg/parse/updater.go
nan-yu: Hmm, it seems we have to update watches in two steps. Before apply, we update watches with known GVKs, and update watches again for those new registered GVKs.

nan-yu: I'm wondering are we overthinking this issue? Even though  `u.apply` and `u.watch` is not atomic, the possibility that an applied object being mutated before the watch is running seems very low.

Regarding the prune issue and CRD/CR in the same commit issue, I'm okay with apply first and then update watches.

karlkfi: The mitigation for this is a lot more complicated than you might think, especially because the mapper needs resetting when new CRDs are added or removed.

For cli-utils, I wrote the DefaultStatusWatcher & ObjectStatusReporter to handle this:
- https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/watcher/object_status_reporter.go#L47
- https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/watcher/default_status_watcher.go

We could use the DefaultStatusWatcher/ObjectStatusReporter here, but it might require significant code changes. And it would duplicate all the watches being made by the applier.

karlkfi: FWIW, the duplication of watches was one of the primary drivers for my suggesting merging the remediator into cli-utils applier and making it do continuous reconciliation, instead of just periodic applies. But that project is pretty significant and hasn't been prioritized.

sdowell: So looking deeper into this, there's a few problems:

1. If the watch [fails to start](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/remediator/watch/filteredwatcher.go#L387) any calls to [Stop will result in a deadlock](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/remediator/watch/filteredwatcher.go#L220), because `start` doesn't let go of the lock until the watch successfully starts. This can be worked around relatively simply by passing in a cancellable context into `startWatch` and calling `cancel` in `Stop`.
2. Errors from starting the watch are [not returned synchronously](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/remediator/watch/manager.go#L206), so we don't actually know if any of the watches successfully started after calling `UpdateWatches`. This means we don't actually have a guarantee that the remediator watches are running before we invoke the applier. Furthermore [the watcher may fail and restart watches](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/remediator/watch/filteredwatcher.go#L298) at any point, which means that we could be missing watch events during watcher restarts.

[1] is worth fixing regardless IMO, however [2] adds quite a bit of complexity to this problem.

sdowell: Regarding [2], does passing in the [resourceVersion](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/remediator/watch/filteredwatcher.go#L298) make it so that we capture events between watch restarts?

pkg/parse/updater.go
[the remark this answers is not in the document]
karlkfi: This problem isn't solvable with the current design. If we update the watches before the apply, they will fail on new resource types and stop watching pruned resources prematurely. If we update the watched after the apply, we might miss events. When I refactored to add remediator pausing during apply, I opted to update the watched after the apply because missing events was the lesser evil.

The only options I know of to mitigate this are:
1. duplicate the watch code in the remediator to act like the applier watches, which watch CRDs and add/remove watches dynamically (more resource usage and complexity)
2. merge the applier and remediator into a single continuous reconciler (b/205185055)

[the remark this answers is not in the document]
sdowell: > For pruning, when the apply (prune) failed, the watch was not stopped before, but it is stopped with the change. Any drift before the next retry won't get remediated.

> If we update the watches before the apply, we miss events for pruned resource types and error on new resource types.

I don't think these statements are strictly true. We pause the remediator and update the declared resource cache before applying, so we will miss events for pruned resources types regardless of when the watches start. We would need to change the logic for the declared resource cache as well if we want to continue remediating pruned objects after a failed apply.

The updated implementation will handle new resource types in the following way: on the first pass of adding a new resource type the watches will fail and the resources will still be applied. Because there was an error setting up the watches, the reconciler will retry. On the second pass, the watches should succeed to be created and the applier will re-run - at which point the remediator can correct for drift with a clean initial state (because the watches were started before the applier ran the second time).

PR #1071, opened 2023-12-08
Add unit test scenario for APIVERSION

victorpras:
Add more APIVERSION failing test scenario for invalid scope and missing data, and add more valid entry for standard scenario.


nan-yu (approved):
/lgtm

PR #1072, opened 2023-12-13
ci: add test runner iam to terraform config

sdowell:
This reflects the current project level permissions which have been manually applied.


nan-yu (approved):
/lgtm

PR #1073, opened 2023-12-14
test: annotate git-server as not safe to evict

sdowell:
The git-server is susceptible to eviction on autopilot, which can cause flakiness for many test cases. This annotation prevents autopilot workloads from being evicted for up to 7 days. The git-server deployment is scoped to a test execution, and therefore should not live longer than a few hours in the worst case.

### DIFF
--- a/e2e/nomostest/git-server.go
+++ b/e2e/nomostest/git-server.go
@@ -35,6 +35,7 @@ const testGitServer = "test-git-server"
 const testGitServerImage = testing.TestInfraArtifactRegistry + "/git-server:v1.0.0"
 const testGitHTTPServer = "http-git-server"
 const testGitHTTPServerImage = testing.TestInfraArtifactRegistry + "/http-git-server:v1.0.0-b3b4984cd"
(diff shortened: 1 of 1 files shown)


e2e/nomostest/git-server.go
```diff
@@ -35,6 +35,7 @@ const testGitServer = "test-git-server"
 const testGitServerImage = testing.TestInfraArtifactRegistry + "/git-server:v1.0.0"
 const testGitHTTPServer = "http-git-server"
 const testGitHTTPServerImage = testing.TestInfraArtifactRegistry + "/http-git-server:v1.0.0-b3b4984cd"
+const safeToEvictAnnotation = "cluster-autoscaler.kubernetes.io/safe-to-evict"
```
karlkfi, line 38: This may not be safe to add to normal clusters, because they allow persistence for much much longer. It's only limited to 7 days on autopilot. So we probably want a check for that.

sdowell: Done

karlkfi (approved):
/lgtm

PR #1075, opened 2023-12-14
test: fix multi sync test repo initialization

sdowell:
Prior to this change the repositories used by the test were initialized before being tracked in RootRepos/NonRootRepos. This meant that if the git-server had to be repaired at any point during the test, the untracked repos would not be recreated. This updates the test to initialize the repos at the same time that they are tracked.


nan-yu (approved):
/lgtm

PR #1076, opened 2023-12-14
fix: remove a false condition that sets the dynamic mode (#1059)

nan-yu:
`hasDynamicNSSelector` should only be set to true when the mode is set to `dynamic` explicitly. It was probably a merging issue that mistakenly sets the flag to true for unknown mode.

It also updates the namespace_selectors_test to avoid mutating the shared variable in two test cases, which causes test flakiness.


sdowell (approved):
/lgtm

PR #1077, opened 2023-12-14
test: auto-rotate service account key & secret

karlkfi:
Updated TestHelmARTokenAuth to include code to auto-rotate expired service account keys and also repair a few edge cases, like no enabled secret versions, invalid secret version, latest secret version disabled/destroyed, etc. So as long as the service account exists and the secret exists, the code should be able to re-generate a valid key on-demand and add it as a new secret version.

Note: Because these steps are not atomic and might be run in parallel in the same project, it's possible that the test might fail due to concurrent regeneration, but it's unlikely due to the order of the steps. It's more likely that a duplicate key is ocationally generated, which shouldn't cause test failure. If this happens frequently, it might exceed the maximum 10 keys per account, but otherwise the keys will eventually be auto-deleted after they expire.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -15,17 +15,23 @@
 package e2e
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
+	"kpt.dev/configsync/e2e"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/artifactregistry"
-	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/iam"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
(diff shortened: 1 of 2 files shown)


e2e/testcases/helm_sync_test.go
```diff
@@ -879,3 +913,148 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
...
 	}
 }
+
+// fetchServiceAccountKeyFile downloads a service account key from google
+// secret manager, using the specified `gsaKeySecretID`. If the key has expired,
+// a new key is generated and stored as a new version of the secret in secret
+// manager. If successful, the old expired key is deleted. Then the path to the
+// new key file is returned. The key file will be deleted when the current
+// test ends.
+func fetchServiceAccountKeyFile(nt *nomostest.NT, projectID, gsaKeySecretID, gsaEmail, gsaName string) (string, error) {
```
sdowell, line 923: Might be worth wrapping this function in a mutex to synchronize threads in a given test execution, in case someone adds another call to this function in the future.

e2e/testcases/helm_sync_test.go
```diff
@@ -879,3 +913,148 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
...
+
+	nt.T.Log("Listing enabled service account key versions in Secret Manager")
+	out, err := nt.Shell.ExecWithDebug("gcloud", "secrets", "versions", "list",
+		gsaKeySecretID,
+		"--filter", "state=ENABLED",
+		"--limit", "1",
+		"--format", "value(name)",
+		"--project", projectID)
+	if err != nil {
+		// TODO: create new key and secret if no secret found
```
sdowell, line 938: Might be worth doing this now to avoid a gotcha when bootstrapping new environments

sdowell (approved):
/lgtm

PR #1078, opened 2023-12-15
test: address edge cases for e2e secret rotation

sdowell:
This change adds the GCP iam needed by the e2e test runner GSA, and adds
some logic to harden the secret rotation for certain edge cases.

Builds on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1077

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -914,6 +915,10 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
 	}
 }
 
+// Synchronize Secret Manager operations across test threads. Note that there
+// can still be conflicts across periodic jobs in Prow since the Secret is shared.
+var secretManagerMux sync.Mutex
+
 // fetchServiceAccountKeyFile downloads a service account key from google
 // secret manager, using the specified `gsaKeySecretID`. If the key has expired,
 // a new key is generated and stored as a new version of the secret in secret
@@ -921,6 +926,8 @@ func helmChartVersion(chartVersion string) nomostest.Sha1Func {
 // new key file is returned. The key file will be deleted when the current
 // test ends.
 func fetchServiceAccountKeyFile(nt *nomostest.NT, projectID, gsaKeySecretID, gsaEmail, gsaName string) (string, error) {
+	secretManagerMux.Lock()
+	defer secretManagerMux.Unlock()
 	// Use a temp file to hold the secret to avoid logging the secret contents
 	gsaKeyFilePath := filepath.Join(nt.TmpDir, fmt.Sprintf("%s.json", gsaName))
 	nt.T.Cleanup(func() {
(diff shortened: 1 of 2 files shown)


e2e/testinfra/terraform/prow/service_accounts.tf
```diff
@@ -23,11 +23,17 @@ resource "google_project_iam_member" "test-runner-iam" {
...
 
   role    = each.value
   member  = "serviceAccount:e2e-test-runner@oss-prow-build-kpt-config-sync.iam.gserviceaccount.com"
   project = data.google_project.project.id
 }
+
+resource "google_service_account_iam_member" "admin-account-iam" {
+  service_account_id = "projects/${data.google_project.project.id}/serviceAccounts/e2e-test-ar-reader@${data.google_project.project.name}.iam.gserviceaccount.com"
+  role               = "roles/iam.serviceAccountKeyAdmin"
+  member             = "serviceAccount:e2e-test-runner@oss-prow-build-kpt-config-sync.iam.gserviceaccount.com"
```
karlkfi, line 38: should the project id be interpolated from the project var?

sdowell: Not for this service account, because the prow project and the test-runner SA are singletons

karlkfi:
You can rebase on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1077 now

e2e/testcases/helm_sync_test.go
```diff
@@ -935,14 +942,27 @@ func fetchServiceAccountKeyFile(nt *nomostest.NT, projectID, gsaKeySecretID, gsa
...
+			return "", err
+		}
+		return gsaKeyFilePath, nil
 	}
 	// For some reason, the "name" value is just the version,
 	// even tho in json the "name" field is the full name.
 	// But what the access command wants is the version. So it's fine.
 	gsaKeySecretVersion := strings.TrimSpace(string(out))
-	if gsaKeySecretVersion == "" {
+	if gsaKeySecretVersion == "" || strings.Contains(gsaKeySecretVersion, "WARNING") {
```
karlkfi, line 965: what's the warning edge case?

sdowell: If you list a secret that exists but doesn't have any versions, it prints a warning message but doesn't return an error

sdowell: Another approach could be trying to parse the output to an integer and failing if it errors

karlkfi: Hmm how about matching on lines that start with WARNING and removing the whole line, just in case there's some other warning we don't know about. 

Or check for exact match of the empty warning and error if there are other unexpected warnings.

sdowell: I don't want this to be too brittle, as we're just checking a warning string on the output for errors. If the warning message changes slightly, I'd prefer this not break. 

I updated this to pass in the `--no-user-output-enabled` flag, which suppresses the warning message. This means that when there are no versions the output should more reliably pass the empty check on this line.

karlkfi: great, thanks.

e2e/testinfra/terraform/prow/service_accounts.tf
```diff
@@ -23,11 +23,17 @@ resource "google_project_iam_member" "test-runner-iam" {
...
     "roles/source.reader",
   ])
 
   role    = each.value
   member  = "serviceAccount:e2e-test-runner@oss-prow-build-kpt-config-sync.iam.gserviceaccount.com"
   project = data.google_project.project.id
 }
+
+resource "google_service_account_iam_member" "admin-account-iam" {
+  service_account_id = "projects/${data.google_project.project.id}/serviceAccounts/e2e-test-ar-reader@${data.google_project.project.name}.iam.gserviceaccount.com"
```
karlkfi, line 36: Why name and not id? Isn't it the ID in the GSA email?

e2e/testinfra/terraform/prow/service_accounts.tf
```diff
@@ -23,11 +23,17 @@ resource "google_project_iam_member" "test-runner-iam" {
     "roles/container.admin",
     "roles/gkehub.admin",
     "roles/monitoring.viewer",
-    "roles/secretmanager.secretAccessor",
+    "roles/secretmanager.admin",
```
nan-yu, line 26: Do we need the `secretmanager.admin` role? The only missing role is `secretmanager.viewer`. We can have both `secretmanager.secretAccessor` and `secretmanager.viewer`.

nan-yu (commented):
Thanks for fixing the TF config for the missing role.

e2e/testinfra/terraform/prow/service_accounts.tf
karlkfi: The code destroys secrets when they're invalid and creates them when they don't exist. So it probably needs admin.

karlkfi (approved):
/lgtm

e2e/testinfra/terraform/prow/service_accounts.tf
sdowell: With https://github.com/GoogleContainerTools/kpt-config-sync/pull/1077 the SA needs a broader range of permissions. It needs permission now to add/disable secret versions when the key expires, and potentially create a new secret if running on a new environment that needs bootstrapping. We could remove the ability to bootstrap the initial secret if we want to keep the permissions more minimal, although that means it needs to be done manually when creating a new test environment.

nan-yu: ACK

nan-yu (approved):
/lgtm

PR #1079, opened 2023-12-15
test: fix multi sync flake with InitGitRepos

sdowell:
The previous fix resolved half of the problem with pushing the initial commit, but InitGitRepos also needs to create the repository on the server before the commit can be pushed. This change registers the RootRepos/NonRootRepos at the beginning of the test, so that if the auto-repair needs to run it will properly call InitGitRepos on all of the expected repositories. A new WatchForAllSyncs predicate is added which makes it easier to wait for a subset of RSyncs.


nan-yu (approved):
/lgtm

PR #1080, opened 2023-12-18
fix: use project_id for GSA email

sdowell:
The email is invalid with id, and expects project_id. Also removes the redundant project prefix.


tiffanny29631 (approved):
/lgtm

karlkfi (approved):
/lgtm

image

PR #1084, opened 2023-12-19
fix: enable user output for version list

sdowell:
Suppressing the user output also disabled output for the happy path where a version exists. This reverts the behavior back to how it was before with an error check on the output.


nan-yu (approved):
/lgtm

PR #1087, opened 2023-12-21
Remove the HierarchyConfig CRD from the CS Operator manifest

haiyanmeng:
This CRD is required by the Config Sync Hierarchical mode. But it does not need to be installed on the cluster.


manifests/operator/kustomization.yaml
```diff
@@ -14,5 +14,4 @@
 
 resources:
 - ../base
-- ../hierarchyconfig-crd.yaml
```
sdowell, line 17: Please also remove the config/tooling for generating this file, since it is no longer used

haiyanmeng: Updated `Makefile.reconcilermanager`. Any other locations I am missing?

haiyanmeng:
@sdowell , I don't see how this CL could break `kpt-config-sync-presubmit`: https://storage.googleapis.com/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/1087/kpt-config-sync-presubmit/1737924437216333824/build-log.txt

Is this a known issue?

sdowell:
> @sdowell , I don't see how this CL could break `kpt-config-sync-presubmit`: https://storage.googleapis.com/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/1087/kpt-config-sync-presubmit/1737924437216333824/build-log.txt
> 
> Is this a known issue?

Seems unrelated to your change. There have been issues with the fake client in the past, it is most likely a flake.

manifests/operator/kustomization.yaml
sdowell: You'll also need to update `manifests/patch/kustomization.yaml` and update the kubebuilder comments at `pkg/api/configmanagement/v1/types.go` to not generate a CRD.

haiyanmeng: How does it look now?

pkg/api/configmanagement/v1/types.go
```diff
@@ -569,54 +569,3 @@ type RepoList struct {
...
 	Items []Repo `json:"items"`
 }
-
-// +genclient
-// +genclient:nonNamespaced
-// +kubebuilder:object:root=true
-// +kubebuilder:resource:scope=Cluster
-
-// HierarchyConfig is used for configuring the HierarchyModeType for managed resources.
-type HierarchyConfig struct {
```
sdowell, line 579: These structs still need to exist, we just don't want to generate the CRD manifests.

manifests/operator/kustomization.yaml
sdowell: Left a comment on the api struct. Also should delete the hierarchconfig-crd.yaml file.

haiyanmeng: Fixed it.

pkg/api/configmanagement/v1/types.go
sdowell: Please remove the kubebuilder comments.

haiyanmeng: Removing the kubebuilder comments would cause `make generate` to remove the `DeepCopyObject` method in zz_generated.deepcopy.go.
```
-// DeepCopyObject is an autogenerated deepcopy function, copying the receiver, creating a new runtime.Object.
-func (in *HierarchyConfig) DeepCopyObject() runtime.Object {
-       if c := in.DeepCopy(); c != nil {
-               return c
-       }
-       return nil
-}
-
```

sdowell: Is there a way to tell controller-gen to generate code but not yaml artifacts? Do we need the generated code?

karlkfi:
What if someone wants to do client-side validation of the HierarchyConfig?

Is that possible if we stop generating/shipping the CRD?

pkg/api/configmanagement/v1/types.go
sdowell: I think you can just update the make target to remove the yaml manifest after generation

haiyanmeng: That is what I have been doing. See the changes to the `configsync-crds` Make rule.

haiyanmeng:
> What if someone wants to do client-side validation of the HierarchyConfig?
> 
> Is that possible if we stop generating/shipping the CRD?

That is a valid concern. My understanding is that if we stop installing the CRD, the users will no longer be able to modify the CRs as suggested in https://cloud.google.com/anthos-config-management/docs/how-to/namespace-scoped-objects#disable_inheritance_for_an_object_type.

@sdowell , WDYT?

sdowell:
> That is a valid concern. My understanding is that if we stop installing the CRD, the users will no longer be able to modify the CRs as suggested in https://cloud.google.com/anthos-config-management/docs/how-to/namespace-scoped-objects#disable_inheritance_for_an_object_type.

I don't think this statement is correct. The CRs are declared in the repo but not applied to the cluster, and that should still be applicable after removing the CRD from the cluster.

> What if someone wants to do client-side validation of the HierarchyConfig? Is that possible if we stop generating/shipping the CRD?

My initial reaction is to not install the CRD on the cluster if there is no intent to create objects of that type on the cluster. That being said, the client-side validation concerns seem worth consideration/analysis.

karlkfi:
My only concern with NOT applying the CRD is that I vaguely recall that some tools download schema from the server so they can do client-side validation. But I personally have not done this.

I don't know if the state of the art has moved forward much in the last 5 years. 

I don't know if we ever made a kpt solution for this, but there's kubeval, kubeconform, and kubectl --dry-run=client. We don't want to break customers using these for client side validation. 

FWIW, it's also possible to use kubectl --dry-run=server to do validation in CI. I don't know if we have any customers doing this, but it's hypothetically possible.

haiyanmeng:
Thanks for the input, @sdowell and @karlkfi .
I will close this PR and create a separate PR for adding  the CRD in the CS OSS manifest.

PR #1090, opened 2024-01-02
Add the HierarchyConfig CRD into the OSS CS manifest

haiyanmeng:
This would allow client-side validation of the HierarchyConfig resources.

Context: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1087


manifests/oss/kustomization.yaml
```diff
@@ -14,4 +14,5 @@
 
 resources:
 - ../base
+- ../hierarchyconfig-crd.yaml
```
sdowell, line 17: If this is going to be a common manifest, please move this to the base.

What is the analysis on whether this CRD needs to be installed on the cluster? It seems preferable to not install it on the cluster if there isn't a use case.

haiyanmeng: > What is the analysis on whether this CRD needs to be installed on the cluster? It seems preferable to not install it on the cluster if there isn't a use case.

I don't have time to do the analysis. If the analysis is a blocker, someone else needs to own it.

karlkfi: I think it largely depends on whether tools read CRDs from a cluster before validation or whether the tools import the CRDs/schemes through other means. But the list of tools we want to support would need to first be assembled.

karlkfi: If we want to skip analysis and be defensive, we can just apply everything and let the user figure out client-side validation. But ofc, even that strategy might be confusing, since users might assume these resources should be or need to be applied to the cluster... It's kind of a lose-lose UX scenario.

haiyanmeng: +1 on applying the CRD.

karlkfi: Can you add a comment to the kustomize file that explains why the CRD is being included, even if the HCs should never be applied?

haiyanmeng: Added a single-line comment.

karlkfi:
Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1092

karlkfi:
Needs rebase

haiyanmeng:
> Needs rebase

Done.

PR #1091, opened 2024-01-02
fix: Add 5m buffer time to askpass

karlkfi:
This should help avoid the credentials expiring before git has a chance to use them.

Fixes: b/316600701

### DIFF
--- a/pkg/askpass/askpass.go
+++ b/pkg/askpass/askpass.go
@@ -27,6 +27,17 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// bufferTime is the approximate amount of time that the credentials should
+// remain valid after being returned by the askpass sidecar.
+//
+// A 5m buffer was chosen as a compromise to avoid git API call errors without
+// significantly increasing credential refresh request volume. It's assumed that
+// credentials are valid for ~60m, so a 5m early refresh should cause less than
+// 10% increase in refresh call volume, while still allowing enough time for the
+// askpass service to respond and git to make at least one API call, possibly
+// multiple and/or retries.
+const bufferTime = 5 * time.Minute
+
 // Server contains server wide state and settings for the askpass sidecar
 type Server struct {
 	Email string
@@ -67,7 +78,7 @@ func (aps *Server) needNewToken() bool {
 		return true
 	}
 
-	if time.Now().After(aps.token.Expiry) {
+	if time.Now().Add(bufferTime).After(aps.token.Expiry) {
 		return true
 	}
 
--- a/pkg/askpass/askpass_test.go
+++ b/pkg/askpass/askpass_test.go
@@ -37,7 +37,8 @@ func TestCachedToken(t *testing.T) {
 		AccessToken:  "0xBEEFCAFE",
 		TokenType:    "Bearer",
 		RefreshToken: "0xDEADC0DE",
-		Expiry:       time.Now().Add(time.Second * 30),
+		// Expiry needs to be long enough to account for early refresh (5m).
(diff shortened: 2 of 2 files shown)


pkg/askpass/askpass.go
```diff
@@ -67,7 +73,7 @@ func (aps *Server) needNewToken() bool {
 		return true
 	}
 
-	if time.Now().After(aps.token.Expiry) {
+	if time.Now().Add(bufferTime).After(aps.token.Expiry) {
```
sdowell, line 76: It seems like we need to subtract rather than add, right?

For example if the token expires at 2:00 PM, we want to refresh the token after 1:55 rather than 2:05.

Also, are there any guarantees the token duration will always be greater than 5 minutes? If the token expires in less than 5 minutes, then the token would constantly be refreshed. Perhaps the buffertime should be a calculated based on the expiration time (e.g. a min, a fraction, etc).

karlkfi: > It seems like we need to subtract rather than add, right?

Adding to "now" is the same as subtracting from the expiry.

>  are there any guarantees the token duration will always be greater than 5 minutes?

I can't find any docs explaining how long the creds last. But for Config Sync this is only ever used with CSR, so it won't have much variance.

If it's an oauth 2 bearer token (aka access token), I think they're always 60m (3600s) for GCP. https://developers.google.com/identity/protocols/oauth2/service-account#jwt-auth

> Perhaps the buffertime should be a calculated based on the expiration time (e.g. a min, a fraction, etc).

We don't know when the credentials where generated, so we don't know how long the full duration is. I don't think we want to use a fraction of the remaining time.

sdowell: Taking a closer look at the oauth library, the Token has a builtin `Valid` method which performs a similar functionality with a expiry delta: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/vendor/golang.org/x/oauth2/token.go#L144

Perhaps we should just replace this with the builtin method?

sdowell: Granted the buffer time is somewhat arbitrary, but I'm thinking 5 minutes may be too aggressive. It seems like some value less than or equal to 1 minute should be sufficient

karlkfi: I considered using 1m, but the check is done before the askpass request has responded and before git has made its request yet. Many http clients default to 30s timeouts so I figured 30s to respond and 30s to request might be too short and wouldn't allow any retries. 

It really depends how long the duration is tho. For 60m, 5m is only 8%. And if it's 3hr, 5m is 3%. My feeling is that even up to a 20% increase in creds requests is worth it to avoid flakey errors. 

If the duration were closer to 10m I would expect to have received a lot more customer issues about this, because it would error for everyone using CSR very regularly.

karlkfi: Valid() is used elsewhere in the code, after our refresh logic.

It's possible there's room to simplify, but I think that logic is actually in a client library. I'll check.

karlkfi: Yeah, Valid is implemented in the Google OAuth client lib. Changing its impl and meaning would require reimplementing upstream things. I'd rather not. 

https://cs.opensource.google/go/x/oauth2/+/master:token.go;drc=a835fc4358f6852f50c4c5c33fddcd1adade5b0a;l=143

sdowell: Ack, that seems reasonable

sdowell (approved):
/lgtm

karlkfi:
Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1092

PR #1092, opened 2024-01-02
chore: bump manifest headers to 2024

sdowell:
The license headers for the manifests are generated as part of the presubmits, and need to be bumped for the new year.

### DIFF
--- a/manifests/cluster-selector-crd.yaml
+++ b/manifests/cluster-selector-crd.yaml
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


karlkfi:
My understanding is that Google policy is to only update copyright  years on file creation, not updates.

Is this change needed to satisfy validation? If so, it seems like maybe the tooling doesn't match the policy.

sdowell:
> My understanding is that Google policy is to only update copyright years on file creation, not updates (go/copyright#the-year).
> 
> Is this change needed to satisfy validation? If so, it seems like maybe the tooling doesn't match the policy.

Yes, this is needed to satisfy the presubmit validation. Currently the CRDs are generated from code and headers are injected, and then the tooling checks if there is a git diff. So the validation now fails since the generated headers have the new year.

karlkfi:
Can you log a bug to fix the automation to preserve the earliest copyright date, instead of always re-generating it?

sdowell:
> Can you log a bug to fix the automation to preserve the earliest copyright date, instead of always re-generating it?

Created a bug

karlkfi (approved):
/lgtm

PR #1093, opened 2024-01-02
chore: ignore self-assignment lint for validation

karlkfi:
This isn't a problem when you lint in the container with the right linter version, but is when you have a different version installed locally. I'd like it to work on my machine without needing to always lint in a container.


karlkfi:
Blocked by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1092

sdowell (approved):
/lgtm

PR #1099, opened 2024-01-04
chore: update golang image to debian bookworm

karlkfi:


Makefile
```diff
@@ -37,7 +37,7 @@ OUTPUT_DIR := $(abspath .output)
 GO_DIR := $(OUTPUT_DIR)/go
 
 # Base image used for all golang containers
-GOLANG_IMAGE := golang:1.21.5-bullseye
+GOLANG_IMAGE := golang:1.21.5-bookworm
```
sdowell, line 40: The `BUILDENV_IMAGE` version below should be bumped so that the `buildenv` image can be rebuilt with this new base image

e.g. `$(TEST_INFRA_REGISTRY)/buildenv:v0.3.1` -> `$(TEST_INFRA_REGISTRY)/buildenv:v0.3.2`

karlkfi: do i need to build it manually?

sdowell: No, it's built/published automatically

karlkfi: Should i bump the buildenv number in a subsequent PR or this PR?

sdowell: IMO should be done in this PR to make sure the new buildenv passes validation/presubmits before changing the base image

sdowell: An example for how it could fail is if there are some OS packages we use which were available in bullseye but not bookworm

sdowell (approved):
/lgtm

PR #1102, opened 2024-01-08
fix: reverse log level ordering for zap/otel-agent

sdowell:
The zap log levels become more verbose as the log level decreases, which is the inverse of how our other log levels operate. This change reverses the order of our logLevel API when translating it to a zap log level. This is intended to create more consistent API behavior between the different containers. The default continues to be INFO, which maps to the integer value of 5 in our new scheme.

### DIFF
--- a/manifests/reposync-crd.yaml
+++ b/manifests/reposync-crd.yaml
@@ -354,12 +354,14 @@ spec:
                           pattern: ^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync|gcenode-askpass-sidecar|otel-agent)$
                           type: string
                         logLevel:
-                          description: logLevel specifies the log level override value
-                            for the container. The default value for git-sync container
-                            is 5, while all other containers will default to 0. Allowed
-                            values are from -1 to 10
+                          description: logLevel specifies the verbosity level of the
+                            logging for a specific container. The "git-sync" and "otel-agent"
+                            containers default to 5, while all other containers default
+                            to 0. Increasing the value of logLevel increases the verbosity
+                            of the logs. Lower severity messages are logged at higher
+                            verbosity. Allowed values are from 0 to 10.
                           maximum: 10
-                          minimum: -1
+                          minimum: 0
                           type: integer
                       required:
                       - containerName
@@ -1403,12 +1405,14 @@ spec:
                           pattern: ^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync|gcenode-askpass-sidecar|otel-agent)$
                           type: string
                         logLevel:
-                          description: logLevel specifies the log level override value
-                            for the container. The default value for git-sync container
-                            is 5, while all other containers will default to 0. Allowed
-                            values are from -1 to 10
+                          description: logLevel specifies the verbosity level of the
+                            logging for a specific container. The "git-sync" and "otel-agent"
+                            containers default to 5, while all other containers default
+                            to 0. Increasing the value of logLevel increases the verbosity
+                            of the logs. Lower severity messages are logged at higher
+                            verbosity. Allowed values are from 0 to 10.
                           maximum: 10
-                          minimum: -1
+                          minimum: 0
                           type: integer
                       required:
                       - containerName
--- a/pkg/api/configsync/v1beta1/resource_override.go
+++ b/pkg/api/configsync/v1beta1/resource_override.go
@@ -160,10 +160,12 @@ type ContainerLogLevelOverride struct {
 	// +kubebuilder:validation:Pattern=^(reconciler|git-sync|hydration-controller|oci-sync|helm-sync|gcenode-askpass-sidecar|otel-agent)$
 	ContainerName string `json:"containerName"`
 
-	// logLevel specifies the log level override value for the container.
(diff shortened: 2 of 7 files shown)


manifests/reposync-crd.yaml
```diff
@@ -355,11 +355,11 @@ spec:
                           type: string
                         logLevel:
                           description: logLevel specifies the log level override value
-                            for the container. The default value for git-sync container
-                            is 5, while all other containers will default to 0. Allowed
-                            values are from -1 to 10
+                            for the container. The default value for the "git-sync"
+                            and "otel-agent" container is 5, while all other containers
+                            will default to 0. Allowed values are from 0 to 10
```
karlkfi, line 360: How about this:

logLevel specifies the verbosity level of the logging for a specific container. By default, each container is configured to log informational and higher severity messages, like warnings and errors. So the "git-sync" and "otel-agent" containers default to 5, while all other containers default to 0. Increasing the value of logLevel increases the verbosity of the logs. Lower severity messages are logged at higher verbosity.

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
```diff
@@ -107,7 +107,9 @@ func mutateContainerLogLevel(c *corev1.Container, override []v1beta1.ContainerLo
 			switch c.Name {
 			case metrics.OtelAgentName:
 				// otel-agent surfaces the log level configuration differently.
-				zapLevel := zapcore.Level(logLevel.LogLevel)
+				// Our log levels range from 0-10, whereas zap ranges from -1 (debug) to 5 (fatal).
```
karlkfi, line 110: i thought 5 was info and fatal was higher?

sdowell: This comment is referring to the underlying integer mapping of the zap library

manifests/reposync-crd.yaml
sdowell: done

pkg/reconcilermanager/controllers/reconciler_container_log_level.go
sdowell: See https://pkg.go.dev/go.uber.org/zap/zapcore#Level

karlkfi: oh, i see. you set anything >= 6 to "debug" (aka -1)

manifests/reposync-crd.yaml
```diff
@@ -354,12 +354,17 @@ spec:
...
                           type: string
                         logLevel:
-                          description: logLevel specifies the log level override value
-                            for the container. The default value for git-sync container
-                            is 5, while all other containers will default to 0. Allowed
-                            values are from -1 to 10
+                          description: logLevel specifies the verbosity level of the
+                            logging for a specific container. By default, each container
+                            is configured to log informational and higher severity
+                            messages, like warnings and errors. The "git-sync" and
```
janetkuo, line 360: > By default, each container is configured to log informational and higher severity messages, like warnings and errors.

Is this true for all containers, even though the default level of git-sync and otel-agent is 5 and others are 0?

sdowell: It looks like this isn't strictly true for `git-sync`: https://github.com/kubernetes/git-sync

```
    -v, --verbose <int>, $GITSYNC_VERBOSE
            Set the log verbosity level.  Logs at this level and lower will be
            printed.  Logs follow these guidelines:

            - 0: Minimal, just log updates
            - 1: More details about updates
            - 2: Log the sync loop
            - 3: More details about the sync loop
            - 4: More details
            - 5: Log all executed commands
            - 6: Log stdout/stderr of all executed commands
            - 9: Tracing and debug messages
```

sdowell: I've removed this line from the description.

pkg/api/configsync/v1beta1/resource_override.go
```diff
@@ -160,10 +160,12 @@ type ContainerLogLevelOverride struct {
...
-	// logLevel specifies the log level override value for the container.
-	// The default value for git-sync container is 5, while all other containers will default to 0.
-	// Allowed values are from -1 to 10
-	// +kubebuilder:validation:Minimum=-1
+	// logLevel specifies the verbosity level of the logging for a specific container.
+	// The "git-sync" and "otel-agent" containers default to 5, while all other containers default to 0.
+	// Increasing the value of logLevel increases the verbosity of the logs.
+	// Lower severity messages are logged at higher verbosity.
+	// Allowed values are from 0 to 10.
+	// +kubebuilder:validation:Minimum=0
```
janetkuo, line 168: How do users know what number maps to what level, e.g. for `otel-agent`, 5 maps to info, 0 maps to fatal, etc.? Do we need to document them all or point to some standard/convention?

sdowell: The general convention here is that a higher log level will be more verbose and print lower severity messages, which is captured by the API description. I think it would be too much information to include the exact mappings in the API description here - we could add a user doc with more information.

janetkuo (approved):
/lgtm

PR #1104, opened 2024-01-11
feat: add CA cert support for oci-sync

sdowell:
This change adds a caCertSecretRef API field for the OCI api which is analogous to the existing git api field. This can be used by users to provide a CA certificate for verifying the authenticity of the OCI registry.

New test scaffolding is added in order to sync images from an in-cluster registry using a contrived CA. This is modeled after the in-cluster git server and is intended to provide a very similar interface for writing tests. The intention is to expand this interface to support other registry providers such as Artifact Registry. Refactoring the existing implementation for Artifact Registry would be a larger scale refactor, and is deferred to a subsequent change.


Makefile
```diff
@@ -409,7 +409,7 @@ clean-helm:
 install-kind: "$(KIND)"
 
 "$(GOBIN)/crane":
-	go install github.com/google/go-containerregistry/cmd/crane@$(CRANE_VERSION)
+	CGO_ENABLED=0 go install github.com/google/go-containerregistry/cmd/crane@$(CRANE_VERSION)
```
karlkfi, line 412: why is CGO_ENABLED=0 needed?

e2e/nomostest/git-server.go
```diff
@@ -44,8 +44,8 @@ func testGitServerSelector() map[string]string {
 	return map[string]string{"app": testGitServer}
 }
 
-func setupGit(nt *NT) error {
-	if *e2e.GitProvider == e2e.Local {
+func setupSyncSources(nt *NT) error {
+	if *e2e.GitProvider == e2e.Local || *e2e.RegistryProvider == e2e.Local {
```
karlkfi, line 48: why OR? Are the Git keys re-used for the local registry?

e2e/nomostest/gitproviders/repository.go
```diff
@@ -621,6 +621,15 @@ func (g *Repository) RenameBranch(current, new string) error {
 	return g.Push(current, "--delete")
 }
 
+// CurrentBranch returns the name of the current branch
+func (g *Repository) CurrentBranch() (string, error) {
+	out, err := g.Git("rev-parse", "--abbrev-ref", "HEAD")
```
karlkfi, line 626: This wont work if on detached HEAD. Do we ever do that? Should we maybe comment it as a gotcha?

e2e/nomostest/new.go
```diff
@@ -438,6 +441,23 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
...
+	nt.T.Cleanup(func() {
+		for _, image := range nt.ociImages {
+			remoteURL, err := nt.RegistryProvider.OCIPushURL(image.Name)
+			if err != nil {
+				nt.T.Fatal(err)
+			}
+			if err := image.Delete(remoteURL); err != nil {
+				nt.T.Error(err)
+			}
+		}
```
karlkfi, line 459: should we clear ociImages after?

e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -49,6 +49,9 @@ type MultiRepo struct {
 	// SkipNonLocalGitProvider will skip the test if run with a GitProvider type other than local.
 	SkipNonLocalGitProvider bool
 
+	// SkipNonLocalRegistryProvider will skip the test if run with a RegistryProvider type other than local.
+	SkipNonLocalRegistryProvider bool
```
karlkfi, line 53: Kinda wish these were just test requirements and not a double negative Skip- Non-, but there's precedent, so i guess it's ok.

e2e/nomostest/git-server.go
```diff
@@ -63,7 +63,15 @@ func setupGit(nt *NT) error {
 			return err
 		}
 		nt.caCertPath = caCertPath
+	}
+	if *e2e.RegistryProvider == e2e.Local {
+		if err := installRegistryServer(nt); err != nil {
```
karlkfi, line 68: any way we can do this in paralle with the git server, since they're both waiting for reconcile?

e2e/nomostest/registry_server.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}
+
+	return nt.Watcher.WatchForCurrentStatus(kinds.Deployment(), TestRegistryServer, TestRegistryNamespace)
+}
+
+// PushOCIImage uses the current file system state of the provided Repository
+// to build an OCI image and push it to the current RegistryProvider. The resulting
+// OCIImage object can be used to set the spec.oci.image field on the RSync.
+func (nt *NT) PushOCIImage(repository *gitproviders.Repository) (*registryproviders.OCIImage, error) {
```
karlkfi, line 78: `BuildAndPushOCIImage` maybe?

Makefile
sdowell: The e2e container doesn't have glibc libraries installed, which causes a linker error when trying to run the crane binary from inside the container (e.g. for the tests).
```
crane: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by crane)
crane: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by crane)
```

e2e/nomostest/registry_server.go
karlkfi: Should this maybe just be something like nt.OCIRegistry.BuildAndPush?

e2e/nomostest/gitproviders/repository.go
sdowell: The `Repository` interface currently does not support entering a detached HEAD state, so we never do this currently. I can add a comment

e2e/nomostest/registry_server.go
```diff
@@ -0,0 +1,358 @@
...
+		core.Namespace(TestRegistryNamespace),
+		core.Labels(testRegistryServerSelector()),
+	)
+	configMap.Data = map[string]string{
+		nginxConf: nginxConfig,
+	}
+	return configMap
+}
+
+func registryDeployment() *appsv1.Deployment {
```
karlkfi, line 157: Can this be a testdata artifact?

e2e/nomostest/registry_server.go
```diff
@@ -0,0 +1,358 @@
...
+					},
+				},
+			},
+		},
+	}
+	return deployment
+}
+
+// See https://distribution.github.io/distribution/recipes/nginx/
+const nginxConfig = `
```
karlkfi, line 256: Can this be a testdata artifact?

e2e/nomostest/registryproviders/image.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// use an auto-incrementing index to create unique file names for tarballs
+var imageIndex int
+
+// NewImage creates a new OCIImage object and associated tarball using the provided
+// Repository. The contents of the git repository will be bundled into a tarball
+// at the artifactDir. The resulting OCIImage object can be pushed to a remote
+// registry using its Push method.
+func NewImage(artifactDir string, shell *testshell.TestShell, repository *gitproviders.Repository, provider RegistryProvider) (*OCIImage, error) {
```
karlkfi, line 33: `BuildImage` maybe? It's not just a Go constructor, so using New- is a little confusing.

e2e/nomostest/registryproviders/image.go
```diff
@@ -0,0 +1,116 @@
...
+	shell              *testshell.TestShell
+}
+
+// FloatingTag returns the floating tag that initially points to this image.
+// This tag is suitable for use on RSync object spec but not for interacting with
+// the image from the test suite.
+// This uses the git branch as a version, such that the tag will be updated
+// whenever a new image is pushed from the git branch. This enables syncing
+// to an OCI image similar to a git branch.
+func (o *OCIImage) FloatingTag() string {
```
karlkfi, line 80: `BranchTag` maybe? The fact that it's floating is important but not descriptive of what it is.

If you really want it in the name, maybe `FloatingBranchTag`?

e2e/nomostest/registryproviders/local.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Type returns the provider type.
+func (l *LocalProvider) Type() string {
+	return e2e.Local
+}
+
+// OCIPushURL returns a URL for pushing images to the remote registry.
+// name refers to the repo name in the format of <NAMESPACE>/<NAME> of RootSync|RepoSync.
+func (l *LocalProvider) OCIPushURL(name string) (string, error) {
```
karlkfi, line 38: OCI prefix on these methods seems a little redundant. Is it informative?

e2e/nomostest/registryproviders/registry_provider.go
```diff
@@ -0,0 +1,41 @@
...
+	// OCISyncURL returns the registry URL for Config Sync to sync from using OCI.
+	// name refers to the repo name in the format of <NAMESPACE>/<NAME> of RootSync|RepoSync.
+	OCISyncURL(name string) string
+}
+
+// NewRegistryProvider creates a RegistryProvider for the specific provider type.
+// This enables writing tests that can run against multiple types of registries.
+func NewRegistryProvider(provider string) RegistryProvider {
+	switch provider {
+	// TODO: Refactor existing Artifact Registry impl to this interface
```
karlkfi, line 37: should we panic in the mean time if it's not local?

manifests/reposync-crd.yaml
```diff
@@ -279,6 +279,20 @@ spec:
                     - gcpserviceaccount
                     - none
                     type: string
+                  caCertSecretRef:
+                    description: caCertSecretRef specifies the name of the secret
+                      where the CA certificate is stored. The creation of the secret
```
karlkfi, line 284: `The creation of the secret should be done out of band by the user` seems weird to note here. Is there precedent? Maybe just cut it.

manifests/reposync-crd.yaml
```diff
@@ -279,6 +279,20 @@ spec:
...
+                      should be done out of band by the user and should store the
+                      certificate in a key named "cert". For RepoSync resources, the
```
karlkfi, line 286: Do we talk about both RootSyncs and RepoSyncs in other fields like this? Or should we use different field descriptions for each?

e2e/nomostest/git-server.go
karlkfi: Seems like maybe the CAs and PKs should be separate. We don't want to require that users use the same CA for both.

pkg/reconcilermanager/controllers/util.go
```diff
@@ -271,6 +272,12 @@ func ociSyncEnvs(opts ociOptions) []corev1.EnvVar {
 		Name:  reconcilermanager.OciSyncWait,
 		Value: fmt.Sprintf("%f", opts.period),
 	})
+	if useCACert(opts.caCertSecretRef) {
+		result = append(result, corev1.EnvVar{
+			Name:  "SSL_CERT_FILE",
```
karlkfi, line 277: Should we use TLS instead of SSL here or is it SSL for some particular reason?

sdowell: The env var name here is consumed directly by the [crypto library](https://pkg.go.dev/crypto/x509#SystemCertPool)

manifests/reposync-crd.yaml
sdowell: Yes, this is written similarly to the existing git field

manifests/reposync-crd.yaml
sdowell: Yes, this is written similarly to the existing git field

e2e/nomostest/registryproviders/local.go
sdowell: The intention is to add `HelmPush` type commands as well with a subsequent PR, which will have different behavior

Makefile
karlkfi: Should we set CGO_ENABLED=0 in the Dockerfile?

e2e/nomostest/git-server.go
karlkfi: Or do we only ever use one or the other and not both in any given test?

e2e/nomostest/registryproviders/registry_provider.go
sdowell: `local` is the only enum option that is supported currently by the test entrypoint, so I don't think it's crucial.

e2e/nomostest/registry_server.go
sdowell: It _can_ be, but I prefer it like this personally. With it inline it's easier to discover from the code and reuse constants IMO

karlkfi: If you need to modify values you can always parse the file into a client.Object and override specific files with constants.

e2e/nomostest/registryproviders/local.go
karlkfi: Helm and OCI methods on the same Registry API? They seem like they should probably have different interfaces. Do they need to be the same?

e2e/nomostest/ntopts/multi_repo.go
sdowell: They could be renamed to `RequireLocal*Provider`. I suggest we refactor that separately

e2e/nomostest/nt.go
```diff
@@ -735,6 +744,35 @@ func (nt *NT) portForwardGitServer() {
 	)
 }
 
+// portForwardRegistryServer forwards the registry-server deployment to a port.
+func (nt *NT) portForwardRegistryServer() {
+	provider := nt.RegistryProvider.(*registryproviders.LocalProvider)
+	resetImages := func(newPort int, podName string) {
```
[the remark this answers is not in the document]
sdowell: Seems reasonable, although I'd prefer to refactor that separately

e2e/nomostest/registryproviders/registry_provider.go
karlkfi: I think it's probably an anti-pattern to use the New- constructor for a builder that returns multiple implementations based on inputs. It might be better to have this be `registryproviders.Build(provider string)` and use the switch to return an error when the provider is unrecognized. Then make the caller figure out how to deal with the error.

e2e/nomostest/registryproviders/image.go
sdowell: Done

e2e/nomostest/registryproviders/image.go
sdowell: Done

e2e/nomostest/registry_server.go
sdowell: Renamed. This isn't pushed into the `RegistryProvider` interface because it's intended as glue between NT and the registry provider.

e2e/nomostest/git-server.go
sdowell: Yeah they can be done in parallel. Updated

e2e/nomostest/new.go
sdowell: It isn't strictly necessary since the next test will start with an empty array, but I added a line to clear it.

e2e/nomostest/git-server.go
sdowell: I'm not sure I understand what you mean - could you elaborate?

Makefile
sdowell: It is set in the dockerfile, the issue here is that `crane` is built outside of the e2e container. So for localhost it would be built on the host machine and in prow it would be built inside the kubekins container.

e2e/nomostest/registryproviders/local.go
sdowell: They don't need to be the same. My expectation is that this will be simpler and more performant. Keep in mind the `RegistryProvider` is mainly just returning the URLs for another interface to consume. In this case the LocalProvider returns an `OCIPushURL` that is consumed by `OCIImage` to push images with `crane`.

In the `helm` scenario, the LocalProvider will return a `HelmPushURL` that will be consumed by a `Helm` (name TBD) struct to push helm packages.

e2e/nomostest/registryproviders/registry_provider.go
sdowell: The behavior here is modeled after the `gitprovider` factory. I would prefer to keep them consistent, but I'm not opposed to refactoring both factories

Makefile
karlkfi: Ahh, I see. Then maybe this should go in another PR to fix it with a comment that says why crane and CGO_ENABLED=0 aren't just in a container.

e2e/nomostest/git-server.go
karlkfi: The only reason you made them share was to avoid duplicating the code, right? Not some technical constraint?

In most cases, the OCI registry and Git registry should be mutually exclusive for tests, right? But if we had some tests for migration they might not be.

It might make sense to just make one set for each to make it clear that they're not related and don't need to be the same. We wouldn't want to accidentally have some tests pass because they're the same, even tho nobody in the real world would have two services use the same private keys, even if they used the same root authority.

sdowell: > The only reason you made them share was to avoid duplicating the code, right? Not some technical constraint?

Yes, they share two things (just for simplicity/convenience and not a technical constraint):
- They live in the same Namespace
- They share a HTTPS certificate signed by the same root CA

> In most cases, the OCI registry and Git registry should be mutually exclusive for tests, right? But if we had some tests for migration they might not be.

I'm not sure what the distinction here is, but we do have tests that switch between git-sync and oci-sync (including the tests in this PR). Most tests just happen to focus mostly on one or the other.

> It might make sense to just make one set for each to make it clear that they're not related and don't need to be the same. We wouldn't want to accidentally have some tests pass because they're the same, even tho nobody in the real world would have two services use the same private keys, even if they used the same root authority.

Sure, I can generate different certs for each service from the same root CA. Although I'm struggling to imagine how this could cause false positives

karlkfi: Risk reduction is about more than just avoiding predictable bugs. It's desirable to make the tests similar to how people are actually going to use the product for multiple reasons. The more realistic the tests, the more likely we are to catch real bugs, rather than synthetic bugs that customers will never see. Avoiding a shortcut now also helps prevent us from ignorantly using the code as a precedent in the future.

Of course, one alternative would be to exhaustively document the shortcut in enough places to be discoverable. Anyone who touches this in the future will need to know why it was done this way, for debugging, and to know why or why not to copy it.

Makefile
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1107

e2e/nomostest/git-server.go
sdowell: Done

e2e/nomostest/registry_server.go
sdowell: Ack, I still prefer this approach

e2e/nomostest/new.go
```diff
@@ -438,6 +445,24 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
...
+	if *e2e.RegistryProvider == e2e.Local {
+		nt.portForwardRegistryServer()
+	}
+	// Images created by the tests should not be persisted after the test finishes.
+	// Cleanup all images before the port forward gets torn down.
+	nt.T.Cleanup(func() {
+		for _, image := range nt.ociImages {
+			remoteURL, err := nt.RegistryProvider.OCIPushURL(image.Name)
+			if err != nil {
+				nt.T.Fatal(err)
```
karlkfi, line 458: if this fails for one image, do we want to continue to delete the rest or bail here?

e2e/nomostest/registry_server.go
```diff
@@ -0,0 +1,383 @@
...
+	// tests to verify basic auth.
+	TestRegistryServerAuthenticated = "test-registry-server-auth"
+	// TestRegistryNamespace is the namespace used for the in-cluster registry
+	TestRegistryNamespace = "test-registry-system"
+	nginxConfigMapName    = "nginx-cm"
+	nginxConf             = "nginx.conf"
+)
+
+func setupRegistry(nt *NT) error {
+	if *e2e.RegistryProvider == e2e.Local {
```
karlkfi, line 62: not really specific to this PR, but should we ping the remote registry to validate auth and fail fast in the case that it's not local?

e2e/nomostest/registryproviders/local.go
karlkfi: I guess I expected a LocalOCIRegistry and a LocalHelmRegistry, with OCIRegistry and HelmRegistry interfaces, so the methods wouldn't need the context in them.

karlkfi: That way we also wouldn't need the SyncSource enum. We could just may the `spec.sourceFormat` as an enum that maps 1:1 with a SourceProvider.

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -1254,64 +1259,99 @@ func TestRepoSyncUpdateCACert(t *testing.T) {
...
+			_, _, testReconciler := setupNSReconciler(t, tc.repoSync)
 
-	// no cert secret is setup to trigger not found error
-	_, _, testReconciler := setupNSReconciler(t, rs, gitSecret)
-	ctx := context.Background()
+			ctx := context.Background()
+
+			// validation should return a not found error
+			err := testReconciler.validateCACertSecret(ctx, tc.repoSync.Namespace, caCertSecret)
+			require.Error(t, err, "Function call should return an error")
```
karlkfi, line 1350: i see you're just enhancing the existing test, but elsewhere we usually have the with and without error cases in the same Test, with an optional error field in the case map. Then you can do `testutil.AssertEqual(t, tc.expectedError, err)` to validate the error. Then you can use `testutil.EqualError(...)` in the test case to wrap a real error to validate both type and and message. 

- https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/kinds/lookup_test.go#L206
- https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/kinds/lookup_test.go#L173

e2e/nomostest/new.go
sdowell: If we can't get the registry URL then it fails fast, because other images will get the same URL/error. We could potentially put in retry logic but I opted for simplicity until there's a signal that this needs more flexible error recovery.

pkg/reconcilermanager/controllers/util.go
karlkfi: Can you pull this out to a const with a description?
Maybe link to the internal const https://cs.opensource.google/go/go/+/master:src/crypto/x509/root_unix.go;l=19

e2e/nomostest/registryproviders/local.go
sdowell: Okay I rewrote this to use an OCIProvider concept rather than `RegistryProvider`, such that we will add a `HelmProvider` variant once the helm-sync api is introduced.

The `SyncSource` enum is still needed because that has to do with the fact that there is a single registry deployment on the cluster that is used for multiple source types. i.e. helm-sync and oci-sync may use the same registry, and that registry serves from a single endpoint/certificate - so both helm-sync and oci-sync would need to use the same endpoint/certificate under the hood.

pkg/reconcilermanager/controllers/util.go
sdowell: Done

pkg/reconcilermanager/controllers/reposync_controller_test.go
sdowell: Done

e2e/nomostest/ntopts/multi_repo.go
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1109

e2e/flags.go
```diff
@@ -104,6 +104,10 @@ var ShareTestEnv = flag.Bool("share-test-env", false,
 var GitProvider = flag.String("git-provider", Local,
 	"The git provider that hosts the Git repositories. Defaults to local.")
 
+// OCIProvider is the provider that hosts the OCI repositories.
+var OCIProvider = flag.String("oci-provider", Local,
+	"The registry provider that hosts the OCI repositories. Defaults to local.")
```
karlkfi, line 109: Seems like the description should specify the other valid options. Or is this a URL?

e2e/nomostest/registryproviders/local.go
karlkfi: What you have here is probably fine.

I think what I was imagining was simply that there would be an OCIRegistry and a HelmRegistry, and the HelmRegistry would wrap the OCIRegistry. Then the test setup would start both, and register cleanup to shut down both, but the OCIRegistry would de-dupe the calls.

e2e/flags.go
sdowell: Local is currently the only valid option. Other options will be added with subsequent PRs (e.g. artifact registry)

karlkfi (approved):

PR #1105, opened 2024-01-12
Increase the default memory/cpu requests for helm-sync on Autopilot clusters

nan-yu:
The Config Sync Autopilot CI failures indicate the helm-sync container is hitting both OOMKilled and `signal: killed` (insufficient CPU) on Autopilot clusters.

- OOMKilled:
  - autopilot regular: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-regular/1745822836569673728
  - autopilot stable: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-stable/1745836426253045760
- insufficient CPU:
  - autopilot rapid: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-rapid/1745826108181319680
  - autopilot rapid-latest: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-rapid-latest/1745819564941250560

This commit increases the default requests for helm-sync on Autopilot only.

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_container_resources.go
+++ b/pkg/reconcilermanager/controllers/reconciler_container_resources.go
@@ -98,10 +98,10 @@ func ReconcilerContainerResourceDefaultsForAutopilot() map[string]v1beta1.Contai
 		},
 		reconcilermanager.HelmSync: {
 			ContainerName: reconcilermanager.HelmSync,
-			CPURequest:    resource.MustParse("150m"),
-			CPULimit:      resource.MustParse("150m"),
-			MemoryRequest: resource.MustParse("256Mi"),
-			MemoryLimit:   resource.MustParse("256Mi"),
+			CPURequest:    resource.MustParse("250m"),
+			CPULimit:      resource.MustParse("250m"),
+			MemoryRequest: resource.MustParse("384Mi"),
+			MemoryLimit:   resource.MustParse("384Mi"),
 		},
 		reconcilermanager.GitSync: {
 			ContainerName: reconcilermanager.GitSync,


karlkfi:
How will this affect the adjustment performed on the first container by autopilot? Will the reconciler container end up with more or less than it does now?

For compute,  Autopilot automatically adjusts your requests to round up to the nearest 250m. And for general-purpose, the CPU:memory ratio must be between 1:1 and 1:6.5, otherwise Autopilot automatically increases the smaller resource. https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#min-max-requests

If this change causes Autopilot to increase the resources allocated to the reconciler, we might want to instead increase the resources to the helm-sync to compensate.

karlkfi:
Due to the way autopilot adjusts memory to match the CPU ratio and CPU must be an increment of 250m, the total memory usually ends up in increments of 256Mi. So to make the math easier to do in your head, I have been using units of 128Mi for the memory requests/limits.

nan-yu:
> Due to the way autopilot adjusts memory to match the CPU ratio and CPU must be an increment of 250m, the total memory usually ends up in increments of 256Mi. So to make the math easier to do in your head, I have been using squares of 2 for the memory requests/limits. Maybe 512Mi would be better than 320Mi? Or even 384 (256 + 128), even tho it's not a square of two.

Increasing from 256Mi to 512Mi seems too much for a single container. It increases a lot when syncing from multiple Helm charts.

I can play with those numbers to accommodate the CPU:memory ratio and the incremental of 250m requirement.

karlkfi:
I think in most cases, helm will use the following containers: reconciler, helm-sync, and otel-agent. So the current total memory is 512+256+64 (832). But since the total CPU is 700+150+10 (860), the CPU will get rounded up to 1000m. Then because it's below the 1:1 ratio, the memory will get rounded up to 1024Mi. So the reconciler currently gets the diff (140m & 192Mi) added to it.

If we want to optimize for reducing adjustment, then we could add some of those resources to the helm-sync instead. For example, we could give helm-sync 250m & 384Mi. Then the total pod resources would still be adjusted to 1000m and 1024Mi, but the helm-sync would have more headroom and we'd still have a little extra in case we want to adjust up the otel-agent later, which there's a customer issue open about.

nan-yu:
That makes sense. In any case when the hydration-controller is enabled along with `helm-sync`. We will have default 1060m and 1068Mi, which will be adjusted to 1250m and 1278Mi, with the diff of 190m and 210Mi, which is less the diff (140m and 192Mi) when `hydration-controller` is not enabled.

250m and 384Mi seem to work for both cases. 

The default is updated.

PR #1106, opened 2024-01-17
test: move the fake test finalizer removal function into a cleanup block

nan-yu:
The fake finalizer should be removed even if the test failed in the middle. Otherwise, the namespace would be stuck in a terminating state. Example: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-regular-bitbucket/1747668032626364416

### DIFF
--- a/e2e/testcases/reconciler_finalizer_test.go
+++ b/e2e/testcases/reconciler_finalizer_test.go
@@ -546,6 +546,29 @@ func TestReconcileFinalizerReconcileTimeout(t *testing.T) {
 	if err := nt.KubeClient.Apply(namespace); err != nil {
 		nt.T.Fatal(err)
 	}
+
+	// Remove the fake finalizer
+	t.Cleanup(func() {
+		namespace = fake.NamespaceObject(namespaceNN.Name)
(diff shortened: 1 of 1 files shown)


e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -546,6 +546,29 @@ func TestReconcileFinalizerReconcileTimeout(t *testing.T) {
...
+	t.Cleanup(func() {
+		namespace = fake.NamespaceObject(namespaceNN.Name)
+		nt.T.Logf("Remove the fake finalizer named %s from Namespace %s", contrivedFinalizer, namespaceNN.Name)
+		if err := nt.KubeClient.Apply(namespace); err != nil {
+			nt.T.Fatal(err)
+		}
+		// With the finalizer removed, the deletion should reconcile
+		tg := taskgroup.New()
+		tg.Go(func() error {
+			return nt.Watcher.WatchForNotFound(kinds.RootSyncV1Beta1(),
```
sdowell, line 560: This should only become not found if we reach the line below that removes the nested RootSync from the git repo, so if we fail before then this assertion may fail. However I think it's fairly low risk and the priority is removing the fake finalizer, so I think this is fine

sdowell (approved):
/lgtm

PR #1107, opened 2024-01-17
ci: set CGO_ENABLED=0 for local crane binary

sdowell:
The crane binary is used by some of the e2e tests, which can be run in a container which does not have glibc. Compiling with CGO causes an error when invoking the binary in such a container.


karlkfi (approved):
/lgtm

PR #1109, opened 2024-01-18
refactor: rename test opt to RequireLocalGitProvider

sdowell:
This name avoids a double negative which helps readability.


karlkfi (approved):
/lgtm

PR #1110, opened 2024-01-20
Configure git-sync timeout from Config Sync

victorpras:
The current `git-sync` timeout is 2 minutes, which should be enough for most cases, but not configurable from Config Sync side. These change will allow users to change git-sync timeout using override API 

b/279565020

### DIFF
--- a/pkg/api/configsync/v1alpha1/resource_override.go
+++ b/pkg/api/configsync/v1alpha1/resource_override.go
@@ -74,6 +74,10 @@ type OverrideSpec struct {
 	// +listMapKey=containerName
 	// +optional
 	LogLevels []ContainerLogLevelOverride `json:"logLevels,omitempty"`
+
+	// gitSyncOverride specify git-sync container override
+	// +optional
+	GitSyncOverride GitSyncOverride `json:"gitSyncOverride,omitempty"`
 }
 
 // RootSyncOverrideSpec allows to override the settings for a RootSync reconciler pod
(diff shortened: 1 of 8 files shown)


pkg/api/configsync/v1alpha1/resource_override.go
```diff
@@ -170,3 +174,14 @@ type ContainerLogLevelOverride struct {
...
+
+// GitSyncOverride specifies the git-sync container override value
+type GitSyncOverride struct {
+	// gitSyncTimeout allows one to configure the timeout for the git-sync container.
+	// Must be no less than 0.
+	// If this field is not provided, the default value is 2 minutes.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	GitSyncTimeout int `json:"gitSyncTimeout,omitempty"`
```
sdowell, line 186: `gitSyncOverride.gitSyncTimeout` seems a bit redundant. I think this could just be `timeout`.

pkg/api/configsync/v1alpha1/resource_override.go
```diff
@@ -74,6 +74,10 @@ type OverrideSpec struct {
 	// +listMapKey=containerName
 	// +optional
 	LogLevels []ContainerLogLevelOverride `json:"logLevels,omitempty"`
+
+	// gitSyncOverride specify git-sync container override
+	// +optional
+	GitSyncOverride GitSyncOverride `json:"gitSyncOverride,omitempty"`
```
sdowell, line 80: Has this API been discussed in a design review already? I have a few concerns:

1. There's already a `gitSyncDepth` field at this level. Do we need a nested `gitSyncOverride` struct? This introduces some API inconsistency.
2. Does `timeout` need to be specific to `git-sync`? Can this concept be generalized to apply to the other *-sync containers as well with a generic API field?

janetkuo:
Closed for clean up. Feel free to reopen if needed.

PR #1111, opened 2024-01-23
feat: add CA cert support for helm-sync

sdowell:
This change adds a caCertSecretRef API field for the helm api which is analogous to the existing git/oci api fields. This can be used by users to provide a CA certificate for verifying the authenticity of the helm repository.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -112,6 +112,10 @@ func newOptStruct(testName, tmpDir string, t nomostesting.NTB, ntOptions ...ntop
 		t.Skip("Test skipped for non-local OCIProvider types")
 	}
 
+	if *e2e.HelmProvider != e2e.Local && optsStruct.RequireLocalHelmProvider {
+		t.Skip("Test skipped for non-local HelmProvider types")
+	}
+
 	if optsStruct.RESTConfig == nil {
 		RestConfig(t, optsStruct)
 		// Increase the QPS for the clients used by the e2e tests.
@@ -217,6 +221,7 @@ func SharedTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		WebhookDisabled:         sharedNt.WebhookDisabled,
 		GitProvider:             sharedNt.GitProvider,
 		OCIProvider:             sharedNt.OCIProvider,
+		HelmProvider:            sharedNt.HelmProvider,
 	}
 
 	if opts.SkipConfigSyncInstall {
@@ -296,6 +301,7 @@ func FreshTestEnv(t nomostesting.NTB, opts *ntopts.New) *NT {
 		WebhookDisabled:         &webhookDisabled,
 		GitProvider:             gitproviders.NewGitProvider(t, *e2e.GitProvider, logger),
 		OCIProvider:             registryproviders.NewOCIProvider(*e2e.OCIProvider),
+		HelmProvider:            registryproviders.NewHelmProvider(*e2e.HelmProvider),
 	}
 
 	// TODO: Try speeding up the reconciler and hydration polling.
@@ -450,7 +456,7 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 		nt.portForwardGitServer()
 	}
 	// set up port forward if using in-cluster registry server
-	if *e2e.OCIProvider == e2e.Local {
+	if *e2e.OCIProvider == e2e.Local || *e2e.HelmProvider == e2e.Local {
 		nt.portForwardRegistryServer()
 	}
 	// Images created by the tests should not be persisted after the test finishes.
@@ -466,6 +472,18 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 			}
 		}
 		nt.ociImages = nil
+	})
+	nt.T.Cleanup(func() {
+		for _, helmPackage := range nt.helmPackages {
+			remoteURL, err := nt.HelmProvider.PushURL(helmPackage.Name)
+			if err != nil {
+				nt.T.Fatal(err)
+			}
+			if err := helmPackage.Delete(remoteURL); err != nil {
+				nt.T.Error(err)
+			}
+		}
+		nt.helmPackages = nil
 	})
 	// The following prerequisites have been met, so we can now push commits
 	// - local repo initialized
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -134,6 +134,11 @@ type NT struct {
 	// crashes, and to tear down images at the end of the test execution.
 	ociImages []*registryproviders.OCIImage
 
+	// helmPackages tracks the images which have been pushed to the registry for the
+	// current test. This is used to re-push images if the in-cluster registry
+	// crashes, and to tear down images at the end of the test execution.
+	helmPackages []*registryproviders.HelmPackage
+
 	// MetricsExpectations tracks the objects expected to be declared in the
 	// source and the operations expected to be performed on them by the set of
 	// RootSyncs and RepoSyncs managed by this test.
@@ -173,6 +178,9 @@ type NT struct {
 
 	// OCIProvider is the provider that hosts the OCI repositories.
 	OCIProvider registryproviders.RegistryProvider
+
+	// HelmProvider is the provider that hosts the helm packages.
+	HelmProvider registryproviders.RegistryProvider
 
 	// RemoteRepositories maintains a map between the repo local name and the remote repository.
 	// It includes both root repo and namespace repos and can be shared among test cases.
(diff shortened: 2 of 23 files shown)


e2e/nomostest/new.go
```diff
@@ -467,6 +473,18 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 		}
 		nt.ociImages = nil
 	})
+	nt.T.Cleanup(func() {
+		for _, helmPackage := range nt.helmPackages {
+			remoteURL, err := nt.HelmProvider.PushURL(helmPackage.Name)
```
karlkfi, line 478: Seems like the name of this method could be something better. It's being used for delete here, not just push.

karlkfi: Perhaps ChartURL?

e2e/nomostest/nt.go
```diff
@@ -749,30 +757,63 @@ func (nt *NT) portForwardGitServer() {
 
 // portForwardRegistryServer forwards the registry-server deployment to a port.
 func (nt *NT) portForwardRegistryServer() {
-	provider := nt.OCIProvider.(*registryproviders.LocalOCIProvider)
-	resetImages := func(newPort int, podName string) {
-		for _, image := range nt.ociImages {
-			// construct push URL with provided port. Calling LocalPort would lead to deadlock
-			if err := image.Push(provider.PushURLWithPort(newPort, image.Name)); err != nil {
-				nt.T.Fatalf("pushing image: %w", err)
+	var onReadyCallbacks []func(int, string)
```
karlkfi, line 760: need a comment about what this does. maybe "The local registry uses ephemeral storage. So re-push all the oci images and helm charts after the registry comes up."

e2e/nomostest/nt.go
```diff
@@ -749,30 +757,63 @@ func (nt *NT) portForwardGitServer() {
...
-		// a new PortForwarder is created for each test.
-		provider.PortForwarder = nil
-	})
-	provider.PortForwarder = nt.newPortForwarder(
+	portForwarder := nt.newPortForwarder(
 		TestRegistryNamespace,
 		TestRegistryServer,
 		fmt.Sprintf(":%d", RegistryHTTPPort),
-		portforwarder.WithOnReadyCallback(resetImages),
+		portforwarder.WithOnReadyCallback(func(newPort int, podName string) {
```
karlkfi, line 789: Is there a possible race condition with the tests trying to use these images & charts before they've all been pushed?

Do we need to block waiting for them all to be pushed before continuing? If so, should we just do it inline here without the callbacks?

e2e/nomostest/registry_server.go
```diff
@@ -126,6 +126,35 @@ func (nt *NT) BuildAndPushOCIImage(repository *gitproviders.Repository) (*regist
 	return image, nil
 }
 
+// BuildAndPushHelmPackage uses the current file system state of the provided Repository
+// to build a helm package and push it to the current HelmProvider. The resulting
+// HelmPackage object can be used to set the spec.oci.image field on the RSync.
+func (nt *NT) BuildAndPushHelmPackage(repository *gitproviders.Repository) (*registryproviders.HelmPackage, error) {
```
karlkfi, line 132: they're called charts, not packages in the helm universe.

e2e/nomostest/new.go
sdowell: The important distinction is that this is the URL that is fit for use from the test framework, as this returns the port forward from localhost. The other URL (SyncURL) returns the Service DNS from inside the cluster for use by helm-sync/oci-sync/git-sync. LMK if you have better ideas for a name

e2e/nomostest/nt.go
sdowell: This callback is called from a different thread by the portforwarder. The point of this is that the registry could crash at any time during the test, and this callback handles the auto-recovery. I don't think there's much to do about synchronizing this in the general case.

e2e/nomostest/registryproviders/helm.go
```diff
@@ -0,0 +1,142 @@
...
+	"kpt.dev/configsync/e2e/nomostest/testshell"
+	"sigs.k8s.io/kustomize/kyaml/copyutil"
+	"sigs.k8s.io/yaml"
+)
+
+// use an auto-incrementing index to create unique file names for tarballs
+var helmIndex int
+
+// BuildHelmPackage creates a new OCIImage object and associated tarball using the provided
+// Repository. The contents of the git repository will be bundled into a tarball
```
karlkfi, line 34: Do we need to do this in the git repo? Can we use a separate directory so we don't include the .git meta files?

karlkfi: Or does it need to be tied to a git repo for some testing reasons? Seems kinda tightly coupled, but maybe this code predates this PR?

e2e/nomostest/registry_server.go
karlkfi: i guess they're packages in the context of config sync... just seems confusing to call them that. maybe it doesn't realy matter...

e2e/nomostest/registryproviders/helm.go
sdowell: This design was deliberate with the intent of minimizing the number of primitives that the tests need to work with. The intent is basically that you can add files to the `Repository` object and then push the artifact to any sync source (git, oci, helm, etc). This way we can also deal with versioning consistently (e.g. tag things with the git branch).

I think this works pretty well for oci from the previous PR, but admittedly feels a bit clunkier with helm because the file/directory structure is very specific to helm.

e2e/nomostest/registry_server.go
sdowell: I don't feel strongly, I'm fine with calling them either

e2e/nomostest/new.go
karlkfi: I guess if we want to keep the method names similar for the helm provider and oci provider, then using "package" is ok. I think maybe we could refactor the names on the providers, but perhaps that should be a follow up PR to avoid making this one larger.

e2e/nomostest/nt.go
karlkfi: Does the proxy block/delay requests if the backend is down?

If so, does that blocking also include the time spent re-pushing the images that are expected to exist?

pkg/helm/helm.go
```diff
@@ -246,15 +248,27 @@ data:
 	return pkgReadWriter.Write(newNodes)
 }
 
+func (h *Hydrator) helm(ctx context.Context, args ...string) ([]byte, error) {
+	var allArgs []string
+	allArgs = append(allArgs, args...)
+	if h.CACertFilePath != "" {
+		allArgs = append(allArgs, "--ca-file", h.CACertFilePath)
+	}
+	out, err := exec.CommandContext(ctx, "helm", allArgs...).CombinedOutput()
```
karlkfi, line 257: do we want the logging from shell.ExecWithDebug here?

e2e/nomostest/new.go
karlkfi: maybe i'll take a a refactoring pass afterwards. it doesn't really need to block this fucntionality.

pkg/helm/helm.go
sdowell: This isn't part of the e2e test scaffolding (where `shell.ExecWithDebug` is from), this is the helm-sync code

I don't think it's always safe to log the args here due to credentials being passed via command line args.

e2e/nomostest/registryproviders/helm.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, fmt.Errorf("getting hash: %w", err)
+	}
+	branch, err := repository.CurrentBranch()
+	if err != nil {
+		return nil, fmt.Errorf("getting branch: %w", err)
+	}
+	// replace / with - to avoid creating nested directories/paths
+	name := strings.Replace(repository.Name, "/", "-", -1)
+	// helm forces a semver, but we can append the branch to create a floating tag
+	version := fmt.Sprintf("v1.0.0-%s", branch)
```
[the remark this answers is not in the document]
sdowell: Not with the current usage (which is just the new tests added with this change), but if there's a need then it can be added as a parameter

e2e/nomostest/nt.go
sdowell: In a way. The [portforwarder blocks calls to LocalPort while the backend is down](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/portforwarder/port_forwarder.go#L142), which is how the test scaffolding gets the proxy port. Any test logic that uses `LocalPort` may need some retry logic to recover if the backend crashes while trying to get the LocalPort.

Before unlocking LocalPort, the portforwarder [invokes each of the registered callbacks](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/portforwarder/port_forwarder.go#L151), which in this case re-pushes the images which are expected to exist.

karlkfi: Ah. I see `update` sets `isReady` first, but then calls `onReadyCallback` inside the same lock.

That should prevent the race condition I was concerned about.

It's messy to have the callback impl calling a PortForwarder method inside the callback tho. I wonder if we could separate the building of the URL path from the building of the address. Building the address needs to be locked but building the path doesn't. Then we could just pass the address into the callback, instead of calling a method to do it. And adding the path could be a package function, instead of a struct method.

karlkfi: You or I could do another PR on this after. Doesn't need to block this one.

e2e/nomostest/nt.go
sdowell: done

karlkfi (approved):
/lgtm

PR #1113, opened 2024-01-26
test: refactor oci/helm tests for registryproviders

sdowell:
This refactors the existing helm-sync/oci-sync tests which push artifacts to use the new RegistryProvider interface. This delivers a few benefits:
- minimize duplicate test scaffolding
- enable running these tests without GKE/AR
- enable running these tests with the presubmits in kind


e2e/nomostest/ntopts/new.go
```diff
@@ -101,6 +101,14 @@ func RequireKind(t testing.NTB) Opt {
 	return func(opt *New) {}
 }
 
+// RequireHelmArtifactRegistry requires the --helm-provider flag to be set to artifact_registry
+func RequireHelmArtifactRegistry(t testing.NTB) Opt {
```
nan-yu, line 105: Do we also need a `RequireOCIArtifictRegistry` option?

e2e/nomostest/registryproviders/artifact_registry.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// ArtifactRegistryOCIProvider provides methods for interacting with the test registry-server
+// using the oci-sync interface.
+type ArtifactRegistryOCIProvider struct {
+	ArtifactRegistryProvider
+}
+
+// Setup performs setup
+func (a *ArtifactRegistryOCIProvider) Setup() error {
```
nan-yu, line 123: The Helm provider sets up the authentication and create the repository.
The OCI provider also needs to set up the authentication.

e2e/nomostest/config_sync.go
```diff
@@ -816,6 +817,128 @@ func RepoSyncObjectV1Beta1(nn types.NamespacedName, repoURL string, sourceFormat
...
+	}
+	switch *e2e.OCIProvider {
+	case e2e.Local:
+		// Local provider requires CA cert because host is not well known
+		rs.Spec.Oci.CACertSecretRef = &v1beta1.SecretReference{
+			Name: PublicCertSecretName(RegistrySyncSource),
+		}
+	case e2e.ArtifactRegistry:
+		// AR provider requires auth because the registry is private
+		rs.Spec.Oci.Auth = configsync.AuthGCPServiceAccount
```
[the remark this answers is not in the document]
sdowell: Done. I'll need to revalidate the tests using this auth method

e2e/nomostest/gitproviders/repository.go
```diff
@@ -499,6 +499,28 @@ func (g *Repository) Copy(sourceDir, destDir string) error {
 	return err
 }
 
+// UseHelmChart copies the files from the provided helm chart to the current
+// repository.
+func (g *Repository) UseHelmChart(chart string) error {
```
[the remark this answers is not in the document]
sdowell: The idea here is that we can use the same `Repository` object as a primitive for creating all types of artifacts (git, oci, helm). There are tests which use the pattern of adding individual files for a helm chart and pushing it, this is just intended as a convenience function to copy over the premade helm charts from the local directory.

e2e/nomostest/registryproviders/artifact_registry.go
sdowell: Done

e2e/flags.go
```diff
@@ -250,6 +281,8 @@ const (
 	GitLab = "gitlab"
 	// CSR indicates using Google Cloud Source Repositories to host the repositories.
 	CSR = "csr"
+	// ArtifactRegistry indicates using Google Artifact Registry to host the repositories.
+	ArtifactRegistry = "artifact_registry"
```
karlkfi, line 285: If we have to type this out sometimes, i suggest "gar" as the typical acronym, similar to "csr".

Or we could spell out csr. If we spell it out i guess I would recommend hyphens, rather than underscores, but I guess it's not really a big deal.

e2e/nomostest/config_sync.go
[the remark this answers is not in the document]
janetkuo: Is it an issue with k8sserviceaccount support in RepoSync or just the test infra configuration?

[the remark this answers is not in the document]
karlkfi: Did you add the right IAM to allow the KSA to use artifactregistry in your project?

- gcpserviceaccount needs you to authorize the GSA. Then the KSA needs IAM to impersonate the GSA.
- k8sserviceaccount just needs you to authorize the KSA directly.

e2e/nomostest/config_sync.go
```diff
@@ -816,6 +817,128 @@ func RepoSyncObjectV1Beta1(nn types.NamespacedName, repoURL string, sourceFormat
...
+		rs.Spec.Oci.GCPServiceAccountEmail = registryproviders.ArtifactRegistryReaderEmail()
+	}
```
karlkfi, line 882: since you have the nt, you can fail the test in the default case where the provider is invalid.

e2e/nomostest/config_sync.go
```diff
@@ -816,6 +817,128 @@ func RepoSyncObjectV1Beta1(nn types.NamespacedName, repoURL string, sourceFormat
...
+func (nt *NT) RootSyncObjectHelm(name string, chart *registryproviders.HelmPackage) *v1beta1.RootSync {
+	rs := RootSyncObjectV1Beta1FromRootRepo(nt, name)
+	rs.Spec.SourceType = string(v1beta1.HelmSource)
+	rs.Spec.Helm = &v1beta1.HelmRootSync{
+		HelmBase: v1beta1.HelmBase{
+			Repo:    nt.HelmProvider.SyncURL(chart.Name),
+			Chart:   chart.Name,
+			Version: chart.Version,
+			Auth:    configsync.AuthNone,
+			Period:  metav1.Duration{Duration: 15 * time.Second},
```
karlkfi, line 896: Maybe move this timeout to a const with a comment as to why we're shortening it and not testing the normal behavior.

Are we testing behavior that normally would take an hour to complete? Is there some way we can do it that matches user behavior better? Or is this the case where we need to make the behavior more complicated and this is just a temporary workaround? If so, maybe add a TODO on the const.

e2e/nomostest/gitproviders/repository.go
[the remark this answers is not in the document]
karlkfi: I had the same confusion looking at the Repository. I expected two interfaces, one for helm and one for OCI. Having them in the same interface feels like leaking implementation details.

e2e/nomostest/config_sync.go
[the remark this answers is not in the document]
sdowell: I was going off of https://github.com/GoogleContainerTools/kpt-config-sync/pull/1121, which didn't add any new provisioning config/logic for IAM. After inspecting the CI project, I see that new IAM must have been added manually for the root-reconciler KSA. I suggest adding this to the provisioning config so that it is automated in the future

e2e/nomostest/ntopts/new.go
karlkfi: Do we need a follow up PR to add RequireOCIArtifictRegistry to the tests that need it? Or are you saying there aren't any?

sdowell: There aren't any currently written in a way that need it.

e2e/nomostest/registryproviders/artifact_registry.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := a.registryLogin(); err != nil {
+		return err
+	}
+	return a.createRepository()
+}
+
+// PushURL returns a URL for pushing images to the remote registry.
+// The name parameter is ignored because helm CLI appends the chart name to the image.
+func (a *ArtifactRegistryHelmProvider) PushURL(_ string) (string, error) {
+	return fmt.Sprintf("oci://%s", a.repositoryAddress()), nil
```
karlkfi, line 204: if this is the same value, you can just call SyncURL here i guess, rather than repeating the format.

e2e/nomostest/registryproviders/local.go
```diff
@@ -19,20 +19,83 @@ import (
...
+		return "", fmt.Errorf("PortForwarder must be set for LocalProvider.localAddress()")
+	}
+	port, err := l.PortForwarder.LocalPort()
+	if err != nil {
+		return "", err
+	}
+	return l.localAddressWithPort(port, name), nil
+}
+
+func (l *LocalProvider) localAddressWithPort(localPort int, name string) string {
```
karlkfi, line 70: seems like these address and image path functions could use some comments about what they're returning and why.

e2e/testcases/oci_sync_test.go
```diff
@@ -442,9 +378,9 @@ func getImageDigest(nt *nomostest.NT, imageName string) (string, error) {
 	return hex, nil
 }
 
-// imageDigestFunc wraps getImageDigest to return a Sha1Func that caches the
+// fetchImageDigestFunc wraps getImageDigest to return a Sha1Func that caches the
 // image digest, to avoid polling the image registry unnecessarily.
-func imageDigestFunc(imageName string) nomostest.Sha1Func {
+func fetchImageDigestFunc(imageName string) nomostest.Sha1Func {
```
karlkfi, line 383: Probably woulda called these funcs `imageDigestFuncByName` and `imageDigestFuncByDigest`. 

It took me a while to figure out why you had renamed cause i found `fetchImageDigestFunc` confusing... It's not fetching the func, it's getting a func to fetch (and cache) the digest.

e2e/testcases/stress_test.go
```diff
@@ -467,61 +467,62 @@ func TestStressMemoryUsageGit(t *testing.T) {
 // 7. IAM for the test runner to write to Artifact Registry repo
 func TestStressMemoryUsageOCI(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.WorkloadIdentity, ntopts.Unstructured,
-		ntopts.StressTest, ntopts.RequireGKE(t),
+		ntopts.StressTest,
```
karlkfi, line 470: Why doesnt this need GKE any more? Isn't it still using ArtifactRegistry? Or does that work from kind and other platforms now?

sdowell: The test can run on other cluster and/or registry types now, depending on what config is provided at runtime.

e2e/nomostest/config_sync.go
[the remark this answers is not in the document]
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1126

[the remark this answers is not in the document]
nan-yu: Yes, those IAM were manually added. #1127 added the configs.

e2e/flags.go
sdowell: Done

e2e/nomostest/config_sync.go
```diff
@@ -816,6 +817,128 @@ func RepoSyncObjectV1Beta1(nn types.NamespacedName, repoURL string, sourceFormat
 	return rs
 }
 
+// UpdateRootSyncSpec updates the RootSync object on the cluster with the provided
+// spec.
+func (nt *NT) UpdateRootSyncSpec(rs *v1beta1.RootSync) error {
```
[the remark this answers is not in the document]
sdowell: Done

e2e/nomostest/config_sync.go
sdowell: The providers should already be validated at this point, but no harm in adding it here. Done

e2e/nomostest/config_sync.go
sdowell: done

e2e/nomostest/registryproviders/artifact_registry.go
sdowell: done

e2e/nomostest/registryproviders/local.go
sdowell: done

e2e/testcases/oci_sync_test.go
sdowell: done

e2e/nomostest/config_sync.go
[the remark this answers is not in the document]
karlkfi: Does it work now?

e2e/nomostest/ntopts/new.go
karlkfi: Sounds like we're missing some tests...

e2e/nomostest/config_sync.go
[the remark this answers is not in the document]
sdowell: Yes It did when I tested with the RepoSync IAM from https://github.com/GoogleContainerTools/kpt-config-sync/pull/1126.

e2e/nomostest/ntopts/new.go
sdowell: Perhaps. There are certain tests for helm-sync which don't have an analog with oci-sync, for example helm-sync supports token auth while oci-sync does not.

There are also still a number of oci-sync tests which are written to pull from a static image whereas these helm-sync tests push a chart during the test. The registry provider type is only applicable when pushing artifacts during the test.

e2e/testcases/stress_test.go
karlkfi: So it still pulls from GAR even when run on kind?

sdowell: That probably would fail on image pull due to permission issues assuming it's a non-public registry, but you could in theory.

I expect the valid permutations would be local registry provider on kind/gke or GAR provider on gke

karlkfi: How does kind get credentials to auth to GCP?

sdowell: It doesn't currently, that's why I said it would fail

e2e/nomostest/ntopts/new.go
karlkfi: Can you file some bugs for this tech debt to make the tests more consistent/robust?

e2e/testcases/stress_test.go
sdowell: If you prefer I can add input validation that prevents the tests from being run with artifact registry on kind

karlkfi: oh sorry, it didnt load the response before i sent the follow up.

karlkfi (approved):
/lgtm

thanks for cleaning it up a bit. we can always clean it up more later if we feel like it.

PR #1114, opened 2024-01-29
fix: update the incorrect error messages

nan-yu:
Some error messages apply to all source types, but these messages are specific to the git source. This commit updates the messages by passing in the source type.


tiffanny29631 (approved):
/lgtm

PR #1115, opened 2024-01-29
refactor: extracts test utils to a common package

nan-yu:
This commit has no functional changes.
It extracts the workload identity related test scaffolding to a common package. 

To avoid cyclic import, it moves the `ValidateError` function from 'testutils/validate.go' to 'testpredicates/predicates.go', and renames `testutils/validate.go` to `testutils.finalizier.go`.

The intent of these changes is to minimize changes to a following PR with BYOID.


janetkuo (approved):
/lgtm

e2e/nomostest/artifactregistry/helm.go
```diff
@@ -165,3 +166,10 @@ func (r *HelmChart) Push() error {
 	}
 	return nil
 }
+
+// HelmChartVersion returns the provided chart version as a function.
+func HelmChartVersion(chartVersion string) nomostest.Sha1Func {
```
sdowell, line 171: This isn't specific to artifact registry, so I don't think this function belongs here. I'm also getting rid of this package with my other PR

nan-yu: /hold

e2e/nomostest/testutils/fleet_membership.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/pkg/reconcilermanager"
+)
+
+// TestCrossProjectFleetProjectID is the fleet host project.
+const TestCrossProjectFleetProjectID = "cs-dev-hub"
+
+// RegisterCluster registers a cluster in a fleet.
+func RegisterCluster(nt *nomostest.NT, fleetMembership, fleetProject, gkeURI string) error {
```
sdowell, line 29: Do these functions need to be public? Looks like they are only intended to be used within this package

e2e/nomostest/testutils/workload_identity_test_spec.go
```diff
@@ -0,0 +1,224 @@
...
+	RootCommitFn  nomostest.Sha1Func
+}
+
+// TestWorkloadIdentity sets up the test skeleton for both GKE WI and Fleet WI.
+// It has the following requirements:
+// 1. GKE cluster with workload identity enabled.
+// 2. The provided Google service account exists.
+// 3. The cluster is registered to a fleet if testing Fleet WI.
+// 4. IAM permission and IAM policy binding are created.
+func TestWorkloadIdentity(t *testing.T, testSpec WITestSpec) {
```
sdowell, line 61: I'm not sure how I feel about this function being moved here, this is moreso the whole testcase implementation rather than a test util.

nan-yu: Agreed that this is the whole testcase implementation. 
The original intent is to extract some utility functions into the `testutils` package, such as `registerCluster`, `describeMembership`, and etc.
However, to avoid import cycle, I moved this function along with other utils.

Technically, it is still a test util which relies on actual testcases to provide the arguments.

Do you have any other suggestions?

e2e/nomostest/artifactregistry/helm.go
nan-yu: ACK. I can find a better place to put the function. If you don't mind, I would like to get this change in first to unblock 1.17.2 release, and then get #1113 merged after.

e2e/nomostest/testutils/fleet_membership.go
nan-yu: Changed it back to package scoped

e2e/nomostest/artifactregistry/helm.go
nan-yu: moved to `e2e/nomostest/nt.go` to be along with other similar Sha1Funcs.

e2e/nomostest/testutils/workload_identity_test_spec.go
nan-yu: /unhold

sdowell: Perhaps we could write these as a single data driven test with multiple test cases, as we've done elsewhere (for [example](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/cli_test.go#L674)). I think something like `e2e/testcases/workload_identity_test.go` would make sense to home all of these workload identity tests.

What is the source of the import cycle?

nan-yu: If you categorize those tests by the workload identity feature, yes, it makes sense with  `e2e/testcases/workload_identity_test.go`.
The testcases are currently divided by different source types.

nan-yu: The import cycle comes from the `ReconcilerPodMissingFWICredsAnnotation` validation function. It is used by both testcases and testutils.

sdowell: I was thinking something like:
```golang
func TestWorkloadIdentity(t *testing.T) {
    testCases := map[string]struct{
        FleetWITest   bool
        CrossProject  bool
        SourceRepo    string
        SourceChart   string
        SourceVersion string
        SourceType    v1beta1.SourceType
        GSAEmail      string
        RootCommitFn  nomostest.Sha1Func
    }{...}
}
```
Primarily to deobfuscate the test logic/assertions from the actual test case.

We could potentially refactor as a subsequent PR, although with the scope of this change I'm concerned about conflicts if we are intending to cherry pick this.

nan-yu: That makes sense. I'll update the current PR to avoid cherrypick conflicts.

nan-yu: done

e2e/nomostest/testutils/hydrated_resources_validation.go
```diff
@@ -0,0 +1,89 @@
...
+func ValidateAllTenants(nt *nomostest.NT, reconcilerScope, baseRelPath string, tenants ...string) {
+	nt.T.Logf("Validate resources are synced for all tenants %s", tenants)
+	ValidateNamespaces(nt, tenants, fmt.Sprintf("path: %s/namespace.yaml\n", baseRelPath))
+	for _, tenant := range tenants {
+		ValidateTenant(nt, reconcilerScope, tenant, baseRelPath)
+	}
+}
+
+// ValidateTenant validates the provided tenant resources under
+// '../testdata/hydration/kustomize-components' are reconciled.
```
sdowell, line 41: nit: These names seem a bit too generic given that the validation is specific to `kustomize-components`. Perhaps we could move these to a `kustomize_components` subpackage, such that the call would look like `kustomize_components.ValidateNamespaces` etc.

sdowell (commented):
A few remaining nits about function naming but mostly lgtm

e2e/nomostest/testutils/hydrated_resources_validation.go
nan-yu: I like it. The call looks self-description after moving into a specific subpackage. Thanks!

sdowell (approved):
/lgtm

e2e/nomostest/testutils/workload_identity_test_spec.go
nan-yu: @karlkfi FYI, CS CI will fail due to this issue and a bunch of bugs will be auto-created and assigned to you as the next onduty.
I will come up with a fix instead of reverting this PR to avoid messy cherrypick history.

sdowell: Ah yeah the `t.Name` must be the same for each of the test cases. We may need to either fix that or revert the change. Sorry about the mixup

nan-yu: should be fixed by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1120. I verified with my clusters.

PR #1116, opened 2024-01-30
fix: get AR authenticator with retries

nan-yu:
The AR authenticator was configured outside of the for loop. If the IAM
policy bindings are changed, the authenticator would NOT be refreshed.
Then it requires to restart the reconciler Pod.
This commit moves the AR authenticator inside the for loop.

### DIFF
--- a/cmd/oci-sync/main.go
+++ b/cmd/oci-sync/main.go
@@ -22,8 +22,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/google/go-containerregistry/pkg/v1/google"
 	"k8s.io/klog/v2/klogr"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/oci"
@@ -82,25 +80,11 @@ func main() {
 		utillog.HandleError(log, true, "ERROR: --timeout must be greater than 0")
 	}
 
-	var auth authn.Authenticator
-	switch configsync.AuthType(*flAuth) {
-	case configsync.AuthNone:
-		auth = authn.Anonymous
-	case configsync.AuthGCPServiceAccount, configsync.AuthK8sServiceAccount, configsync.AuthGCENode:
-		a, err := google.NewEnvAuthenticator()
-		if err != nil {
-			utillog.HandleError(log, true, "ERROR: failed to get the authentication with type %q: %v", *flAuth, err)
-		}
-		auth = a
-	default:
-		utillog.HandleError(log, true, "ERROR: unsupported authentication type %q", *flAuth)
-	}
-
 	initialSync := true
 	failCount := 0
 	for {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(*flSyncTimeout))
-		if err := oci.FetchPackage(ctx, *flImage, *flRoot, *flDest, auth); err != nil {
+		if err := oci.FetchPackage(ctx, log, *flAuth, *flImage, *flRoot, *flDest); err != nil {
 			if *flMaxSyncFailures != -1 && failCount >= *flMaxSyncFailures {
 				// Exit after too many retries, maybe the error is not recoverable.
 				log.Error(err, "too many failures, aborting", "failCount", failCount)
--- a/pkg/oci/fetcher.go
+++ b/pkg/oci/fetcher.go
@@ -24,15 +24,36 @@ import (
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/google"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"golang.org/x/net/context"
 	"k8s.io/klog/v2"
+	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/util"
+	utillog "kpt.dev/configsync/pkg/util/log"
 )
 
+// authenticator returns an authn.Authenticator that generates access tokens.
+func authenticator(authType string, logger *utillog.Logger) (authenticator authn.Authenticator, err error) {
+	switch configsync.AuthType(authType) {
+	case configsync.AuthNone:
+		return authn.Anonymous, nil
+	case configsync.AuthGCPServiceAccount, configsync.AuthK8sServiceAccount, configsync.AuthGCENode:
+		return google.NewEnvAuthenticator()
+	default:
+		utillog.HandleError(logger, true, "ERROR: unsupported authentication type %q", authType)
+	}
+	return nil, nil
+}
+
 // FetchPackage fetches the package from the OCI repository and write it to the destination.
-func FetchPackage(ctx context.Context, imageName, ociRoot, rev string, auth authn.Authenticator) error {
+func FetchPackage(ctx context.Context, logger *utillog.Logger, authType, imageName, ociRoot, rev string) error {
+	auth, err := authenticator(authType, logger)
+	if err != nil {
+		return fmt.Errorf("failed to get the authentication with type %q: %w", authType, err)
+	}
+
 	image, err := PullImage(imageName, remote.WithContext(ctx), remote.WithAuth(auth))
 	if err != nil {
 		return err


e2e/nomostest/iam/account.go
```diff
@@ -39,3 +44,32 @@ func DescribeServiceAccount(nt *nomostest.NT, gsaEmail, projectID string) ([]byt
...
+	if err != nil {
+		return fmt.Errorf("adding IAM policy binding to an IAM service account: %w", err)
+	}
+	return nil
+}
+
+// UnimpersonateGSA removes the IAM policy binding from the IAM service account
+func UnimpersonateGSA(nt *nomostest.NT, fleetProject, gsaEmail string, ksaRef types.NamespacedName) error {
+	out, err := iamPolicyBinding(nt, "remove-iam-policy-binding", gsaEmail, WIUserRole, getWIUser(fleetProject, ksaRef))
+	if err != nil && !strings.Contains(string(out), "Policy binding with the specified principal, role, and condition not found") {
```
janetkuo, line 64: Output is subject to change and could break our tests. Does it provide a flag for ignoring not found errors?

nan-yu: Unfortunately, it doesn't provide such a flag. Another option is to only check for the `not found` string, but it seems not very specific.

janetkuo: That's fine. We'll need to update test code if the gcloud output changes and breaks us.

e2e/testcases/helm_sync_test.go
```diff
@@ -396,15 +396,29 @@ image:
 // are set. Namespace-scoped resources with their namespace field set should be deployed to their set namespace; namespace-scoped
 // resources with their namespace field not set should be deployed to the default namespace.
 // This test will work only with following pre-requisites:
-// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for accessing images in Artifact Registry.
+// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with.
```
janetkuo, line 399: Why's the line about role deleted?

e2e/testcases/helm_sync_test.go
```diff
@@ -434,31 +448,38 @@ func TestHelmDefaultNamespace(t *testing.T) {
...
-// 2. The Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for access image in Artifact Registry.
-// 3. An IAM policy binding is created between the Google service account and the Kubernetes service accounts with the `roles/iam.workloadIdentityUser` role.
-//
-//	gcloud iam service-accounts add-iam-policy-binding --project=${GCP_PROJECT} \
-//	   --role roles/iam.workloadIdentityUser \
-//	   --member "serviceAccount:${GCP_PROJECT}.svc.id.goog[config-management-system/root-reconciler]" \
-//	   e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com
-//
-// 4. The following environment variables are set: GCP_PROJECT, GCP_CLUSTER, GCP_REGION|GCP_ZONE.
+// 2. The Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created.e following environment variables are set: GCP_PROJECT, GCP_CLUSTER, GCP_REGION|GCP_ZONE.
```
janetkuo, line 451: typo around `created.e`

e2e/testcases/helm_sync_test.go
nan-yu: Because the permission and IAM policy binding are created during the test.
This is because gcpserviceaccount and k8sserviceaccount tests are going to create and delete those bindings before and after test, so we can't always assume they always exist.

e2e/testcases/helm_sync_test.go
nan-yu: done

e2e/testcases/helm_sync_test.go
janetkuo: I see, I thought this comment was about what the test's going to do.

e2e/nomostest/projects/iam_policy_binding.go
```diff
@@ -0,0 +1,49 @@
...
+	out, err := iamPolicyBinding(nt, "add-iam-policy-binding", principle, role)
+	nt.T.Logf("Add IAM Policy binding: %s", string(out))
+	if err != nil {
+		return fmt.Errorf("adding IAM policy binding for a project: %w", err)
+	}
+	return nil
+}
+
+// RemoveIAMPolicyBinding removes an IAM policy binding for a project
+func RemoveIAMPolicyBinding(nt *nomostest.NT, principle, role string) error {
```
sdowell, line 36: `principal`

e2e/testcases/workload_identity_test.go
```diff
@@ -286,3 +278,136 @@ func truncateStringByLength(s string, l int) string {
...
+		return "roles/artifactregistry.reader", nil // OCI images on AR
+	case v1beta1.HelmSource:
+		return "roles/artifactregistry.reader", nil // Helm charts on AR
+	default:
+		return "", fmt.Errorf("invalid source type %q", sourceType)
+	}
+}
+
+// grantIAMPermissions sets up IAM policy binding to grant IAM permissions
+func grantIAMPermissions(nt *nomostest.NT, principle, role string) {
```
sdowell, line 392: The test-runner GSA doesn't have permission to manage IAM currently, and I'm not sure that's something we want to add due to privilege escalation.

Can this be managed with the terraform config?

nan-yu: Terraform config is a one-off configuration. The IAM permissions are created and deleted by each testcase.
I think we may need to grant the test-runner GSA permission to manage IAM.

e2e/testcases/helm_sync_test.go
```diff
@@ -396,24 +397,41 @@ image:
 // are set. Namespace-scoped resources with their namespace field set should be deployed to their set namespace; namespace-scoped
 // resources with their namespace field not set should be deployed to the default namespace.
 // This test will work only with following pre-requisites:
-// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for accessing images in Artifact Registry.
+// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created.
 func TestHelmDefaultNamespace(t *testing.T) {
```
sdowell, line 401: This conflicts with several of my changes in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1113. I would have preferred to get those refactoring changes in first, but if necessary I can rebase on this change

e2e/testcases/workload_identity_test.go
sdowell: Why do we need to create/delete with every test case? Can we use multiple GSAs?

e2e/testcases/helm_sync_test.go
nan-yu: This PR needs to be cherry-picked on to the release branch, but #1113 probably doesn't have to.
To minimize changes in a patch release, I would rather #1113 to be rebased any merged after.

sdowell: Yeah, that makes sense. Unfortunately I'll have to manually revalidate those tests for each of the gke/provider permutations after rebasing 😢

One of the costs for cherry picking a feature

e2e/testcases/helm_sync_test.go
```diff
@@ -434,31 +452,41 @@ func TestHelmDefaultNamespace(t *testing.T) {
...
+
 	chart, err := artifactregistry.PushHelmChart(nt, privateSimpleHelmChart, privateSimpleHelmChartVersion)
 	if err != nil {
 		nt.T.Fatalf("failed to push helm chart: %v", err)
 	}
 
 	nt.T.Log("Update RootSync to sync from a private Artifact Registry")
+	// Set the helm re-pulling duration to 5s instead of relying on the default 1h,
+	// particularly in scenarios where the new IAM policy bindings may not have
+	// been fully effective.
```
janetkuo, line 487: nit: Here 5s was set for pulling the latest chart version. Would you keep that information here as well?

e2e/testcases/helm_sync_test.go
```diff
@@ -396,24 +397,41 @@ image:
...
 	if err != nil {
 		nt.T.Fatalf("failed to push helm chart: %v", err)
 	}
 
 	nt.T.Log("Update RootSync to sync from a private Artifact Registry")
-	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "git": null, "helm": {"repo": "%s", "chart": "%s", "version": "%s", "auth": "gcpserviceaccount", "gcpServiceAccountEmail": "%s", "namespace": "", "deployNamespace": ""}}}`,
-		v1beta1.HelmSource, chart.Image.RepositoryOCI(), chart.Image.Name, chart.Image.Version, gsaARReaderEmail()))
+	// Set the helm re-pulling duration to 5s instead of relying on the default 1h,
+	// particularly in scenarios where the new IAM policy bindings may not have
+	// been fully effective.
```
janetkuo, line 432: It might be more clear to say:
```suggestion
	// because updates to IAM policy bindings doesn't trigger a reconciliation.
```
Suggesting changing other similar comments too.

e2e/testcases/workload_identity_test.go
nan-yu: The test case creates/deletes these bindings for two purposes:
1. Identify the minimal bindings required for each auth type (either gcpserviceaccount or k8sserviceaccount).
2. Add validations for the KNV2004 errors. There're actually two types of  permission errors: permissions not granted, and IAM bindings not created for the WI. As a side-effect, these validations align the testcases with the documentation. After these changes, we need to update the documentation for each invalid scenario.

e2e/testcases/helm_sync_test.go
nan-yu: Agreed and sorry for the extra costs.

If not cherry picking the feature in a patch release, it will block other services :disappointed:

e2e/testcases/helm_sync_test.go
nan-yu: done

e2e/testcases/helm_sync_test.go
nan-yu: removed

e2e/nomostest/projects/iam_policy_binding.go
nan-yu: updated

e2e/testcases/workload_identity_test.go
sdowell: I'm still a bit concerned about the risk. I imagine the test could verify the various error modes by switching between multiple GSAs. Is there a reason this wouldn't work?

nan-yu: Switching between multiple GSAs doesn't seem to be a normal user flow and will trigger a Pod restart.
By using the same GSA with IAM bindings created/deleted, this PR identifies the repulling issue in oci-sync, and helm-sync requires a more frequent re-pulling period.

sdowell: It just occurred to me a larger issue with this is that this strategy won't work when multiple tests are running against the same project in parallel, as they do in CI. I expect this would cause a lot of test flakiness due to tests stepping on each other

nan-yu: That's right. I was testing all the tests in sequence.
What about creating a unique GSA for each test, and then delete it after?

sdowell: Is it necessary that we test re-polling in an e2e test? Perhaps a unit test would suffice.

Creating a unique GSA for each test could probably work for CI.

nan-yu: Hmm, re-polling test can be done in a unit test, but there is probably no way to distinguish the following KNV2004 errors:
1. missing IAM binding between GSA and KSA: https://github.com/GoogleContainerTools/kpt-config-sync/blob/d4d2f694645ec092f7dcb44a582b7ec3ebf107e4/e2e/testcases/workload_identity_test.go#L231
2. missing IAM permission for GSA: https://github.com/GoogleContainerTools/kpt-config-sync/blob/d4d2f694645ec092f7dcb44a582b7ec3ebf107e4/e2e/testcases/workload_identity_test.go#L250

It maybe okay as we will recommend using BYOID over GSA for simplicity.

> Creating a unique GSA for each test could probably work for CI.

It works with GSA, but not BYOId unless we create a unique RSync for each test.

Given the above reasons, let me change it back by assuming all permissions are pre-granted.

nan-yu: The validation for different permission errors can be extracted to a separate PR.

This PR is only for fixing AR authenticator with retries.
It would be tricky to add a unit test because we will need to verify the main process won't exit abnormally (the previous behavior) when it fails to get a valid authenticator.

nan-yu: This is related to BYOID, but not a requirement.
The issue only happens when the permission is missing initially, and then granted later. The workaround is to restart the Pod.

Given that. it doesn't need to be cherrypicked to the 1.17 branch, but can wait until 1.18.0.

PR #1117, opened 2024-01-31
Aggregate kustomize metrics labels

tiffanny29631:
Remove the configsync.sync labels from Monarch pipeline for Kustomize metrics.

### DIFF
--- a/e2e/testcases/otel_collector_test.go
+++ b/e2e/testcases/otel_collector_test.go
@@ -30,6 +30,7 @@ import (
 	"google.golang.org/genproto/googleapis/api/monitoredres"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/iam"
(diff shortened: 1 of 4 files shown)


e2e/testcases/otel_collector_test.go
```diff
@@ -96,11 +101,17 @@ func TestOtelCollectorDeployment(t *testing.T) {
 
 	startTime := time.Now().UTC()
 
-	nt.T.Log("Adding test commit after otel-collector is started up so multiple commit hashes are processed in pipelines")
-	namespace := fake.NamespaceObject("foo")
-	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/ns.yaml", namespace))
-	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Adding foo namespace"))
-	if err := nt.WatchForAllSyncs(); err != nil {
+	nt.T.Log("Add the kustomize components root directory to enable kustomize metrics")
```
nan-yu, line 104: Should we validate kustomize metrics are exported successfully?
I guess without the changes, we can still get other metrics, but not kustomize metrics.

tiffanny29631: The Kustomize metrics are only exported to cloud monarch, and iirc there is quite a latency for the metrics from WI enabled clusters to show up in Cloud Monarch, let me first see if the test can pull metrics from cloud monarch and then see how soon it is.

tiffanny29631: Cloud monarch might not be accessible from the test in OSS, it's a internal service that does not provide public APIs. I can add a case where old configuration _creates_ the error and latest config clears it out.

karlkfi: Should we test this in nomos-operator so we have access to Cloud Monarch?

tiffanny29631: There is a go/mash that we may be able to leverage for listing metrics in cloud monarch. I'm not sure if running in acm-ci would grant more access than from the oss, and the change might not be trivial. For getting this fix into 1.17.2 I recommend testing the other way around - that no error is expected when Kustomize component is running, and expecting errors with old config that doesn't aggregate.

tiffanny29631: Was researching for a feasible way to fetch the internal metrics - there are PII data involved so access control is strict. They do have stubby client that is not fancily documented for querying the metrics.

tiffanny29631: Listing metrics itself is not enough for this case either, we need to confirm the data stream in a certain time frame for a project.

nan-yu (approved):
/lgtm

PR #1120, opened 2024-02-01
test: fix test failures caused by sub tests running in parallel

nan-yu:
`workload_identity_test.go` defines a set of sub tests in a table driven format. Each sub test enables parallel execution with `t.Parallel()`. A go routine is created for each sub test. However, when it launches the go routines, the test case variables (the `tc` variable in the for loop iterator) are evaluated when the go routines run. Hence, by the time the go routine is actually executed, the for loop has completed, and each test ends up using the variables for the last test case.

To fix this issue, this commit rebinds the `tc` variable inside the loop. This creates a copy for the variable and go routine uses the local copy for execution.

It also adds the hubv1 schema because the test gets the Membership object to validate the Fleet workload identity credentials. loop


sdowell (approved):
/lgtm

PR #1121, opened 2024-02-01
feature: add BYOID support for OCI artifacts and Helm charts

nan-yu:
This commit integrates BYOID support into Config Sync by supporting a new auth type for OCI and Helm: `k8sserviceaccount`. It requires GKE or Fleet Workload Identity to be enabled. The KSA along with WI will be exchanged to a federated token which can be used to authenticate to Ubermint-enabled GCP services directly.


pkg/validate/raw/validate/source_spec_validator.go
```diff
@@ -142,6 +142,10 @@ func OciSpec(oci *v1beta1.Oci, rs client.Object) status.Error {
 	// will fail to apply.
 	switch oci.Auth {
 	case configsync.AuthGCENode, configsync.AuthNone:
+	case configsync.AuthK8sServiceAccount:
+		if oci.GCPServiceAccountEmail != "" {
+			return IllegalGCPSAEmail(v1beta1.OciSource, rs)
```
sdowell, line 147: It seems like this should return IllegalGCPSAEmail for _any_ auth type that is _not_ GCPServiceAccountEmail, but this only returns for AuthK8sServiceAccount. Is that intentional?

nan-yu: The presence of GCPServiceAccountEmail will interfere with the  AuthK8sServiceAccount auth type, but not other auth types.
The check is a MUST for  AuthK8sServiceAccount, but I agree it is more consistent if other auth types have the same check.

nan-yu: It turns out a bigger change, and I extracted into a separate PR: #1122.

e2e/testcases/workload_identity_test.go
```diff
@@ -287,3 +294,52 @@ func truncateStringByLength(s string, l int) string {
...
+// migrateFromGSAtoKSA tests the scenario of migrating from impersonating a GSA
+// to leveraging KSA+WI (a.k.a, BYOID/Ubermint).
+func migrateFromGSAtoKSA(nt *nomostest.NT, rs *v1beta1.RootSync, ksaRef types.NamespacedName, fleetWITest bool) {
+	nt.T.Log("Update RootSync auth type from gcpserviceaccount to k8sserviceaccount")
+	oldGeneration := rs.Generation
+	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"%s": {"auth": "k8sserviceaccount", "gcpServiceAccountEmail": ""}}}`, rs.Spec.SourceType))
+
+	// Validations
+	nt.T.Log("Validate the GSA annotation is removed from the RootSync's service account")
+	nomostest.Wait(nt.T, "wait for GSA annotation to be removed from KSA", nt.DefaultWaitTimeout, func() error {
```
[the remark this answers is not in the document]
nan-yu: done

nan-yu:
There is no dependency for this PR. It is ready for another look.

e2e/testcases/workload_identity_test.go
```diff
@@ -287,3 +293,52 @@ func truncateStringByLength(s string, l int) string {
...
+	nt.T.Log("Validate RootSync generation and observedGeneration is updated")
+	require.NoError(nt.T,
+		nt.Watcher.WatchObject(kinds.RootSyncV1Beta1(), rs.Name, rs.Namespace,
+			[]testpredicates.Predicate{
+				testpredicates.GenerationEquals(oldGeneration + 1),
+				testpredicates.HasObservedLatestGeneration(nt.Scheme),
+			}))
+}
+
+func validateAllResourcesReconciled(nt *nomostest.NT, sourceType v1beta1.SourceType, sourceChart, syncDir string, rootCommitFn nomostest.Sha1Func) {
```
sdowell, line 326: nit: return error from these functions and invoke fatal from caller

nan-yu: Some errors are marked as error (failed, but the test can continue execution), such as validating if a managed resource is reconciled.
Some are marked as fatal and exit immediately, such as CS sync failures.

Handling them inside the function makes it possible to distinguish between the two scenarios. If no strong objections, I would rather keep as is.

sdowell: I tend to prefer failing fast on all such validation errors, in my experience it makes debugging faster/simpler. But no strong objection if you prefer to keep it this way

nan-yu: Fair enough. `e2e/nomostest/kustomizecomponents/validate.go` used `nt.T.Error` because it validated each reconciled resource separately, and it was easy to collect all reconcile failures in the past, but we don't see those failures anymore. I also changed them for consistency.

e2e/testcases/workload_identity_test.go
```diff
@@ -255,27 +261,33 @@ func TestWorkloadIdentity(t *testing.T) {
...
+				strings.HasPrefix(tc.sourceRepo, "oci://"+nomostesting.ARHost) {
+				if err := migrateFromGSAtoKSA(nt, rs, ksaRef, tc.fleetWITest); err != nil {
+					nt.T.Fatal()
 				}
-			} else {
-				err := nt.WatchForAllSyncs(nomostest.WithRootSha1Func(tc.rootCommitFn),
-					nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: tenant}))
-				if err != nil {
-					nt.T.Fatal(err)
+				if err := validateAllResourcesReconciled(nt, tc.sourceType, tc.sourceChart, tenant, tc.rootCommitFn); err != nil {
```
sdowell, line 288: This means that reconciler-manager has reconciled the reconciler Deployment, but do we know that the `*-sync` container was able to sync with k8sserviceaccount credentials? It seems like this may just pass immediately since the package was already synced using gcpserviceaccount. Perhaps we should push a new version and/or update the RSync spec to sync a new version?

e2e/testcases/workload_identity_test.go
```diff
@@ -255,27 +261,33 @@ func TestWorkloadIdentity(t *testing.T) {
...
-			if tc.sourceType == v1beta1.HelmSource {
-				err := nt.WatchForAllSyncs(nomostest.WithRootSha1Func(tc.rootCommitFn),
-					nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: tc.sourceChart}))
-				if err != nil {
-					nt.T.Fatal(err)
-				}
-				if err := nt.Validate(fmt.Sprintf("my-coredns-%s", tc.sourceChart), "coredns", &appsv1.Deployment{}); err != nil {
-					nt.T.Error(err)
+			if err := validateAllResourcesReconciled(nt, tc.sourceType, tc.sourceChart, tenant, tc.rootCommitFn); err != nil {
+				nt.T.Fatal()
```
sdowell, line 279: pass in `err` to `Fatal`

e2e/testcases/workload_identity_test.go
nan-yu: Inside the `migrateFromGSAtoKSA` function, I explicitly checked the fwi credentials in the reconciler Pod, and the RSync's generation/observedGeneration. The check here confirms the RSync status is synced.

e2e/testcases/workload_identity_test.go
nan-yu: oops, done

e2e/testcases/workload_identity_test.go
sdowell: I don't think that means the new reconciler Pod has successfully synced though, does it? I think that just means that the `reconciler-manager` updated the `reconciler` Pod to match the latest `RSync` spec. I don't think we guarantee here that the updated `reconciler` Pod can successfully sync.

nan-yu: When `reconciler-manager` updates the Deployment, the Deployment status will change from `ready` to `inprogress`, which will then cause RSync.status.condition[Reconciling] to be changed to true. If the `reconciler` Pod can't successfully sync, `WatchForAllSyncs` should fail.

nan-yu: I see the confusion with the internal status change. I can make it sync a new image/helm chart version to avoid the confusion.

sdowell: To be clear I'm not just talking about the Deployment reconciling here, I'm talking about the new `reconciler` container successfully syncing from the source. From my reading of the code there is still a race condition where the Deployment is current/ready but hasn't done a sync loop yet. Since the `lastSyncedCommit` from the previous Pod (using `gcpserviceaccount`) still matches the `rootCommitFn`, I don't think `WatchForAllSyncs` is going to wait for the new sync container to run.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -125,11 +125,13 @@ func (r *reconcilerBase) upsertServiceAccount(
 		if len(refs) > 0 {
 			childSA.OwnerReferences = refs
 		}
-		// Update annotation when Workload Identity is enabled on a GKE cluster.
+		// Set annotation when impersonating a GSA on a Workload Identity enabled cluster.
 		// In case, Workload Identity is not enabled on a cluster and spec.git.auth: gcpserviceaccount,
 		// the added annotation will be a no-op.
```
janetkuo, line 130: Would you update this comment as well given that the added annotation won't be no-op in some cases and needs to be removed?

e2e/testcases/workload_identity_test.go
```diff
@@ -277,6 +294,14 @@ func TestWorkloadIdentity(t *testing.T) {
 				}
 				kustomizecomponents.ValidateAllTenants(nt, string(declared.RootReconciler), "../base", tenant)
 			}
+
+			// Migrate from gcpserviceaccount to k8sserviceaccount
+			if strings.HasPrefix(tc.sourceRepo, nomostesting.ARHost) ||
+				strings.HasPrefix(tc.sourceRepo, "oci://"+nomostesting.ARHost) {
```
janetkuo, line 300: Would it be more robust if we check whether `rs` auth is gcpserviceaccount?

e2e/testcases/workload_identity_test.go
nan-yu: Yes, I was also talking about the `reconciling` condition in the RootSync, which indicates the `reconciler` Pod status.

`WatchForAllSyncs` does check the reconciling condition (`reconciler` Pod/container status): https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/wait_for_sync.go#L244

I think I see where the potential race condition is with the following flow:
1. RSync auth type is changed
2. `reconciler-manager` updates the `reconciler` Deployment
3. `reconciler` Deployment spec is updated, but the status changes from `Ready` to `InProgress`, which is not current, and returns an error: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/rootsync_controller.go#L255
4. `reconciler-manager` updates the RSync status by setting the `Reconciling` condition: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/rootsync_controller.go#L353 and then updates the RSync observedGeneration: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/rootsync_controller.go#L1109

At this time, `WatchForAllSyncs` will return an error if the `reconciler` Pod is not ready. 
The race condition happens when the `reconciler` Pod becomes ready but not finishes the syncing yet.

nan-yu: Updated the spec to validate new resources can be reconciled with k8sserviceaccount.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: done

e2e/testcases/workload_identity_test.go
nan-yu: I don't think that is necessary. The previous step sets the auth type to `gcpserviceaccount` for all test cases. If we add the check, then it is always true in the current implementation. Plus, we also validate the GSA annotation is added to the KSA.

janetkuo: What's this repo URL check for? I originally thought it's for making sure we only run `migrateFromGSAtoKSA` when we're already using GSA as auth.

nan-yu: It guarantees that we only migrate from GSA to KSA for OCI/Helm with Artifact Registry. The test also includes Git source with CSR, which will be excluded with the check.

janetkuo: If so, we should check the source type to be OCI & Helm, and checking auth is GSA. Checking repo url isn't that robust given that it's subject to change and we could skip this test without knowing.

janetkuo: Alternatively we could just immediately return in `migrateFromGSAtoKSA` when source type is Git.

nan-yu: Forgot to mention another caveat. The test table also includes OCI images with GCR (although they can be removed when GCR is official deprecated after May 15, 2024). Besides the source type, we also need to check the host.

Helm charts can also be hosted elsewhere other than AR. It would be safe/easy to check if the source repo's host is AR.

sdowell: Perhaps it would be best to just be explicit on the testcase spec with a `testK8sServiceAccount` type of field

nan-yu: That sounds good.

nan-yu: done

janetkuo (approved):
/lgtm

sdowell (approved):
/lgtm

PR #1122, opened 2024-02-01
refactor the source spec validation

nan-yu:
This commit covers more failure scenarios with the redudant fields validaiton.

### DIFF
--- a/pkg/api/configsync/register.go
+++ b/pkg/api/configsync/register.go
@@ -14,7 +14,12 @@
 
 package configsync
 
-import "time"
+import (
+	"strings"
+	"time"
+
+	"golang.org/x/exp/slices"
+)
 
 const (
 	// GroupName is the name of the group of configsync resources.
@@ -89,6 +94,20 @@ const (
 	DefaultHelmReleaseNamespace = "default"
 )
 
+// SourceType specifies the type of the source of truth.
+type SourceType string
+
+const (
+	// GitSource represents the source type is Git repository.
+	GitSource SourceType = "git"
+
+	// OciSource represents the source type is OCI package.
+	OciSource SourceType = "oci"
+
+	// HelmSource represents the source type is Helm repository.
+	HelmSource SourceType = "helm"
+)
+
 // AuthType specifies the type to authenticate to a repository.
 type AuthType string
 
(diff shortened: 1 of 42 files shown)


pkg/api/configsync/register.go
```diff
@@ -120,3 +139,36 @@ const (
 	// declared to be created by the reconciler.
 	NamespaceStrategyExplicit NamespaceStrategy = "explicit"
 )
+
+var supportedAuthTypes = map[SourceType][]string{
```
sdowell, line 143: Minor nit, but perhaps this could be written as a function to guarantee immutability

e2e/testcases/reconciler_manager_test.go
```diff
@@ -591,7 +591,7 @@ func TestManagingReconciler(t *testing.T) {
 
 	// test case 7: the reconciler-manager should mount the git-creds volumes again if the auth type requires a git secret
 	nt.T.Log("Switch the auth type gcpserviceaccount to ssh")
-	nt.MustMergePatch(rs, `{"spec":{"git":{"auth":"ssh","secretRef":{"name":"git-creds"}}}}`)
```
sdowell, line 594: I agree with this stricter validation, but just wanted to point out that this could be temporarily disruptive for some users if they have gcpServiceAccountEmail set for other auth types

pkg/validate/raw/validate/source_spec_validator.go
```diff
@@ -295,17 +296,27 @@ func MissingSecretRef(sourceType v1beta1.SourceType, o client.Object) status.Err
...
 			kind, sourceType, configsync.AuthGCPServiceAccount, gcpSASuffix).
 		BuildWithResources(o)
 }
 
+// IllegalGCPSAEmail reports that a RootSync/RepoSync declares an auth type that
+// doesn't allow gcpServiceAccountEmail to be declared.
+func IllegalGCPSAEmail(sourceType configsync.SourceType, o client.Object) status.Error {
+	kind := o.GetObjectKind().GroupVersionKind().Kind
+	return invalidSyncBuilder.
+		Sprintf("%ss that spec.%s.auth is not %q must not specify spec.%s.gcpServiceAccountEmail",
```
sdowell, line 312: I think this may be clearer without a double negative. For example, `%ss must only specify spec.%s.gcpServiceAccountEmail when using the %s auth type`

e2e/testcases/reconciler_manager_test.go
nan-yu: I was also thinking about it. It sounds like we don't have a unified rule to validate the spec.
Currently, the secret ref is allowed **only** when it is required by certain auth types, but gcpServiceAccountEmail doesn't have such requirement.
On the other hand, spec.git/spec.oci/spec.helm can co-exist even it is not the corresponding source type.

For secret ref, it makes sense to be restrictive because the presence will trigger a copy of the Secret in the c-m-s Namespace. Due to the restriction, this PR makes a consistent behavior between secret ref and gcpServiceAccountEmail. From that perspective, I feel it is okay to ask users to update their configs to follow the same rule.

pkg/validate/raw/validate/source_spec_validator.go
nan-yu: SG. done

pkg/api/configsync/register.go
nan-yu: SG. done

pkg/api/configsync/register.go
```diff
@@ -120,3 +139,43 @@ const (
...
+			string(AuthCookieFile),
+			string(AuthGCENode),
+			string(AuthToken),
+			string(AuthNone),
+			string(AuthGCPServiceAccount),
+		}
+	case OciSource:
+		return []string{
+			string(AuthGCPServiceAccount),
+			string(AuthToken),
```
janetkuo, line 158: We don't support token for OCI

nan-yu: Oh, yes. Removed.

e2e/testcases/reconciler_manager_test.go
janetkuo: Making validation stricter is considered breaking. Agreed that this is inconsistent, but we should try our best to avoid breaking users. Unless allowing gcpServiceAccountEmail has huge downside, we should just keep allowing it.

nan-yu: What about also loosening the restriction on the secretRef validation? Basically, any field that is not needed will be ignored.
That won't break any users, but it makes all fields be validated in a consistent way.

janetkuo: Loosening it has the downside of causing the Secret to be copied over to c-m-s as you mentioned, but I think that might be fine if we prefer keeping the rule consistent.

nan-yu: We can copy it over only when it is used by certain auth types.

sdowell: Assuming this is going to be cherry picked, I don't think we need to alter the behavior with this change. I think it would be preferable to minimize the changes for the cherry pick

If desired we can change the behavior for the next minor release

nan-yu: Yes, I was going to say it doesn't need to be cherrypicked if we don't add more validations.

This can wait until the next minor release.

PR #1123, opened 2024-02-03
test: add test coverage for missing permissions

nan-yu:
There are multiple issues with missing permission, though they all share the same KNV2004 error code.
This commit adds test coverage for the following scenarios:
1. A GSA doesn't exist when using gcpserviceaccount
2. A GSA exists, but it is not impersonated with a KSA.
3. A GSA exists and is impersonated, but it doesn't have reader permission.
4. A WI+KSA doesn't have reader permission.

### DIFF
--- a/e2e/testcases/permission_test.go
+++ b/e2e/testcases/permission_test.go
@@ -1,0 +1,387 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e"
+	"kpt.dev/configsync/e2e/nomostest"
+	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/e2e/nomostest/testpredicates"
+	"kpt.dev/configsync/e2e/nomostest/testutils"
+	"kpt.dev/configsync/e2e/nomostest/workloadidentity"
+	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/importer/filesystem"
+	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
+	"kpt.dev/configsync/pkg/status"
+	"kpt.dev/configsync/pkg/testing/fake"
+)
+
+// gsaRestrictedEmail returns a GSA email that has no permissions and no IAM bindings as a WI user.
+func gsaRestrictedEmail() string {
+	return fmt.Sprintf("e2e-test-restricted-user@%s.iam.gserviceaccount.com", *e2e.GCPProject)
+}
+
+// gsaRestrictedImpersonatedEmail returns a GSA email that has no permissions, but has an IAM binding as a WI user.
+func gsaRestrictedImpersonatedEmail() string {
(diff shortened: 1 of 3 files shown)


e2e/testcases/permission_test.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			nt := nomostest.New(t, nomostesting.WorkloadIdentity,
+				ntopts.Unstructured,
+				ntopts.RequireGKE(t))
+
+			if err := workloadidentity.ValidateEnabled(nt); err != nil {
```
karlkfi, line 320: Do we need WI enabled for KSA auth to work?

e2e/testcases/permission_test.go
```diff
@@ -0,0 +1,387 @@
...
+				}
+			}
+			nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/rs-byoid.yaml", rs))
+			nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Add a RootSync with restricted k8sserviceaccount"))
+
+			ksaRef := types.NamespacedName{
+				Namespace: configsync.ControllerNamespace,
+				Name:      core.RootReconcilerName(rs.Name),
+			}
+			require.NoError(nt.T,
```
karlkfi, line 368: most of the e2e tests use `nt.Must()` instead of require.

e2e/testcases/workload_identity_test.go
```diff
@@ -417,3 +354,102 @@ func migrateFromGSAtoKSA(nt *nomostest.NT, rs *v1beta1.RootSync, ksaRef types.Na
...
+		sourceSpec.sourceVersion = chart.Version
+		sourceSpec.rootCommitFn = nomostest.HelmChartVersionShaFn(chart.Version)
+	}
+	return nil
+}
+
+// mustConfigureRootSync updates RootSync to sync with the provided auth.
+// It reuses the RootSync instead of creating a new one so that test resources
+// can be cleaned up after the test.
+func mustConfigureRootSync(nt *nomostest.NT, rs *v1beta1.RootSync, tenant string, sourceSpec *sourceSpec) {
```
karlkfi, line 437: mustPatchRootSync maybe?

e2e/testcases/permission_test.go
```diff
@@ -0,0 +1,387 @@
...
+
+			spec := &sourceSpec{
+				sourceType:    tc.sourceType,
+				sourceRepo:    tc.sourceRepo,
+				sourceChart:   tc.sourceChart,
+				sourceVersion: tc.sourceVersion,
+				gsaEmail:      tc.gsaEmail,
+			}
+			mustConfigureRootSync(nt, rs, tenant, spec)
+			nt.WaitForRootSyncSourceError(rs.Name, status.SourceErrorCode, tc.expectedErrMsg)
```
karlkfi, line 255: All these auth errors makes me wonder if we shouldnt add a new error code. I would normally expect the problem with a source error to be in the source, but these are all auth problems fetching the source.

e2e/testcases/permission_test.go
```diff
@@ -0,0 +1,387 @@
...
+	// The exact error message returned for Fleet WI is:
+	// credential refresh failed: auth URL returned status 500, body: \"error retrieveing TokenSource.Token: oauth2/google: status code 403: {\\n  \\\"error\\\": {\\n    \\\"code\\\": 403,\\n    \\\"message\\\": \\\"Permission 'iam.serviceAccounts.getAccessToken' denied on resource (or it may not exist).\\\",\\n    \\\"status\\\": \\\"PERMISSION_DENIED\\\",\\n    \\\"details\\\": [\\n      {\\n        \\\"@type\\\": \\\"type.googleapis.com/google.rpc.ErrorInfo\\\",\\n        \\\"reason\\\": \\\"IAM_PERMISSION_DENIED\\\",\\n        \\\"domain\\\": \\\"iam.googleapis.com\\\",\\n        \\\"metadata\\\": {\\n          \\\"permission\\\": \\\"iam.serviceAccounts.getAccessToken\\\"\\n        }\\n      }\\n    ]\\n  }\\n}\\n\\n\"
```
karlkfi, line 56: This error is really messy with all the multi-line json escaping. Is there any way we can clean up the error a bit for the user to make it more readable?

karlkfi: - `error retrieving TokenSource.Token` is from our askpass sidecar code.
- `oauth2/google: status code %d: %s` is from one of 3 places in the oauth2/google internal library.

I wonder if we could add some code to the askpass sidecar to detect and parse a JSON payload in the error string and do some formatting to pretty-print it so it's more readable like the GKE WI error.

karlkfi: Or even better, maybe we can modify the oauth2 library to parse it for us...

e2e/testcases/permission_test.go
karlkfi: This is just in our test setup though, right? 

Does the cluster actually need WI enabled in order to use BYOID? 

Any k8s cluster can inject a KSA into a Pod. 

Do we need to test BYOID on non-WI clusters?

karlkfi: Nevermind. I misread your response.

nan-yu:
I accidentally pushed from the wrong fork branch. I'll push a new patch after addressing these comments.

PR #1127, opened 2024-02-05
Add required IAM settings for the test infra

nan-yu:
The new k8sserviceaccount requires reader permissions to be granted to the KSA, along with the workload identity pool. This commit adds IAM settings as a one-off config.

### DIFF
--- a/e2e/testinfra/terraform/common/service_accounts.tf
+++ b/e2e/testinfra/terraform/common/service_accounts.tf
@@ -78,3 +78,15 @@ resource "google_service_account_iam_member" "e2e_metric_writer_gke_binding" {
   role               = "roles/iam.workloadIdentityUser"
   member             = "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-monitoring/default]"
 }
+
+# Grant source reader permissions to the RootSync's KSA with GKE workload identity.
+resource "google_project_iam_member" "root-reconciler-wi-sa-iam" {
+  for_each = toset([
+    "roles/source.reader",
+    "roles/artifactregistry.reader",
+    "roles/storage.objectViewer",
+  ])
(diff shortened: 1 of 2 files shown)


e2e/testinfra/terraform/common/service_accounts.tf
```diff
@@ -78,3 +78,25 @@ resource "google_service_account_iam_member" "e2e_metric_writer_gke_binding" {
...
+}
+
+resource "google_project_iam_member" "root-reconciler-fwi-sa-iam" {
+  for_each = toset([
+    "roles/source.reader",
+    "roles/artifactregistry.reader",
+    "roles/storage.objectViewer",
+  ])
+  role    = each.value
+  member             = "serviceAccount:cs-dev-hub.svc.id.goog[config-management-system/root-reconciler]"
```
sdowell, line 100: The `cs-dev-hub` IAM should probably be specific to the prow environment rather than in `common`. I don't think we can assume any dev environment will be able to set up IAM with this hard coded project.

nan-yu: That means dev environments won't be able to run the e2e tests with Fleet WI.

sdowell: Yes, and that's the case already today. The dev environment is intended to be provisionable/usable by anyone, including an external contributor.

If we want to make fleet WI identity testable in a more generic sense, we would need to parametrize the fleet project in a number of places instead of hard coding `cs-dev-hub`. I think that is a bigger change that is out of scope here.

nan-yu: Make sense and moved to the `prow` directory.

sdowell (approved):
/lgtm

PR #1130, opened 2024-02-06
fix: use google-cloud-cli package instead of sdk

sdowell:
The package name has transitioned to cli instead of sdk, and the sdk package is currently broken.


tiffanny29631 (approved):
/lgtm

PR #1131, opened 2024-02-06
fix: use google-cloud-cli package instead of sdk (#1130)

tiffanny29631:
The package name has transitioned to cli instead of sdk, and the sdk package is currently broken.


nan-yu (approved):
/lgtm

PR #1135, opened 2024-02-08
test: fix test failures with invalid image digest

nan-yu:
`workload_identity_test.go` should use the `imageDigestFuncByName` function to fetch the image digest.


sdowell (approved):
/lgtm

PR #1136, opened 2024-02-08
fix: use more limited securityContext for pods

sdowell:
This restricts some of the default security context fields to follow principle of least privilege.


karlkfi (approved):
/lgtm

PR #1137, opened 2024-02-08
test: move the fake test finalizer removal function into a cleanup b lock (#1106)

nan-yu:
The fake finalizer should be removed even if the test failed in the middle. Otherwise, the namespace would be stuck in a terminating state. Example: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-regular-bitbucket/1747668032626364416


nan-yu:
Similar issues occurred on the release CI. 
Example:
- https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-rapid-release/1755686994614161408

sdowell (approved):
/lgtm

PR #1138, opened 2024-02-13
test: refactor helm provider to not depend on gitproviders

nan-yu:
Currently, it calls for a `UseHelmChart` function before building and pushing a Helm chart. The function resides inside the gitproviders package, which copies the test helm chart directory to the git local directory. Ideally, if the source type is changed from Git to Helm, it is not necessary to update the git repository.

This commit refactors the code to separate the helm repo dir from the git repo dir. When building and pushing a helm chart, it is optional to provide a test helm chart dir, any K8s objects, or a chart version.

It also makes similar changes to the oci provider.

### DIFF
--- a/e2e/nomostest/registryproviders/helm.go
+++ b/e2e/nomostest/registryproviders/helm.go
@@ -21,8 +21,11 @@ import (
 	"strconv"
 	"strings"
 
-	"kpt.dev/configsync/e2e/nomostest/gitproviders"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
 	"kpt.dev/configsync/e2e/nomostest/testshell"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/kustomize/kyaml/copyutil"
 	"sigs.k8s.io/yaml"
 )
@@ -31,7 +34,10 @@ import (
 var helmIndex int
 
 type helmOptions struct {
-	version string
+	version      string
+	sourceChart  string
+	chartObjects []client.Object
+	scheme       *runtime.Scheme
 }
 
 // HelmOption is an optional parameter when building a helm chart.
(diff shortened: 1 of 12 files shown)


e2e/nomostest/registryproviders/helm.go
```diff
@@ -44,37 +50,69 @@ func HelmChartVersion(version string) func(options *helmOptions) {
 	}
 }
 
+// HelmChartDir builds the chart with the specified chart input directory.
+func HelmChartDir(dir string) func(options *helmOptions) {
```
sdowell, line 54: I think I'd prefer this interface just take the chart name rather than an arbitrary dir path. If this function assumes that all charts are in the `testdata/helm-charts` directory I think that will nudge future authors to keep things consistent.

e2e/testcases/workload_identity_test.go
```diff
@@ -261,6 +270,16 @@ func TestWorkloadIdentity(t *testing.T) {
 				tc.sourceChart = chart.Name
 				tc.sourceVersion = chart.Version
 				tc.rootCommitFn = nomostest.HelmChartVersionShaFn(chart.Version)
+			} else if tc.sourceType == v1beta1.OciSource && tc.requireOCIArtifactRegistry {
+				image, err := nt.BuildAndPushOCIImage(nomostest.RootSyncNN(rs.Name),
+					registryproviders.ImageInputDir("../testdata/hydration/kustomize-components"),
```
sdowell, line 275: This should use `tc.sourceRepo`

e2e/testcases/workload_identity_test.go
```diff
@@ -261,6 +270,16 @@ func TestWorkloadIdentity(t *testing.T) {
...
+					registryproviders.ImageVersion(tc.sourceVersion))
+				if err != nil {
+					nt.T.Fatalf("failed to push oci image: %v", err)
+				}
+
+				tc.sourceRepo = image.FloatingTag()
```
sdowell, line 281: Why is this variable value changing from the local directory path to an image tag? I think we should have two separate variables to keep this clear

e2e/testcases/workload_identity_test.go
```diff
@@ -261,6 +270,16 @@ func TestWorkloadIdentity(t *testing.T) {
 				tc.sourceChart = chart.Name
 				tc.sourceVersion = chart.Version
 				tc.rootCommitFn = nomostest.HelmChartVersionShaFn(chart.Version)
+			} else if tc.sourceType == v1beta1.OciSource && tc.requireOCIArtifactRegistry {
```
sdowell, line 273: I don't think the `requireOCIArtifactRegistry` check is needed here, this variable should only really need to be used by the test scaffolding to skip the test

nan-yu: The intent is to reuse the flag. Agreed it seems to be hacky.
Alternatively, we'll need another flag for building and pushing the OCI image. The sourceType isn't enough because the OCI image can be hosted in GCR.

sdowell: If that's the case, we might consider adding a `GCR` provider type to `registryproviders`. That's a bigger change though so I think this is fine for now, albeit a bit hacky. Can you leave a comment with some context?

nan-yu: Sure. We can remove the check along with GCR tests once GCR is end of life.

e2e/testcases/workload_identity_test.go
nan-yu: Added another variable for the sourcePackage

e2e/nomostest/registryproviders/helm.go
nan-yu: Make sense. Added a comment for the parent directory of the source chart and source package.

e2e/testcases/workload_identity_test.go
nan-yu: missed it. done

e2e/testcases/workload_identity_test.go
nan-yu: comment added

sdowell (approved):
/lgtm

PR #1139, opened 2024-02-21
test: prevent eviction of test pods

sdowell:
Autopilot now allows setting the safe-to-evict annotation to request an extended runtime of Pods. This change sets the annotation on each of the test components which use ephemeral storage, which should reduce flakiness. These Deployments are deleted at the end of test execution, which will allow scale down.

### DIFF
--- a/e2e/nomostest/git-server.go
+++ b/e2e/nomostest/git-server.go
@@ -127,9 +127,6 @@ func gitDeployment() *appsv1.Deployment {
 		core.Namespace(testGitNamespace),
 		core.Labels(testGitServerSelector()),
 	)
-	if *e2e.GKEAutopilot {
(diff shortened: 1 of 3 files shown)


e2e/nomostest/git-server.go
```diff
@@ -127,9 +127,6 @@ func gitDeployment() *appsv1.Deployment {
 		core.Namespace(testGitNamespace),
 		core.Labels(testGitServerSelector()),
 	)
-	if *e2e.GKEAutopilot {
-		deployment.SetAnnotations(map[string]string{safeToEvictAnnotation: "false"})
```
sdowell, line 131: Note: this was setting the annotation on the deployment rather than the pod template

tiffanny29631 (approved):
/lgtm

PR #1140, opened 2024-02-22
fix: stop validating root secrets

nan-yu:
Config Sync doesn't generate a copy of root Secret, so no need to validate the generated secret name length.

b/326430718


sdowell (approved):
/lgtm

PR #1142, opened 2024-02-22
chore: removing the old errors package.

mikebz:
codebase leaned on a package that is no longer maintained.  This is a cleanup task.

Instead of a massive replace there was a thought to create our own proxy for error reporting, but I think this would be a tax we'd have to maintain for a long while so it's probably better to just do the replacement right now.

### DIFF
--- a/cmd/nomos/parse/find_files.go
+++ b/cmd/nomos/parse/find_files.go
@@ -15,10 +15,10 @@
 package parse
 
 import (
+	"fmt"
 	"os/exec"
 	"strings"
 
-	"github.com/pkg/errors"
 	"kpt.dev/configsync/pkg/importer/filesystem/cmpath"
 )
 
@@ -32,7 +32,7 @@ import (
 func FindFiles(dir cmpath.Absolute) ([]cmpath.Absolute, error) {
 	out, err := exec.Command("find", dir.OSPath(), "-type", "f", "-not", "-path", "*/\\.git/*").CombinedOutput()
 	if err != nil {
-		return nil, errors.Wrap(err, string(out))
+		return nil, fmt.Errorf("%s: %w", string(out), err)
 	}
 	files := strings.Split(string(out), "\n")
 	var result []cmpath.Absolute
--- a/cmd/nomos/status/status.go
+++ b/cmd/nomos/status/status.go
@@ -16,6 +16,7 @@ package status
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -25,7 +26,6 @@ import (
 	"text/tabwriter"
 	"time"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
 	"kpt.dev/configsync/cmd/nomos/flags"
@@ -77,12 +77,12 @@ func SaveToTempFile(ctx context.Context, contexts []string) (*os.File, error) {
 	printStatus(ctx, writer, clientMap, names)
 	err = tmpFile.Close()
 	if err != nil {
-		return tmpFile, errors.Wrap(err, "failed to close status file writer with error")
+		return tmpFile, fmt.Errorf("failed to close status file writer with error: %w", err)
 	}
 
 	f, err := os.Open(tmpFile.Name())
 	if err != nil {
-		return tmpFile, errors.Wrap(err, "failed to open the file for reading")
+		return tmpFile, fmt.Errorf("failed to open the file for reading: %w", err)
 	}
 
 	return f, nil
@@ -102,9 +102,8 @@ var Cmd = &cobra.Command{
 
 		clientMap, err := ClusterClients(cmd.Context(), flags.Contexts)
 		if err != nil {
-			// If "no such file or directory" error, unwrap and display before exiting
-			if unWrapped := errors.Cause(err); os.IsNotExist(unWrapped) {
-				return errors.Wrapf(err, "failed to create client configs")
+			if errors.Is(err, os.ErrNotExist) {
+				return fmt.Errorf("failed to create client configs: %w", err)
 			}
 
 			klog.Fatalf("Failed to get clients: %v", err)
@@ -171,7 +170,7 @@ func printStatus(ctx context.Context, writer *tabwriter.Writer, clientMap map[st
 
 	currentContext, err := restconfig.CurrentContextName()
 	if err != nil {
-		fmt.Printf("Failed to get current context name with err: %v\n", errors.Cause(err))
+		fmt.Printf("Failed to get current context name with err: %v\n", err)
 	}
 
 	// Now we write everything at once. Processing and then printing helps avoid screen strobe.
--- a/cmd/nomos/version/version.go
+++ b/cmd/nomos/version/version.go
@@ -284,19 +284,15 @@ func tabulate(es []entry, out io.Writer) {
 	w := util.NewWriter(out)
 	defer func() {
 		if err := w.Flush(); err != nil {
-			// nolint:errcheck
 			fmt.Fprintf(os.Stderr, "error on Flush(): %v", err)
 		}
 	}()
-	// nolint:errcheck
 	fmt.Fprintf(w, format, "CURRENT", "CLUSTER_CONTEXT_NAME", "COMPONENT", "VERSION")
 	for _, e := range es {
 		if e.err != nil {
-			// nolint:errcheck
 			fmt.Fprintf(w, format, e.current, e.name, e.component, fmt.Sprintf("<error: %v>", e.err))
 			continue
 		}
-		// nolint:errcheck
 		fmt.Fprintf(w, format, e.current, e.name, e.component, e.version)
 	}
 }
--- a/cmd/nomos/version/version.go
+++ b/cmd/nomos/version/version.go
@@ -16,6 +16,7 @@ package version
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -23,7 +24,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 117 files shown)


cmd/nomos/parse/find_files.go
```diff
@@ -32,7 +32,7 @@ import (
 func FindFiles(dir cmpath.Absolute) ([]cmpath.Absolute, error) {
 	out, err := exec.Command("find", dir.OSPath(), "-type", "f", "-not", "-path", "*/\\.git/*").CombinedOutput()
 	if err != nil {
-		return nil, errors.Wrap(err, string(out))
+		return nil, fmt.Errorf("%s: %w", out, err)
```
karlkfi, line 35: Still need the string cast for byte array. The %s will see it as an array and try to stringify the array instead of casting. This is a pretty common gotcha and might be elsewhere too.

mikebz: Good catch.  Fixed some of them and will need to think where else my regex replaced them.

karlkfi: Run git diff and grep for "string("

cmd/nomos/status/status.go
```diff
@@ -171,7 +170,7 @@ func printStatus(ctx context.Context, writer *tabwriter.Writer, clientMap map[st
 
 	currentContext, err := restconfig.CurrentContextName()
 	if err != nil {
-		fmt.Printf("Failed to get current context name with err: %v\n", errors.Cause(err))
+		fmt.Printf("Failed to get current context name with err: %v\n", err)
```
karlkfi, line 173: cause or full error?

e2e/nomostest/git-server.go
```diff
@@ -138,6 +135,9 @@ func gitDeployment() *appsv1.Deployment {
 		Template: corev1.PodTemplateSpec{
 			ObjectMeta: v1.ObjectMeta{
 				Labels: testGitServerSelector(),
+				Annotations: map[string]string{
```
karlkfi, line 138: why was this moved?

mikebz: agreed that it's weird that it's showing up on the diff.  This is the incoming change from main.

cmd/nomos/parse/find_files.go
mikebz: I think we are good.  Here is the git diff output:
```
λ git diff main mb_err_pkg | grep "string("
-               return nil, errors.Wrap(err, string(out))
+               return nil, fmt.Errorf("%s: %w", string(out), err)
        files := strings.Split(string(out), "\n")
-                               t.Fatal(errors.Wrap(err, string(out)))
+                               t.Fatal(fmt.Errorf("%s: %w", string(out), err))
-                                               t.Fatal(errors.Wrap(err, string(out)))
+                                               t.Fatal(fmt.Errorf("%s: %w", string(out), err))
-                                       t.Fatal(errors.Wrap(err, string(out)))
+                                       t.Fatal(fmt.Errorf("%s: %w", string(out), err))
                        name, err, string(out))
        operations := strings.Fields(string(out))
                        operation, err, string(out))
-                       return errors.Errorf("failed to delete cluster %s: %v\nstdout/stderr:\n%s", name, err, string(out))
+                       return fmt.Errorf("failed to delete cluster %s: %v\nstdout/stderr:\n%s", name, err, string(out))
-               return errors.Errorf("failed to create cluster %s: %v\nstdout/stderr:\n%s", name, err, string(out))
+               return fmt.Errorf("failed to create cluster %s: %v\nstdout/stderr:\n%s", name, err, string(out))
-               return errors.Errorf("failed to get credentials: %v\nstdout/stderr:\n%s", err, string(out))
+               return fmt.Errorf("failed to get credentials: %v\nstdout/stderr:\n%s", err, string(out))
-               return errors.Errorf("failed to refresh access_token: %v\nstdout/stderr:\n%s", err, string(out))
+               return fmt.Errorf("failed to refresh access_token: %v\nstdout/stderr:\n%s", err, string(out))
                        clusterName, err, string(out))
        clusters := strings.Fields(string(out))
        match := re.FindStringSubmatch(string(bytes))
-               return "", errors.Wrap(err, string(out))
+               return "", fmt.Errorf("%s: %w", string(out), err)
        if strings.Contains(string(out), "\"type\": \"error\"") {
                return "", errors.New(string(out))
-                       errs = multierr.Append(errs, errors.Wrap(err, string(out)))
+                       errs = multierr.Append(errs, fmt.Errorf("%s: %w", string(out), err))
                        errs = multierr.Append(errs, errors.New(string(out)))
-               return "", errors.Wrap(err, string(out))
+               return "", fmt.Errorf("%s: %w", string(out), err)
-               return "", errors.Wrap(err, string(out))
+               return "", fmt.Errorf("%s: %w", string(out), err)
        return string(out), nil
-               return -1, errors.Wrap(err, string(out))
+               return -1, fmt.Errorf("%s: %w", string(out), err)
-               return "", errors.Wrap(err, string(out))
+               return "", fmt.Errorf("%s: %w", string(out), err)
        if !strings.Contains(string(out), fmt.Sprintf("\"name\":\"%s\"", repoName)) {
                return "", errors.New(string(out))
-               return "", errors.Wrap(err, string(out))
+               return "", fmt.Errorf("%s: %w", string(out), err)
-                               errs = multierr.Append(errs, errors.Wrap(err, string(out)))
+                               errs = multierr.Append(errs, fmt.Errorf("%s: %w", string(out), err))
                        if !strings.Contains(string(out), "\"message\":\"202 Accepted\"") {
-                       errs = multierr.Append(errs, errors.Wrap(err, string(out)))
+                       errs = multierr.Append(errs, fmt.Errorf("%s: %w", string(out), err))
                if !strings.Contains(string(out), "\"message\":\"202 Accepted\"") {
-                       return errors.Errorf("unexpected response in DeleteRepoByID: %s", string(out))
+                       return fmt.Errorf("unexpected response in DeleteRepoByID: %s", string(out))
-                       return nil, errors.Wrap(err, string(out))
+                       return nil, fmt.Errorf("%s: %w", string(out), err)
-               return errors.Errorf("failed to copy directory: %s", string(out))
+               return fmt.Errorf("failed to copy directory: %s", string(out))
        return strings.TrimSpace(string(out)), nil
                if string(actual) != value {
-               return out, errors.Wrapf(err, "invoking helm: %s", string(out))
+               return out, fmt.Errorf("invoking helm: %s: %w", string(out), err)
        return string(bytes), nil
                newValue := string(newObj.Data[string(configsync.AuthSSH)])
mikebz-ct src/kpt-config-sync λ 
```

e2e/nomostest/git-server.go
sdowell: I suggest cleaning up the proposed commit history, this commit comes from another PR

cmd/nomos/version/version.go
```diff
@@ -257,7 +256,7 @@ type entry struct {
 func entries(vs map[string]vErr) []entry {
 	currentContext, err := restconfig.CurrentContextName()
 	if err != nil {
-		fmt.Printf("Failed to get current context name with err: %v\n", errors.Cause(err))
+		fmt.Printf("Failed to get current context name with err: %v\n", err)
```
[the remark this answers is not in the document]
mikebz: Good question.  I did think about this one for a bit.  It's not clear to me why we would want to ignore the error at the top of the stack in case these are wrapped.  The danger of calling the new equivalent of the Cause() is that you might just get a `nil` https://pkg.go.dev/errors#Unwrap.

In this case, the original author didn't leave any trace of why they chose to ignore the wrapping error.  In most circumstances I would want to see the full error.

cmd/nomos/status/status.go
mikebz: commented above, good question.

e2e/nomostest/git-server.go
mikebz: I will be squashing on merge.  It's not clear why this is showing up as a diff against main.  https://github.com/GoogleContainerTools/kpt-config-sync/blob/426697e1a2fd0b4298f47fbb4373a0af2b4eb91c/e2e/nomostest/git-server.go#L138

sdowell: By default the auto-submit will perform a squash, but the squashed commit message with all of these commits will be pretty confusing. Would be preferable to squash before the change is submitted so that we can review the commit message

mikebz: let me try to make it all work and make everyone happy and then I will work on the commit history stuff... 😅

cmd/nomos/version/version.go
```diff
@@ -96,12 +96,11 @@ of the "nomos" client binary for debugging purposes.`,
 func allKubectlConfigs() (map[string]*rest.Config, error) {
 	allCfgs, err := restconfig.AllKubectlConfigs(flags.ClientTimeout)
 	if err != nil {
-		// Unwrap the "no such file or directory" error for better readability
-		if unWrapped := errors.Cause(err); os.IsNotExist(unWrapped) {
-			err = unWrapped
+		var pathErr *os.PathError
```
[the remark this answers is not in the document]
mikebz: interesting question, of course best to verify it with a test, but I think creating this test case would be pretty challenging.  I don't even really know if we can ever hit this situation.  Regardless, someone put this in so we can just keep it.   In golang code:
https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/os/error.go;l=90

It looks the types of errors that package is looking for 3 types:
```
// underlyingError returns the underlying error for known os error types.
func underlyingError(err error) error {
	switch err := err.(type) {
	case *PathError:
		return err.Err
	case *LinkError:
		return err.Err
	case *SyscallError:
		return err.Err
	}
	return err
}
```
Meanthing that PathError is the most likely type to expect when we are opening a non-existing file.

Other code snippets that open non existing files get a PathError as well:
https://go.dev/play/p/3XXM5JUiq4P

```
func main() {
    if _, err := os.Open("non-existing"); err != nil {
        var pathError *os.PathError
        if errors.As(err, &pathError) {
            fmt.Println("Failed at path:", pathError.Path)
        } else {
            fmt.Println(err)
        }
    }
}
```

e2e/nomostest/git-server.go
mikebz: Some squashed probably more to go

cmd/nomos/version/version.go
```diff
@@ -96,12 +96,11 @@ of the "nomos" client binary for debugging purposes.`,
...
+		if errors.As(err, &pathErr) {
+			err = pathErr
 		}
 
-		// nolint:errcheck
```
janetkuo, line 104: Does this need to be removed? Looks unrelated to this change.

cmd/nomos/version/version.go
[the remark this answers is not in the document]
janetkuo: This LGTM. If there's concern, we can remove this logic given that this simply makes the returned error more readable.

janetkuo (approved):

cmd/nomos/version/version.go
mikebz: removed.

PR #1145, opened 2024-02-23
chore: update the base image of the test-git-server

nan-yu:
The test scaffolding uses
https://github.com/jkarlosb/git-server-docker/blob/v1.0.0/Dockerfile to set up a test-git-server. However, that repo uses git 2.8.3, which doesn't support the `--initial-branch` option for `git init`.

We need to set up a different default branch rather than `master`. This commit updates the Dockerfile by updating the alpine base image from 3.4 to 3.19, which comes with git 2.43.0.

It also updates the image tag by appending the commit hash.

### DIFF
--- a/Makefile.build
+++ b/Makefile.build
@@ -199,7 +199,7 @@ config-sync-manifest-local: docker-registry config-sync-manifest
 # e2e/nomostest/git-server.go to reflect the version change
 GIT_SERVER_DOCKER := $(OUTPUT_DIR)/git-server-docker
 GIT_SERVER_RELEASE := v1.0.0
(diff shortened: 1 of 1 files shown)


karlkfi:
We might want to fork that and build it async so i can get scanned and patched like the other dependencies. Doing this way is expedient, but not particularly secure.

karlkfi:
This seems fine for now.

nan-yu:
> We might want to fork that and build it async so i can get scanned and patched like the other dependencies. Doing this way is expedient, but not particularly secure.

Make sense. I'll create an internal issue to track the work, but it will probably go to backlog with low priority.

nan-yu:
Internal task created, and this PR still needs an approval.

sdowell (approved):
/lgtm

tiffanny29631 (approved):
/lgtm

PR #1146, opened 2024-02-23
test: add a test to confirm the git-sync not pulling latest issue

nan-yu:
`git-sync` v4.1.0 couldn't pull the latest commit from a branch when the branch has multiple remotes and the branch in different remotes are out of sync.

This commit adds a test to confirm the behavior. A subsequent commit will update git-sync to v4.2.0 to fix this issue.

### DIFF
--- a/e2e/nomostest/git-server.go
+++ b/e2e/nomostest/git-server.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"kpt.dev/configsync/e2e"
+	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/kinds"
@@ -32,7 +33,8 @@ import (
 
 const testGitNamespace = "config-management-system-test"
 const testGitServer = "test-git-server"
-const testGitServerImage = testing.TestInfraArtifactRegistry + "/git-server:v1.0.0"
+
+const testGitServerImage = testing.TestInfraArtifactRegistry + "/git-server:v1.0.0-68df304c"
 const testGitHTTPServer = "http-git-server"
 const testGitHTTPServerImage = testing.TestInfraArtifactRegistry + "/http-git-server:v1.0.0-b3b4984cd"
 const safeToEvictAnnotation = "cluster-autoscaler.kubernetes.io/safe-to-evict"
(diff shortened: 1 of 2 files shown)


e2e/testcases/git_sync_test.go
```diff
@@ -0,0 +1,76 @@
...
+	nt.T.Logf("Mitigation: set spec.git.branch to HEAD to pull the latest commit")
+	nomostest.SetGitBranch(nt, configsync.RootSyncName, "HEAD")
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+	if err := nt.Validate("hello", "", &corev1.Namespace{}); err != nil {
+		nt.T.Fatal(err)
+	}
+	if err := nt.Validate(configsync.RootSyncName, configmanagement.ControllerNamespace,
+		&v1beta1.RootSync{}, testpredicates.RootSyncLastSyncedCommitNotEquals(initialSyncedCommit)); err != nil {
```
tiffanny29631, line 57: Should the test look at git ls-remote and compare equal to status.LastSyncedCommit or checking SHA updated is sufficient?

sdowell: I would suggest we get the SHA of the new commit from the local git repo and use that here (`"add Namespace"`)

```golang
nt.RootRepos[configsync.RootSyncName].Hash()
```

e2e/nomostest/git-server.go
```diff
@@ -32,7 +33,9 @@ import (
 
 const testGitNamespace = "config-management-system-test"
 const testGitServer = "test-git-server"
-const testGitServerImage = testing.TestInfraArtifactRegistry + "/git-server:v1.0.0"
+
+// TODO: update the image tag with commit SHA of https://github.com/GoogleContainerTools/kpt-config-sync/pull/1145.
+const testGitServerImage = testing.TestInfraArtifactRegistry + "/git-server:v1.0.0-patch"
```
sdowell, line 38: gentle reminder: this can be updated now

nan-yu: updated

e2e/testcases/git_sync_test.go
nan-yu: `nt.WatchForAllSyncs()` in line 50 uses the default `DefaultRootSha1Fn` function to validate the `LastSyncedCommit` field. The `DefaultRootSha1Fn` function uses `nt.RootRepos[configsync.RootSyncName].Hash()`. I don't think we need to reimplement the `DefaultRootSha1Fn` function explicitly here.

The extra validation guarantees that `nt.RootRepos[configsync.RootSyncName].Hash()` returns a different SHA other than the initial one.

sdowell: In that case, can we use `WatchForAllSyncs` rather than `RootSyncLastSyncedCommitNotEquals`? That seems like a more strict/explicit check

nan-yu: We have both `WatchForAllSyncs` and `RootSyncLastSyncedCommitNotEquals`. 
The first check validates `LastSyncedCommit` is the same as  `nt.RootRepos[configsync.RootSyncName].Hash()`.
The second check validates `nt.RootRepos[configsync.RootSyncName].Hash()` is updated explicitly.

nan-yu: It also explicitly shows that the mitigation successfully pulls the latest commit.

sdowell: My main point is that the second assertion seems a bit redundant, but it's not a blocker

sdowell (approved):
/lgtm

e2e/testcases/git_sync_test.go
nan-yu: It is redundant because the validation is a bit implicit. I'm using this extra validation to explicitly show that the `lastSyncedCommit`  field is updated.

sdowell: Assuming you have commit A -> B, and then:
1. Assert that the currently synced commit is B
2. Then assert that the currently synced commit is not A

The second assertion seems redundant given that it is a weaker assertion than the first

sdowell: In what case would `B != A`? I interpret that as a given here, since the `DefaultRootSha1Fn` is using the commit hash of the local HEAD

nan-yu: Make sense as we define the `Hash()` function.
My original intent is to prevent any unexpected git behavior, such as `git ls-remote` not returning the latest commit, but in this case, it seems I'm overthinking.
Removed the redundant validation.

PR #1147, opened 2024-02-23
fix: update git-sync to v4.2.1 to fix a pulling issue

nan-yu:
There is a bug in git-sync v4.1.0. When branches in different remotes
are out of sync, `git-sync` fetches the commit SHA from the last line,
which may not be the latest. This leads to an issue that Config Sync
couldn't pull the latest commit from HEAD.

The issue was fixed in v4.2.0 by
https://github.com/kubernetes/git-sync/pull/845.
This commit updates git-sync to v4.2.1 to include the fix.

It also bumps the debian-base to latest version for CVE fixes.

b/325341042


nan-yu:
This PR depends on #1146

PR #1148, opened 2024-02-26
e2e: Add registry login/logout to test setup

karlkfi:
The gcloud auth token only lasts 1 hour, so we need to re-auth for each test that uses helm or oci.

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -402,6 +402,7 @@ func TestHelmDefaultNamespace(t *testing.T) {
 	nt := nomostest.New(t,
 		nomostesting.SyncSource,
 		ntopts.Unstructured,
+		ntopts.RequireHelmProvider,
 	)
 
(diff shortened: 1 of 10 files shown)


e2e/testcases/helm_sync_test.go
```diff
@@ -730,6 +733,7 @@ func TestHelmARTokenAuth(t *testing.T) {
 		nomostesting.SyncSource,
 		ntopts.Unstructured,
 		ntopts.RequireGKE(t),
+		ntopts.RequireHelmProvider,
 		ntopts.RequireHelmArtifactRegistry(t),
```
sdowell, line 737: Could have `RequireHelmArtifactRegistry` imply `RequireHelmProvider` as well, although that might require a bit more refactoring

karlkfi: Done

sdowell (approved):
/lgtm

PR #1149, opened 2024-02-27
test: fix helm client e2e auth

karlkfi:
- Add HelmClient & OCIClient with per-env config files to
  avoid conflicts due to config changes and/or authn/z.
- Add login/logout to the Helm & OCI providers that delegates to
  the HelmClient & OCIClient
- Use gcloud credHelper for docker, helm, and crane
- Logout of crane & helm to force use of gclound credHelper
- Add provider login to registry port-forwarding, in case we need
  it later for local registry auth.
- Move kubectl and gcloud shell commands to their own files
- Remove shell Docker/Helm/Git wrappers. Just use ExecWithDebug.
  They weren't being used consistently anyway.
- Add Env to TestShell to make it easier to add environment variables
  to specific sub-shells, for use in HelmClient & OCIClient, which
  have their own configs.
- Refactor Helm & OCI providers to have helm-specific and oci-specific
  methods. The new methods are more explicit about what they return
  and should make it harder to use the wrong one accidentally.


e2e/nomostest/nt.go
```diff
@@ -755,70 +755,6 @@ func (nt *NT) portForwardGitServer() {
 	)
 }
 
-// portForwardRegistryServer forwards the registry-server deployment to a port.
-func (nt *NT) portForwardRegistryServer() {
```
sdowell, line 759: I feel I'd prefer to keep this close in proximity to the similar methods so that it's easier to discover. Is there a reason this code needs to be moved, or can we call `portForwardRegistryServer` from `setupRegistryClient`?

e2e/nomostest/registry_server.go
```diff
@@ -97,32 +98,136 @@ func setupRegistry(nt *NT) error {
...
+	// For local registries, set up port-forwarding to the in-cluster registry server.
+	if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local || opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
+		// The local registry uses ephemeral storage. So login and re-push all
+		// the oci images and helm charts after the registry recovers from a crash.
+		// Before the test, nt.ociImages & nt.helmPackages will be empty.
+		var onReadyCallbacks []func(int, string)
+		if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local {
+			provider := nt.OCIProvider.(*registryproviders.LocalOCIProvider)
+			setupFn := func(newPort int, podName string) {
+				if err := provider.Login(); err != nil {
```
nan-yu, line 126: Does it need to login with local provider?

e2e/nomostest/registry_server.go
```diff
@@ -97,32 +98,136 @@ func setupRegistry(nt *NT) error {
...
-			if err := nt.HelmProvider.Logout(); err != nil {
-				nt.T.Error(err)
+
+	// For local registries, set up port-forwarding to the in-cluster registry server.
+	if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local || opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
+		// The local registry uses ephemeral storage. So login and re-push all
+		// the oci images and helm charts after the registry recovers from a crash.
+		// Before the test, nt.ociImages & nt.helmPackages will be empty.
+		var onReadyCallbacks []func(int, string)
+		if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local {
```
nan-yu, line 123: nit: we can check either `opts.RequireOCIProvider` or `*e2e.OCIProvider == e2e.Local`. It's redundant to check both as line 118 guarantees the other one.

e2e/nomostest/nt.go
karlkfi: Login needs to happen after Ready and before Push, and portForwardRegistryServer sets up Pushes from the callbacks that run when the registry Pod is Ready. 

Today this isn't critical because the local registry Login is a no-op, but if we ever add auth for the local registry, the login needs to happen after each time the registry is restarted, after it's ready and after the port-forwarding is ready.

e2e/nomostest/registry_server.go
karlkfi: Not today, but it shouldn't need to know that the local Login is a no-op.

e2e/nomostest/registry_server.go
karlkfi: Not true. It's possible for the provider to be local but the current test not to require OCI, if the current test requires Helm, for example.

e2e/nomostest/nt.go
sdowell: What I'm suggesting is just a refactoring, e.g.:
```golang
func setupRegistryClient(nt *NT, opts *ntopts.New) error {
    ...
    nt.portForwardRegistryServer(opts)
    ...
}
```

karlkfi: Login needs to be performed at different times depending on whether there's port-forwarding or not.

e2e/nomostest/registry_server.go
sdowell: FWIW the in-cluster registry deployment already supports basic auth for testing, but surfaces a separate unauthenticated port for the test framework to push to for convenience. So this may be an implementation detail, but I don't think the test scaffolding should ever need to worry about it

karlkfi: If we use auth, can we skip the port-forwarding?

e2e/nomostest/nt.go
sdowell: I understand that, here's the complete snippet I had in mind to clear ambiguity:
```golang
// portForwardRegistryServer forwards the registry-server deployment to a port.
func (nt *NT) portForwardRegistryServer(opts *ntopts.New) {
	// The local registry uses ephemeral storage. So re-push all the oci images
	// and helm charts after the registry recovers from a crash.
	var onReadyCallbacks []func(int, string)
	if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local {
		provider := nt.OCIProvider.(*registryproviders.LocalOCIProvider)
		resetImages := func(newPort int, podName string) {
			for _, image := range nt.ociImages {
                if err := provider.Login(); err != nil {
			        nt.T.Fatalf("login: %v", err)
			    }
				// construct push URL with provided port. Calling LocalPort would lead to deadlock
				if err := image.Push(provider.PushURLWithPort(newPort, image.Name)); err != nil {
					nt.T.Fatalf("pushing image: %w", err)
				}
			}
		}
		onReadyCallbacks = append(onReadyCallbacks, resetImages)
	}
	if opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
		provider := nt.HelmProvider.(*registryproviders.LocalHelmProvider)
		resetPackages := func(newPort int, podName string) {
            if err := provider.Login(); err != nil {
			    nt.T.Fatalf("login: %v", err)
			}
			for _, pkg := range nt.helmPackages {
				// construct push URL with provided port. Calling LocalPort would lead to deadlock
				if err := pkg.Push(provider.PushURLWithPort(newPort, pkg.Name)); err != nil {
					nt.T.Fatalf("pushing image: %w", err)
				}
			}
		}
		onReadyCallbacks = append(onReadyCallbacks, resetPackages)
	}
	portForwarder := nt.newPortForwarder(
		TestRegistryNamespace,
		TestRegistryServer,
		fmt.Sprintf(":%d", RegistryHTTPPort),
		portforwarder.WithOnReadyCallback(func(newPort int, podName string) {
			for _, fn := range onReadyCallbacks {
				fn(newPort, podName)
			}
		}),
	)
	if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local {
		provider := nt.OCIProvider.(*registryproviders.LocalOCIProvider)
		nt.T.Cleanup(func() {
			// clear PortForwarder after each test. at this point it has been stopped
			// a new PortForwarder is created for each test.
			provider.PortForwarder = nil
		})
		provider.PortForwarder = portForwarder
	}
	if opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
		provider := nt.HelmProvider.(*registryproviders.LocalHelmProvider)
		nt.T.Cleanup(func() {
			// clear PortForwarder after each test. at this point it has been stopped
			// a new PortForwarder is created for each test.
			provider.PortForwarder = nil
		})
		provider.PortForwarder = portForwarder
	}
	nt.startPortForwarder(
		TestRegistryNamespace,
		TestRegistryServer,
		portForwarder,
	)
}

...

func setupRegistryClient(nt *NT, opts *ntopts.New) error {
    if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local || opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
        nt.portForwardRegistryServer()
    }
	// For non-local registries, just login before each test that uses the provider.
	if opts.RequireOCIProvider && *e2e.OCIProvider != e2e.Local {
		if err := nt.OCIProvider.Login(); err != nil {
			return err
		}
	}
	if opts.RequireHelmProvider && *e2e.HelmProvider != e2e.Local {
		if err := nt.HelmProvider.Login(); err != nil {
			return err
		}
	}

	// Register this cleanup after startPortForwarder to ensure this cleanup is
	// executed before the port-forwarder is shut down.
	nt.T.Cleanup(func() {
		if opts.RequireHelmProvider {
			// Delete all charts & images pushed during the test.
			for _, pkg := range nt.helmPackages {
				if err := pkg.Delete(); err != nil {
					nt.T.Error(err)
				}
			}
			nt.helmPackages = nil
			// Logout to reset the local client credentials.
			if err := nt.HelmProvider.Logout(); err != nil {
				nt.T.Error(err)
			}
		}
		if opts.RequireOCIProvider {
			// Delete all charts & images pushed during the test.
			for _, image := range nt.ociImages {
				if err := image.Delete(); err != nil {
					nt.T.Error(err)
				}
			}
			nt.ociImages = nil
			// Logout to reset the local client credentials.
			if err := nt.OCIProvider.Logout(); err != nil {
				nt.T.Error(err)
			}
		}
	})

	return nil
}
```

e2e/nomostest/registry_server.go
sdowell: No, these are separate problems. The port forwarding is so that the test scaffolding can push to the in-cluster registry

sdowell: Similar to how the test scaffolding pushes to the in-cluster git server using the local port forward

karlkfi: Oh right, because we didn't want to set up ingress and we want to delay calls when it's unhealthy.

e2e/nomostest/nt.go
karlkfi: Fine.

I don't like all the spaghetti caused by these callbacks, or that we need to put login in two different places, but cleaning that up seems like it would take a lot more refactoring and I'm not sure how to do it yet.

karlkfi: This also means we need to pass the test options into an NT method, which feels kind of wrong. Usually for the NT methods the options have already been resolved into the new NT. But I don't really want to copy the opts variables into NT variables...

e2e/nomostest/registryproviders/artifact_registry.go
```diff
@@ -121,33 +121,17 @@ type ArtifactRegistryOCIProvider struct {
...
 	return nil
 }
 
 // Logout of the registry with the crane client
 func (a *ArtifactRegistryOCIProvider) Logout() error {
-	_, err := a.shell.ExecWithDebug("crane", "auth", "logout", a.registryHost())
-	if err != nil {
-		return fmt.Errorf("running logout command: %w", err)
-	}
+	// Never log out of docker clients.
```
sdowell, line 133: Should we just remove the logout methods/interfaces?

karlkfi: Assuming we dont need to log into the local registries, I'll just move the "login" to the env setup.

karlkfi: Oh wait. I can't. It depends on the location, which is a provider field...

sdowell: I think you should be able to move login back to the `Setup` method so that it's done once at the beginning of the test suite. The provider is scoped to the test suite rather than per testcase

e2e/nomostest/registry_server.go
```diff
@@ -97,32 +97,70 @@ func setupRegistry(nt *NT) error {
 	return nil
 }
 
-// setupRegistryClient handles registry authentication.
-// This is necessary on each test because gcloud auth tokens expire after 1h and
-// some test suites take longer than 1h.
+// setupRegistryClient handles registry authentication and port-forwarding.
+//
+// Re-auth is necessary on each test because gcloud auth tokens expire after 1h
```
sdowell, line 102: This comment is no longer accurate with the usage of the credential helper

e2e/nomostest/registryproviders/artifact_registry.go
karlkfi: I'm gonna add client configs to NT and just make them all independent per environment under TmpDir so we dont have to worry about them conflicting

e2e/nomostest/nt.go
karlkfi: Alright. I think I finally found a way to refactor this.

I added a new `ProxiedRegistryProvider` interface that has ProxyAddress, Restore, and Reset methods. This way the details of what's happening is hidden in the proxy-specific providers without causing deadlocks. As a bonus, the Cleanup is now separate for the proxy providers and non-proxy providers. So the Restore and Reset methods are both called from `portForwardRegistryServer` and `setupRegistryClient` only needs to call Login/Logout for the non-proxied providers.

e2e/nomostest/registry_server.go
karlkfi: This Login is now moved into `ProxiedRegistryProvider.Restore`. So `portForwardRegistryServer` doesn't need to know whether login is necessary or not or how to do it.

e2e/nomostest/registryproviders/artifact_registry.go
karlkfi: Login/Logout has been refactored. Please review. I think they can be left in the new refactor because they have comments that explain why it's not needed, and the caller doesn't need to know.

e2e/nomostest/registry_server.go
karlkfi: Fixed.

karlkfi:
/retest

Looks like TestReconcileRepoSyncLifecycleValidToInvalid is flaky.

e2e/nomostest/registry_server.go
```diff
@@ -97,31 +97,43 @@ func setupRegistry(nt *NT) error {
...
+// OCI provider is only configured if RequireOCIProvider is true.
 func setupRegistryClient(nt *NT, opts *ntopts.New) error {
-	nt.T.Cleanup(func() {
-		if opts.RequireHelmProvider || opts.RequireLocalHelmProvider {
-			if err := nt.HelmProvider.Logout(); err != nil {
-				nt.T.Error(err)
-			}
+	// For local registries, set up port-forwarding to the in-cluster registry server.
+	// portForwardRegistryServer handles login and logout too.
+	if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local || opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
```
sdowell, line 114: Can we rename `RequireOCIProvider`/`RequireHelmProvider` to differentiate them from the other `Require*` options? The other `Require` options indicate to skip the test (e.g. `RequireLocalHelmProvider`), which is functionally quite different from this option.

I'm thinking something along the lines of
- `IsHelmTest`/`IsOCITest`
- `UsesHelmProvider`/`UsesOCIProvider`
- etc

e2e/nomostest/testshell/helm_client.go
```diff
@@ -0,0 +1,123 @@
...
+// Login uses gcloud to configure docker auth credentials used by helm and crane.
+//
+// Also explicitly calls `helm registry logout` and `crane auth logout` to
+// remove any saved tokens for the specified registry, as these take precedence.
+//
+// Files modified:
+// - `${DOCKER_CONFIG}/config.json` (crane/docker)
+// - `${HELM_CONFIG_HOME}/registry/config.json` (helm)
+func (hc *HelmClient) Login(registryHost string) error {
+	if out, err := hc.Helm("registry", "logout", registryHost); err != nil {
```
sdowell, line 85: Is the explicit logout still necessary now that a config path is being set in the shell's environment?

e2e/nomostest/registry_server.go
karlkfi: Yeah, I get that it's a little confusing. My thinking is that the test defines its requirements, and the test infra either injects it or skips it.

sdowell: That's fair, works for me

e2e/nomostest/testshell/helm_client.go
karlkfi: As long as the cleanup happens and the environment isn't re-used while dirty, then yeah, we shouldn't need to log out before logging it. 

Is that happening? 

I think the TmpDir doesn't actually get nuked when a test fails when running --debug. If you don't manually delete it, will it use the same one next time based on the name of the environment/cluster? Or is it randomized?

e2e/nomostest/config_sync.go
```diff
@@ -824,10 +824,15 @@ func RepoSyncObjectV1Beta1(nn types.NamespacedName, repoURL string, sourceFormat
 
 // RootSyncObjectOCI returns a RootSync object that syncs the provided OCIImage.
 func (nt *NT) RootSyncObjectOCI(name string, image *registryproviders.OCIImage) *v1beta1.RootSync {
+	imageURL, err := image.RemoteAddressWithTag()
+	if err != nil {
+		nt.T.Fatalf("OCIImage.SyncURLWithTag: %v", err)
```
sdowell, line 829: nit: `RemoteAddressWithTag` here and below

e2e/nomostest/testshell/helm_client.go
karlkfi: removed

e2e/nomostest/config_sync.go
karlkfi: fixed

sdowell (approved):
/lgtm

PR #1150, opened 2024-02-27
fix: stop validating root secrets (#1140)

nan-yu:
Config Sync doesn't generate a copy of root Secret, so no need to validate the generated secret name length.

b/326430718


nan-yu:
There're merge conflicts while cherrypicking #1140 because `pkg/reconcilermanager/controllers/rootsync_controller.go` was also updated to support CA certificate for OCI/Helm: https://github.com/GoogleContainerTools/kpt-config-sync/commit/3dd1f8dc7a0622c875a2744068f140d53fea739c#diff-87f46982b7989d2ed226f9f1007d2327aae710a78e09c68bb4575ba18f0e340c.

We need to manually resolve the merge conflicts.

sdowell (approved):
/lgtm

PR #1153, opened 2024-02-28
test: update e2e test to fix gitlab push error

nan-yu:
Gitlab doesn't allow pushing a branch with `refs/remotes/upstream/*`. It failed with a `deny updating a hidden ref` error. This commit changes the remote branch to just `upstream/main`, which is `refs/heads/upstream/main`.

### DIFF
--- a/e2e/testcases/git_sync_test.go
+++ b/e2e/testcases/git_sync_test.go
@@ -35,7 +35,7 @@ func TestMultipleRemoteBranchesOutOfSync(t *testing.T) {
 	}
 
 	nt.T.Log("Create an extra remote tracking branch")
-	nt.Must(nt.RootRepos[configsync.RootSyncName].Push("HEAD:refs/remotes/upstream/main"))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Push("HEAD:upstream/main"))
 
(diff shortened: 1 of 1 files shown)


nan-yu:
This should fix the ` ! [remote rejected] HEAD -> upstream/main (deny updating a hidden ref)` failure in the gitlab CI.

e2e/testcases/git_sync_test.go
```diff
@@ -35,7 +35,7 @@ func TestMultipleRemoteBranchesOutOfSync(t *testing.T) {
 	}
 
 	nt.T.Log("Create an extra remote tracking branch")
-	nt.Must(nt.RootRepos[configsync.RootSyncName].Push("HEAD:refs/remotes/upstream/main"))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Push("HEAD:upstream/main"))
```
sdowell, line 38: I believe this may not be functionally the same. IIUC In this context `upstream/main` will just be the name of a head/branch on the remote, whereas before we were pushing a remote ref. If the fix was reverted, would this test still reproduce the bug?

sdowell (approved):
/lgtm

PR #1155, opened 2024-02-29
Add test to confirm the required RBAC for managing namespace resources

nan-yu:
One of the Config Sync customers complains that the RBAC for Namespace reconciler only works with ClusterRole with RoleBinding, but not Role with RoleBinding. This commit adds test to validate Role + RoleBinding also works for Namespace reconciler RBAC.

It also iterates through all required permissions for managing namespace-scoped resources in a namespace reconciler.

### DIFF
--- a/e2e/testcases/namespace_repo_test.go
+++ b/e2e/testcases/namespace_repo_test.go
@@ -20,10 +20,14 @@ import (
 	"testing"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


karlkfi:
> One of the Config Sync customers complains that the RBAC for Namespace reconciler only works with ClusterRole with RoleBinding, but not Role with RoleBinding.

Was the customer correct?

nan-yu:
> > One of the Config Sync customers complains that the RBAC for Namespace reconciler only works with ClusterRole with RoleBinding, but not Role with RoleBinding.
> 
> Was the customer correct?

Nope. The customer was wrong, but our documentation doesn't clarify it and we don't have test to cover the Role+RoleBinding scenario.
So this commit adds the test coverage and I'm planning to update the public doc.

sdowell (approved):
/lgtm

PR #1156, opened 2024-03-05
test: fix registry cleanup to delete images after

sdowell:
A recent refactor updated the cleanup logic to only delete images in cleanup for local registries. The registry state needs to be reset regardless of the registry type.

### DIFF
--- a/e2e/nomostest/registry_server.go
+++ b/e2e/nomostest/registry_server.go
@@ -114,22 +114,28 @@ func setupRegistryClient(nt *NT, opts *ntopts.New) error {
 	if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local || opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
 		nt.portForwardRegistryServer(opts.RequireHelmProvider, opts.RequireOCIProvider)
 	}
-	// For non-local registries, just login before each test and logout after.
-	if opts.RequireOCIProvider && *e2e.OCIProvider != e2e.Local {
+	// Login before each test, then reset the registry and logout after the test.
+	if opts.RequireOCIProvider {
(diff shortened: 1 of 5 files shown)


e2e/nomostest/registry_server.go
```diff
@@ -114,22 +114,28 @@ func setupRegistryClient(nt *NT, opts *ntopts.New) error {
 	if opts.RequireOCIProvider && *e2e.OCIProvider == e2e.Local || opts.RequireHelmProvider && *e2e.HelmProvider == e2e.Local {
 		nt.portForwardRegistryServer(opts.RequireHelmProvider, opts.RequireOCIProvider)
 	}
-	// For non-local registries, just login before each test and logout after.
-	if opts.RequireOCIProvider && *e2e.OCIProvider != e2e.Local {
+	// Login before each test, then reset the registry and logout after the test.
```
sdowell, line 117: Given that the login/logout is idempotent, I wrote this to just allow redundant login for local registries as it keeps the control flow simpler

e2e/nomostest/registryproviders/artifact_registry.go
```diff
@@ -199,12 +210,26 @@ func (a *ArtifactRegistryOCIProvider) DeleteImage(imageName, digest string) erro
 	return a.deleteImage(imageName, digest)
 }
 
+// Reset the state of the registry by deleting images pushed during the test.
+func (a *ArtifactRegistryOCIProvider) Reset() error {
```
sdowell, line 214: JFYI the reason the cleanup was factored the way it was before was to avoid sprawl of reimplementing cleanup for each registry variant. But I'd prefer to take a surgical approach and get things working before another refactor

karlkfi:
/retest

Looks like a metrics flake maybe

nan-yu (approved):
/lgtm

PR #1159, opened 2024-03-08
test: fix flake with TestNamespaceRepo

sdowell:
The test used a short timeout to wait for the Deployment to become current, which can be flaky on autopilot. The test does not really care whether the Deployment reconciles, so the assertion is changed to a simple existence check. Additionally, if the test fails in the middle it may not have permission to manage/delete the Deployment. A cleanup block is added to ensure the Deployment can be deleted in cleanup.

### DIFF
--- a/e2e/testcases/namespace_repo_test.go
+++ b/e2e/testcases/namespace_repo_test.go
@@ -112,6 +112,10 @@ func TestNamespaceRepo_Centralized(t *testing.T) {
 }
 
 func validateRepoSyncRBAC(nt *nomostest.NT, ns string, nsRepo *gitproviders.Repository, configureRBAC configureRBACFunc) {
+	nt.T.Cleanup(func() {
+		// Grant full permission to manage the Deployment in case the test fails early
+		configureRBAC(nt, ns, []string{"*"})
+	})
 	nt.T.Log("Add a deployment to the namespace repo without RBAC")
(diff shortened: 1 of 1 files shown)


e2e/testcases/namespace_repo_test.go
```diff
@@ -147,9 +151,8 @@ func validateRepoSyncRBAC(nt *nomostest.NT, ns string, nsRepo *gitproviders.Repo
 	if err := nt.WatchForAllSyncs(); err != nil {
 		nt.T.Fatal(err)
 	}
-	if err := nt.Watcher.WatchForCurrentStatus(kinds.Deployment(), "hello-world", ns,
-		testwatcher.WatchTimeout(30*time.Second)); err != nil {
-		nt.T.Fatalf("deployment hello-word not current: %v", err)
+	if err := nt.Validate("hello-world", ns, &appsv1.Deployment{}); err != nil {
+		nt.T.Fatalf("deployment hello-word not found: %v", err)
```
tiffanny29631, line 155: nit:world

tiffanny29631 (commented):
/lgtm

e2e/testcases/namespace_repo_test.go
sdowell: fixed

nan-yu (approved):
/lgtm

PR #1160, opened 2024-03-12
Update csr e2e tests to push sources during test

nan-yu:
This commit makes the csr e2e tests to be hermetic.

With the CSR spec being configured at the test scaffoloding layer, this 
commit also includes the following changes: 
- add a '--workloadidentity' flag to run workload_identity_test.go if 
  and only if the flag is specified. It doesn't require the check before 
  because the RootSync syncs from a local test git-server initially. 
  After the change, the test requires --git-provider=csr. Hence, it 
  requires a separate Prow job to run these tests. 
- move all GCENode test into a single file. 
- unify the test flags to make sure particular tests run if and only if 
  the corresponding flag is specified. 
- remove the unused --load flag. 
 
When this commit is merged, we need to update the prow configs: 
- gcenode CI 
  ``` 
  - <<: *config-sync-standard-job 
  name: kpt-config-sync-standard-regular-gcenode 
  annotations: 
    testgrid-dashboards: googleoss-kpt-config-sync-main 
    testgrid-tab-name: standard-regular-gcenode 
  spec: 
    <<: *config-sync-standard-job-spec 
    containers: 
    - <<: *config-sync-standard-job-container 
      args: 
      - 'GCP_SUBNETWORK=prow-e2e-subnetwork-usw1-11' 
      - 'GKE_RELEASE_CHANNEL=regular' 
      - 'E2E_CLUSTER_PREFIX=standard-regular-gcenode' 
      - 'E2E_NUM_CLUSTERS=1' 
      - 'E2E_GIT_PROVIDER=csr' 
      - 'E2E_ARGS=--gcenode' 
  ```

- workloadidentity CI: 
  ``` 
  - <<: *config-sync-standard-job 
  name: kpt-config-sync-standard-regular-workloadidentity 
  annotations: 
    testgrid-dashboards: googleoss-kpt-config-sync-main 
    testgrid-tab-name: standard-regular-workloadidentity 
  spec: 
    <<: *config-sync-standard-job-spec 
    containers: 
    - <<: *config-sync-standard-job-container 
      args: 
      - 'GCP_SUBNETWORK=prow-e2e-subnetwork-usw1-11' 
      - 'GKE_RELEASE_CHANNEL=regular' 
      - 'E2E_CLUSTER_PREFIX=standard-regular-workloadidentity'
      - 'E2E_NUM_CLUSTERS=1'
      - 'E2E_GIT_PROVIDER=csr'
      - 'E2E_ARGS=--workloadidentity'
  ``` 
 
- stress CI and KCC CI 
  The test regex filter (-run=TestStress* and -run=TestKCC*) becomes 
  optional because non-stress or non-KCC tests will be skipped.


e2e/testcases/workload_identity_test.go
```diff
@@ -63,39 +65,40 @@ func TestWorkloadIdentity(t *testing.T) {
 		testKSAMigration            bool
 		requireHelmArtifactRegistry bool
 		requireOCIArtifactRegistry  bool
+		requireCSR                  bool
```
sdowell, line 68: Need to set `requireCSR` on the relevant test cases below

nan-yu: Good catch. I have it in #1161. Just added here.

e2e/nomostest/gitproviders/cloud_source_repository.go
```diff
@@ -0,0 +1,158 @@
...
+			slice = slice[0:maxLen]
+		}
+		truncated = append(truncated, slice)
+	}
+	return strings.Join(truncated, delimiter)
+}
+
+// CSR repo name may contain between 3 and 63 lowercase letters, digits and hyphens.
+// sanitizeRepoName replaces all slashes with hyphens, and truncate the name.
+func sanitizeRepoName(repoPrefix, name string) string {
```
tiffanny29631, line 138: Do we care about the repo names? Can we just remove all non-letter chars from the concatenated string, or it might cause collision?

nan-yu: The repo name is in the format of `cs-e2e-[CLUSTER_NAME]-[RSYNC_NAMESPACe]-[RSYNC-NAME]-[RSYNC_NAME_LENGTH]`. There're not many non-letter chars. It can still go over the length limit if CLUSTER_NAME is very long. 

The function truncates each hyphen-separated slice, which is still easy to map back to the cluster name and RSync.

e2e/nomostest/git-server.go
```diff
@@ -71,7 +72,7 @@ func setupGit(nt *NT) error {
 			nt.describeNotRunningTestPods(testGitNamespace)
 			return fmt.Errorf("waiting for git-server Deployment to become available: %w", err)
 		}
-	} else {
+	case e2e.Bitbucket, e2e.GitLab:
```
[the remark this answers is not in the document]
nan-yu: Historically, the idea was to run all combinations for auth types and git providers. However, that was considered low priority. Plus, the test permutation could increase a lot.
Hence, we ended up having a separate test case for testing CSR with the recommended auth types: gcenode and gcpserviceaccount.

e2e/testcases/csr_auth_test.go
```diff
@@ -60,24 +46,34 @@ func gsaCSRReaderEmail() string {
 // https://cloud.google.com/anthos-config-management/docs/how-to/installing-config-sync#git-creds-secret
 func TestGCENode(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.Unstructured,
-		ntopts.RequireGKE(t), ntopts.GCENodeTest)
+		ntopts.RequireGKE(t), ntopts.GCENodeTest,
+		ntopts.RequireCloudSourceRepository(t),
```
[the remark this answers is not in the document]
nan-yu: I see your point. Basically, the gcenode testcase requires specifying `ntopts.GCENodeTest` to indicate it tests the `gcenode` auth type, and also requires CSR/AR by specifying `ntopts.RequireCloudSourceRepository(t)`.

I think we can remove the confusion by changing `ntopts.GCENodeTest` to `ntopts.RequiresGCENode` and checking if `*e2e.GceNode` is true.

```
func RequireGceNode(t testing.NTB) Opt {
	if !*e2e.GceNode {
		t.Skip("The --gcenode flag must be set to `true` to run this test.")
	}
	return func(opt *New) {}
}
```

So we will have both `ntopts.RequiresGCENode` and `ntopts.RequireCloudSourceRepository(t)` on `TestGCENode`.
Similarly, we'll have `ntopts.RequiresGCENode` and `ntopts.RequireOCIArtifactRegistry` on `TestGCENodeOCI`, and 
`ntopts.RequiresGCENode` and `ntopts.RequireHelmArtifactRegistry` on `TestHelmGCENode`.

The benefit of having them separately is that it is clear that this sepecific test requires a gcenode with WI disabled, and requires source provider to be a GCP service.

We can also do it in `main_test.go` as you suggested, but we still need a `ntopts` to check `*e2e.GceNode`. And the condition needs to cover all source providers, e.g.
```
if *e2e.GceNode == true && *e2e.GitProvider != e2e.CSR && *e2e.OCIProvider != e2e.GAR && *e2e.HelmProvider != e2e.GAR {
	errs = multierr.Append(errs, fmt.Errorf("GCENode test can only run on the CSR git provider, or AR OCI provider, or AR Helm provider."))
}
```

e2e/nomostest/config_sync.go
```diff
@@ -666,10 +670,21 @@ func RootSyncObjectV1Alpha1(name, repoURL string, sourceFormat filesystem.Source
...
-		SecretRef: &v1alpha1.SecretReference{
+	}
+	switch *e2e.GitProvider {
+	case e2e.CSR:
+		// CSR provider requires auth because the repository is private
+		if *e2e.GceNode {
+			rs.Spec.Git.Auth = configsync.AuthGCENode
+		} else {
+			rs.Spec.Git.Auth = configsync.AuthGCPServiceAccount
+			rs.Spec.Git.GCPServiceAccountEmail = gitproviders.CSRReaderEmail()
```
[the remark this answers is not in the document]
nan-yu: >  I think it's a smell how there isn't any gcenode specific setup/assertions in TestGCENode

We do have an assertion for the gcenode tests: `workloadidentity.ValidateDisabled(nt)`. Maybe the `e2e.GceNode` flag name is confusing? All CIs are running on GCE nodes, but gcenode tests require WI to be disabled. What about renaming to something like `--wi=disabled`?

>  If --gcenode --git-provider=csr is provided, I should in theory be able to run any of the generic git-sync tests with that config.

I'm not sure if I agree with this. We just need a single test to validate gcenode + CSR works. We don't need to run the entire test suite to test all features with gcenode+CSR.

> If the idea is that we don't want to run the whole test suite for the CI job, then we have other options to filter which tests are run (either by test name regex or test group).

Yes. I think we can rename `TestHelmGCENode` to `TestGCENodeHelm` and reuse  `-run=TestGCENode*` to filter these gcenode tests.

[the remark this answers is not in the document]
sdowell: > We do have an assertion for the gcenode tests: workloadidentity.ValidateDisabled(nt). Maybe the e2e.GceNode flag name is confusing? All CIs are running on GCE nodes, but gcenode tests require WI to be disabled. What about renaming to something like --wi=disabled?

I think `workloadidentity.ValidateDisabled(nt)` is perhaps redundant and/or should just be handled by the test scaffolding. The test scaffolding already handles the disabling of workload identity on cluster creation when `--gcenode` is set. I don't think there's any need to rename the parameter.

> I'm not sure if I agree with this. We just need a single test to validate gcenode + CSR works. We don't need to run the entire test suite to test all features with gcenode+CSR.

I'm not suggesting that we necessarily create a CI job that runs the whole test suite. But with these abstractions in place I don't see a technical reason that I shouldn't be able to run any generic `git-sync` test using gcenode+CSR. For example if I wanted to run some one-off test from my workstation with that permutation.

> Yes. I think we can rename TestHelmGCENode to TestGCENodeHelm and reuse -run=TestGCENode* to filter these gcenode tests.

How about running the `sync-source` test group?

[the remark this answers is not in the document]
nan-yu: > The test scaffolding already handles the disabling of workload identity on cluster creation when --gcenode is set. 

Cool. I wasn't aware of that. Then agreed that we don't need the assertion.

> How about running the sync-source test group?

Then do we run all tests in the `sync-souce` test group with gcenode CSR/AR?

e2e/nomostest/gitproviders/cloud_source_repository.go
nan-yu: Simplified to append a hash for truncation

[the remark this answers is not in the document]
nan-yu: Simplified to append a hash for truncation

e2e/testcases/csr_auth_test.go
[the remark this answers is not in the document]
nan-yu: see https://github.com/GoogleContainerTools/kpt-config-sync/pull/1160#discussion_r1529156562

e2e/flags.go
```diff
@@ -95,6 +91,10 @@ var KCC = flag.Bool("kcc", false,
 var GceNode = flag.Bool("gcenode", false,
 	"If true, run test with 'gcenode' auth type.")
 
+// WorkloadIdentity enables running the e2e tests for 'gcpserviceaccount' and 'k8sserviceaccount' auth types
+var WorkloadIdentity = flag.Bool("workloadidentity", false,
```
sdowell, line 95: Does this imply we're going to add another CI job permutation for `--workload-identity`? It seems like it can be implied to run the WI tests when `TestCluster==gke` and `GceNode==false`

nan-yu: These WI tests need to run when `--test-cluster=gke` and `--gcenode=false`, and `--git-provider=csr`. 
It would need a new CI job with these flags, similarly as how the gcenode CI is configured. We probably don't need a permutation to run on all channels.

e2e/nomostest/config_sync.go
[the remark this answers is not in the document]
nan-yu: Forgot to reply your comment regarding `I think workloadidentity.ValidateDisabled(nt) is perhaps redundant and/or should just be handled by the test scaffolding.`.

I think we still need the validation because we may run the test with a pre-provisioned GKE cluster, which may have WI enabled.

e2e/nomostest/gitproviders/cloud_source_repository_test.go
```diff
@@ -0,0 +1,62 @@
...
+		testName     string
+		repoPrefix   string
+		repoName     string
+		expectedName string
+	}{
+		{
+			testName:     "RepoSync test-ns/repo-sync",
+			repoPrefix:   "test",
+			repoName:     "test-ns/repo-sync",
+			expectedName: "cs-e2e-test-test-ns-repo-sync-19dcbc51",
```
sdowell, line 34: I think we only need the truncation/hashing when the repo name exceeds the maximum length. Such that for this case the `expectedName` would be `cs-e2e-test-test-ns-repo-sync`.

e.g.
```golang
    ...
    if len(fullName) < repoNameMaxLen {
        return fullName
    }
    fullName = fullName[:repoNameMaxLen-1-repoNameHashLen]
    return fmt.Sprintf("%s-%s", fullName), hashStr)
```

e2e/nomostest/new.go
```diff
@@ -78,29 +78,40 @@ func newOptStruct(testName, tmpDir string, t nomostesting.NTB, ntOptions ...ntop
 		opt(optsStruct)
 	}
 
-	e2eTest := !optsStruct.LoadTest && !optsStruct.StressTest
-	if !*e2e.E2E && e2eTest {
+	if !*e2e.E2E {
```
sdowell, line 81: nit: I think this could be moved to input validation in `main_test.go`

e2e/nomostest/gitproviders/cloud_source_repository_test.go
nan-yu: We can't because we need to replace all slashes with hyphens.
The first two testcases explain why do we need the hash even the repo name doesn't exceed the maximum length.
`cs-e2e-test-test-ns-repo-sync` can represent RepoSync `test-ns/repo-sync` or `test/ns-repo-sync`.

sdowell: Ack, good point - I missed that. This looks good in that case

e2e/nomostest/new.go
nan-yu: Removed. `main_test.go` already has a check for it.

e2e/flags.go
sdowell: Ack we need to set `--git-provider=csr` specifically to run the git/WI tests, however there is still overlap for the `workloadidentity` auth types for other source providers (e.g. GAR with oci-sync/helm-sync).

Here I'm just talking about this new `workloadidentity` flag which means we need to create new testgrid permutations to specifically run any of the workload identity tests. I'm thinking it would be simpler to omit this flag and run the `workloadidentity` tests based on the existing options rather than adding new permuations.

sdowell: This is essentially the inverse of the `gcenode` flag, so having both seems like it might create redundant permutation

nan-yu: I see. So you want oci/helm + wi to run with the current CI jobs, but only have a new job to run git + wi.

I was thinking all `workloadidentity` tests running on a separate job, including git, oci, and helm. We'll need a separate job anyway. This option runs all `workloadidentity` tests consistently.

This new flag allows us to select all `workloadidentity` tests for the new job, and skip them for other CI jobs.

nan-yu: > This is essentially the inverse of the `gcenode` flag, so having both seems like it might create redundant permutation

It is not the inverse of the `gcenode` flag. We'll need 3 types of CI jobs:
- a job with the `--gcenode` flag that only runs the `ntopts.GCENodeTest` tests.
- a job with `--gcenode=false`, `--git-provider=local`, `--oci-provider=gar`  and `--helm-provider=gar` that runs  other testcases with local test-git-server. (oci/helm + wi tests won't be selected)
- a job with `--gcenode=false`, `--git-provider=csr`, `--oci-provider=gar`  and `--helm-provider=gar` that only runs the `ntopts.WorkloadIdentityTest` tests.

sdowell: To run all of the `workloadidentity` tests with this new permutation you would need to provide `--helm-provider=gar --oci-provider=gar --git-provider=csr --workloadidentity --test-cluster=gke`, which is already a lot of parameters. Also, it's already assumed for all of the other tests that if the helm/oci provider is GAR that it will default to workloadidentity auth.

Without this parameter, the only new job permutation we need is one that sets `--git-provider=csr`.

nan-yu: Without this parameter, how do you filter out the git+wi test from other jobs?

sdowell: Based on whether the git provider supports workload identity, i.e. `RequireCloudSourceRepository`. This would be similar to how it's currently handled for helm/oci with `RequireHelmArtifactRegistry`/`RequireOCIArtifactRegistry`

nan-yu: Right. The PR has it in place: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1160/files#diff-8d37aad37a787f98c6132f35743b0be4045a4470a4f6554baa6015b7d6a0b66dR209.

Based on the existing Prow configuration (--git-provider=local), the git+wi test is selected and would fail after this PR because of `ntopts.RequireCloudSourceRepository(t)`.
We would need a way to filter out the git+wi test on existing CI jobs.

sdowell: If `--git-provider=local` I expect that these tests would be skipped rather than fail, right?

My understanding for the `ntopts.RequireCloudSourceRepository(t)` is that they run if `--git-provider=csr` and are skipped otherwise.

nan-yu: Ah, I see. Somehow I thought these requireXXX flags would fail the test if it didn't match.
Then I agree, we don't need the flag. Will update the PR.

nan-yu: updated

e2e/testinfra/README.md
```diff
@@ -28,22 +25,6 @@ value is `cs-dev-hub`.
 
 ## Usage
 
-
-1. Push test OCI images to private registry, for example,
-    ```bash
-    GCP_PROJECT=your-gcp-project-name make push-test-oci-images-private
```
sdowell, line 34: It looks like we might still need this step to push GCR images, right?

sdowell (commented):
One final comment but LGTM

e2e/testinfra/README.md
nan-yu: Good catch. Updated.

sdowell (approved):
/lgtm

PR #1161, opened 2024-03-12
Add test coverage for RepoSync auth types

nan-yu:
b/325119647

### DIFF
--- a/e2e/testcases/gcenode_test.go
+++ b/e2e/testcases/gcenode_test.go
@@ -23,13 +23,13 @@ import (
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/kustomizecomponents"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/policy"
 	"kpt.dev/configsync/e2e/nomostest/registryproviders"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/e2e/nomostest/testutils"
 	"kpt.dev/configsync/e2e/nomostest/workloadidentity"
 	"kpt.dev/configsync/pkg/api/configsync"
-	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/declared"
 	"kpt.dev/configsync/pkg/testing/fake"
 )
@@ -53,7 +53,9 @@ func TestGCENodeCSR(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.Unstructured,
 		ntopts.RequireGKE(t), ntopts.GCENodeTest,
 		ntopts.RequireCloudSourceRepository(t),
-	)
+		ntopts.NamespaceRepo(testNs, configsync.RepoSyncName),
+		ntopts.RepoSyncPermissions(policy.AllAdmin()), // NS reconciler manages a bunch of resources.
+		ntopts.WithDelegatedControl)
 
 	if err := workloadidentity.ValidateDisabled(nt); err != nil {
 		nt.T.Fatal(err)
@@ -71,9 +73,23 @@ func TestGCENodeCSR(t *testing.T) {
 		}
 	}`)
 
-	err := nt.WatchForAllSyncs(nomostest.WithRootSha1Func(nomostest.RemoteRootRepoSha1Fn),
+	nt.T.Log("Add the namespace-repo directory to RepoSync's repo")
+	repoSync := fake.RepoSyncObjectV1Beta1(testNs, configsync.RepoSyncName)
+	repoSyncRef := nomostest.RepoSyncNN(testNs, configsync.RepoSyncName)
+	nt.Must(nt.NonRootRepos[repoSyncRef].Copy("../testdata/hydration/namespace-repo", "."))
+	nt.Must(nt.NonRootRepos[repoSyncRef].CommitAndPush("add DRY configs to the repository"))
+	nt.MustMergePatch(repoSync, `{
+		"spec": {
+			"git": {
+				"dir": "namespace-repo"
+			}
+		}
+	}`)
+
+	err := nt.WatchForAllSyncs(
 		nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{
 			nomostest.DefaultRootRepoNamespacedName: "kustomize-components",
+			repoSyncRef:                             "namespace-repo",
 		}))
 	if err != nil {
 		nt.T.Fatal(err)
@@ -82,6 +98,7 @@ func TestGCENodeCSR(t *testing.T) {
 	if err := testutils.ReconcilerPodMissingFWICredsAnnotation(nt, nomostest.DefaultRootReconcilerName); err != nil {
 		nt.T.Fatal(err)
 	}
+	kustomizecomponents.ValidateTenant(nt, repoSyncRef.Namespace, repoSyncRef.Namespace, "base")
 }
 
 // TestGCENodeOCI tests the `gcenode` auth type for the OCI image.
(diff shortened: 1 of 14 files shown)


e2e/testcases/csr_auth_test.go
```diff
@@ -60,25 +47,52 @@ func gsaCSRReaderEmail() string {
 // https://cloud.google.com/anthos-config-management/docs/how-to/installing-config-sync#git-creds-secret
 func TestGCENode(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.SyncSource, ntopts.Unstructured,
-		ntopts.RequireGKE(t), ntopts.GCENodeTest)
+		ntopts.RequireGKE(t), ntopts.GCENodeTest,
+		ntopts.RequireCloudSourceRepository(t),
```
sdowell, line 51: What about the logic in this test is specific to CSR? It seems like this test case could run on other git providers as well. Also, note that we don't currently have any CI jobs which set `--gcenode --git-provider=csr`.

Since the `gcenode` logic is moved to the test scaffolding, I think with a little refactoring all of the tests could be validating `gcenode` with `--git-provider=csr`. This would give greater test coverage for gcenode/csr.
- We could update the job config [here](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml#L316) to set `--git-provider=csr` 
- Update the option logic such that for `gcenode` we run all git tests for CSR and run all oci/helm tests for GCR/AR

nan-yu: The `gcenode` auth type is specific to Google services like CSR, AR. It should not run on other git providers.

My thought was to update the job config and make all `GCENode` testcases run with `--git-provider=csr --oci-provider=gar --helm-provider=gar`.

sdowell: > The gcenode auth type is specific to Google services like CSR, AR. It should not run on other git providers.

Yes, this is why I mentioned that we should set this on the job config.

My point is that the logic inside the testcase doesn't seem specific to GCENode or CSR - the abstraction is handled inside the test scaffolding. So the `TestGCENode` testcase itself isn't really descriptive of a GCENode/CSR test - if you removed the `ntopts.GCENodeTest` option this test case would probably still pass on other providers.

My refactoring proposal is aimed at getting much greater test coverage for GCENode/CSR/GCR/AR.

nan-yu: Do we need to run all git tests for CSR with `gcenode`? I agree that we can get much greater test coverage for `gcenode` and CSR, but a lot of testcases validate features that are irrelevant with the auth type. Plus, `gcenode` requires WI to be disabled. If we run all git tests, we'll need another set of GKE clusters with WI disabled.

sdowell: I think my confusion comes from the proposed changes to the test scaffolding, for example in `RootSyncObjectV1Beta1`. That scaffolding is generic and is used by all tests - so by moving the CSR handling logic in there it seems to imply that any generic git test should be runnable on CSR (with or without gcenode). There's not really anything special about this `TestGCENode` testcase, presumably you could exercise similar functionality with any of the other generic git-sync tests.

That's not to say I'm opposed to moving this logic to generic scaffolding - I think this makes the tests more flexible/intuitive. I agree that we probably don't need to run all of the tests for CSR/gcenode, but I'm trying to think of a more intuitive way to structure this. Maybe the `gcenode` job could run a subset of tests by selecting a test group (e.g. `sync-source`)?

nan-yu: Just noticed that the current `gcenode` job only runs git test with CSR and oci with AR, but not helm with AR: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml#L330.

Instead of defining a separate test group, we can just use the `--gcenode` flag and `ntopts.GCENodeTest` to select the gcenode tests exclusively on gcenode cluster with WI disabled.
e.g.
```
- <<: *config-sync-standard-job
  name: kpt-config-sync-standard-regular-gcenode
  annotations:
    testgrid-dashboards: googleoss-kpt-config-sync-main
    testgrid-tab-name: standard-regular-gcenode
  spec:
    <<: *config-sync-standard-job-spec
    containers:
    - <<: *config-sync-standard-job-container
      args:
      - 'GCP_SUBNETWORK=prow-e2e-subnetwork-usw1-11'
      - 'GKE_RELEASE_CHANNEL=regular'
      - 'E2E_CLUSTER_PREFIX=standard-regular-gcenode'
      - 'E2E_NUM_CLUSTERS=1'
      - 'E2E_GIT_PROVIDER=csr'
      - 'E2E_ARGS=--gcenode'
```

sdowell: AFAICT that job should be running on helm with AR (at least, the `E2E_HELM_PROVIDER=helm` option is set).

To be clear I'm not suggesting that we define a new test group. Since this change is originating from the parent PR, I think we should move this discussion to that PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1160#discussion_r1525530326

nan-yu: > AFAICT that job should be running on helm with AR (at least, the E2E_HELM_PROVIDER=helm option is set).

The current prow config uses a regex to filter  tests `-run=TestGCENode`: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml#L330C29-L330C45. Hence, the Helm test with AR is not selected. We should fix that.

> To be clear I'm not suggesting that we define a new test group. Since this change is originating from the parent PR, I think we should move this discussion to that PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1160#discussion_r1525530326

Replied your comments in the other PR.

e2e/testcases/gcenode_test.go
```diff
@@ -71,9 +73,23 @@ func TestGCENodeCSR(t *testing.T) {
 		}
 	}`)
 
-	err := nt.WatchForAllSyncs(nomostest.WithRootSha1Func(nomostest.RemoteRootRepoSha1Fn),
+	nt.T.Log("Add the namespace-repo directory to RepoSync's repo")
+	repoSync := fake.RepoSyncObjectV1Beta1(testNs, configsync.RepoSyncName)
+	repoSyncRef := nomostest.RepoSyncNN(testNs, configsync.RepoSyncName)
+	nt.Must(nt.NonRootRepos[repoSyncRef].Copy("../testdata/hydration/namespace-repo", "."))
```
sdowell, line 79: I'm curious about the decision to add hydration to these test cases. My understanding was that the primary motivation of this change is to add test coverage for more auth types with RepoSync. Is there also intent to add more test coverage of hydration?

sdowell: Oh I see, this test was already using hydration for RootSync. This seems less unexpected to me in that case

nan-yu: Agreed these tests don't require hydration. Historically, we chose an existing CSR repo, which happened to be a DRY repo. Now we can create the CSR repo in a hermetic way, it could be any repo.

e2e/testcases/workload_identity_test.go
```diff
@@ -53,168 +63,234 @@ import (
...
-			crossProject:  false,
-			sourcePackage: "hydration/kustomize-components",
-			sourceType:    v1beta1.GitSource,
-			gsaEmail:      gitproviders.CSRReaderEmail(),
-			rootCommitFn:  nomostest.RemoteRootRepoSha1Fn,
-			requireCSR:    true,
+			name:         "Authenticate to Git repo on CSR with GKE WI",
+			fleetWITest:  false,
+			crossProject: false,
+			rootSrcCfg:   sourceConfig{pkg: "hydration/kustomize-components", dir: "kustomize-components", commitFn: nomostest.RemoteRootRepoSha1Fn},
```
sdowell, line 84: nit: I expected `dir: "."` here. It looks like this is copying the nested directory over. Is that intentional? I feel this would be more intuitive if the contents were placed into the top level directory of the repo

e2e/testcases/workload_identity_test.go
```diff
@@ -267,105 +344,236 @@ func TestWorkloadIdentity(t *testing.T) {
...
+	for _, mutator := range mutators {
+		mutator(repoSyncHelm)
+	}
+	if err = nt.KubeClient.Apply(repoSyncHelm); err != nil {
+		return nil, err
+	}
+	return chart, nil
+}
+
+func rootSyncUsingOCI(nt *nomostest.NT, rsRef types.NamespacedName, sc sourceConfig, mutators ...rootSyncMutator) (rsyncValidateMeta, error) {
```
sdowell, line 515: nit: I think the function names here (`r*SyncUsing*`) are a bit imprecise, particularly since they are scoped to the `testcases` package. That being said, I'm not really sure what to name these given that they are making several assumptions that seem pretty specific to this test case

I think something like `SetupRootSyncUsingOCIDelegatedControl` might more accurately describe the behavior, although even then the inputs/outputs still seem fairly bespoke to this test

e2e/testcases/workload_identity_test.go
nan-yu: It is intentional. The default sync directory is not `.`, but `acme`:  https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/gitproviders/repository.go#L61.

We could remove existing contents under `acme` and then copy all contents under `kustomize-components` to the `acme` directory. It feels like clearer to use a different sync directory.

e2e/testcases/workload_identity_test.go
```diff
@@ -267,105 +344,236 @@ func TestWorkloadIdentity(t *testing.T) {
...
-				Namespace: configsync.ControllerNamespace,
-				Name:      core.RootReconcilerName(rs.Name),
-			}
-			nt.T.Log("Validate the GSA annotation is added to the RootSync's service account")
+			rootReconcilerName := core.RootReconcilerName(rootSync.Name)
+			nsReconcilerName := core.NsReconcilerName(repoSync.Namespace, repoSync.Name)
+			nt.T.Log("Validate the GSA annotation is added to the RSync's service accounts")
 			require.NoError(nt.T,
-				nt.Watcher.WatchObject(kinds.ServiceAccount(), ksaRef.Name, ksaRef.Namespace, []testpredicates.Predicate{
+				nt.Watcher.WatchObject(kinds.ServiceAccount(), rootReconcilerName, configsync.ControllerNamespace, []testpredicates.Predicate{
```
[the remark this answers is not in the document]
nan-yu: SG

e2e/testcases/workload_identity_test.go
sdowell: Ack, this seems reasonable

e2e/testcases/workload_identity_test.go
sdowell: SGTM

sdowell (approved):
/lgtm

PR #1164, opened 2024-03-13
Check for AdmissionWebhook during validation

tiffanny29631:
This change checks for the existence of Admission Webhook during validation, if not available then compute of declared fields is skipped and corresponding annotation is removed from objects.


tiffanny29631:
Change includes unit test that verifies the check. Working on e2e tests.

karlkfi:
Maybe I'm missing something obvious, but can you explain why this change is needed?

tiffanny29631:
re 

> Maybe I'm missing something obvious, but can you explain why this change is needed?

go/cs-stop-loading-openapi for 1.18. I'm converting the PR to WIP soon as it requires a bit more change. With the current solution the reposync has trouble accessing Webhook.

sdowell:
> With the current solution the reposync has trouble accessing Webhook.

For this particularly problem, the solution I would suggest is to have `reconciler-manager` watch for the Webhook and inject a parameter (e.g. environment variable) into the reconciler Deployment

tiffanny29631:
> > With the current solution the reposync has trouble accessing Webhook.
> 
> For this particularly problem, the solution I would suggest is to have `reconciler-manager` watch for the Webhook and inject a parameter (e.g. environment variable) into the reconciler Deployment

Exactly, thanks for the suggestion

cmd/reconciler/main.go
```diff
@@ -102,6 +102,8 @@ var (
 			configsync.NamespaceStrategyImplicit, configsync.NamespaceStrategyExplicit, configsync.NamespaceStrategyImplicit))
 
 	dynamicNSSelectorEnabled = flag.Bool("dynamic-ns-selector-enabled", util.EnvBool(reconcilermanager.DynamicNSSelectorEnabled, false), "")
+
+	webhookEnabled = flag.Bool("webhook-enabled", util.EnvBool(reconcilermanager.WebhookEnabled, false), "")
```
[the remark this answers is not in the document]
tiffanny29631: It's passed into reconciler.Options before reconciler.Run(reconciler.Options), unless there is another place I missed?

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -203,6 +211,14 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 	return controllerruntime.Result{}, nil
 }
 
+func (r *RepoSyncReconciler) checkAdmissionWebhook(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
tiffanny29631: Done

cmd/reconciler/main.go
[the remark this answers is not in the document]
nan-yu: Yes, the flag is passed into reconciler.Options, but its initial value is not set. It is always false with the current implementation.

When the reconciler-manager gets the value, it should pass the value to the reconciler via the `reconcilermanager.WebhookEnabled` env variable.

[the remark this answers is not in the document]
tiffanny29631: Done! Added in the latest push

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -426,6 +433,9 @@ func (r *RootSyncReconciler) Register(mgr controllerruntime.Manager, watchFleetM
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
 		// TODO: is it possible to watch with a label filter?
 		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}},
+			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRootSync),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
+		Watches(&source.Kind{Type: &admissionv1.ValidatingWebhookConfiguration{}},
 			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRootSync),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}))
```
nan-yu, line 440: You can't reuse the `mapObjectToRootSync` function here.
Basically, the function should map the `admission-webhook.configsync.gke.io` ValidatingWebhookConfiguration object to all RootSync objects.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -474,6 +481,9 @@ func (r *RepoSyncReconciler) Register(mgr controllerruntime.Manager, watchFleetM
 		// Watch RoleBindings in all namespaces, because RoleBindings are created
 		// in the namespace of the RepoSync. Only maps to existing RepoSyncs.
 		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}},
+			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRepoSync),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})).
+		Watches(&source.Kind{Type: &admissionv1.ValidatingWebhookConfiguration{}},
 			handler.EnqueueRequestsFromMapFunc(r.mapObjectToRepoSync),
```
nan-yu, line 487: You can't reuse the `mapObjectToRepoSync` function here.
Basically, the function should map the `admission-webhook.configsync.gke.io` ValidatingWebhookConfiguration object to all RepoSync objects.

You'll need a separate mapper.

pkg/reconcilermanager/controllers/rootsync_controller.go
tiffanny29631: Done, added a new map function.

pkg/reconcilermanager/controllers/reposync_controller.go
tiffanny29631: Done!

e2e/nomostest/webhook.go
```diff
@@ -44,39 +43,29 @@ func StopWebhook(nt *NT) {
 		return
 	}
 	webhookName := configuration.Name
-	webhookGK := "validatingwebhookconfigurations.admissionregistration.k8s.io"
+	webhookGK := "ValidatingWebhookConfiguration"
```
nan-yu, line 46: You can use `kinds.ValidatingWebhookConfiguration().Kind` instead.

e2e/nomostest/webhook.go
```diff
@@ -44,39 +43,29 @@ func StopWebhook(nt *NT) {
...
-
-	err = nt.Watcher.WatchObject(kinds.ValidatingWebhookConfiguration(),
-		webhookName, "",
-		[]testpredicates.Predicate{
-			testpredicates.HasAnnotation(metadata.WebhookconfigurationKey, metadata.WebhookConfigurationUpdateDisabled),
-		},
-		testwatcher.WatchTimeout(30*time.Second))
-	if err != nil {
-		nt.T.Fatal(err)
+	err := nt.KubeClient.Get(webhookName, "", &admissionv1.ValidatingWebhookConfiguration{})
```
nan-yu, line 48: No need to Get before Delete. You can delete the webhook config, and then check if the error is NOT_FOUND.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,84 @@ metadata:
...
-	if _, err := nt.Shell.Kubectl("replace", "-f", filepath.Join(nt.TmpDir, "webhook.yaml")); err != nil {
-		nt.T.Fatalf("failed to replace the admission webhook %v", err)
-	}
+	nomostest.StopWebhook(nt)
 
-	// Verify that the webhook is disabled.
-	if _, err := nt.Shell.Kubectl("delete", "ns", "hello"); err != nil {
-		nt.T.Fatalf("failed to run `kubectl delete ns hello` %v", err)
+	tg := taskgroup.New()
+	tg.Go(func() error {
```
nan-yu, line 171: Why do we need a taskgroup? It only runs a single validation and waits for the validation result.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,84 @@ metadata:
...
+		return nt.Validate(
```
nan-yu, line 172: The reconciler Deployment may not be reconciled immediately. A direct validation may result in test flakiness. You may want to watch for the update.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,84 @@ metadata:
...
+	})
+	if err := tg.Wait(); err != nil {
+		nt.T.Fatal(err)
 	}
-
-	// Remove the annotation for disabling webhook
-	if _, err := nt.Shell.Kubectl("annotate", "ValidatingWebhookConfiguration", "admission-webhook.configsync.gke.io", "configsync.gke.io/webhook-configuration-update-"); err != nil {
-		nt.T.Fatalf("failed to remove the annotation in the admission webhook %v", err)
+	nt.T.Logf("Check if the namespace resourceversion has changed after deleting Admission Webhook")
+	err := nt.Watcher.WatchObject(kinds.Namespace(), "hello", "",
```
nan-yu, line 183: Line 170 starts a taskgroup with a single task. You may want to starts another task to watch the Namespace.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,84 @@ metadata:
...
+		[]testpredicates.Predicate{
+			testpredicates.ResourceVersionNotEquals(nt.Scheme, version),
+		})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.T.Logf("Check if the test namespace has dropped the declared-field annotation")
+	if err := nt.Validate("hello", "", &corev1.Namespace{}, testpredicates.MissingAnnotation(metadata.DeclaredFieldsKey)); err != nil {
```
nan-yu, line 191: You can move this validation to line 183 with an additional predicate.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,84 @@ metadata:
...
+		nt.T.Fatal(err)
 	}
 
-	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespaces/test/ns.yaml", fake.NamespaceObject("test")))
-	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add another Namespace"))
-	if err := nt.WatchForAllSyncs(); err != nil {
+	helloNS = &corev1.Namespace{}
+	if err := nt.KubeClient.Get("hello", "", helloNS); err != nil {
```
nan-yu, line 196: The Get request is to get the resourceVersion. I don't think we care about the resourceVersion that much. I think validating the update of the declared-fields annotation should be good enough.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,84 @@ metadata:
...
+	nt.T.Logf("Check if the namespace resourceversion has changed after installing Admission Webhook")
+	err = nt.Watcher.WatchObject(kinds.Namespace(), "hello", "",
+		[]testpredicates.Predicate{
+			testpredicates.ResourceVersionNotEquals(nt.Scheme, version),
+		})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+	nt.T.Logf("Check if the test namespace has restored the declared-field annotation")
+	if err := nt.Validate("hello", "", &corev1.Namespace{}, testpredicates.HasAnnotationKey(metadata.DeclaredFieldsKey)); err != nil {
```
nan-yu, line 228: Similarly, this validation can be merged with `nt.Watcher.WatchObject`.

pkg/parse/namespace.go
```diff
@@ -88,11 +88,19 @@ func (p *namespace) parseSource(ctx context.Context, state sourceState) ([]ast.F
...
 		DynamicNSSelectorEnabled: false,
+		WebhookEnabled:           p.WebhookEnabled,
 	}
 	options = OptionsForScope(options, p.Scope)
 
 	objs, err = validate.Unstructured(ctx, p.Client, objs, options)
 
+	if !p.WebhookEnabled {
+		for _, obj := range objs {
+			klog.Infof("Removing %s annotation as Admission Webhook is disabled", metadata.DeclaredFieldsKey)
```
nan-yu, line 99: You may want to move this log outside of the for loop to avoid spamming the logs.

pkg/parse/root.go
```diff
@@ -142,6 +143,13 @@ func (p *root) parseSource(ctx context.Context, state sourceState) ([]ast.FileOb
 		objs, err = validate.Hierarchical(objs, options)
 	}
 
+	if !p.WebhookEnabled {
+		for _, obj := range objs {
+			klog.Infof("Removing %s annotation as Admission Webhook is disabled", metadata.DeclaredFieldsKey)
```
nan-yu, line 148: ditto

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -187,6 +188,12 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		r.logger(ctx).V(3).Info("Sync deletion timestamp detected")
 	}
 
+	enabled, err := r.isWebhookEnabled(ctx)
+	if err != nil {
+		return controllerruntime.Result{}, err
```
nan-yu, line 193: For consistency, we need to record a metric for the reconciliation error and add a log for the error.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -651,6 +661,14 @@ func (r *RepoSyncReconciler) mapSecretToRepoSyncs(secret client.Object) []reconc
 	return requests
 }
 
+func (r *RepoSyncReconciler) mapAdmissionWebhookToRepoSyncs(admissionWebhook client.Object) []reconcile.Request {
+	sRef := client.ObjectKeyFromObject(admissionWebhook)
+	if sRef.Name == "admission-webhook.configsync.gke.io" {
```
nan-yu, line 666: You can use the constant in https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/webhook/configuration/constants.go#L30.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -651,6 +661,14 @@ func (r *RepoSyncReconciler) mapSecretToRepoSyncs(secret client.Object) []reconc
 	return requests
 }
 
+func (r *RepoSyncReconciler) mapAdmissionWebhookToRepoSyncs(admissionWebhook client.Object) []reconcile.Request {
+	sRef := client.ObjectKeyFromObject(admissionWebhook)
```
nan-yu, line 665: You can use `admissionWebhook.GetName()` to replace `sRef.Name`.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -527,7 +537,7 @@ func (r *RepoSyncReconciler) mapMembershipToRepoSyncs() func(client.Object) []re
 			if apierrors.IsNotFound(err) {
 				klog.Info("Fleet Membership not found, clearing membership cache")
 				r.membership = nil
-				return r.requeueAllRepoSyncs()
+				return r.requeueAllRepoSyncs("membership")
```
nan-yu, line 540: Use `fleetMembershipName`.

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -543,11 +553,11 @@ func (r *RepoSyncReconciler) mapMembershipToRepoSyncs() func(client.Object) []re
 			return nil
 		}
 		r.membership = m
-		return r.requeueAllRepoSyncs()
+		return r.requeueAllRepoSyncs("membership")
```
nan-yu, line 556: ditto

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -158,6 +159,14 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		r.logger(ctx).V(3).Info("Sync deletion timestamp detected")
 	}
 
+	enabled, err := r.isWebhookEnabled(ctx)
+	if err != nil {
+		return controllerruntime.Result{}, err
+	}
+	if r.webhookEnabled != enabled {
+		r.webhookEnabled = enabled
```
nan-yu, line 167: Inconsistency between rootsync_controller.go and reposync_controller.go.
It's okay to assign without the equality check.

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -158,6 +159,14 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		r.logger(ctx).V(3).Info("Sync deletion timestamp detected")
 	}
 
+	enabled, err := r.isWebhookEnabled(ctx)
+	if err != nil {
+		return controllerruntime.Result{}, err
```
nan-yu, line 164: ditto for the metric and logs.

e2e/testcases/admission_test.go
tiffanny29631: Would watchForCurrentStatus work for this check? I'm using it just to give enough head room for the reconciliation, similar to checking the resource version change in the test namespace - only check for the annotation _after_ the version is updated.

e2e/testcases/admission_test.go
tiffanny29631: Adding the annotation check as another task would fail straight away, I'm not sure if `watch` fits better for this case, but waiting for the new root-reconciler to come up -> check test object updated -> check annotation is gone seems necessary

e2e/nomostest/webhook.go
tiffanny29631: Done

e2e/nomostest/webhook.go
tiffanny29631: Done

pkg/parse/namespace.go
tiffanny29631: Done

pkg/parse/root.go

pkg/reconcilermanager/controllers/reposync_controller.go
tiffanny29631: Done

pkg/reconcilermanager/controllers/reposync_controller.go
tiffanny29631: Done

pkg/reconcilermanager/controllers/reposync_controller.go
tiffanny29631: Done

e2e/testcases/admission_test.go
nan-yu: > Would watchForCurrentStatus work for this check?

`watchForCurrentStatus` doesn't work because the reconciler Deployment may not be updated yet. Also it doesn't allow you to pass in a predicate.

> I'm using it just to give enough head room for the reconciliation

Even though the new reconciler Deployment is current, it is not guaranteed that the parse-apply-watch loop has completed. And `nt.WatchForAllSyncs` doesn't help either because the commit SHA is the same. 

> similar to checking the resource version change in the test namespace - only check for the annotation after the version is updated.

The resource version can be changed by various reasons. It is possible that the resource version changed, but the version has not been updated yet. I would suggest watch for the version update directly.

nan-yu: So for the reconciler Deployment, you can actually have

```
tg.Go(func() error {
		predicates := []testpredicates.Predicate{
			testpredicates.StatusEquals(nt.Scheme, kstatus.CurrentStatus),
			testpredicates.DeploymentMissingEnvVar(reconcilermanager.Reconciler, reconcilermanager.WebhookEnabled),
		}
		return nt.Watcher.WatchObject(kinds.Deployment(),
			core.RootReconcilerName(configsync.RootSyncName), configsync.ControllerNamespace, predicates)
	})
```

For the test Namespace, you can have 
```
tg.Go(func() error {
		return nt.Watcher.WatchObject(kinds.Namespace(), "hello", "",
			[]testpredicates.Predicate{
				testpredicates.MissingAnnotation(metadata.DeclaredFieldsKey),
			})
	})
```

tiffanny29631: Done! Thanks for the explanation!

e2e/testcases/admission_test.go
tiffanny29631: Using watcher now.

e2e/testcases/admission_test.go
tiffanny29631: Resolved in earlier comment.

e2e/testcases/admission_test.go
tiffanny29631: Done

e2e/testcases/admission_test.go
```diff
@@ -141,55 +147,96 @@ metadata:
...
+	if err := nomostest.InstallWebhook(nt); err != nil {
 		nt.T.Fatal(err)
 	}
-
 	nomostest.WaitForWebhookReadiness(nt)
 
-	// Verify that the webhook is now enabled
-	if _, err := nt.Shell.Kubectl("delete", "ns", "test"); err == nil {
-		nt.T.Fatal("got `kubectl delete ns hello` success, want return err")
+	nt.T.Logf("Check declared-fields annotation is missing")
```
nan-yu, line 211: "Check declared-fields annotation is missing" -> "Check declared-fields annotation is re-populated"

e2e/testcases/admission_test.go
```diff
@@ -141,55 +147,96 @@ metadata:
...
+			testpredicates.StatusEquals(nt.Scheme, kstatus.CurrentStatus),
+			testpredicates.DeploymentMissingEnvVar(reconcilermanager.Reconciler, reconcilermanager.WebhookEnabled),
+		}
+		return nt.Watcher.WatchObject(kinds.Deployment(),
+			core.RootReconcilerName(configsync.RootSyncName), configsync.ControllerNamespace, predicates)
+	})
+	tg.Go(func() error {
+		return nt.Watcher.WatchObject(kinds.Namespace(), "hello", "",
+			[]testpredicates.Predicate{
+				testpredicates.MissingAnnotation(metadata.DeclaredFieldsKey),
```
nan-yu, line 181: It may be worth adding a validation for the presence of the annotation before stopping the webhook, so we know the annotation is removed instead of missing initially.

pkg/parse/namespace.go
```diff
@@ -88,11 +88,19 @@ func (p *namespace) parseSource(ctx context.Context, state sourceState) ([]ast.F
...
 		AllowAPICall:             false,
 		DynamicNSSelectorEnabled: false,
+		WebhookEnabled:           p.WebhookEnabled,
 	}
 	options = OptionsForScope(options, p.Scope)
 
 	objs, err = validate.Unstructured(ctx, p.Client, objs, options)
 
+	if !p.WebhookEnabled {
+		klog.Infof("Removing %s annotation as Admission Webhook is disabled", metadata.DeclaredFieldsKey)
```
nan-yu, line 98: I would suggest making the log at level 3. Otherwise, the log would show up for every sync when the webhook is disabled.

pkg/parse/run.go
```diff
@@ -456,11 +457,20 @@ func parseSource(ctx context.Context, p Parser, trigger string, state *reconcile
 	}
 
 	start := time.Now()
-	objs, sourceErrs := p.parseSource(ctx, state.cache.source)
+	var sourceErrs status.MultiError
+
+	objs, errs := p.parseSource(ctx, state.cache.source)
+	if !p.options().WebhookEnabled {
+		for _, obj := range objs {
+			klog.Infof("Removing %s annotation as Admission Webhook is disabled", metadata.DeclaredFieldsKey)
```
nan-yu, line 465: Moving it outside of the for loop and make it log at level 3.

pkg/parse/root.go
```diff
@@ -142,6 +143,13 @@ func (p *root) parseSource(ctx context.Context, state sourceState) ([]ast.FileOb
 		objs, err = validate.Hierarchical(objs, options)
 	}
 
+	if !p.WebhookEnabled {
+		klog.Infof("Removing %s annotation as Admission Webhook is disabled", metadata.DeclaredFieldsKey)
```
nan-yu, line 147: klog.V(3).Infof

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -187,6 +189,13 @@ func (r *RepoSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		r.logger(ctx).V(3).Info("Sync deletion timestamp detected")
 	}
 
+	enabled, err := r.isWebhookEnabled(ctx)
+	if err != nil {
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+		return controllerruntime.Result{}, err
```
nan-yu, line 195: return controllerruntime.Result{}, fmt.Errorf("failed to get the admission webhook configuration: %w", err).

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -158,6 +160,13 @@ func (r *RootSyncReconciler) Reconcile(ctx context.Context, req controllerruntim
 		r.logger(ctx).V(3).Info("Sync deletion timestamp detected")
 	}
 
+	enabled, err := r.isWebhookEnabled(ctx)
+	if err != nil {
+		metrics.RecordReconcileDuration(ctx, metrics.StatusTagKey(err), start)
+		return controllerruntime.Result{}, err
```
nan-yu, line 166: return controllerruntime.Result{}, fmt.Errorf("failed to get the admission webhook configuration: %w", err).

e2e/nomostest/webhook.go
```diff
@@ -44,39 +42,24 @@ func StopWebhook(nt *NT) {
 		return
 	}
 	webhookName := configuration.Name
-	webhookGK := "validatingwebhookconfigurations.admissionregistration.k8s.io"
+	webhookGK := kinds.ValidatingWebhookConfiguration().Kind
```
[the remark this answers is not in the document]
tiffanny29631: Cleaned up a bit.

e2e/nomostest/webhook.go
```diff
@@ -44,39 +42,24 @@ func StopWebhook(nt *NT) {
...
-			testpredicates.HasAnnotation(metadata.WebhookconfigurationKey, metadata.WebhookConfigurationUpdateDisabled),
-		},
-		testwatcher.WatchTimeout(30*time.Second))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
-
-	out, err = nt.Shell.Kubectl("delete", webhookGK, webhookName)
-	if err != nil {
+	out, err := nt.Shell.Kubectl("delete", webhookGK, webhookName)
```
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/testcases/admission_test.go
```diff
@@ -141,55 +147,96 @@ metadata:
...
-	}
-
-	// Remove the annotation for disabling webhook
-	if _, err := nt.Shell.Kubectl("annotate", "ValidatingWebhookConfiguration", "admission-webhook.configsync.gke.io", "configsync.gke.io/webhook-configuration-update-"); err != nil {
-		nt.T.Fatalf("failed to remove the annotation in the admission webhook %v", err)
+	//helloNS := &corev1.Namespace{}
+	//if err := nt.KubeClient.Get("hello", "", helloNS); err != nil {
+	//	nt.T.Fatal(err)
+	//}
+	//version := helloNS.ResourceVersion
```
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/testcases/admission_test.go
```diff
@@ -141,55 +147,96 @@ metadata:
...
-		nt.T.Fatalf("failed to create a tmp file %v", err)
-	}
-
-	// Recreate the admission webhook
-	if _, err := nt.Shell.Kubectl("replace", "-f", filepath.Join(nt.TmpDir, "webhook.yaml")); err != nil {
-		nt.T.Fatalf("failed to replace the admission webhook %v", err)
-	}
-
-	// Verify that the webhook is disabled.
-	if _, err := nt.Shell.Kubectl("delete", "ns", "hello"); err != nil {
```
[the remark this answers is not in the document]
tiffanny29631: Resumed the drift prevention check.

e2e/testcases/admission_test.go
tiffanny29631: Done!

pkg/parse/namespace.go
tiffanny29631: Just realized this segment is repeated in namespace.go, root.go and run.go, cleaned up so it's done in run.go once. Fixed log level.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,82 @@ metadata:
...
-
 	nt := nomostest.New(t, nomostesting.DriftControl)
 
+	// Test starts with Admission Webhook already installed
+	nomostest.WaitForWebhookReadiness(nt)
+
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespaces/hello/ns.yaml", fake.NamespaceObject("hello")))
-	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add Namespace"))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespaces/world/ns.yaml", fake.NamespaceObject("world")))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add Namespaces"))
```
nan-yu, line 158: Extra indentation needs to be removed.

e2e/testcases/admission_test.go
```diff
@@ -141,55 +148,82 @@ metadata:
...
-`)
-
 	nt := nomostest.New(t, nomostesting.DriftControl)
 
+	// Test starts with Admission Webhook already installed
+	nomostest.WaitForWebhookReadiness(nt)
+
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespaces/hello/ns.yaml", fake.NamespaceObject("hello")))
-	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add Namespace"))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespaces/world/ns.yaml", fake.NamespaceObject("world")))
```
nan-yu, line 157: Do we need a separate Namespace? When the `hello` Namespace is deleted without drift prevention, it should be recreated by the remediator.

nan-yu (commented):
Just 2 minor comments, otherwise look good.

e2e/testcases/admission_test.go
tiffanny29631: Done

e2e/testcases/admission_test.go
tiffanny29631: Reverted to `hello` only.

nan-yu (approved):
/lgtm

nan-yu:
@tiffanny29631 Can we unhold this PR?

PR #1167, opened 2024-03-15
removed ctrl.Log

mikebz:
old techdebt (b/264312480) to remove the ctrl.Log

### DIFF
--- a/cmd/reconciler-manager/main.go
+++ b/cmd/reconciler-manager/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


cmd/reconciler-manager/main.go
```diff
@@ -92,10 +93,10 @@ func main() {
 		setupLog.Error(err, "failed to build dynamic client")
 		os.Exit(1)
 	}
-	watchFleetMembership := fleetMembershipCRDExists(dynamicClient, mgr.GetRESTMapper())
+	watchFleetMembership := fleetMembershipCRDExists(dynamicClient, mgr.GetRESTMapper(), &setupLog)
```
tiffanny29631, line 96: Adding an argument here is not necessary as the function can use the global var declared on L49

mikebz: I believe L49 is declared within main, I have removed the setupLog from globals.  I don't think you can just grab variables up the stack, but maybe I am wrong?

tiffanny29631: Yeah you are right, overlooked the removal.

tiffanny29631 (approved):
/lgtm

PR #1170, opened 2024-03-21
Fix issue with helm not being executable

Camila-B:
This PR makes sure that helm is executable within the helm-sync container


nan-yu (approved):
/lgtm

PR #1172, opened 2024-03-22
Log ssh-keygen output on failure

karlkfi:
It seems like keygen is failing, but not erroring. Log the output if the file wasn't created, to help debug.


nan-yu (approved):
/lgtm

PR #1174, opened 2024-03-22
test: fix CI failures when running with --git-provider=csr

nan-yu:
https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2228 changes the default git provider from local to csr, which also changes the defalt git auth from ssh to gcpserviceaccount, which causes a bunch of testing failures and missing permissions.

This commit updates the tests to set the secret for non-csr git providers and also adds the missing IAM bindings.

### DIFF
--- a/e2e/testcases/git_sync_test.go
+++ b/e2e/testcases/git_sync_test.go
@@ -19,7 +19,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"kpt.dev/configsync/e2e/nomostest"
-	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/api/configmanagement"
 	"kpt.dev/configsync/pkg/api/configsync"
@@ -41,22 +40,13 @@ func TestMultipleRemoteBranchesOutOfSync(t *testing.T) {
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespaces/hello/ns.yaml", fake.NamespaceObject("hello")))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add Namespace"))
 
-	nt.T.Logf("Mitigation: set spec.git.branch to HEAD to pull the latest commit")
-	nomostest.SetGitBranch(nt, configsync.RootSyncName, "HEAD")
+	nt.T.Logf("Verify git-sync can pull the latest commit with the default branch and revision")
 	// WatchForAllSyncs validates RootSync's lastSyncedCommit is updated to the
 	// local HEAD with the DefaultRootSha1Fn function.
 	if err := nt.WatchForAllSyncs(); err != nil {
 		nt.T.Fatal(err)
 	}
 	if err := nt.Validate("hello", "", &corev1.Namespace{}); err != nil {
-		nt.T.Fatal(err)
-	}
-
-	nt.T.Logf("Verify git-sync can pull the latest commit with the default branch and revision")
-	nomostest.SetGitBranch(nt, configsync.RootSyncName, gitproviders.MainBranch)
-	// WatchForAllSyncs validates RootSync's lastSyncedCommit is updated to the
-	// local HEAD with the DefaultRootSha1Fn function.
-	if err := nt.WatchForAllSyncs(); err != nil {
 		nt.T.Fatal(err)
 	}
 
--- a/e2e/testcases/invalid_auth_test.go
+++ b/e2e/testcases/invalid_auth_test.go
@@ -15,21 +15,39 @@
 package e2e
 
 import (
+	"fmt"
 	"testing"
 
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/metrics"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
 	"kpt.dev/configsync/pkg/status"
+	"kpt.dev/configsync/pkg/testing/fake"
 )
 
 func TestInvalidAuth(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.ACMController)
 
-	// Update RootSync to sync from a github repo.
-	// The ssh key only works for test-git-server, not github, so it will get a permission error.
-	nomostest.SetGitRepo(nt, configsync.RootSyncName, "git@github.com:test/not-exist")
+	// Update RootSync to sync from a GitHub repo with an ssh key.
+	// The ssh key only works for test-git-server, not GitHub, so it will get a permission error.
+	rs := fake.RootSyncObjectV1Beta1(configsync.RootSyncName)
+	nt.MustMergePatch(rs, fmt.Sprintf(`{
+		"spec": {
+			"git": {
+				"repo": "%s",
+				"auth": "%s",
+				"secretRef": {
+					"name": "%s"
+				}
+			}
+		}
+	}`, "git@github.com:config-sync-examples/not-exist", configsync.AuthSSH, controllers.GitCredentialVolume))
+
+	if err := nomostest.SetupFakeSSHCreds(nt, rs.Kind, nomostest.RootSyncNN(rs.Name), configsync.AuthSSH, controllers.GitCredentialVolume); err != nil {
+		nt.T.Fatal(err)
+	}
 	nt.WaitForRootSyncSourceError(configsync.RootSyncName, status.SourceErrorCode, "Permission denied")
 
 	rootSyncNN := nomostest.RootSyncNN(configsync.RootSyncName)
--- a/e2e/testcases/proxy_test.go
+++ b/e2e/testcases/proxy_test.go
@@ -27,6 +27,7 @@ import (
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/reconcilermanager/controllers"
 	"kpt.dev/configsync/pkg/testing/fake"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -55,6 +56,9 @@ func TestSyncingThroughAProxy(t *testing.T) {
 	nt.T.Log("Set auth type to cookiefile")
 	nt.MustMergePatch(rs, `{"spec": {"git": {"auth": "cookiefile"}}}`)
 	nt.T.Log("Verify the secretRef error")
+	if err = nomostest.SetupFakeSSHCreds(nt, rs.Kind, nomostest.RootSyncNN(rs.Name), configsync.AuthCookieFile, controllers.GitCredentialVolume); err != nil {
+		nt.T.Fatal(err)
+	}
 	nt.WaitForRootSyncStalledError(rs.Namespace, rs.Name, "Validation", `git secretType was set as "cookiefile" but cookie_file key is not present`)
 
 	nt.T.Log("Set auth type to token")
--- a/e2e/testinfra/terraform/modules/service_account/service_account.tf
+++ b/e2e/testinfra/terraform/modules/service_account/service_account.tf
@@ -24,15 +24,58 @@ resource "google_service_account" "gcp_sa" {
 }
 
 resource "google_service_account_iam_member" "k8s_sa_binding" {
+  for_each = toset([
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/root-reconciler]", # The default RootSync used in e2e test.
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns]", # The default RepoSync used in e2e test.
+    # RootSync and RepoSync used in TestNomosStatusNameFilter.
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/root-reconciler-crontab-sync]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-bookinfo-bookinfo-repo-sync-18]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-bookinfo-crontab-sync-12]",
+    # RootSync and RepoSync used in TestMultiSyncs_Unstructured_MixedControl.
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/root-reconciler-rr1]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/root-reconciler-rr2]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/root-reconciler-rr3]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns-nr1-3]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns-nr2-3]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns-nr3-3]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns-nr4-3]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns-nr5-3]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-ns-2-nr1-3]",
+    # RootSync used in TestConflictingDefinitions_RootToRoot.
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/root-reconciler-root-test]",
+    # RepoSync used in TestConflictingDefinitions_RootToRoot.
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns-rs-test-1-9]",
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns-rs-test-2-9]",
+    # RepoSync used in TestNamespaceRepo_Delegated, TestDeleteRepoSync_Delegated_AndRepoSyncV1Alpha1, TestDeleteRepoSync_Centralized_AndRepoSyncV1Alpha1, TestManageSelfRepoSync, TestDeleteNamespaceReconcilerDeployment, TestReconcilerManagerRootSyncCRDMissing.
+    "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-bookstore]",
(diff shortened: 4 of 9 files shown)


e2e/testinfra/terraform/modules/service_account/service_account.tf
```diff
@@ -24,15 +24,50 @@ resource "google_service_account" "gcp_sa" {
 }
 
 resource "google_service_account_iam_member" "k8s_sa_binding" {
+  for_each = toset([
```
karlkfi, line 27: Do we really want to be managing these with terraform?
It might be better to manage them with K8s RBAC in the e2e framework. That way it auto-applies and can be test-specific.

nan-yu: I tried that approach earlier in a different PR, but then got pushed back because we don't want to grant `e2e-test-runner` permissions to create IAM bindings.

e2e/testcases/proxy_test.go
```diff
@@ -55,6 +58,25 @@ func TestSyncingThroughAProxy(t *testing.T) {
 	nt.T.Log("Set auth type to cookiefile")
 	nt.MustMergePatch(rs, `{"spec": {"git": {"auth": "cookiefile"}}}`)
 	nt.T.Log("Verify the secretRef error")
+	secret := fake.SecretObject(controllers.GitCredentialVolume, core.Namespace(configsync.ControllerNamespace))
```
karlkfi, line 61: could use some comments here to explain the branching code

nan-yu: It is also needed by other tests, so I moved it to a function in config_sync.go with more descriptions.

nan-yu:
Finally got all failed tests pass on my test clusters.

e2e/testinfra/terraform/modules/service_account/service_account.tf
karlkfi: Why does it need IAM tho? Can't they just be RBAC?

nan-yu: These IAM bindings allow R*Sync KSAs to be able to impersonate a GSA at the project level.

We can't mange them with K8s RBAC at the GKE cluster level.

nan-yu: If we really want them to be auto-applied and test-specific, we have to create the IAM bindings in the test. Then the prow job (`e2e-test-runner`) needs to have privileged permissions to create IAM bindings, which we want to avoid.

e2e/testcases/git_sync_test.go
```diff
@@ -41,8 +40,7 @@ func TestMultipleRemoteBranchesOutOfSync(t *testing.T) {
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/namespaces/hello/ns.yaml", fake.NamespaceObject("hello")))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add Namespace"))
 
-	nt.T.Logf("Mitigation: set spec.git.branch to HEAD to pull the latest commit")
```
nan-yu, line 44: The mitigation part is removed because the issue was fixed in git-sync, so we don't need the mitigation.
Also, the mitigation doesn't work for CSR because CSR's default remote branch is `master`, and fetching the HEAD of a non-existent remote branch returns an error.

For my own record:

##  `.git/refs` folder

### CSR
```
$ tree cs-e2e-cs-e2e-nanyu-cluster-1-config-management-system-516336e8/.git/refs/
cs-e2e-cs-e2e-nanyu-cluster-1-config-management-system-516336e8/.git/refs/
├── heads
│   └── main
└── tags

3 directories, 1 file
```

### GitLab

```
$ tree config-management-system-root-test-7eaaa576-efa9-44fb-8912-b66890e572eb/.git/refs/
config-management-system-root-test-7eaaa576-efa9-44fb-8912-b66890e572eb/.git/refs/
├── heads
│   └── main
├── remotes
│   └── origin
│       └── HEAD
└── tags

5 directories, 2 files
```

## `git branch -a`

### CSR

```
$ git -C cs-e2e-cs-e2e-nanyu-cluster-1-config-management-system-516336e8/ branch -a
* main
  remotes/origin/main
  remotes/origin/upstream/main
```

### GitLabl

```
$ git -C config-management-system-root-test-7eaaa576-efa9-44fb-8912-b66890e572eb/ branch -a
* main
  remotes/origin/HEAD -> origin/main
  remotes/origin/main
```

e2e/testinfra/terraform/modules/service_account/service_account.tf
karlkfi: oh, I see. These are for workloadIdentityUser, not cluster permissions. My mistake.

It seems unfortunate that we need to modify this TF and manually apply it any time we want to test in a new namespace or with a new RootSync.

Maybe we should have a pool of these so we don't need so many IAM bindings...

karlkfi: This means we're giving all the RootSyncs and RepoSyncs permission to impersonate the same GSA, with whatever IAM that GSA has. That's not really how customers are going to use WI...

It might be safer to follow the principle of least privilege and only give each test the permissions it needs. Tho that would mean giving the test GSA permission to modify IAM in the project.

karlkfi: Personally, I think giving the test GSA IAM to modify IAM is probably fine, as long as it's only scoped to the test project. It already has permission to do all sorts of other expensive things. Giving it project admin won't make it appreciably more risky at this point.

e2e/testinfra/terraform/modules/service_account/service_account.tf
```diff
@@ -24,15 +24,58 @@ resource "google_service_account" "gcp_sa" {
...
```
sdowell, line 27: I hadn't considered the implications of needing to create WI bindings for each permutation of reconciler KSAs across the test suite. This would probably be a bit easier to manage if we implemented more standardized naming of Namespaces/RSyncs across the tests. However, this has me reconsidering whether we should take a step back and go with your original proposal to have a one-off job for testing CSR+WI.

karlkfi: I suspect it's fine to just give the e2e test GSA project admin and have the tests set up a new GSA with WI IAM for every new RootSync/RepoSync.

There's not really an appreciable difference in risk between having GCE admin and project admin. Either one can exfiltrate secrets from the project and spend lots of money.

nan-yu: > whether we should take a step back and go with your original proposal to have a one-off job for testing CSR+WI.

My original concern about change the default git provider to CSR are longer running duration and potential failures. This PR will fix my second concern. For the longer running duration. I picked two CI jobs and compared the duration. Surprisingly, the CSR running duration is slightly shorter than the local test-git-server with a subset of testcases.

With these two concerns being addressed and the benefit of more test coverage with CSR, plus one less job (test cluster) to maintain, I'm inclined to go with the current job configs.

Agreed this long WI bindings list doesn't look nice, but this is a one-off setup for the test projects.

We can create a pool of these WI bindings, or implement a standardized naming of Namespaces/RSyncs, but I would think that requires more changes for the test structure without a clear benefit.

nan-yu: Testing running duration for CSR and the local test-git-server:


Testcase | CSR duration | local git-server duration
-- | -- | --
Total Duration | 13846.91 | 14168.28
TestCreateAPIServiceAndEndpointInTheSameCommit | 589.88 | 663.64
TestSyncDeploymentAndReplicaSet | 243.54 | 211.82
TestDisableWebhookConfigurationUpdate | 78.5 | 90.37
TestAddUpdateDeleteLabel | 151.39 | 118.5
TestAdmission | 55.56 | 69.15
TestObjectInCMSNamespace | 47.01 | 59.18
TestAddUpdateDeleteAnnotation | 128.31 | 128.71
TestReconcilerResilientToFlakyAPIService | 60.12 | 72.28
TestNamespaceDefaultCanBeManaged | 181.81 | 194.46
TestAdoptClientSideAppliedResource | 77.05 | 59.31
TestDeleteManagedFieldsWithIgnoreMutationAnnotation | 68.96 | 81.14
TestNamespacePolicyspaceConversion | 84.15 | 95
TestNamespaceGatekeeperSystemCanBeManaged | 143.33 | 186.4
TestNamespaceRepo_Delegated | 266.66 | 317.67
TestRolebindingsUpdated | 61.95 | 89.33
TestConflictingDefinitions_RootToRoot | 185.69 | 217.33
TestMultipleVersions_RoleBinding | 213.52 | 250.42
TestNamespaceGarbageCollection | 67.22 | 95.22
TestNamespaceKubePublicCanBeManaged | 182.42 | 135.18
TestMultipleVersions_CustomResourceV1 | 121.25 | 129.14
TestInlineClusterSelectorFormat | 338.89 | 480.22
TestMultipleVersions_CustomResourceV1Beta1 | 39 | 40
TestSyncFromNomosHydrateOutputYAMLDir | 48.01 | 60.18
TestSyncFromNomosHydrateUnstructuredOutputWithClusterSelectorJSONFlat | 78.29 | 61.15
TestNamespaceKubeSystemCanBeManaged | 138.45 | 134.51
TestSyncFromNomosHydrateOutputJSONDir | 47.62 | 45.42
TestSyncLabelsAndAnnotationsOnKubeSystem | 128.73 | 130.58
TestObjectReactsToChangeInLegacyClusterSelector | 149.79 | 190.49
TestClusterSelectorForCRD | 238.56 | 309.73
TestReconcilerFinalizer_Orphan | 50.51 | 62.16
TestSyncFromNomosHydrateUnstructuredOutputWithClusterSelectorYAMLFlat | 47.86 | 59.65
TestImporterIgnoresNonSelectedCustomResource | 89.08 | 67.08
TestReconcilerFinalizer_Foreground | 52.2 | 62.42
TestDoNotRemoveManagedByLabelExceptForConfigManagement | 47.35 | 46.57
TestClusterSelectorOnNamespace | 320.82 | 327.82
TestClusterSelectorAnnotationConflict | 92.78 | 70.29
TestMustRemoveCustomResourceWithDefinitionV1 | 273.03 | 191.2
TestAddAndRemoveCustomResourceV1 | 272.56 | 248.87
TestRevertClusterRole | 60.22 | 69.04
TestClusterSelectorOnObject | 216.58 | 345.36
TestClusterRoleLifecycle | 129.06 | 130.26
TestDontDeleteAllNamespace | 216.01 | 275.65
TestChangeCustomResourceDefinitionSchema | 116.44 | 147.39
TestObjectReactsToChangeInInlineClusterSelector | 180.64 | 130.34
TestTargetingDifferentResourceQuotasToDifferentCluster | 126.41 | 119.31
TestDeclareImplicitNamespace | 184.13 | 133.94
TestMustRemoveCustomResourceWithDefinitionV1Beta1 | 40.28 | 40.03
TestMustRemoveUnManagedCustomResourceV1 | 118.79 | 130.45
TestPublicHelm | 254.01 | 263.07
TestAddUpdateNamespaceScopedCRDV1 | 167.27 | 239.5
TestAddUpdateRemoveClusterScopedCRDV1 | 107.31 | 239.36
TestDeclaredFieldsPod | 109.25 | 121.06
TestAddAndRemoveCustomResourceV1Beta1 | 38.96 | 38.08
TestConstraintTemplateAndConstraintInSameCommit | 112.55 | 186.45
TestSyncUpdateCustomResource | 40.24 | 35.95
TestOtelCollectorDeployment | 208.87 | 197.63
TestLargeCRD | 80.26 | 78.22
TestCRDDeleteBeforeRemoveCustomResourceV1 | 120.29 | 147.81
TestAddUpdateRemoveClusterScopedCRDV1Beta1 | 40.05 | 39.11
TestPublicOCI | 105.05 | 104.22
TestAddUpdateNamespaceScopedCRDV1Beta1 | 38.34 | 39.75
TestNamespaceSelectorUnstructuredFormat | 344.55 | 443.09
TestHelmConfigMapOverride | 129.44 | 127.46
TestHelmWatchConfigMap | 120.24 | 115.33
TestSwitchFromGitToOciDelegated | 117.82 | 108.71
TestHydrateRemoteResource | 541.57 | 209.03
TestHelmLatestVersion | 109.85 | 111.88
TestHelmVersionRange | 120.75 | 118.42
TestHelmDefaultNamespace | 76.48 | 73.29
TestOtelCollectorGCMLabelAggregation | 78.78 | 75.64
TestNamespaceStrategyMultipleRootSync | 250.99 | 186.19
TestHydrateKustomizeComponent | 235.95 | 160.76
TestNamespaceStrategy | 149.22 | 173.58
TestHelmARTokenAuth | 136.92 | 138.1
TestNamespaceSelectorHierarchicalFormat | 94.05 | 76.4
TestHydrateHelmComponent | 275.94 | 239.93
TestHelmEmptyChart | 64.98 | 64.92
TestNamespaceLabelAndAnnotationLifecycle | 159.14 | 158.5
TestHydrateHelmOverlay | 191.63 | 164.14
TestManagementDisabledNamespace | 234.27 | 225.1
TestDeclareNamespace | 59.89 | 58.41
TestDependencyWithReconciliation | 918.15 | 914.68
TestNamespaceExistsAndDeclared | 92.06 | 66.82
TestReconcileFinalizerReconcileTimeout | 175.7 | 163.51
TestNamespaceEnabledAnnotationNotDeclared | 48.75 | 47.48
TestDeleteRepoSync_Delegated_AndRepoSyncV1Alpha1 | 91.38 | 80.31
TestSurfaceFightError | 236.71 | 249.63
TestPreventDeletionNamespace | 209.34 | 219.47
TestReconcilerManagerRootSyncCRDMissing | 188.49 | 166.59
TestSyncFailureAfterSuccessfulSync | 121.01 | 126.34

- Local test-git-server: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-rapid/1770850582882422784
- CSR: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-rapid/1771937766527995904

e2e/testcases/invalid_auth_test.go
```diff
@@ -15,21 +15,39 @@
...
+		"spec": {
+			"git": {
+				"repo": "%s",
+				"auth": "%s",
+				"secretRef": {
+					"name": "%s"
+				}
+			}
+		}
+	}`, "git@github.com:test/not-exist", configsync.AuthSSH, controllers.GitCredentialVolume))
```
sdowell, line 46: nit: Perhaps we should use an org we control to reduce chance of unexpected changes (e.g. `GoogleContainerTools/not-exist`)

e2e/testinfra/terraform/modules/service_account/service_account.tf
karlkfi: CSR being faster is probably exclusively because it doesn't need to spin up the local hit server Deployment and wait for it to be ready. 

My primary concern here is maintainability tho. With this TF config, every time you add a new test you're going to need to update this TF and apply it. We're gonna forget and it's gonna break CI. Plus, when we try to run in a different project we're gonna forget to apply this and it won't work. Or worse we may accidentally apply it to the wrong project, as we've done before. 

Basically, this is more manual steps that will cause unexpected CI failures and make test changes cost more. 

I'd rather just give the e2e test runner project admin and have it manage any infra it needs.

e2e/testinfra/terraform/modules/service_account/service_account.tf
karlkfi: resolving because discussion is in the other thread

sdowell (commented):
lgtm with a nit. I'll give Karl a chance to review

/lgtm

e2e/testinfra/terraform/modules/service_account/service_account.tf
nan-yu: > CSR being faster is probably exclusively because it doesn't need to spin up the local git server Deployment and wait for it to be ready.

The above table shows the duration of each test case. The spin-up of the local git server Deployment is not included because it is already ready before each test case.

I think there're two questions here.
1. Whether to run CIs with `CSR` as the default Git provider, or with the local test-git-server. With the test duration and this PR, I think we all agree to pick CSR.
2. Whether to set up WI bindings in the TF config or give the e2e test runner project admin and have it manage any infra it needs. I would prefer to set up in the TF config because the e2e tests are running in parallel on different clusters. Some KSA are shared across multiple testcases. We may run into a case where one test case is creating the WI binding for a KSA, but the other test case is cleaning up it for the same KSA.

sdowell: Agreed that tests running in parallel against the same project makes it infeasible to manage workload identity IAM on a per-test basis

e2e/testcases/invalid_auth_test.go
nan-yu: done

e2e/testinfra/terraform/modules/service_account/service_account.tf
karlkfi: Can we name the GSAs differently for each parallel cluster so they dont overlap?

nan-yu: Then we need to manage the lifecycle of multiple GSAs, which is even more work than managing these WI bindings.
Moreover, the test running on which cluster is kind of random, so it would be tricky to assign a GSA for a test case.

karlkfi: > Then we need to manage the lifecycle of multiple GSAs, which is even more work than managing these WI bindings.

Yes, it would be a little more work, but not that much, and only up front work. It can be fully automated in the test frameworks so that the TF doesn't need to be updated/applied when new test changes are made later.

> Moreover, the test running on which cluster is kind of random, so it would be tricky to assign a GSA for a test case.

We just need to create the GSAs & IAM before the test and clean them up after, like we do with other test resources.

The name of the GSA can include the name of the test environment, the test name, and name of the namespace, and the name of the KSA. This will probably be too long, but that's ok because there's a standard pattern for shortening names in GCP: hash the full name and append the short hash to the end of the truncated name. That way the name is unique and reproducible.

nan-yu: So you're proposing creating and deleting a GSA for each test case? I'm wondering how much overhead it may add.
That also requires e2e-test-runner to be granted the `roles/iam.serviceAccountAdmin` role, which includes `iam.serviceAccounts.create` and `iam.serviceAccounts.setIamPolicy` permissions.

@sdowell Any concerns with the more privileged permissions? If you're also on board with the proposal, I can update this PR.

karlkfi: Yes, it will take a few seconds to provision and delete GSA and IAM, But probably not much long than it takes to set up the port-forwarding to the local git server.

karlkfi: Occurs to me that the extra time spent could be caused by the local got server being evicted and rescheduled, causing the port-forward proxy to block git calls until it's up again.

sdowell: > @sdowell Any concerns with the more privileged permissions? If you're also on board with the proposal, I can update this PR.

Yes I still have the same concerns regarding privilege escalation. My preference would be the current approach that @nan-yu proposed.

nan-yu: Discussed offline with @sdowell and @karlkfi . We decided to go with this proposal to fix the CI and then circle back to this later if it is too much pain to keep updating the TF configs for new KSAs.

karlkfi:
/lgtm

For the record, I objected to putting the IAM in TF because I believe it will make for more test failures and gotchas and more work for developers, but I was outvoted in favor of the principle of least privileged. Given that, this is fine, for now. We can circle back when it becomes an annoyance for the team.

e2e/testinfra/terraform/modules/service_account/service_account.tf
nan-yu: Still waiting for an LGTM and approval.

nan-yu:
> /lgtm
> 
> For the record, I objected to putting the IAM in TF because I believe it will make for more test failures and gotchas and more work for developers, but I was outvoted in favor of the principle of least privilege. Given that, this is fine, for now. We can circle back when it becomes an annoyance for the team.

Thanks for reviewing it. Can you also approve it?

karlkfi (approved):

PR #1175, opened 2024-03-25
Add retry with backoff on error to helm-sync

Camila-B:
b/321790360

### DIFF
--- a/cmd/helm-sync/main.go
+++ b/cmd/helm-sync/main.go
@@ -76,6 +76,16 @@ var (
 		"the password or personal access token to use for helm authantication")
 )
 
+func errorBackoff() wait.Backoff {
+	return wait.Backoff{
+		Duration: 1 * time.Second,
+		Factor:   2,
+		Steps:    math.MaxInt,
+		Jitter:   0.1,
+		Cap:      util.WaitTime(*flWait),
+	}
+}
+
 func main() {
 	utillog.Setup()
 	log := utillog.NewLogger(klogr.New(), *flRoot, *flErrorFile)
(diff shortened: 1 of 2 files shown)


cmd/helm-sync/main.go
```diff
@@ -108,6 +110,14 @@ func main() {
 
 	initialSync := true
 	failCount := 0
+
+	backoff := wait.Backoff{
```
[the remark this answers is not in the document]
Camila-B: Done

cmd/helm-sync/main.go
```diff
@@ -141,11 +151,15 @@ func main() {
 				os.Exit(1)
 			}
 
+			step := backoff.Step()
```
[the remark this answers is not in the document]
Camila-B: Done

sdowell:
> /label tide/merge-method-squash

JFYI squash is the default, so typically no need to add this label

cmd/helm-sync/main.go
```diff
@@ -108,6 +120,9 @@ func main() {
 
 	initialSync := true
 	failCount := 0
+	step := time.Duration(0)
```
sdowell, line 123: nit: no need to declare this variable up here, can be declared in the loop

Camila-B: Got it, thanks

cmd/helm-sync/main.go
```diff
@@ -141,11 +155,15 @@ func main() {
 				os.Exit(1)
 			}
 
+			step := backoff.Step()
+
 			failCount++
```
sdowell, line 160: This has some implications on the duration that this process will error loop, since it will exit once `failCount >= *flMaxSyncFailures`. This is because the sleep duration will be less for the first dozen or so iterations (until it reaches the cap). 

I think this is probably still a reasonable duration to spend error looping before exiting, so I don't expect this should cause any problems. WDYT @nan-yu?

nan-yu: Agreed. I don't think the abnormal exit duration matters that much if the helm-sync container is already in a failure state.

sdowell: SG, thanks for confirmation

sdowell (approved):
/lgtm

PR #1179, opened 2024-03-27
test: properly skip tests that use test wrapper

sdowell:
The tests that use the test wrapper directly instead of the nomostest entrypoint were not properly being skipped. This moves the skip logic to the shared entrypoint.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -57,7 +57,7 @@ const fileMode = os.ModePerm
 const NomosE2E = "nomos-e2e"
 
 // newOptStruct initializes the nomostest options.
-func newOptStruct(testName, tmpDir string, t nomostesting.NTB, ntOptions ...ntopts.Opt) *ntopts.New {
+func newOptStruct(testName, tmpDir string, ntOptions ...ntopts.Opt) *ntopts.New {
 	// TODO: we should probably put ntopts.New members inside of NT use the go-convention of mutating NT with option functions.
 	optsStruct := &ntopts.New{
 		Name:   testName,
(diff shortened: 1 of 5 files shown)


karlkfi:
When are you supposed to use nomostest.NewTestWrapper vs nomostest.New, and what's the difference?

sdowell:
> When are you supposed to use nomostest.NewTestWrapper vs nomostest.New, and what's the difference?

The main difference currently is that `nomostest.New` creates and/or connects to a k8s cluster and performs all of the cluster setup. These tests which use the test wrapper directly seem to skip those steps as an optimization since they don't interact with a k8s cluster at all.

I'm not attached to this interface/naming, the primary motivation is to make sure the test rules are applied consistently (e.g. skip flags and test feature flags). Another option could be to add an `ntopt` which skips cluster creation/setup, although I felt that may be more confusing since the `NT` struct that is returned is all about interacting with the test cluster

sdowell:
/assign @nan-yu 

FYI after we removed the test regex some of these tests are no longer being skipped. For example see https://testgrid.k8s.io/googleoss-kpt-config-sync-main#standard-rapid-latest-stress

nan-yu:
This fixes the skip logic for tests using the test wrapper, but tests using neither the test wrapper nor the nomostest entrypoint still get selected, e.g. `e2e/nomostests/gitproviders/cloud_source_repository_test.go`.

It is the only unit test in the `e2e` directory. We don't have a good way to skip them. Should we remove it from the e2e folder? It is okay for this specific test because it's simple, but may not work if we add more unit tests later.

nan-yu (approved):
/lgtm

PR #1180, opened 2024-03-27
test: set memory request and limit for stress test

sdowell:


nan-yu (approved):
/lgtm

e2e/testcases/stress_test.go
```diff
@@ -224,9 +228,34 @@ func TestStressLargeRequest(t *testing.T) {
...
+		MemoryLimit:   resource.MustParse("1500Mi"),
+	}
+	if *e2e.GKEAutopilot {
+		reconcilerOverride.MemoryRequest = resource.MustParse("1500Mi")
+	}
+	rootSync.Spec = v1beta1.RootSyncSpec{
+		SourceFormat: string(filesystem.SourceFormatUnstructured),
+		Git: &v1beta1.Git{
+			Repo:      "https://github.com/config-sync-examples/crontab-crs",
+			Branch:    "main",
```
karlkfi, line 243: is main default now with the latest git-sync? @nan-yu

nan-yu: The short answer is no.

`git-sync` v4 uses a different flag `--ref` to specify a branch, a tag or a commit hash. It also consolidates the legacy `--branch` and `--rev` flags: https://github.com/kubernetes/git-sync/blob/v4.2.1/main.go#L358.

If none of these flags are set, it fetches the HEAD of the default remote branch, which may or may not be `main`.

But in Config Sync, if none of these flags are set, we force to use `master` as the default branch of backward compatibility.

PR #1186, opened 2024-03-28
[chore] Bump Helm to v3.14.3-gke.1

tiffanny29631:


Makefile
```diff
@@ -60,7 +60,7 @@ KUSTOMIZE_VERSION := v5.3.0-gke.1
 KUSTOMIZE := $(BIN_DIR)/kustomize
 KUSTOMIZE_STAGING_DIR := $(OUTPUT_DIR)/third_party/kustomize
 
-HELM_VERSION := v3.13.3-gke.3
+HELM_VERSION := v3.14.3-gke.1
```
sdowell, line 63: The presubmits should fail and tell you this, but you'll need to also update the value here: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/hydrate/tool_util.go#L53

tiffanny29631: Done!

nan-yu (approved):

PR #1191, opened 2024-04-05
Fix a race condition on reconciler finalizer removal

nan-yu:
Previously, the reconciler is responsible for removing the reconciler finalizer from the RSync object, but if the reconciler fails to update RSync to remove the finalizer (e.g. the reconciler is deleted or the reconciler doesn't have permissions to update RSync), RSync would stuck in deletion forever.

This commit updates the reconciler-manager to remove the reconciler finalizer to unblock RSync deletion if RSync's Namespace is terminating. A side-effect of this update is that managed resouces couldn't be deleted even though the deletion propagation is set to Foreground, but since the reconciler couldn't function, managed resources couldn't be deleted anyway.

b/332940374


karlkfi:
/hold

I need to think about this carefully and I don't really have time today.

I think there's some problems with deletion ordering that might crop up in this situation. My general suggestion would be for users not to delete namespaces with RootSyncs in them before carefully deleting the RootSync, but I understand that might be counter intuitive or surprising since everyone takes namespace garbage collection for granted.

A potentially better solution might be to add webhook that blocks namespace deletion when a RepoSync is present. Or perhaps a finalizer on the Namespace that executes before the namespace garbage collector.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -617,10 +617,41 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
...
+		//   might get deleted prior to RootSync, resulting in the finalizer being
+		//   unable to be removed. Consequently, RootSync would become stuck in the
+		//   deletion process.
+		// - For RepoSync, if its Namespace is terminating, the RoleBinding for the
+		//   namespace reconciler could be deleted and fail to be re-created.
+		//   Therefore, the reconciler would lack the necessary permissions to
+		//   remove the reconciler finalize from the RepoSync object, leading to
+		//   RepoSync being stuck in deletion.
+		// Therefore, if the Namespace is terminating, the reconciler-manager
+		// deletes the reonciler finalizer.
```
janetkuo, line 641:
```suggestion
		// deletes the reconciler finalizer.
```

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -617,10 +617,41 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
...
+			r.logger(ctx).Error(err, "Get sync object Namespace failed")
+			return err
+		}
+		// The removal of the reconciler finalizer might encounter issues in both
+		// RootSync and RepoSync.
+		// - For RootSync, if its Namespace is terminating, the reconciler Deployment
+		//   might get deleted prior to RootSync, resulting in the finalizer being
+		//   unable to be removed. Consequently, RootSync would become stuck in the
+		//   deletion process.
+		// - For RepoSync, if its Namespace is terminating, the RoleBinding for the
```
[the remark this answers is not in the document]
nan-yu: I'm okay with applying this logic for RepoSync first, but I think it is also safe to apply to RootSync as well.
Agreed that it's reasonable to expect that things won't function correctly if removing the Config Sync namespace, but the presence of the finalizer would cause the RootSync and the system Namespace to get stuck in terminating, which leaves an unclean environment. Since things are already broken, it seems acceptable to me to make it break in a clean way. Plus, this is a very unusual edge case.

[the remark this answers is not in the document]
janetkuo: This sounds reasonable to me. My comment is mostly to address @karlkfi's concern. The RepoSync case is pretty straightforward (when the namespace is gone you don't actually need a finalizer to clean up things that live in the namespace), so at least that part won't have problems with deletion ordering crop up that @karlkfi is concerned about.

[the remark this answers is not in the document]
sdowell: I imagine there may be some cases where a user might prefer for the reconciler finalizer to complete before the namespace finalizer. The primary example that comes to mind is deletion ordering, where perhaps it is required that objects are deleted in the order specified by `depends-on`. In this case, the namespace finalizer will not honor the deletion ordering in the same way as the reconciler finalizer.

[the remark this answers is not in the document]
nan-yu: Honoring the deletion ordering is important when pruning resources in normal cases, but I'm not sure if the deletion ordering matters that much if everything is eventually deleted. The namespace finalizer will prevent objects from being removed until all items are removed from the corresponding finalizers, so there is ordering to some extent, just not the same as reconciler finalizer.

Regarding @karlkfi 's proposal:
> A potentially better solution might be to add webhook that blocks namespace deletion when a RepoSync is present. Or perhaps a finalizer on the Namespace that executes before the namespace garbage collector.

- `perhaps a finalizer on the Namespace that executes before the namespace garbage collector` won't work because the Namespace may not be managed by Config Sync. Config Sync may fight with other controllers when adding the finalizer.
- `add webhook to block namespace deletion when a RepoSync is present ` doesn't make sense in terms of the fleet tenancy integration example. The fleet Namespace is created/deleted independently with  RepoSyncs. The webhook will add more frictions for cascading deletion.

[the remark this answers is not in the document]
sdowell: I'm not entirely convinced that the behavior of the namespace finalizer will always be equivalent to the behavior of letting the reconciler finalizer run before the namespace reconciler. For example:
- Imagine objects which have side effects outside the cluster which need to be pruned in a specific order, for example KCC resources.
- Imagine a controller similar to a RepoSync reconciler in the cluster which has CRs/finalizers and RoleBindings in the namespace. By letting the namespace finalizer run first the RoleBinding may be deleted first, leading to a similar race condition as the current issue with the RepoSync reconciler.

At a high level, the problems I could see arising reduce to:
- Not honoring deletion ordering leading to other side effects (e.g. resources outside the cluster)
- Not honoring deletion ordering leading to a race condition which causes the namespace finalizer to become stuck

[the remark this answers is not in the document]
nan-yu: > I'm not entirely convinced that the behavior of the namespace finalizer will always be equivalent to the behavior of letting the reconciler finalizer run before the namespace reconciler.

That makes sense. We should have a mechanism to allow the reconciler finalizer to run if possible, and only let the namespace finalizer run when the reconciler is not functioning.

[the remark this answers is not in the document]
nan-yu: What about adding additional checks for the reconciler state besides checking the Namespace terminating state?
So the reconciler-manager will only remove the reconciler finalizer when
1. the RSync's Namespace is terminating, **AND**
2. the reconciler Deployment is not available, **OR** the RepoSync has a STALLED condition (we can even make it more explicit that the STALLED condition has a RoleBinding upsert error).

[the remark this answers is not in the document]
janetkuo: Per discussion with @nan-yu and @sdowell offline, finalizer on RepoSync doesn't stop managed resources from being deleted by namespace controller, so there's no point for ns reconciler to try to maintain deletion ordering when Namespace itself is being terminated. We'll go with this code logic for removing RepoSync finalizer when Namespace is being terminated.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -639,16 +650,65 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
...
+// A potential reason for update failure is the Namespace is terminating.
+//   - RootSync: if its Namespace is terminating, the reconciler Deployment
+//     might get deleted prior to RootSync, resulting in the finalizer being
+//     unable to be removed. Consequently, RootSync would become stuck in the
+//     deletion process.
+//   - RepoSync: if its Namespace is terminating, the RoleBinding for the
+//     namespace reconciler could be deleted and fail to be re-created.
+//     Therefore, the reconciler would lack the necessary permissions to
+//     remove the reconciler finalizer from the RepoSync object, leading to
+//     RepoSync being stuck in deletion.
```
janetkuo, line 678: Suggest updating the comment to make it clear that this logic is only handled for RepoSync.

nan-yu: done

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -617,12 +617,13 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
 	// Else - the object is being deleted.
 
 	if controllerutil.ContainsFinalizer(syncObj, metadata.ReconcilerFinalizer) {
-		// The object is being deleted, but the reconciler finalizer is still running.
-		// Wait for the reconciler finalizer to complete.
-		r.logger(ctx).Info("Waiting for Reconciler Finalizer to finish")
-		return nil
+		removed, err := r.removeReconcilerFinalizer(ctx, syncObj)
```
karlkfi, line 620: This method name makes this confusing to read. It doesn't always remove the finalizer. Sometimes it just returns and waits for a new event to say it was removed.

Also, the function itself has different logic for the rootsync and reposync. Can we instead push the type-specific logic into the reposync and rootsync controllers? Maybe inject a finalizer strategy that implements an interface method?

It would be nice to keep the code in the base class common to both impls.

karlkfi:
After digging into this a bit more my concern is this:

Removing the finalizer on the RepoSync when its parent Namespace is being deleted DOES solve the case where deletion ordering is unimportant. 

The problem with this solution is that it prevents the reconciler from even attempting to solve the harder problem where deletion ordering IS important.

Additionally, in both cases, the ResourceGroup inventory is left behind.

My counter-proposal would be to only remove the finalizer if there aren't any explicit dependencies defines in the resources being managed by the RepoSync reconciler.

To do that, the reconciler itself would need to stay in charge of removing the RSync finalizer, because only the reconciler has access to the state of the objects without having to query them all. Plus, having only one code path in charge of a specific update is a good idea anyway to avoid race conditions and simplify debugging. Since the reconciler already deletes its RSync's finalizer, we could just skip the Destroy call all together and let the namespace garbage collector handle it.

This proposal has one notable downside: The reconciler deployment needs a RoleBinding to have permission to remove the RepoSync finalizer, and that RoleBinding is also deleted by the namespace garbage collector. To avoid this I have two ideas:
1. Add a webhook (or a new endpoint on the admission-controller) to block deletion of the RoleBinding until after the reconciler Deployment is NotFound.
2. Add a finalizer to the RoleBinding until after the reconciler Deployment is NotFound.

Both of these options have pros and cons, but the webhook could be more generally useful. For example:
- A webhook could block deletion of Namespaces if they contain any objects with incoming dependencies, allowing the RepoSync reconciler to delete package contents in the right order first.
- A webhook could block deletion of any object with an incoming dependency, enforcing deletion ordering more widely.

In the case of namespace garbage collection, temporary deletion errors are ok, because it will just keep retrying, giving the RepoSync reconciler a chance to do it's work.

In the case where the RepoSync reconciler needs an additional user-managed RoleBinding to manage other resource types, the user could add a dependency on the RepoSync to make it depend on their RoleBinding. This way, the user's RoleBinding would also be protected the same way the default RoleBinding is protected by the webhook.

nan-yu:
> Removing the finalizer on the RepoSync when its parent Namespace is being deleted DOES solve the case where deletion ordering is unimportant.

That is the intent of this fix.

> The problem with this solution is that it prevents the reconciler from even attempting to solve the harder problem where deletion ordering IS important.

It might be helpful to clarify that `depends-on` and deletion order shouldn't be treated as identical concepts. If deletion ordering IS important, users should leverage custom finalizers on their resources. An example is CR depends on CRD, but the deletion of CRD should not be blocked by CR.

> Additionally, in both cases, the ResourceGroup inventory is left behind.

This is not correct. ResourceGroup is in the same namespace as the RepoSync. It would be deleted as well.

nan-yu:
> In the case of namespace garbage collection, temporary deletion errors are ok, because it will just keep retrying, giving the RepoSync reconciler a chance to do it's work.

In this case, no deletion ordering is guaranteed.

> In the case where the RepoSync reconciler needs an additional user-managed RoleBinding to manage other resource types, the user could add a dependency on the RepoSync to make it depend on their RoleBinding. This way, the user's RoleBinding would also be protected the same way the default RoleBinding is protected by the webhook.

It is also not guaranteed that RoleBinding and RepoSync are managed in the same repo, and I don't think Config Sync supports external dependency.

karlkfi:
> ResourceGroup is in the same namespace as the RepoSync. It would be deleted as well.

I stand corrected. You're right, only the RootSync ResourceGroups are in the `config-management-system`. 

In that case, the RepoSync & ResourceGroup being deleted at approximately the same time by the namespace garbage collector is probably good enough.

However, the other related problem is that the UI & CLI will show the RepoSync/package as deleted, but the package contents may not have all been fully deleted if they're blocked by a finalizer. So the user won't be able to use the Console Package UI or `nomos status` to watch the status of those objects if they have finalizers.

> It might be helpful to clarify that depends-on and deletion order shouldn't be treated as identical concepts.

Yes. The primary issue here is with explicit dependencies, with the depends-on annotation, not implicit dependencies, like namespaces and CRDs.

> In this case, no deletion ordering is guaranteed.

The namespace garbage collector doesn't ensure dependency deletion ordering, but we do with RSync uninstall (with the deletion-propagation annotation). In this case the two policies conflict and we lose. But it doesn't have to be that way. Just because it is this way doesn't mean it should be or that users want it this way. We can make it better!

> It is also not guaranteed that RoleBinding and RepoSync are managed in the same repo, and I don't think Config Sync supports external dependency.

I'm not implying they would be or need to be. My proposal is that the webhook detect depends-on annotations and protect the dependency from deletion, even if the object wasn't synced by Config Sync. That way the reconciler-manager can add one on the reconciler Deployment -> default RoleBinding and the user can add one for RepoSync -> user RoleBinding.

In the future, we might want to figure out how to allow multiple clients to manage separate depends-on annotations on the same object to define different dependencies, but for this use case we don't need that yet.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: sounds good. I'll change to use interface and type-specific implementation.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -639,16 +640,69 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
...
+				fmt.Sprintf("failed to update %s to remove the reconciler finalizer", r.syncKind))
+			r.logger(ctx).Error(err, "The reconciler finalizer removal failed")
+			return false, err
+		}
+		r.logger(ctx).Info("The reconciler finalizer removal successful")
+		return true, nil
+	}
+
+	// RepoSync's Namespace is active, then wait for the reconciler finalizer to complete.
+	r.logger(ctx).Info("Waiting for the namespace reconciler finalizer to finish")
```
janetkuo, line 702: nit: We don't need to remove finalizer if the namespace is active, so it's fine not to log it. Suggest a slight change to make the code more clean:

```go
if syncObjNamespace.Status.Phase != corev1.NamespaceTerminating {
  return false, nil
}

// <move L687-698 here>
```

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -639,16 +640,69 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
...
+// The reconciler adds the finalizer when deletion propagation is set to
+// `Foreground`, and removes it when RSync has a deletionTimestamp.
+// When the RepoSync's Namespace is terminating, the RoleBinding for the
+// namespace reconciler could be deleted and fail to be re-created.
+// Therefore, the namespace reconciler would lack the necessary permissions to
+// remove the reconciler finalizer from the RepoSync object, leading to
+// RepoSync being stuck in deletion. This function allows the reconciler-manager
+// to delete the reconciler finalizer in this case.
+// It returns whether the reconciler finalizer has been removed.
+// If removed, the controller can proceed to tear down other resources.
```
janetkuo, line 672: The logic fixes the RoleBinding race, but the reason we do this is because we know Namespace controller will handle the garbage collection (deletion can't be cancelled). 

Suggest updating the comment L663-672 to:
```
// When the RepoSync's Namespace is terminating, Namespace controller will
// handle the deletion propagation of all resources managed by the RepoSync.
// Therefore, remove the RepoSync finalizer to avoid duplicated work.
// This also fixes an issue with a race when Namespace controller deletes
// the reconciler's RoleBinding needed for removing the finalizer. 
```

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -639,16 +640,69 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
...
+func (r *reconcilerBase) removeReconcilerFinalizer(ctx context.Context, syncObj client.Object) (bool, error) {
```
janetkuo, line 673: How about
```suggestion
func (r *reconcilerBase) removeFinalizerOnNSTermination(ctx context.Context, syncObj client.Object) (bool, error) {
```

nan-yu: I am changing it to a `reconcilerFinalizerHandler` interface with a `handleReconcilerFinalizer` function.

The RootSync handler is a no-op and waits for the reconciler finalizer to finish.
The RepoSync handler checks whether Namespace is terminating. If yes, it removes the finalizer. Otherwise, it waits for finalizer to finish.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: Added a `reconcilerFinalizerHandler` interface and a `handleReconcilerFinalizer` function.
The name is a bit long and I'm open for better names.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: done

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: done

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: updated

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -617,12 +619,13 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
...
-		// Wait for the reconciler finalizer to complete.
-		r.logger(ctx).Info("Waiting for Reconciler Finalizer to finish")
-		return nil
+		removed, err := r.handleReconcilerFinalizer(ctx, syncObj)
+		if !removed || err != nil {
+			return err
+		}
 	}
 
+	// The 'configsync.gke.io/reconciler' finalizer is removed successfully.
```
janetkuo, line 628: This comment is slightly confusing because the `if` block in L621 and L628 look so similar. 
The reconciler finalizer is removed in the previous block, and in this block we check/handle the reconciler-manager finalizer.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -617,12 +619,13 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
 	// Else - the object is being deleted.
 
 	if controllerutil.ContainsFinalizer(syncObj, metadata.ReconcilerFinalizer) {
-		// The object is being deleted, but the reconciler finalizer is still running.
-		// Wait for the reconciler finalizer to complete.
-		r.logger(ctx).Info("Waiting for Reconciler Finalizer to finish")
```
janetkuo, line 622: Should we keep the log and comment when reconciler finalizer isn't removed (L624)?

mikebz:
Hi everyone!  Sorry for chiming in on this when I have not been involved in much of the technical reasoning.  We have an upcoming code freeze and therefore need to evaluate some of these decision in the light that we have deadlines pending.

Question:
1. Do we have ready alternatives to this?
2. The UX of namespace being deleted and removing all resources out of order could lead to issues, is this a risk we can take on right now and think about it later?
3. Is this change irreversible?  Meaning will customers take dependency on this behavior and then we are locked into this choice?

If the answers are:
1. We don't have alternatives at hand.
2. The risk is small (meaning someone who was relying on Deployment A to be deleted before Deployment B, and are upset that this is happening when namespace is deleted).
3. We can improve the behavior later.

I suggest merging this in and using the few days we have before the code freeze for additional testing.  It doesn't seem like we want to ship the bug that we currently have "RSync would stuck in deletion forever."

janetkuo (approved):
Just some nits and LGTM otherwise

karlkfi:
I don't really understand why y'all are in a hurry to get this half-fix in with side effects, for an issue that's existed for years, right before a code freeze, but I agree that the side effect is less bad than the current behavior.

I don't think this will be un-reversable. It's just more tech debt that will have to be reverted if we decide to later apply the more complete solution.

Anyway, I'll make another review pass today.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: done

janetkuo:
A few comments:
1. Deletion ordering is orthogonal to RSync garbage collection, because deletion can be performed by other controllers/users in the cluster, and it could happen before the RSync is being deleted. Reconciler can make sure it deletes things in order, but it can't stop others from deleting things out of order. 
1. Apply ordering (depends-on) doesn't necessary imply deletion ordering. An example is that you want to apply CRD before applying CR, but we don't necessary want to block CRD from being deleted just because CR hasn't been deleted. 
1. When a Namespace is being terminated, Namespace controller already handles resource deletion of everything managed by the RepoSync. It's safe and easier for reconcilers to let Namespace controller handle garbage collection.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: removed and added logs if not removed.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -101,6 +101,8 @@ type reconcilerBase struct {
 
 	// syncKind is the kind of the sync object: RootSync or RepoSync.
 	syncKind string
+
+	reconcilerFinalizerHandler
```
karlkfi, line 105: huh. i didn't know it was possible to have multiple "parent" structs with inherited methods. TIL.

Might make more sense to move this to the top with loggingController though.

Alternatively, do the strategy methods actually need to be available on the reconcilerBase, or would it be better to just keep them private so the caller can't run them?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -639,11 +645,11 @@ func (r *reconcilerBase) setupOrTeardown(ctx context.Context, syncObj client.Obj
 		controllerutil.RemoveFinalizer(syncObj, metadata.ReconcilerManagerFinalizer)
 		if err := r.client.Update(ctx, syncObj); err != nil {
 			err = status.APIServerError(err,
-				fmt.Sprintf("failed to update %s to remove finalizer", r.syncKind))
-			r.logger(ctx).Error(err, "Finalizer removal failed")
+				fmt.Sprintf("failed to update %s to remove the reconciler-manager finalizer", r.syncKind))
+			r.logger(ctx).Error(err, "The reconciler-manager finalizer removal failed")
```
karlkfi, line 649: nit: i like to start logs with verbs or nouns if possible, to make them easier to scan for visually. Something like "Failed to remove reconciler-manager finalizer" or "Removal of reconciler-manager finalizer failed".

not a big deal though.

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
```diff
@@ -0,0 +1,94 @@
...
+// If the reconciler is unable to update R*Sync to remove the finalizer, R*Sync
+// would get stuck in the deletion process.
+// To unblock the deletion, the reconciler-manager acts as a delegator to remove
+// R*Sync's finalizer when needed.
+type reconcilerFinalizerHandler interface {
+	// handleReconcilerFinalizer handles the `configsync.gke.io/reconciler`
+	// finalizer when the reconciler is not able to remove it from R*Sync.
+	handleReconcilerFinalizer(context.Context, client.Object) (bool, error)
+}
+type rootReconcilerFinalizerHandler struct {
```
karlkfi, line 41: nit: line break

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
```diff
@@ -0,0 +1,94 @@
...
+// The reconciler adds the finalizer when deletion propagation is set to
+// `Foreground`, and removes it when R*Sync has a deletionTimestamp.
+// If the reconciler is unable to update R*Sync to remove the finalizer, R*Sync
+// would get stuck in the deletion process.
+// To unblock the deletion, the reconciler-manager acts as a delegator to remove
+// R*Sync's finalizer when needed.
+type reconcilerFinalizerHandler interface {
+	// handleReconcilerFinalizer handles the `configsync.gke.io/reconciler`
+	// finalizer when the reconciler is not able to remove it from R*Sync.
+	handleReconcilerFinalizer(context.Context, client.Object) (bool, error)
```
karlkfi, line 39: nit: I generally recommend using public methods on interfaces to make them more testable in other packages if you need to inject a fake.

If the field on the reconcilerBase is a private field and not a parent struct that should be sufficient to prevent callers form executing the method.

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
```diff
@@ -0,0 +1,94 @@
...
+		r.logger(ctx).Error(err, "Get sync object Namespace failed")
+		return false, err
+	}
+
+	if syncObjNamespace.Status.Phase != corev1.NamespaceTerminating {
+		// RepoSync's namespace is active, wait for the reconciler finalizer to complete.
+		return false, nil
+	}
+
+	// RepoSync's namespace is terminating, remove the reconciler finalizer.
```
karlkfi, line 82: I think we might want a log message for this case, so we can identify when it's happened when debugging.

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
```diff
@@ -0,0 +1,94 @@
...
+	if !controllerutil.RemoveFinalizer(syncObj, metadata.ReconcilerFinalizer) {
+		r.logger(ctx).Info("The reconciler finalizer is not found")
+		return true, nil
+	}
+	if err := r.client.Update(ctx, syncObj); err != nil {
```
karlkfi, line 87: Would it make more sense to patch the finalizer removal?

Erroring on update conflict and retrying might be ok, but patching seems potentially faster.

karlkfi: I see it was already updating. So that's fine. Nevermind.

nan-yu: I used a consistent style as removing the reconciler-manager finalizer: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/reconciler_base.go#L640.

Agreed that patching might be more efficient. We can change them altogether in a separate PR without any functionality change.

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
karlkfi: Another potentially simpler option for this might be use just inject a strategy function and not a struct. Then you wouldn't need the rootReconcilerFinalizerHandler impl at all. You could could just skip it when the strategy function is nil.

karlkfi:
/lgtm

I suspect we'll eventually need to revert this later when we add a more general solution that works for deletion dependencies and RootSyncs, but for now it seems like an improvement for RepoSyncs at least.

I left some minor comments, but overall it LGTM.

I would also suggest including the NamespaceTerminating part of https://github.com/GoogleContainerTools/kpt-config-sync/pull/1194 in this PR, so that they can be pulled out together, if necessary, since they cause the same side effect. But it's not necessarily a blocker.

Feel free to remove the hold when the comments are addressed.

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
nan-yu: I would prefer to use a struct as a place holder for the rootReconcilerFinalizerHandler.
If we ever need to update the rootReoncilerFinalizerHandler, we just need to update the implementation, instead of passing a different strategy function.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: done, keep the function private to reconcilerBase.

pkg/reconcilermanager/controllers/reconciler_base.go
nan-yu: rephrased

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
nan-yu: done

pkg/reconcilermanager/controllers/reconciler_finalizer_handler.go
nan-yu: done

nan-yu:
/unhold

#1194 is related, but a different issue. It shouldn't block this.

karlkfi (approved):

PR #1192, opened 2024-04-08
fix: only instantiate ValueConverter when webhook is enabled

nan-yu:
When instantiating the ValueConverter, it also calls the Refresh function to pull fresh schemas from the OpenAPI discovery endpoint, which is very memory-intensive. This commit changes it to only instantiate the converter when webhook is enabled, so it will reduce memory usage when the webhook is not used.

b/295389452

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -213,11 +213,6 @@ func Run(opts Options) {
 		klog.Fatalf("Instantiating Remediator: %v", err)
 	}
 
-	converter, err := declared.NewValueConverter(discoveryClient)
-	if err != nil {
-		klog.Fatalf("Instantiating converter: %v", err)
-	}
-
 	// Configure the Parser.
 	var parser parse.Parser
 	fs := parse.FileSource{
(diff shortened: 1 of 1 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -254,6 +248,15 @@ func Run(opts Options) {
 			Remediator: rem,
 		},
 	}
+	// Only instantiate the converter when the webhook is enabled because the
+	// instantiation pulls fresh schemas from the openapi discovery endpoint.
+	if opts.WebhookEnabled {
+		parseOpts.Converter, err = declared.NewValueConverter(discoveryClient)
```
sdowell, line 254: It looks like this line will be hit now, since the Converter will be `nil`: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/validate/raw/hydrate/declared_field_hydrator.go#L41

Assuming this is the intent, the log message should be updated

nan-yu: Nope. The converter would be nil only when the webhook is disabled. When webhook is disabled, `DeclaredFields` won't be invoked: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/validate/raw/raw.go#L83.

sdowell: Ack, sorry I missed that

nan-yu: FYI, this also needs to go into the 1.18 release as without this, OpenAPI will still be loaded at the instantiation time.

Sorry for not capture this earlier.

This change has been verified with memory usage reduction.

PR #1194, opened 2024-04-12
Improvements for the NamespaceSelector dynamic mode

nan-yu:
- It only initializes `nsControllerState` when dynamicNSSelectorEnabled is true, so that pkg/parse/run.go can stop `nsEventTimer` when nsControllerState is nil.
- When namespace_hydrator selects dynamic/on-cluster Namespaces, it filters out terminating Namespaces because Kubernetes doesn't allow creating new content in terminating Namespaces.

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -257,13 +257,19 @@ func Run(opts Options) {
 		}
 	}
 
-	nsControllerState := namespacecontroller.NewState()
+	var nsControllerState *namespacecontroller.State
 	if opts.ReconcilerScope == declared.RootReconciler {
 		rootOpts := &parse.RootOptions{
 			SourceFormat:             opts.SourceFormat,
 			NamespaceStrategy:        opts.NamespaceStrategy,
 			DynamicNSSelectorEnabled: opts.DynamicNSSelectorEnabled,
-			NSControllerState:        nsControllerState,
+		}
+		if opts.DynamicNSSelectorEnabled {
+			// Only set nsControllerState when dynamic NamespaceSelector is enabled on
+			// RootSyncs.
+			// RepoSync can't manage NamespaceSelectors.
+			nsControllerState = namespacecontroller.NewState()
+			rootOpts.NSControllerState = nsControllerState
 		}
 		parser = parse.NewRootRunner(parseOpts, rootOpts)
 	} else {
--- a/pkg/validate/scoped/hydrate/namespace_hydrator.go
+++ b/pkg/validate/scoped/hydrate/namespace_hydrator.go
@@ -143,7 +143,7 @@ func buildSelectorMap(ctx context.Context, c client.Client, objs *objects.Scoped
 		var selectedNamespaces []string
 		// Select dynamic/on-cluster Namespaces that match the NamespaceSelector's labels.
 		for _, ns := range nsList.Items {
-			if selector.Matches(labels.Set(ns.GetLabels())) {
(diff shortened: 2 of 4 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -257,13 +257,15 @@ func Run(opts Options) {
...
 
-	nsControllerState := namespacecontroller.NewState()
+	var nsControllerState *namespacecontroller.State
 	if opts.ReconcilerScope == declared.RootReconciler {
 		rootOpts := &parse.RootOptions{
 			SourceFormat:             opts.SourceFormat,
 			NamespaceStrategy:        opts.NamespaceStrategy,
 			DynamicNSSelectorEnabled: opts.DynamicNSSelectorEnabled,
-			NSControllerState:        nsControllerState,
+		}
```
karlkfi, line 266: Does the state need to be initialized for RepoSyncs or just RootSyncs?

nan-yu: Just RootSyncs because RepoSyncs can't manage NamespaceSelector.

karlkfi: Gotcha. Can you add an inline comment here to explain?
I understand it now, I just want to make sure it's clear for someone without context why we're only setting it in this 2nd level if statement.

pkg/validate/scoped/hydrate/namespace_hydrator.go
```diff
@@ -143,7 +143,7 @@ func buildSelectorMap(ctx context.Context, c client.Client, objs *objects.Scoped
 		var selectedNamespaces []string
 		// Select dynamic/on-cluster Namespaces that match the NamespaceSelector's labels.
 		for _, ns := range nsList.Items {
-			if selector.Matches(labels.Set(ns.GetLabels())) {
+			if ns.Status.Phase != corev1.NamespaceTerminating && selector.Matches(labels.Set(ns.GetLabels())) {
```
karlkfi, line 146: Is there a case when these were previously selected, but now the namespace is being deleted that would cause the objects to be removed from the UI & ResourceGroup before they were deleted?

Doesn't this cause the same side effect as https://github.com/GoogleContainerTools/kpt-config-sync/pull/1191?

If so, I think it'd make more sense to include this part of the change in that PR so that those namespace deletion changes are all together. Otherwise, if we need to revert them later it would also revert the unrelated nsControllerState fix in this PR.

nan-yu: > Is there a case when these were previously selected, but now the namespace is being deleted that would cause the objects to be removed from the UI & ResourceGroup before they were deleted?

The Namespace finalizer should delete the previously selected resources. It is an improvement because the resource should not be selected before, and would never become managed by Config Sync (i.e. show in the UI & ResourceGroup). Hence, we should not care if they're deleted or not. It is the terminating Namespace finalizer's job.

> Doesn't this cause the same side effect as https://github.com/GoogleContainerTools/kpt-config-sync/pull/1191?

It is different. #1191 aims to unblock RepoSync deletion, while this change applies to managed resources. Since the resources are dynamically selected by the NamespaceSelector, it should be okay to unselect them when the Namespace is terminating.
I don't think the improvement would break anything.

pkg/reconciler/reconciler.go
nan-yu: done, also added a comment to the field in `RootOptions`.

karlkfi (approved):
/lgtm

Thanks for getting this fix in so quick!

PR #1196, opened 2024-04-23
Add memory test doc for 1.18

tiffanny29631:


docs/memory-usage-reduce-1160-1180.md
```diff
@@ -0,0 +1,34 @@
+# Config Sync Memory Usage Reduction v1.16 vs v1.18
+
+Config Sync v1.18 contains change to stop loading OpenAPI for schema validations when Config Sync admission webhook is not enabled. Here are some test results that shows the reduction in the memory usage.
```
karlkfi, line 3: Can you explain the problem first?

How does using OpenAPI increase the memory? In what use cases?

docs/memory-usage-reduce-1160-1180.md
```diff
@@ -0,0 +1,34 @@
...
+| AutoPilot 1.29 | 230   | 160   | -30.4% |
+| Standard 1.27  | 260   | 170   | -34.6% |
+| Standard 1.28  | 220   | 140   | -36.4% |
+| Standard 1.29  | 210   | 140   | -33.3% |
+
+### Test Setup
+
+#### Test Repository
+
+[Config Sync Quickstart](https://github.com/GoogleCloudPlatform/anthos-config-management-samples/tree/main/config-sync-quickstart)
```
karlkfi, line 19: How many objects is this syncing? How many different resource types? How many namespace?

These numbers also affect memory usage. So having tests with varying amounts would help show how much reduction is gained at various scales. Is the % reduction constant or dependent on other variables?

karlkfi: Does memory usage scale just with the CRDs managed by the current reconciler or also with any CRDs on the cluster?

FWIW, the applier watch cache memory usage increases with any objects on the cluster in an overlapping resources and namespace (it's actually even more complicated than this due to RootSyncs often using cluster-scoped watches). So I'm trying to figure out how isolated your tests are on those dimentions.

tiffanny29631: The usage in this case scales with the any CRDs on the cluster, not the CRDs managed by the reconciler. With the test CRDs on the AutoPilot clusters the reconciler will get OOM killed with the default memory request / limit.

karlkfi: How many CRDs can it handle with the default resources on autopilot?

Does the size of the CRDs matter?

docs/memory-usage-reduce-1160-1180.md
janetkuo: @nan-yu did profiling previously and loading the OpenAPI consumes a sizable fraction of the memory.

janetkuo (approved):
/lgtm

janetkuo:
This doc is good given that it includes the scope of tests being performed. We can add more types of benchmarking tests later.

docs/memory-usage-reduce-1160-1180.md
tiffanny29631: We could enhance the dataset with your work in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1187 later.

PR #1199, opened 2024-04-26
Add event handler to Applier/Destroyer

karlkfi:
Use an injected event handler function to avoid needing to cache errors in the supervisor. Instead, cache errors in the Updater, which already has a lock for this.

This event handler simplifies the Applier/Destroyer API and makes it easy to add more events in the future, like for inventory updates.

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1204

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -68,28 +68,47 @@ var (
 type Applier interface {
 	// Apply creates, updates, or prunes all managed resources, depending on
 	// the new desired resource objects.
-	// Returns the set of GVKs which were successfully applied and any errors.
+	// Error events are sent to the eventHandler.
+	// Returns the status of the applied objects and statistics of the sync.
 	// This is called by the reconciler when changes are detected in the
 	// source of truth (git, OCI, helm) and periodically.
-	Apply(ctx context.Context, desiredResources []client.Object) status.MultiError
-	// Errors returns the errors encountered during apply.
-	// This method may be called while Destroy is running, to get the set of
-	// errors encountered so far.
-	Errors() status.MultiError
+	Apply(ctx context.Context, eventHandler func(Event), desiredResources []client.Object) (ObjectStatusMap, *stats.SyncStats)
 }
 
 // Destroyer is a bulk client for deleting all the managed resource objects
 // tracked in a single ResourceGroup inventory.
 type Destroyer interface {
 	// Destroy deletes all managed resources.
-	// Returns any errors encountered while destroying.
+	// Error events are sent to the eventHandler.
+	// Returns the status of the destroyed objects and statistics of the sync.
 	// This is called by the reconciler finalizer when deletion propagation is
 	// enabled.
-	Destroy(ctx context.Context) status.MultiError
-	// Errors returns the errors encountered during destroy.
-	// This method may be called while Destroy is running, to get the set of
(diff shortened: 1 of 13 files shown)


pkg/applier/applier.go
```diff
@@ -589,83 +602,52 @@ func (a *supervisor) applyInner(ctx context.Context, objs []client.Object) (map[
...
-// Errors returns the errors encountered during the last apply or current apply
-// if still running.
-// Errors implements the Applier and Destroyer interfaces.
-func (a *supervisor) Errors() status.MultiError {
-	a.errorMux.RLock()
-	defer a.errorMux.RUnlock()
-
-	// Return a copy to avoid persisting caller modifications
-	return status.Append(nil, a.errs)
+func (a *supervisor) sendErrorEvent(err error, eventHandler func(Event)) {
```
nan-yu, line 634: The `sendErrorEvent` function doesn't need to be a member of the supervisor.

pkg/applier/applier.go
```diff
@@ -69,28 +68,47 @@ var (
...
+	// Error events are sent to the eventHandler.
+	// Returns the status of the applied objects and statistics of the sync.
 	// This is called by the reconciler when changes are detected in the
 	// source of truth (git, OCI, helm) and periodically.
-	Apply(ctx context.Context, desiredResources []client.Object) (map[schema.GroupVersionKind]struct{}, status.MultiError)
-	// Errors returns the errors encountered during apply.
-	// This method may be called while Destroy is running, to get the set of
-	// errors encountered so far.
-	Errors() status.MultiError
+	Apply(ctx context.Context, eventHandler func(Event), desiredResources []client.Object) (ObjectStatusMap, *stats.SyncStats)
```
nan-yu, line 75: What about renaming `eventHandler` to `errorEventHandler` or `errorHandler` to distinguish it from `supervisorEventHandler`?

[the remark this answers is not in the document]
karlkfi: That's what it was doing before, but returning them lets us test them.

karlkfi: I plan on adding inventory events to it, errors are just the first pass.

But yes, having two event handlers in the same file is a bit confusing. Open to other suggestions though. 

I was thinking maybe the supervisorEventHandler could be merged with the supervisor, but the supervisorEventHandler has a destroy flag I don't really want to pass to all the methods individually.

pkg/applier/applier.go
karlkfi: fixed.

pkg/applier/applier.go
nan-yu: > I plan on adding inventory events to it, errors are just the first pass.

I thought other inventory events would be handled by the `supervisorEventHandler`? What are other inventory events?

> but the supervisorEventHandler has a destroy flag I don't really want to pass to all the methods individually.

Only the `processWaitEvent` function reads the destroy flag. If we merge `supervisorEventHandler` with the supervisor, the flag only needs to be passed to one function, which seems not bad.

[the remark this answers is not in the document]
nan-yu: Sounds good.

karlkfi: done

nan-yu (approved):
/lgtm

PR #1203, opened 2024-05-01
chore: update owners

mikebz:


OWNERS_ALIASES
nan-yu, line 1: We also need to add [Camila-B](https://github.com/Camila-B) to the maintainers list?

mikebz: sure, but let's not stop this PR and just do a separate one for her when she is ready.

sdowell (approved):
/lgtm

PR #1204, opened 2024-05-01
Fix declared.Resources to be consistently ordered 

karlkfi:
- Resources' DeclaredUnstructureds, DeclaredObjects, DeclaredGVKs,
  and getObjectSet were previously iterating over a Go map, causing the
  output to be randomly ordered. This change replaces the map with an
  OrderedMap that can be iterated in insert order. This makes it
  easier to test classes that use declared.Resources, including
  Applier.Apply and parse.parseAndUpdate.
- Add fakes for Applier, Remediator, and ConfigParser
- Add test validation for Apply, Remediate, and Parse inputs and outputs
- Fix fake.RootSync/RepoSync to avoid populating the status. These
  return a FileObject, which never have a status when created by the
  parser.
- Fix filesystem.AsCoreObjects to return nil when empty
- Replace kpt.dev/configsync/pkg/util/ordered with
  github.com/elliotchance/orderedmap/v2 which supports generics and
  actually has unit tests.

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1205
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1206
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1208
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1209


karlkfi:
/hold

I'm gonna break this down some more.

pkg/parse/root_test.go
```diff
@@ -107,46 +78,54 @@ func gitSpec(repo string, auth configsync.AuthType) core.MetaMutator {
...
+					FileObjects: []ast.FileObject{
+						fake.Role(core.Namespace("foo")),
+					},
+				},
 			},
-			want: []ast.FileObject{
-				fake.UnstructuredAtPath(kinds.Namespace(),
-					"",
-					core.Name("foo"),
+			expectedObjsToApply: []ast.FileObject{
```
karlkfi, line 107: These look more changed that they are. I just re-ordered them to match the new consistent ordering, with added implicit namespaces last.

karlkfi: I also updated the `SyncTokenAnnotationKey` annotation value to match the commit added by using a populated `state.cache.source` as input.

karlkfi:
/unhold

Extracted and merged 4 other PRs. This should be manageable now.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/davecgh/go-spew v1.1.1
+	github.com/elliotchance/orderedmap/v2 v2.2.0
```
sdowell, line 16: nit: This library does not seem to have much activity. Are we comfortable depending on it?

karlkfi: The one we used before didn't have any tests at all...

It's just a little utility. I don't really expect frequent updates if it works.

karlkfi: But if you've got a better option, let me know!

karlkfi: Actually, I was wrong. It did have tests, they just weren't vendored, which makes sense. 

That said, it's updated even less frequently.

karlkfi: Other options include:
- https://github.com/goombaio/orderedmap
- https://github.com/samdolan/go-ordered-map

Both older and not using generics

sdowell (approved):
/lgtm

PR #1206, opened 2024-05-01
Unwrap single status.MultiErrors

karlkfi:
- Fix Applier/Destroyer.addError from wrapping/appending nil errors
- Add status.Wrap for wrapping status.Errors without an existing MultiError.
- Fix status.Append to avoid wrapping errors that are already a MultiError
- Add testerrors.AssertEqual to simplify testing status.Errors. Ideally we would just fix statis.Error.Is, but that would require fixing all our test assertions. This approach is a bit easier, so we can fix assertions a little at a time to make them more strict.
- Backfilled unit tests for ErrorForResourceWithResource

### DIFF
--- a/pkg/applier/applier_test.go
+++ b/pkg/applier/applier_test.go
@@ -18,18 +18,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"reflect"
-	"strings"
 	"testing"
 	"time"
 
 	"github.com/GoogleContainerTools/kpt/pkg/live"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/applier/stats"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/declared"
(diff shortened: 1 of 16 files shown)


nan-yu (approved):
/lgtm

pkg/applier/applier_test.go
```diff
@@ -139,23 +135,6 @@ func TestApply(t *testing.T) {
 				formApplyEvent(event.ApplyPending, testObj2, nil),
 			},
 			expectedError: ErrorForResourceWithResource(errors.New("unknown type"), idFrom(testID), testObj),
-			expectedCSEs: []v1beta1.ConfigSyncError{
```
sdowell, line 142: Does this still verify the `resources` fields? I don't think this is all surfaced by the `error` interface, so it seems like this might be tested - although I haven't yet dug into the `testutils` code from cli-utils

/hold

sdowell: This also reduces the test to basically asserting `foo() == foo()`, where `foo` is the real code path. The test assertions should be written such that `foo() == expectedValue`. Reading this test it's not really clear what the output should look like

karlkfi: The test here is asserting that Applier.Apply returns the error received from a specific event and wraps it appropriately with the resource.

sdowell: With this test I could change the underlying implementation of `ErrorForResourceWithResource` to no longer include the ResourceRef and the test would still pass

karlkfi: Yes, but this test isn't testing that ErrorForResourceWithResource exposes the right fields/methods. It's just testing that Applier.Apply returned the expected error. 

If you want to confirm ErrorForResourceWithResource has a ResourceRef field/method, that should be a test of ErrorForResourceWithResource itself.

sdowell: Can you add some individual tests for that then, since the coverage is being removed here?

karlkfi: Fine, but we have zero tests for error builders and dozens of usages of BuildWithResources already. ErrorForResourceWithResource just wraps BuildWithResources the same way everything else does.

sdowell: Agreed, which is why I asked for these assertions to be added to these tests in the first place. I find the error builders themselves pretty obfuscated and untested.

This bugfix was specifically for the ResourceRef fields on the RSync status, so I wanted to make sure there were assertions in place which verify the behavior and prevent regression

karlkfi: Done

sdowell: Thank you!

sdowell (approved):
/lgtm
/unhold

karlkfi:
Fixed the e2e error. Need new LGTM

karlkfi:
Fixed the e2e harder, for real this time.

PR #1208, opened 2024-05-02
Fix status errors to work with errors.Is

karlkfi:
- errors.Is performs a basic equality comparison on the input errors which panics if the errors are not comparable. This effectively requires errors to use pointers so that the pointer can be used for equality comparison.
- Update all status errors to use pointers. This has the added benefit of not repeatedly copying error field values every time an error is returned to a function caller.
- Update testerrors.AssertEqual to support asymetric equality comparison, which is performed by go-cmp.Equal when using the EquateErrors option.

### DIFF
--- a/pkg/status/management_conflict_error.go
+++ b/pkg/status/management_conflict_error.go
@@ -64,19 +64,19 @@ type managementConflictErrorImpl struct {
 	currentManager string
 }
 
-var _ ManagementConflictError = managementConflictErrorImpl{}
+var _ ManagementConflictError = &managementConflictErrorImpl{}
 
-func (m managementConflictErrorImpl) ConflictingManager() string {
+func (m *managementConflictErrorImpl) ConflictingManager() string {
 	return m.currentManager
 }
 
-func (m managementConflictErrorImpl) CurrentManagerError() ManagementConflictError {
+func (m *managementConflictErrorImpl) CurrentManagerError() ManagementConflictError {
 	return ManagementConflictErrorBuilder.
 		Sprint(currentErrorMsg(m.newManager, m.newManager)).
 		BuildWithConflictingManagers(m.resource, m.newManager, m.currentManager)
 }
 
-func (m managementConflictErrorImpl) ConflictingManagerError() ManagementConflictError {
+func (m *managementConflictErrorImpl) ConflictingManagerError() ManagementConflictError {
 	return ManagementConflictErrorBuilder.
 		Sprintf("The %q reconciler detects a management conflict for a resource declared in another repository. "+
 			"Remove the declaration for this resource from either the current repository, or the repository managed by %q.",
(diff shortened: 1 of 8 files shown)


pkg/status/management_conflict_error.go
```diff
@@ -64,58 +64,61 @@ type managementConflictErrorImpl struct {
...
 }
 
-func (m managementConflictErrorImpl) Body() string {
+func (m *managementConflictErrorImpl) Body() string {
 	return formatBody(m.underlying.Body(), "\n\n", formatResources(m.resource))
 }
 
-func (m managementConflictErrorImpl) Is(target error) bool {
+func (m *managementConflictErrorImpl) Is(target error) bool {
+	if target == nil {
```
nan-yu, line 120: Do we need to check if `target == nil` here? I thought `m.underlying.Is(target)` should have it checked as part of `baseErrorImpl`.

pkg/testing/testerrors/testerrors.go
```diff
@@ -29,12 +31,47 @@ import (
...
+		err: err,
+	}
+}
+
+type equalError struct {
+	err error
+}
+
+func (e *equalError) Error() string {
+	return fmt.Sprintf("EqualError{Type: %T, Error: %q}", e.err, e.err)
```
tiffanny29631, line 60: Type and error both e.err?

pkg/testing/testerrors/testerrors.go
```diff
@@ -29,12 +31,47 @@ import (
...
+func (e *equalError) Is(target error) bool {
+	if target == nil {
+		return false
+	}
+	// Unwrap EqualErrors to allow asymmetric comparison.
+	if ee, ok := target.(*equalError); ok {
+		return e.Is(ee.err)
+	}
+	return reflect.TypeOf(e.err) == reflect.TypeOf(target) &&
+		e.err.Error() == target.Error()
```
tiffanny29631, line 72: Can this be erros.Is

pkg/status/management_conflict_error.go
karlkfi: If we assume baseErrorImpl is always at the bottom AND baseErrorImpl.Is allows/checks nil AND we don't change it in the future to add more validation, then yes, the check in the other classes would be redundant.

However, I'd like to make the validation more strict in the future, and I ran into a bunch of edge cases where I needed to do the nil check before any other type-specific validation. So I figured I'd just add these here preemptively. It's also good to have them so they're not tightly coupled to baseErrorImpl.

pkg/testing/testerrors/testerrors.go
karlkfi: %T prints the type of the error. %q prints the Error() string message.

sdowell (approved):
/lgtm

pkg/testing/testerrors/testerrors.go
karlkfi: Maybe, but it would mean something different, and require all errors compared to be pointers.

This code was copied from cli-utils. I just added the type check above this line to allow unwrapping for comparison of two wrapped equalErrors.

The intent is to allow comparison of errors where the top most type is correct and the message is correct, but the middle structure doesn't need to be exactly the same. That way, you can optionally use fmt.Errorf("") for wrapped errors and still confirm that the error message is complete, without needing to reverse-engineer all the different error types that got wrapped.

PR #1209, opened 2024-05-02
Test errors more strictly

karlkfi:
- Use testutil.AssertEqual in more places to validate both the error type and the error message.
- Update test expectations with more correct errors
- Fix a case of error wrapping with %v instead of %w
- Remove TestRoot_ParseErrorsMetricValidation, which used to test a metric that has been removed.
- Fix TestSortErrors which was expecting an incorrect ordering.
- Simplify status.Append to only allocate a slice and multiError when necessary.
- Add RetriableError.Is to make errors.Is work


nan-yu (approved):
/lgtm

PR #1211, opened 2024-05-06
Add e2e NT.ClusterVersion detection

karlkfi:
- Get the cluster version from the discovery client, parse it, and save it in the NT environment for use by tests.
- Update the v1beta1 CRD/RBAC check to use NT.ClusterVersion
- Update tests using the v1beta1 check to skip their test, not just return as successful.
- In the future, we can use the cluster version to check for autopilot resource bursting. But since resource bursting is temporarily disabled by default, we can't validate it today.

### DIFF
--- a/e2e/testcases/custom_resource_definitions_test.go
+++ b/e2e/testcases/custom_resource_definitions_test.go
@@ -118,13 +118,10 @@ func mustRemoveCustomResourceWithDefinition(nt *nomostest.NT, crd client.Object)
 }
 func TestMustRemoveCustomResourceWithDefinitionV1Beta1(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.Reconciliation1)
-	support, err := nt.SupportV1Beta1CRDAndRBAC()
-	if err != nil {
-		nt.T.Fatal("failed to check the supported CRD versions")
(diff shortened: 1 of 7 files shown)


e2e/testcases/custom_resource_definitions_test.go
```diff
@@ -118,13 +118,10 @@ func mustRemoveCustomResourceWithDefinition(nt *nomostest.NT, crd client.Object)
...
 func TestMustRemoveCustomResourceWithDefinitionV1Beta1(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.Reconciliation1)
-	support, err := nt.SupportV1Beta1CRDAndRBAC()
-	if err != nil {
-		nt.T.Fatal("failed to check the supported CRD versions")
-	}
-	if support {
-		mustRemoveCustomResourceWithDefinition(nt, anvilV1Beta1CRD())
+	if !nt.SupportV1Beta1CRDAndRBAC() {
+		nt.T.Skip("Kubernetes v1.22 and later do not support the v1beta1 CRD API")
```
sdowell, line 122: v1.22 and below is no longer in our testing matrix, so we could also consider removing these tests

karlkfi: Suggest filing a bug for that.

PR #1212, opened 2024-05-06
chore: update k8s to 1.28

sdowell:


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -519,7 +522,7 @@ func (r *RepoSyncReconciler) watchConfigMaps(rs *v1beta1.RepoSync) error {
 		klog.Infoln("Adding watch for ConfigMaps in namespace ", rs.Namespace)
 		ctrlr := *r.controller
 
-		if err := ctrlr.Watch(&source.Kind{Type: withNamespace(&corev1.ConfigMap{}, rs.Namespace)},
+		if err := ctrlr.Watch(source.Kind(r.cache, withNamespace(&corev1.ConfigMap{}, rs.Namespace)),
```
nan-yu, line 525: We didn't use cache for watches before. I think this also helps make watches more efficient.

pkg/webhook/validate_test.go
```diff
@@ -159,7 +159,7 @@ func TestValidator_Handle(t *testing.T) {
 				core.Annotation(csmetadata.DeclaredFieldsKey, `{"f:metadata":{"f:labels":{"f:app.kubernetes.io/managed-by":{}},"f:annotations":{"f:configmanagement.gke.io/managed":{},"f:configsync.gke.io/manager":{}}},"f:rules":{}}`),
 			),
 			user: configSyncNamespaceReconciler("bookstore", repoSyncName),
-			deny: metav1.StatusReasonUnauthorized,
+			deny: metav1.StatusReasonForbidden,
```
nan-yu, line 162: Do you know why the deny code is changed? Forbidden is quite different from unauthorized.

sdowell: It looks like the logic is changed in controller runtime. To be fair `forbidden` is more accurate than `unauthorized` in this scenario:

- Forbidden means that the user has valid credentials but is not allowed to perform the action
- Unauthorized means that the user does not have valid credentials

nan-yu: make sense

nan-yu (approved):
/lgtm

pkg/hydrate/tool_util.go
```diff
@@ -326,7 +326,11 @@ func ValidateOptions(ctx context.Context, rootDir cmpath.Absolute, apiServerTime
 }
 
 func newClientClient(cfg *rest.Config) (client.Client, error) {
-	mapper, err := apiutil.NewDynamicRESTMapper(cfg)
+	httpClient, err := rest.HTTPClientFor(cfg)
```
karlkfi, line 329: Is there a way to pass this httpclient through to the client.New so it uses the same one? Or does the client.Client not use a rest.Client?

sdowell: They both use `http.Client`. Looks like it can be passed into [client.Options](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/vendor/sigs.k8s.io/controller-runtime/pkg/client/client.go#L42) when invoking `client.New`. If not provided through options, it [calls the same function](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/vendor/sigs.k8s.io/controller-runtime/pkg/client/client.go#L138) to instantiate a new `http.Client`.

So we could probably pass this in to avoid duplicate instantiation

PR #1214, opened 2024-05-07
chore: use trusted golang base image

sdowell:
This updates the golang base image to a trusted base image built by google.


tiffanny29631 (approved):
/lgtm

PR #1219, opened 2024-05-10
Reset declared_resource to 0 when commit updates

tiffanny29631:
The current OpenCensus library keeps a metric stream that is uniquely identified by commit alive even though no new values are issued.

This makes the Otel Collector metricstransform processor receives all time value when calculating the max and gives false output.

The temporary work around is to reset the stream of previous commit to 0 when parser sees new commit.

b/321875474

### DIFF
--- a/e2e/testcases/reconciler_finalizer_test.go
+++ b/e2e/testcases/reconciler_finalizer_test.go
@@ -32,6 +32,7 @@ import (
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/e2e/nomostest/policy"
+	e2eretry "kpt.dev/configsync/e2e/nomostest/retry"
 	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
@@ -523,21 +524,25 @@ func TestReconcileFinalizerReconcileTimeout(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 	nt.T.Cleanup(func() {
-		namespace := &corev1.Namespace{}
-		if err := nt.KubeClient.Get(namespaceNN.Name, namespaceNN.Namespace, namespace); err != nil {
-			if apierrors.IsNotFound(err) { // Happy path - exit
-				return
+		_, err := e2eretry.Retry(30*time.Second, func() error {
+			namespace := &corev1.Namespace{}
+			if err := nt.KubeClient.Get(namespaceNN.Name, namespaceNN.Namespace, namespace); err != nil {
+				if apierrors.IsNotFound(err) { // Happy path - exit
+					return nil
+				}
+				return err // unexpected error
 			}
-			nt.T.Error(err) // unexpected error
-			return
-		}
-		if testutils.RemoveFinalizer(namespace, nomostest.ConfigSyncE2EFinalizer) {
-			// The test failed to remove the finalizer. Remove to enable deletion.
-			if err := nt.KubeClient.Update(namespace); err != nil {
-				nt.T.Error(err)
-			} else {
+			if testutils.RemoveFinalizer(namespace, nomostest.ConfigSyncE2EFinalizer) {
+				// The test failed to remove the finalizer. Remove to enable deletion.
+				if err := nt.KubeClient.Update(namespace); err != nil {
+					return err
+				}
 				nt.T.Log("removed finalizer in test cleanup")
 			}
+			return nil
+		})
+		if err != nil {
+			nt.T.Fatal(err)
 		}
 	})
 	// Add a fake finalizer to the namespace to block deletion
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	"fmt"
 	"testing"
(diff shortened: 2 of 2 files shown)


pkg/declared/resources.go
```diff
@@ -68,6 +68,16 @@ func (r *Resources) Update(ctx context.Context, objects []client.Object, commit
 
 	// Record the declared_resources metric, after parsing but before validation.
 	metrics.RecordDeclaredResources(ctx, commit, len(newObjects))
+	if oldCommit != commit {
+		// Reset the stream value of previous commit to 0 to provide correct input
+		// for the Otel Collector metricstransform processor's MAX aggregation.
+		// This is a temporary fix for issue b/321875474, ensuring the metrics
+		// accurately reflect decreases in declared resources. Eventually we should
+		// migrate to otel-collector-go to use the async gauge so that older streams
+		// won't get updated when there's no longer new values.
```
karlkfi, line 77: Without this "reset", how long does the collector hold onto the old value and continue reporting it?

Is there another way to remove the metric, instead of setting the value to zero?

pkg/parse/run.go
```diff
@@ -491,6 +491,7 @@ func parseAndUpdate(ctx context.Context, p Parser, trigger string, state *reconc
 	klog.V(3).Info("Parser starting...")
 	sourceErrs := parseSource(ctx, p, trigger, state)
 	klog.V(3).Info("Parser stopped")
+	oldCommit := state.sourceStatus.commit
```
karlkfi, line 494: Does this get set from the RSync status or will it be the empty string when the reconciler first starts up?

pkg/declared/resources.go
tiffanny29631: The old values will on get cleared when the reconciler restarts, which is not regular and user needs to manually force it as of now to correct the metric data.

I haven't seen a good way to remove older streams, iiuc we'll need to retain all previous commits to be able to identify the historic streams, which would be unreasonable for some rapid source updating clients. One good way to solve this organically is as stated in comment - migrate to Otel Collector metric SDK and use the async gauge.

pkg/parse/run.go
tiffanny29631: I can add a check for empty string when resetting, are you indicating that the reset when reconciler first started might create a new stream with empty commit?

tiffanny29631:
WIP adding e2e tests

tiffanny29631:
e2e test added to check the resource counts. Also changed the test to look for full list of Config Sync and Resource Group Controller metrics when checking the availability to prevent accidental name change - except the three metrics that are empty without error conditions: `resource_conflict`, `resource_fight`, `internal_error`. We could add more e2e test to create error scenario and verify them in a later PR.

pkg/parse/run.go
karlkfi: Yeah. If it's empty string, we don't know what the previous commit was and can't reset it. We should probably skip sending the reset metric in this case. But it leaves open a gap when if the commit changes when the reconciler is down the metric won't be reset. This could easily happy during upgrade or rescheduling, especially if the repo moves quickly or scheduling is slow (like autopilot) or broken for some other reason.

karlkfi: However, if `state.sourceStatus.commit` is populated from the RSync status before this, then it should work ok. As long as we skip metric reset when the oldCommit is empty.

karlkfi:
> except the three metrics that are empty without error conditions: resource_conflict, resource_fight, internal_error. 

Is there an easy way to initialize these to zero? That would make them show up in Cloud Monitoring, even if there haven't been any errors.

pkg/parse/run.go
tiffanny29631: Breaking down by case:
- A new commit comes in when reconciler down during upgrade - no metric would be sent anyways, and the previous streams will be killed due to the pod restart
- Got oldCommit as empty string - not likely, but maybe when the reconciler first starts up and performing first sync? I added a check for empty string during reset.

tiffanny29631: The state is updated in line 282 before running ParseAndUpdate where the fix takes place, if that's where the sourceStatus gets updated from Rsync state then the oldCommit should be legit?

tiffanny29631:
> > except the three metrics that are empty without error conditions: resource_conflict, resource_fight, internal_error.
> 
> Is there an easy way to initialize these to zero? That would make them show up in Cloud Monitoring, even if there haven't been any errors.

Most solutions are hacky like making the counter increase by 0 to create a fake update, we could probably set it when parser starts.

tiffanny29631:
re @karlkfi the metrics with this issue are the counters, where they only gets recorded when an increment comes in. Just did some test with the aforementioned workaround then Prometheus verification for those metrics started failing, somehow the value vectors were not 0 even though it was the only value given.

tiffanny29631:
Counter seems to be incrementing anyways, even though 0 was given `Metric recorded: { "Name": "internal_errors_total", "Value": 0, "Tags": { {source {cs_counter_initialization {{-1}}}} } }`, not sure if it's due to the int64 value been converted to float when converting to Measurement or the recorder is treating .

Strangely Prometheus reads different number from GCM: Prometheus counts internal_error to 11 http://screen/4BaVpwRsdgghCKc, while GCM counts to 4 http://screen/4nysc8wDM9os5Nu.

It might not be a good option to mock the initial value until further research.

karlkfi:
Can you write up a short doc on this. Or schedule a meeting. I'm confused now. Need to understand the problem(s), possible solutions(s), and tradeoffs.

tiffanny29631:
Added one on calendar, we can circle back on it tomorrow.

PR #1220, opened 2024-05-14
test: use longer waits for managed field tests

sdowell:
The 30 second wait time used by these assertions works most of the time, but can be brittle in certain scenarios - for example if the reconciler Pod is rescheduled.


nan-yu (approved):
/lgtm

karlkfi (approved):
/lgtm

Thanks! The default timeout is the right way to go here, because it's already adjusted for autopilot, which is especially slow to schedule new pods because it often needs to bring up a new node or even a whole new instance group.

PR #1222, opened 2024-05-16
test: print RootSync for truncateSourceError

sdowell:
This is intended to help with debugging when this predicate fails.


nan-yu (approved):
/lgtm

PR #1227, opened 2024-05-16
test: use SSA to create parent RootSync

sdowell:
This instantiates the base RootSync using SSA, which ensures that subsequent SSA calls will use the same field manager and update fields appropriately. This fixes the current TestStressLargeRequest failures.

This also ensures the RootSync spec is reset for every test, which is a desirable property.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -471,7 +471,12 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
 	// understand the Repo and RootSync types as ConfigSync is now installed.
 	nt.RenewClient()
 
-	// Create the RootSync if it doesn't exist, and wait for it to be Synced.
+	// Initialize the base RootSync using SSA for field management
+	rs := RootSyncObjectV1Beta1FromRootRepo(nt, configsync.RootSyncName)
+	if err := nt.KubeClient.Apply(rs); err != nil {
+		nt.T.Fatal(err)
+	}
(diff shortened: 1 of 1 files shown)


karlkfi (approved):
/lgtm

e2e/nomostest/new.go
```diff
@@ -471,7 +471,12 @@ func setupTestCase(nt *NT, opts *ntopts.New) {
...
 	nt.RenewClient()
 
-	// Create the RootSync if it doesn't exist, and wait for it to be Synced.
+	// Initialize the base RootSync using SSA for field management
+	rs := RootSyncObjectV1Beta1FromRootRepo(nt, configsync.RootSyncName)
+	if err := nt.KubeClient.Apply(rs); err != nil {
+		nt.T.Fatal(err)
+	}
+	// Wait for Config Sync to be ready and the base RootSync to be synced.
 	if err := WaitForConfigSyncReady(nt); err != nil {
```
nan-yu, line 480: `WaitForConfigSyncReady` calls `ValidateMultiRepoDeployments`, which creates the RootSync again. Should we remove the CREATE call from `ValidateMultiRepoDeployments`?

sdowell: Yes I think we can probably remove the CREATE call, however `WaitForConfigSyncReady` is called by other code paths - so I figured it's safer to just start with this change since we probably want to cherry pick

PR #1228, opened 2024-05-17
Add tests to cover the behaviour of backoff and jitter on retry

Camila-B:
These tests verify that:
1. Retry attempts stop when the maximum retry limit (12) has been reached
2. Exponential backoff is used
3. The retryTimer doesn't block the events triggered by other timers


pkg/parse/run.go
```diff
@@ -88,11 +98,30 @@ func Run(ctx context.Context, p Parser, nsControllerState *namespacecontroller.S
...
 	defer nsEventTimer.Stop()
 
-	state := &reconcilerState{
-		backoff:     defaultBackoff(),
-		retryTimer:  retryTimer,
-		retryPeriod: opts.RetryPeriod,
+	var state *reconcilerState
+	var runFn RunFn
+
+	if runOpts != nil {
```
karlkfi, line 104: to simplify this method, I'd recommend using a DefaultRunOpts to build the RunOps. Then make the RunOpts a non-pointer in the func receiver so it's required and have the caller call DefaultRunOpts to build the ops when executing the func.

pkg/parse/run.go
```diff
@@ -66,8 +67,17 @@ const (
 	RenderingNotRequired string = "Rendering not required but is currently enabled"
 )
 
+// RunOpts are the options used when calling Run
+type RunOpts struct {
+	runFn           RunFn
```
karlkfi, line 72: IME, func type names usually end in -Func. That way variables and fields can use -Fn without colliding. In this case there's no collision because the type is public, but I'd still recommend using -Func for consistency.

pkg/parse/run_test.go
```diff
@@ -422,3 +430,65 @@ func TestRun(t *testing.T) {
 		})
 	}
 }
+func TestBackoffRetryCount(t *testing.T) {
```
karlkfi, line 433: newline between funcs, should fail linting

pkg/parse/run.go
Camila-B: Done

pkg/parse/run.go
Camila-B: Done

pkg/parse/run_test.go
```diff
@@ -55,7 +56,7 @@ const (
 	symLink = "rev"
 )
 
-func newParser(t *testing.T, fs FileSource, renderingEnabled bool) Parser {
+func newParser(t *testing.T, fs *FileSource, renderingEnabled bool, retryPeriod time.Duration, pollingPeriod time.Duration) Parser {
```
karlkfi, line 59: Passing nil to a constructor is a code smell. It makes it harder to read the constructor invocation without looking at the constructor implementation. 

Similar to the DefaultRunOpts, I would suggest using an EmptyFiles or FakeFiles or some other such Files impl or constructor for the case you're adding, with a well defined behavior, instead of adding conditionals into the constructor.

If you absolutely need optional arguments in a constructor, it's usually preferable to have multiple clearly named constructors. But I don't think that's the case here, since it's just a test that doesn't get to a branch that needs the Files.

Camila-B: Done. I added a function in `run_test.go` that creates an empty FileSource

pkg/parse/run_test.go
```diff
@@ -422,3 +427,84 @@ func TestRun(t *testing.T) {
...
+		}
+
+		if testIsDone(reimportCount, retryCount) {
+			cancelFn()
+		}
+	}
+}
+
+func emptyFileSource() FileSource {
+	return FileSource{
```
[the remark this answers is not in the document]
Camila-B: Good to know, thanks!

karlkfi (approved):
/lgtm

PR #1229, opened 2024-05-20
test: use SSA to create parent RootSync (#1227)

tiffanny29631:
This instantiates the base RootSync using SSA, which ensures that subsequent SSA calls will use the same field manager and update fields appropriately. This fixes the current TestStressLargeRequest failures.

This also ensures the RootSync spec is reset for every test, which is a desirable property.


sdowell (approved):
/lgtm

PR #1230, opened 2024-05-21
Add retry when cleanup test finalizers

tiffanny29631:


karlkfi (approved):
/lgtm

Thanks!

tiffanny29631:
I tried manually patching the persisting deployments but it doesn't help with the RootSync object, is there a good way to clean it up?

PR #1231, opened 2024-05-21
e2e: attempt to uninstall packages in cleanup

karlkfi:
If uninstall fails or times out, continue with existing more aggressive cleanup. This may take longer, but should give Config Sync a chance to clean up objects following reverse dependency ordering. Hopefully, this will help avoid race conditions and deadlocks.

### DIFF
--- a/e2e/nomostest/clean.go
+++ b/e2e/nomostest/clean.go
@@ -38,6 +38,7 @@ import (
 	"kpt.dev/configsync/pkg/api/configmanagement"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/reconcilermanager"
@@ -45,6 +46,7 @@ import (
 	"kpt.dev/configsync/pkg/util"
 	"kpt.dev/configsync/pkg/webhook/configuration"
 	"sigs.k8s.io/cli-utils/pkg/common"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -68,6 +70,18 @@ func Clean(nt *NT) error {
 		nt.T.Logf("[CLEANUP] Test environment cleanup took %v", elapsed)
 	}()
 
+	// If the reconciler-manager is running, attempt to uninstall packages.
+	// If not running, continue with more aggressive cleanup.
+	// This should help avoid deletion ordering edge cases.
+	running, err := isReconcilerManagerHealthy(nt)
+	if err != nil {
+		return err
+	}
+	if running {
+		if err := uninstallAllPackages(nt); err != nil {
+			nt.T.Errorf("[CLEANUP] failed to uninstall packages: %v", err)
+		}
+	}
 	// Delete lingering APIService, as this will cause API discovery to fail
 	if err := deleteTestAPIServices(nt); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


e2e/nomostest/clean.go
```diff
@@ -126,6 +133,88 @@ func Clean(nt *NT) error {
...
+		if !obj.GetDeletionTimestamp().IsZero() {
+			nt.T.Logf("[CLEANUP] already terminating %s object %s ...", gvk.Kind, nn)
+		} else {
+			nt.T.Logf("[CLEANUP] deleting %s object %s ...", gvk.Kind, nn)
+			if err := nt.KubeClient.Delete(obj, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+				if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
+					// skip waiting
+					continue
+				}
+				return fmt.Errorf("unable to delete %s object %s: %w",
```
sdowell, line 194: The early return here and on line 183 means we wouldn't wait for any other elements of the `taskgroup` to complete before returning. I think we want to continue and defer returning the error, so that we wait for other goroutines to complete before exiting the function.

This is what I have in mind to elegantly group the errors:
```golang
			if err := nt.KubeClient.Delete(obj, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
				if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
					// skip waiting
					continue
				}
                tg.Go(func() error {
				  return fmt.Errorf("unable to delete %s object %s: %w",
					  gvk.Kind, nn, err)
                })
                continue
			}
```

e2e/nomostest/clean.go
```diff
@@ -126,6 +133,88 @@ func Clean(nt *NT) error {
...
+	nt.T.Logf("[CLEANUP] listing %s objects...", gvk.Kind)
+	rsObjList := &unstructured.UnstructuredList{}
+	rsObjList.SetGroupVersionKind(kinds.ListGVKForItemGVK(gvk))
+	if err := nt.KubeClient.List(rsObjList, client.InNamespace(configmanagement.ControllerNamespace)); err != nil {
+		return fmt.Errorf("listing %s: %v", gvk.Kind, err)
+	}
+	// Find the unmanaged packages
+	var rsObjs []*unstructured.Unstructured
+	for _, item := range rsObjList.Items {
+		if core.GetAnnotation(&item, metadata.ResourceManagementKey) == metadata.ResourceManagementEnabled {
```
sdowell, line 172: I think the intent is `core.GetAnnotation(&item, metadata.ResourceManagementKey) != metadata.ResourceManagementEnabled` to get *unmanaged* objects

e2e/nomostest/clean.go
```diff
@@ -126,6 +133,88 @@ func Clean(nt *NT) error {
...
+			}
+		}
+		tg.Go(func() error {
+			nt.T.Logf("[CLEANUP] Waiting for deletion or finalizer failure of %s object %s ...", gvk.Kind, nn)
+			return nt.Watcher.WatchObject(gvk, nn.Name, nn.Namespace,
+				[]testpredicates.Predicate{
+					testpredicates.Or(nt.Logger,
+						testpredicates.ObjectNotFoundPredicate(nt.Scheme),
+						testpredicates.StatusEquals(nt.Scheme, kstatus.FailedStatus),
+						testpredicates.RootSyncHasConditionStatus(v1beta1.RootSyncReconcilerFinalizerFailure, metav1.ConditionTrue),
```
tiffanny29631, line 205: Would this predicate apply to RepoSyncs as well?

e2e/nomostest/clean.go
karlkfi: Done.

e2e/nomostest/clean.go
karlkfi: Fixed.

e2e/nomostest/clean.go
karlkfi: Fixed. Made it generic.

e2e/nomostest/clean.go
```diff
@@ -126,6 +133,94 @@ func Clean(nt *NT) error {
...
+		} else {
+			nt.T.Logf("[CLEANUP] Deleting %s object %s ...", gvk.Kind, nn)
+			if err := nt.KubeClient.Delete(obj, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+				if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
+					// skip waiting
+					nt.T.Logf("[CLEANUP] Confirmed removal of %s object %s", gvk.Kind, nn)
+					continue
+				}
+				// return an error, but don't prevent other deletes or waits
+				tg.Go(func() error {
```
sdowell, line 193: I think we need to `continue` after this, otherwise it will add the wait for not found task to the wait group on line 199. If the delete operation failed, I don't think we want to wait for the object to be not found.

e2e/nomostest/clean.go
```diff
@@ -126,6 +133,95 @@ func Clean(nt *NT) error {
...
+		}
+		tg.Go(func() error {
+			nt.T.Logf("[CLEANUP] Waiting for removal or failure of %s object %s ...", gvk.Kind, nn)
+			err := nt.Watcher.WatchObject(gvk, nn.Name, nn.Namespace,
+				[]testpredicates.Predicate{
+					testpredicates.Or(nt.Logger,
+						testpredicates.ObjectNotFoundPredicate(nt.Scheme),
+						testpredicates.StatusEquals(nt.Scheme, kstatus.FailedStatus),
+						testpredicates.HasConditionStatus(nt.Scheme,
+							string(v1beta1.RootSyncReconcilerFinalizerFailure), corev1.ConditionTrue),
```
sdowell, line 208: This string should probably change based on whether the GVK is RootSync or RepoSync. It happens that the `RootSyncReconcilerFinalizerFailure`/`RepoSyncReconcilerFinalizerFailure` constants currently have the same value, but I don't think this should be assumed given they are separate constants. Alternatively, we could consolidate the constants to something like `ReconcilerFinalizerFailure` if the intent is for them to have the same value.

karlkfi: Pulled out the conditions above the loop and added a switch for this.

e2e/nomostest/clean.go
karlkfi: Added

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1353,6 +1386,49 @@ func ClusterRoleBindingSubjectNamesEqual(subjects ...string) func(o client.Objec
 	}
 }
 
+// And returns a Predicate that succeeds only if all of the Predicates succeed.
+// If any one of the specified Predicates errors, the And Predicate errors.
+// Logs the number of predicates, if successful.
+func And(logger *testlogger.TestLogger, predicates ...Predicate) Predicate {
```
sdowell, line 1392: Is `And` needed? This is basically the default behavior if you pass in a list of predicates right?

e2e/nomostest/clean.go
sdowell: Looks better, thanks

e2e/nomostest/testpredicates/predicates.go
karlkfi: Removed

e2e/nomostest/clean.go
```diff
@@ -68,6 +70,11 @@ func Clean(nt *NT) error {
 		nt.T.Logf("[CLEANUP] Test environment cleanup took %v", elapsed)
 	}()
 
+	if err := uninstallAllPackages(nt); err != nil {
```
sdowell, line 73: I think this is going to fail whenever Clean is called [here](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/new.go#L351), because the reconciler-manager Deployment will be not found. This will impact periodics because they set `--create-clusters=lazy`

karlkfi: Pulled out an `isReconcilerManagerHealthy` func to check that first without failing the test when it's unhealthy/uninstalled.

sdowell (approved):
/lgtm

PR #1232, opened 2024-05-22
[metric] #3 Reset the declared_resources metric when at a new commit [restored]

tiffanny29631:
See the falsely closed PR https://github.com/GoogleContainerTools/kpt-config-sync/pull/1219 for context.

### DIFF
--- a/e2e/testcases/otel_collector_test.go
+++ b/e2e/testcases/otel_collector_test.go
@@ -258,12 +258,60 @@ func TestGCMMetrics(t *testing.T) {
 	if err != nil {
 		nt.T.Fatal(err)
 	}
-	_, err = retry.Retry(60*time.Second, func() error {
+	_, err = retry.Retry(120*time.Second, func() error {
 		var err error
 		for _, metricType := range GCMMetricTypes {
 			descriptor := fmt.Sprintf("%s/%s", GCMMetricPrefix, metricType)
 			it := listMetricInGCM(ctx, nt, client, startTime, descriptor)
 			err = multierr.Append(err, validateMetricInGCM(nt, it, descriptor, nt.ClusterName))
+		}
+		return err
+	})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Adding test namespace")
+	namespace := fake.NamespaceObject("foo")
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/ns.yaml", namespace))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Adding foo namespace"))
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Checking resource related metrics after adding test resource")
+	resourceMetrics := []string{
+		csmetrics.DeclaredResourcesName,
+		rgmetrics.ResourceCountName,
+		rgmetrics.ReadyResourceCountName,
+	}
+	_, err = retry.Retry(120*time.Second, func() error {
+		var err error
+		for _, metricType := range resourceMetrics {
+			descriptor := fmt.Sprintf("%s/%s", GCMMetricPrefix, metricType)
+			it := listMetricInGCM(ctx, nt, client, startTime, descriptor)
+			err = multierr.Append(err, validateMetricInGCM(nt, it, descriptor, nt.ClusterName, metricHasValue(3)))
+		}
+		return err
+	})
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Remove the test resource")
+	nt.Must(nt.RootRepos[configsync.RootSyncName].Remove("acme/ns.yaml"))
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Remove the test namespace"))
+	if err := nt.WatchForAllSyncs(); err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Checking resource related metrics after removing test resource")
+	_, err = retry.Retry(120*time.Second, func() error {
+		var err error
+		for _, metricType := range resourceMetrics {
+			descriptor := fmt.Sprintf("%s/%s", GCMMetricPrefix, metricType)
+			it := listMetricInGCM(ctx, nt, client, startTime, descriptor)
+			err = multierr.Append(err, validateMetricInGCM(nt, it, descriptor, nt.ClusterName, metricHasLastestValue(2)))
 		}
 		return err
 	})
@@ -448,6 +496,29 @@ func metricDoesNotHaveLabel(label string) metricValidatorFunc {
 			return fmt.Errorf("expected metric to not have label, but found %s=%s", label, value)
 		}
 		return nil
+	}
+}
+
+func metricHasValue(value int64) metricValidatorFunc {
+	return func(series *monitoringpb.TimeSeries) error {
+		points := series.GetPoints()
+		for _, point := range points {
+			if point.GetValue().GetInt64Value() == value {
+				return nil
+			}
+		}
+		return fmt.Errorf("expected metric to have value %v but did not find in response", value)
+	}
+}
+
+func metricHasLastestValue(value int64) metricValidatorFunc {
+	return func(series *monitoringpb.TimeSeries) error {
+		points := series.GetPoints()
+		lastPoint := points[len(points)-1]
+		if lastPoint.GetValue().GetInt64Value() == value {
+			return nil
+		}
+		return fmt.Errorf("expected metric to have latest value %v but did not find in response", value)
 	}
 }
 
--- a/pkg/declared/resources.go
+++ b/pkg/declared/resources.go
@@ -42,10 +42,14 @@ type Resources struct {
 	objectMap *orderedmap.OrderedMap[core.ID, *unstructured.Unstructured]
 	// commit of the source in which the resources were declared
 	commit string
+	// previousCommit is the preceding commit to the commit
+	previousCommit string
 }
 
 // Update performs an atomic update on the resource declaration set.
 func (r *Resources) Update(ctx context.Context, objects []client.Object, commit string) ([]client.Object, status.Error) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
 	// First build up the new map using a local pointer/reference.
 	newSet := orderedmap.NewOrderedMap[core.ID, *unstructured.Unstructured]()
 	newObjects := []client.Object{}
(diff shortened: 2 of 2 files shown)


tiffanny29631:
Context of using rate() instead of present_over_time() in PromQL: https://github.com/prometheus/prometheus/issues/8255 the first proposal of adding a present_over_time() function to represent if a timeserie exists in the time frame.

In our case the error metrics (i.e. resource_fights_total) will always exist even though the value of it does not change when no resource fights are happening, which will make the present_over_time() always returning 1 http://screen/4wdqXrJqAUFP7PZ. 

Using rate() instead can give us an idea whether a fight was recorded within the queried time frame - if no fights are recorded, the record rate should go back to 0.

e2e/nomostest/prometheus_metrics.go
```diff
@@ -344,56 +344,54 @@ func metricResourceFightsHasValueAtLeast(nt *NT, syncLabels prometheusmodel.Labe
 		labels := prometheusmodel.LabelSet{
 			prometheusmodel.LabelName(ocmetrics.KeyComponent.Name()): prometheusmodel.LabelValue(ocmetrics.OtelCollectorName),
 		}.Merge(syncLabels)
-		// ResourceFightsView counts the total number of ResourceFights, so we don't need to aggregate
-		query := fmt.Sprintf("%s%s", metricName, labels)
 		if value == 0 {
-			// Tolerate missing metrics when expecting a zero value.
-			// Don't allow any value other than zero.
-			return metricExistsWithValueOrDoesNotExist(ctx, nt, v1api, query, float64(value))
+			query := fmt.Sprintf("rate(%s%s[1m])", metricName, labels)
```
karlkfi, line 348: Does this mean "within the last minute"?

e2e/nomostest/prometheus_metrics.go
```diff
@@ -344,56 +344,54 @@ func metricResourceFightsHasValueAtLeast(nt *NT, syncLabels prometheusmodel.Labe
...
 func metricResourceConflictsHasValueAtLeast(nt *NT, syncLabels prometheusmodel.LabelSet, commitHash string, value int) MetricsPredicate {
 	return func(ctx context.Context, v1api prometheusv1.API) error {
 		metricName := ocmetrics.ResourceConflictsView.Name
 		metricName = fmt.Sprintf("%s%s", prometheusConfigSyncMetricPrefix, metricName)
+		if value == 0 {
+			labels := prometheusmodel.LabelSet{
+				prometheusmodel.LabelName(ocmetrics.KeyComponent.Name()): prometheusmodel.LabelValue(ocmetrics.OtelCollectorName),
+				prometheusmodel.LabelName(ocmetrics.KeyCommit.Name()):    ocmetrics.Initialization,
+			}.Merge(syncLabels)
+			query := fmt.Sprintf("rate(%s%s[30s])", metricName, labels)
```
karlkfi, line 368: Why use 30s here and 1m for the other call? Seems like they should be the same, and maybe configurable by serializing a time.Durration argument.

e2e/nomostest/prometheus_metrics.go
```diff
@@ -344,56 +344,54 @@ func metricResourceFightsHasValueAtLeast(nt *NT, syncLabels prometheusmodel.Labe
...
-		// ResourceFightsView counts the total number of ResourceFights, so we don't need to aggregate
-		query := fmt.Sprintf("%s%s", metricName, labels)
 		if value == 0 {
-			// Tolerate missing metrics when expecting a zero value.
-			// Don't allow any value other than zero.
-			return metricExistsWithValueOrDoesNotExist(ctx, nt, v1api, query, float64(value))
+			query := fmt.Sprintf("rate(%s%s[1m])", metricName, labels)
+			return metricExistsWithValue(ctx, nt, v1api, query, 0)
 		}
+		query := fmt.Sprintf("%s%s", metricName, labels)
```
karlkfi, line 351: why are we using rate for `metricExistsWithValue` but not `metricExistsWithValueAtLeast`?

pkg/metrics/otel.go
```diff
@@ -129,7 +131,7 @@ processors:
             aggregation_type: max
       - include: declared_resources
         action: update
-        new_name: current_declared_resources
+        new_name: declared_resources
```
karlkfi, line 134: Can this be a different PR?

pkg/metrics/otel.go
```diff
@@ -96,7 +96,9 @@ processors:
   filter/cloudmonitoring:
     metrics:
       include:
-        match_type: regexp
+        # Use strict match type to ensure metrics like 'kustomize_resource_count'
+        # is excluded
+        match_type: strict
```
karlkfi, line 101: Can this be a different PR?

pkg/metrics/metrics.go
```diff
@@ -34,73 +65,73 @@ var (
...
 	DeclaredResources = stats.Int64(
-		"declared_resources",
+		DeclaredResourcesName,
 		"The number of declared resources parsed from Git",
 		stats.UnitDimensionless)
 
 	// ApplyOperations metric measures the number of applier apply events.
 	ApplyOperations = stats.Int64(
-		"apply_operations",
+		ApplyOperationsName,
```
karlkfi, line 98: This looks like it changes the name to include the "_total" suffix. Was that intentional? Same with the other totals

pkg/metrics/record.go
```diff
@@ -27,6 +27,9 @@ import (
 	"kpt.dev/configsync/pkg/status"
 )
 
+// Initialization is the tag value to initiate a Count
+const Initialization = "cs_counter_initialization"
```
karlkfi, line 31: Seems weird to just have one of these. might make sense as an enum of operations? Or maybe we should just remove the dead arg, since we don't send it any more.

karlkfi: Shouldn't RecordResourceFight send 0 instead of 1 when initializing?

pkg/declared/resources.go
```diff
@@ -45,7 +45,7 @@ type Resources struct {
 }
 
 // Update performs an atomic update on the resource declaration set.
-func (r *Resources) Update(ctx context.Context, objects []client.Object, commit string) ([]client.Object, status.Error) {
+func (r *Resources) Update(ctx context.Context, objects []client.Object, commit string, oldCommit string) ([]client.Object, status.Error) {
```
karlkfi, line 48: What if we store the commit in the Resources? That way when Update is called we have the old commit on the struct and the new commit coming in as an arg.

e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
```diff
@@ -0,0 +1,326 @@
+# Copyright 2022 Google LLC
```
karlkfi, line 1: Do we need this as a test artifact? Can it be generated from code so we make sure it's the same as what a user gets? Or is there something different about it for testing?

e2e/nomostest/prometheus_metrics.go
tiffanny29631: That is my understanding

e2e/nomostest/prometheus_metrics.go
tiffanny29631: I was manually tuning each so forgot to align the interval, we could pass an argument duration but then need to convert it to the verbal time with unit.. Or we just use seconds to make things easier.

e2e/nomostest/prometheus_metrics.go
tiffanny29631: When it falls in value == 0 check,  the rsync summary passed in contains no conflict / fight / internal error. In this case we want the recording rate of this metric to be strictly 0 to indicate that no conflict / fight / internal error happening.

When the given value is >0, the aforementioned error has happened and is gonna increment as retry happens, so we verify that the actual counter is at least the number of times that we think it's happening. Rate in this case would just report a number like 0.87/s, which doesn't mean much when verifying.

pkg/metrics/otel.go
tiffanny29631: Sure let me modify it a bit, along with the constify

pkg/metrics/metrics.go
tiffanny29631: If you look at view.go it was appending the `_total` suffix before, and that's actually where the metric names are finalized when exporting. The const-fication puts them in one place to rule out the confusion. I've checked both Kmetrics and resource group controller, they two don't have the name discrepancy between metric.go and view.go. Agreed that pulling this out to a separate PR may be better, will cleanup a bit along with the declared_resources chagne.

pkg/parse/run.go
```diff
@@ -78,6 +78,12 @@ type RunFunc func(ctx context.Context, p Parser, trigger string, state *reconcil
 
 // Run keeps checking whether a parse-apply-watch loop is necessary and starts a loop if needed.
 func Run(ctx context.Context, p Parser, nsControllerState *namespacecontroller.State, runOpts RunOpts) {
+	// Pre-initialize count metrics to ensure visibility, even if errors haven't occurred.
+	// Special label indicates initial state. Metrics are cumulative and used for rate aggregation
+	// to detect ongoing errors. Zero value means no recent errors.
+	metrics.RecordResourceFight(ctx, metrics.Initialization)
+	metrics.RecordInternalError(ctx, metrics.Initialization)
+	metrics.RecordResourceConflict(ctx, metrics.Initialization)
```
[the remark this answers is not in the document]
tiffanny29631: I think the initial intention was to make these metrics appear no matter the error conditions have happened or not, otherwise the metrics will be missing until the error condition has happened at least once. And yes, recording them will plus 1 to the counter value.

e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
tiffanny29631: This ConfigMap is a bit different where it takes off the metric filtering and exports all applicable metric to GCM to test the overall behavior, it's definitely different from what the user would get. The user faced behavior is tested in `TestOtelCollectorDeployment`.

e2e/nomostest/prometheus_metrics.go
karlkfi: time.Duration.String() should give you want you need. Prometheus is written in Go anyway.

pkg/parse/run.go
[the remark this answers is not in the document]
karlkfi: This will make these metrics all blip to a rate > 0 whenever the reconciler starts. I don't know if that's a great tradeoff...

Is there a way to set the value to zero and then use a rate query that only matches value=1?

e2e/nomostest/prometheus_metrics.go
karlkfi: I think if we set the value=0 when initializing the metrics and then modify the query to only match value=1 before the rate, then we can unify these.

pkg/parse/run.go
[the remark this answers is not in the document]
tiffanny29631: Yes correct, with this method the rate will go up then resume 0 when reconciler starts. I think we've been through this: once we `set`, it will increment the metric value by 1 whether we set a 0 or not. We could fallback to previous when we tolerate the metric being missing. Or it's a bigger change, we make it a lastValue and give it a meaningful count.

e2e/nomostest/prometheus_metrics.go
tiffanny29631: I'm confused, we are not setting value=0. We could do match value=1 then check rate=0, but then we still need to check for valueAtLeast when there's fight happening.

karlkfi: What I'm saying is that we COULD set value=0 when we initialize the metric. Then it would be possible to ignore those records when querying the rate, by only querying for value=1.

e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
karlkfi: > it takes off the metric filtering and exports all applicable metric to GCM

Do we need to test the metrics that are normally filtered?

If not, we can just use the same ConfigMap the customer gets, right?

tiffanny29631: In theory we supports viewing all the metrics we listed in the public doc in GCM if user is experienced enough to adjust the filter configuration, so it's better to make sure they all work in this test.

pkg/declared/resources.go
tiffanny29631: How to know if a commit is the first one? Will the field be empty when a Resource is first created

e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
karlkfi: Is there any way we can make this into a feature flag that we can enable dynamically, instead of hard-coding another permutation here?

I'm trying to avoid needing to manually keep them in sync.

tiffanny29631: It's possible to generate test ConfigMap and patch the data field as needed

e2e/nomostest/prometheus_metrics.go
tiffanny29631: This part has been moved to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1237, will be making change accordingly.

e2e/nomostest/prometheus_metrics.go
tiffanny29631: This part has been moved to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1237, will be making change accordingly.

pkg/parse/run.go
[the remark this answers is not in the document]
tiffanny29631: This part has been moved to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1237, will be making change accordingly.

e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
tiffanny29631: Will clean up in another PR.

pkg/metrics/record.go
tiffanny29631: This part has been moved to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1237, will be making change accordingly.

pkg/declared/resources.go
karlkfi: Empty string is the default value for any string field.

tiffanny29631: Done! Added a separate field in Resources.

pkg/declared/resources.go
```diff
@@ -42,6 +42,8 @@ type Resources struct {
 	objectMap *orderedmap.OrderedMap[core.ID, *unstructured.Unstructured]
 	// commit of the source in which the resources were declared
 	commit string
+	// previousCommit is the preceding commit to the commit
+	previousCommit string
```
karlkfi, line 46: The objectMap and commit are locked by the mutex. I think we need to do the same for the previousCommit to avoid race conditions. 

But looking at the code, I don't quite understand why it's using a setter and getter. Maybe whoever wrote it (maybe me) wanted to avoid touching all the other methods. 

I think the right way to fix this would be to remove the getter/setter, Lock the Update, and RLock all the other methods.

e2e/testcases/otel_collector_test.go
```diff
@@ -451,6 +499,29 @@ func metricDoesNotHaveLabel(label string) metricValidatorFunc {
 	}
 }
 
+func metricHasValue(value int64) metricValidatorFunc {
+	return func(serie *monitoringpb.TimeSeries) error {
```
karlkfi, line 503: what's a "serie"? missing an s?

e2e/testcases/otel_collector_test.go
```diff
@@ -270,6 +270,54 @@ func TestGCMMetrics(t *testing.T) {
...
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Checking resource related metrics after adding test resource")
+	resourceMetrics := []string{
+		csmetrics.DeclaredResourcesName,
+		rgmetrics.ResourceCountName,
+		rgmetrics.ReadyResourceCountName,
+	}
+	_, err = retry.Retry(60*time.Second, func() error {
```
karlkfi, line 288: `1*time.Minute` but is that long enough?

pkg/declared/resources.go
```diff
@@ -68,6 +70,17 @@ func (r *Resources) Update(ctx context.Context, objects []client.Object, commit
 
 	// Record the declared_resources metric, after parsing but before validation.
 	metrics.RecordDeclaredResources(ctx, commit, len(newObjects))
+	if r.previousCommit != commit && r.previousCommit != "" {
+		// Reset the stream value of previous commit to 0 to provide correct input
```
karlkfi, line 74: > Reset the stream value of previous commit to 0 to provide correct input for the Otel Collector metricstransform processor's MAX aggregation.

So when we added the commit label to the metric it started keeping the latest value for each commit, instead of the latest value for all commits, right? 

What does that have to do with MAX aggregation?

pkg/declared/resources.go
```diff
@@ -68,6 +70,17 @@ func (r *Resources) Update(ctx context.Context, objects []client.Object, commit
...
+		// for the Otel Collector metricstransform processor's MAX aggregation.
+		// This is a temporary fix for issue b/321875474, ensuring the metrics
+		// accurately reflect decreases in declared resources. Eventually we should
+		// migrate to otel-collector-go to use the async gauge so that older streams
```
karlkfi, line 78: > Eventually we should migrate to otel-collector-go to use the async gauge so that older streams won't get updated when there's no longer new values.

Does this mean that with the current gauge metric it will just keep accumulating metrics every time there's a new commit label value and report all of them forever? 

If so, that seems like a pretty problematic memory leak...

pkg/declared/resources.go
tiffanny29631: We could keep the getter / setter to keep appearance clean and just extract the lock to the calling function level? My preference is to have separate getter/setter for each member but it could be just too much overhead.

e2e/testcases/otel_collector_test.go
tiffanny29631: Running the test stand alone shows the 1min timeout is enough, could  make it 120s to reduce possible noice in CI.

e2e/testcases/otel_collector_test.go
tiffanny29631: Done

e2e/testcases/otel_collector_test.go
tiffanny29631: Done

pkg/declared/resources.go
tiffanny29631: My understanding is that the MAX aggregation would be seeing the `latest value for each commit` since the reconciler pod starts, which means a previous max value persists when user removes objects from SoT.

pkg/declared/resources.go
tiffanny29631: I would need to print and check the logs of the timeseries getting written out from otel-agent to see how exactly it's keeping all the streams alive, but the outcome was clear http://b/321875474#comment28.

And if it's a consistent refresh, it does sound like a mem leak or metering fraud, which is not going to be fixed with the sunset of Opencensus. This could be a good reason to migrate to Open Telemetry SDK.

pkg/declared/resources.go
karlkfi: The reason they were put in the same setter is because they need to be updated together atomicly, in the same lock. Same goes for the previousCommit.

And now with previousCommit, the read and write need to be in the same lock. So it doesn't make sense to have a setter.

pkg/declared/resources.go
karlkfi: Oh, right, we're removing the commit label before sending it to GCM, that's why MAX matters here. So it's not expensive for GCM users, because it's aggregated down to one record, but it's probably expensive for Prometheus users. And it's probably a memory leak in the otel-collector and maybe the otel-agent, which buffers them all.

I'm beginning to think that this workaround maybe isn't worth doing, since it's tangential to the actual problem, and we'll have to rip it out later when we actually fix it.

tiffanny29631: It's less risk of a fix for a patch release as we are getting questions from cx.

tiffanny29631: I've attached the TODO b/ in the comment as a reminder.

pkg/declared/resources.go
karlkfi, line 74: Suggested comment:

For Cloud Monitoring, we have configured otel-collector to remove the commit label, to reduce cardinality, but this aggregation uses the max value (b/321875474). So in order for the latest commit to be chosen as the max value, we reset the previous commit value to zero.
TODO: Remove this workaround after migrating to the otel-collector metrics client and switching from gauge to async gauge

pkg/declared/resources.go
tiffanny29631: Okie, I've lifted the r/w lock out of the getter/setter.

pkg/declared/resources.go
tiffanny29631: Done!

pkg/declared/resources.go
```diff
@@ -68,95 +70,85 @@ func (r *Resources) Update(ctx context.Context, objects []client.Object, commit
...
+	if r.previousCommit != commit && r.previousCommit != "" {
+		// For Cloud Monitoring, we have configured otel-collector to remove the
+		// commit label, to reduce cardinality, but this aggregation uses the max
+		// value (b/321875474). So in order for the latest commit to be chosen as
+		// the max value, we reset the previous commit value to zero.
+		// TODO: Remove this workaround after migrating to the otel-collector metrics client and switching from gauge to async gauge
+		metrics.RecordDeclaredResources(ctx, r.previousCommit, 0)
+	}
+
+	r.mutex.Lock()
```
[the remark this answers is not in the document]
tiffanny29631: Hmm yeah just did and tried with the same failing test, still not helping. `TestDontDeleteAllNamespaces`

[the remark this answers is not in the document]
tiffanny29631: Oh it's the order of deleteNS and update members. This time it should work.

karlkfi (approved):
/lgtm

PR #1234, opened 2024-05-28
[metric] #1 Turn metric names to const

tiffanny29631:
Some Config Sync metrics were appended _total suffix in the view file where the metric name finalizes, which creates confusion.

This change puts the naming in one place to reduce confusion.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -16,16 +16,47 @@ package metrics
 
 import "go.opencensus.io/stats"
 
+const (
+	// APICallDurationName is the name of API duration metric
+	APICallDurationName = "api_duration_seconds"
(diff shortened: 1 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -16,16 +16,47 @@ package metrics
 
 import "go.opencensus.io/stats"
 
+const (
+	// APICallDurationName is the name of API duration metric
+	APICallDurationName = "api_duration_seconds"
```
karlkfi, line 21: Do we hard code these names in the e2e tests? Or are we using things like `APICallDuration.Name()`?

tiffanny29631: They are matched by the .Name() functions, not the string value.

karlkfi (approved):
/lgtm

PR #1237, opened 2024-05-28
[WIP] Initialize count metrics on reconciler start

tiffanny29631:
The Count views won't show up in query results until data is sent when error
condition happens.

This change initialize these metrics and adjust the e2e test to check
for Rate aggregation.

The public documentation should also be revised to recommend the Rate
aggregation for this type of metrics.


tiffanny29631:
> Q: is this still relevant or abandoned?

b/348053947

tiffanny29631:
Since this approach may result in inaccurate data being sent on reconciler start/restart, and customers are already informed about the current metric behavior, closing this PR for now. We should revisit this issue and address it properly as part of the migration from OpenCensus to OpenTelemetry.

PR #1238, opened 2024-05-30
test: log cluster hash and version along with the cluster name

nan-yu:
b/343553185

### DIFF
--- a/e2e/nomostest/clusters/kind.go
+++ b/e2e/nomostest/clusters/kind.go
@@ -136,6 +136,11 @@ func (c *KindCluster) Connect() error {
 	return c.provider.ExportKubeConfig(c.Name, c.KubeConfigPath, false)
 }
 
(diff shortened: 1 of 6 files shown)


e2e/nomostest/clusters/kind.go
```diff
@@ -136,6 +136,11 @@ func (c *KindCluster) Connect() error {
 	return c.provider.ExportKubeConfig(c.Name, c.KubeConfigPath, false)
 }
 
+// Hash returns empty string for kind cluster
+func (c *KindCluster) Hash() (string, error) {
+	return "", nil
```
karlkfi, line 141: maybe something like "<KinD>" or something so it's not just empty when logged.

nan-yu: done

karlkfi (approved):
/lgtm

PR #1240, opened 2024-05-30
feat: add OSS migration logic

sdowell:
This change updates the nomos migrate command to support migrating from a ConfigManagement installation to a standalone Config Sync installation. This builds on the existing logic, such that first it will migrate from mono-repo to multi-repo if applicable.

### DIFF
--- a/cmd/nomos/migrate/configmanagement.go
+++ b/cmd/nomos/migrate/configmanagement.go
@@ -1,0 +1,208 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migrate
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"kpt.dev/configsync/cmd/nomos/status"
+	"kpt.dev/configsync/cmd/nomos/util"
+	"kpt.dev/configsync/pkg/api/configmanagement"
+	"kpt.dev/configsync/pkg/kinds"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+)
+
+func migrateConfigManagement(ctx context.Context, cc *status.ClusterClient, kubeCtx string) error {
+	isOSS, err := util.IsOssInstallation(ctx, cc.ConfigManagement, cc.Client, cc.K8sClient)
+	if err != nil {
+		return err
+	}
+	if isOSS {
+		printNotice("The cluster is already running as an OSS installation.")
+		return nil
+	}
+
+	fmt.Printf("Removing ConfigManagement on cluster %q ...\n", kubeCtx)
+	cmYamlFile, err := saveConfigManagementOperatorYaml(ctx, cc, kubeCtx)
+	if err != nil {
+		return err
+	}
+	printHint(`Resources for the ConfigManagement operator have been saved in a temp folder. If the migration process is terminated, it can be recovered manually by running the following commands:
+  kubectl delete deployment -n config-management-system config-management-operator --ignore-not-found --cascade=foreground && \
+  (kubectl patch configmanagement config-management --type="merge" -p '{"metadata":{"finalizers":[]}}' ; \
+  kubectl delete configmanagement config-management --cascade=orphan --ignore-not-found) && \
+  kubectl delete -f %s --ignore-not-found`, cmYamlFile)
+
+	if dryRun {
+		dryrunConfigManagement()
+		return nil
+	}
+	if err := executeConfigManagementMigration(ctx, cc); err != nil {
+		return err
+	}
+	return nil
+}
+
+func dryrunConfigManagement() {
+	printInfo(deletingConfigManagement)
+}
+
+// getOperatorObjects returns a list of objects associated with the ConfigManagement
+// operator installation. The objects are returned in the order in which they
+// should be applied. They should be deleted in the reverse order.
+func getOperatorObjects() []*unstructured.Unstructured {
+	var objs []*unstructured.Unstructured
+
+	crd := &unstructured.Unstructured{}
+	crd.SetGroupVersionKind(kinds.CustomResourceDefinitionV1())
+	crd.SetName(util.ConfigManagementCRDName)
+	objs = append(objs, crd)
+
+	cm := &unstructured.Unstructured{}
+	cm.SetGroupVersionKind(kinds.ConfigManagement())
+	cm.SetName(util.ConfigManagementName)
+	objs = append(objs, cm)
+
+	sa := &unstructured.Unstructured{}
+	sa.SetGroupVersionKind(kinds.ServiceAccount())
+	sa.SetName(util.ACMOperatorDeployment)
+	sa.SetNamespace(configmanagement.ControllerNamespace)
+	objs = append(objs, sa)
+
+	cr := &unstructured.Unstructured{}
+	cr.SetGroupVersionKind(kinds.ClusterRole())
+	cr.SetName(util.ACMOperatorDeployment)
+	objs = append(objs, cr)
+
+	crb := &unstructured.Unstructured{}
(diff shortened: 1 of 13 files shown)


sdowell:
> Would be good to have a doc that explains how to use this and what it does

This doc exists: https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/how-to/migrate-multi-repo

Do you mean a doc in the repo?

nan-yu:
After reviewing the ConfigManagement object, it appears most fields have been handled gracefully by the migration flow, except for `clusterName` and `ConfigSyncLogLevel`.

The operator updates the reconciler Deployment directly by appending the `--cluster-name` argument to the container args. However, the reconciler Deployment manifest in the ConfigMap is not updated.

This could cause issues if the reconciler-management Pod restarts. I think we need to handle that in the migration logic. Similarly for the `ConfigSyncLogLevel`.

sdowell:
> After reviewing the ConfigManagement object, it appears most fields have been handled gracefully by the migration flow, except for `clusterName` and `ConfigSyncLogLevel`.
> 
> The operator updates the reconciler Deployment directly by appending the `--cluster-name` argument to the container args. However, the reconciler Deployment manifest in the ConfigMap is not updated.
> 
> This could cause issues if the reconciler-management Pod restarts. I think we need to handle that in the migration logic. Similarly for the `ConfigSyncLogLevel`.

The cluster name and log level arguments are set on the `reconciler-manager` Deployment, so this should be robust to `reconciler-manager` Pod restarts. I think the only case where the arguments would be lost is if the `reconciler-manager` Deployment itself was deleted after a migration.

nan-yu:
> > After reviewing the ConfigManagement object, it appears most fields have been handled gracefully by the migration flow, except for `clusterName` and `ConfigSyncLogLevel`.
> > The operator updates the reconciler Deployment directly by appending the `--cluster-name` argument to the container args. However, the reconciler Deployment manifest in the ConfigMap is not updated.
> > This could cause issues if the reconciler-management Pod restarts. I think we need to handle that in the migration logic. Similarly for the `ConfigSyncLogLevel`.
> 
> The cluster name and log level arguments are set on the `reconciler-manager` Deployment, so this should be robust to `reconciler-manager` Pod restarts. I think the only case where the arguments would be lost is if the `reconciler-manager` Deployment itself was deleted after a migration.

Ah, that's right. The reconciler-manager propagates the parameters to the reconciler. 

In terms of reconciler-managment Deployment being deleted, I think that's out of the scope of migration.

cmd/nomos/migrate/configmanagement.go
```diff
@@ -0,0 +1,233 @@
...
+			err := sc.Client.Get(ctx, key, obj)
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		})
+		if err != nil {
+			return err
+		}
+		// If this object is the Deployment, wait for all of its Pods to be deleted
```
nan-yu, line 174: It uses the `foreground` propagation policy, so Pod should have been terminated before the Deployment is gone. Do we need to check again?

e2e/testcases/cli_test.go
```diff
@@ -1276,3 +1285,311 @@ func TestApiResourceFormatting(t *testing.T) {
...
+	nt := nomostest.New(t, nomostesting.NomosCLI, ntopts.SkipConfigSyncInstall)
+
+	nt.T.Cleanup(func() {
+		// Restore state of Config Sync installation after test
+		if err := nomostest.InstallConfigSync(nt); err != nil {
+			nt.T.Fatal(err)
+		}
+	})
+	nt.T.Cleanup(func() {
+		// Delete the ConfigManagement operator in case the test failed early.
```
nan-yu, line 1299: This deletes the operator only, but will leave other resources (service account, cluster role and cluster rolebinding) around. We can delete `cmOperatorYaml` first and then re-install Config Sync.

e2e/testcases/cli_test.go
```diff
@@ -1276,3 +1285,311 @@ func TestApiResourceFormatting(t *testing.T) {
...
+
+	cmObj := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "configmanagement.gke.io/v1",
+			"kind":       "ConfigManagement",
+			"metadata": map[string]interface{}{
+				"name": "config-management",
+			},
+			"spec": map[string]interface{}{
+				"enableMultiRepo": true,
```
nan-yu, line 1454: We can start with monorepo mode initially.

cmd/nomos/migrate/configmanagement.go
sdowell: Surprisingly from my testing it doesn't seem like this waits for the Pods to be terminated. When using `kubectl delete --cascase=foreground` it does seem to wait, I'm not sure why this client doesn't wait.

e2e/testcases/cli_test.go
sdowell: It's intentional to start with multi-repo mode here, because this causes the ACM operator to take ownership of the multi-repo objects. This means when it switches to mono-repo mode the multi-repo objects will be deleted.

If this starts with monorepo mode initially, then the multi-repo objects will still be installed alongside the mono-repo objects. The alternative is to write some code to manually clean up all of the multi-repo objects, but this seemed simpler.

e2e/testcases/cli_test.go
sdowell: It's generally not as simple as just deleting `cmOperatorYaml`, because there are things like finalizers on ConfigManagement which can stall deletion.

The happy path is that this cleanup is a no-op because the migration was successful. This is a best-effort cleanup for the unhappy path where `nomos migrate` failed for some reason. A comprehensive cleanup logic would look something like the existing logic in `nomos migrate`, which I think is overkill for the purposes of test cleanup here. If the nomos-operator Deployment is not running then there shouldn't be any fights with the standalone Config Sync installed by the tests, which should be sufficient for the tests to continue.

cmd/nomos/migrate/configmanagement.go
janetkuo: This `Delete` always sets policy to background:
https://github.com/sdowell/kpt-config-sync/blob/814150dd136ed9c92c4fee48cb57e5626ab3c4b7/pkg/syncer/client/client.go#L81-L83
```go
// Delete deletes the given obj from Kubernetes cluster and records prometheus metrics.
// This automatically sets the propagation policy to always be "Background".
func (c *Client) Delete(ctx context.Context, obj client.Object, opts ...client.DeleteOption) status.Error {
```

sdowell:
> It's because this Delete always sets policy to background:
https://github.com/sdowell/kpt-config-sync/blob/814150dd136ed9c92c4fee48cb57e5626ab3c4b7/pkg/syncer/client/client.go#L81-L83

@janetkuo I don't think this explains it, because the client created for `nomos migrate` is the [direct controller-runtime Client](https://github.com/sdowell/kpt-config-sync/blob/814150dd136ed9c92c4fee48cb57e5626ab3c4b7/cmd/nomos/status/client.go#L485).

The only place that the syncer.client is created is [in the reconciler](https://github.com/sdowell/kpt-config-sync/blob/814150dd136ed9c92c4fee48cb57e5626ab3c4b7/pkg/reconciler/reconciler.go#L164)

cmd/nomos/migrate/configmanagement.go
```diff
@@ -0,0 +1,226 @@
...
+
+// executeConfigManagementMigration performs the migration from ConfigManagement
+// installation to a standalone OSS Config Sync installation. It's assumed that
+// the ConfigManagement installation is already running in the multi-repo mode.
+func executeConfigManagementMigration(ctx context.Context, sc *status.ClusterClient) error {
+	printInfo(waitingForRootSyncCRD)
+	if err := waitForMultiRepoCRDsToBeEstablished(ctx, sc.Client); err != nil {
+		return err
+	}
+	printInfo("The Config Sync CRDs have been established")
```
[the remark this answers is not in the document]
sdowell: Done

cmd/nomos/migrate/configmanagement.go
sdowell: Fixed a bug in the NotFound check above, and removed the specific Pod check

cmd/nomos/migrate/migrate.go
```diff
@@ -45,23 +45,28 @@ import (
...
+	monorepoMigrateDir         = "nomos-migrate-monorepo"
+	configManagementMigrateDir = "nomos-migrate-configmanagement"
+	rootSyncYamlFile           = "root-sync.yaml"
+	cmOrigYAMLFile             = "cm-original.yaml"
+	cmOperatorYAMLFile         = "config-management-operator.yaml"
+	cmMultiYAMLFile            = "cm-multi.yaml"
 
 	updatingConfigManagement    = "Updating the ConfigManagement object ..."
-	waitingForRootSyncCRD       = "Waiting for the RootSync CRD to be established ..."
+	waitingForRootSyncCRD       = "Waiting for ConfigSync CRDs to be established ..."
```
nan-yu, line 56: Suggest renaming the variable to match the content.

sdowell: done

e2e/testcases/cli_test.go
sdowell: done

nan-yu (approved):
/lgtm

PR #1241, opened 2024-05-30
Use field managers for all K8s API calls

karlkfi:
- Add component-specific field managers: configsync.gke.io/resource-group-controller configsync.gke.io/nomos-cli configsync.gke.io/reconciler-manager configsync.gke.io (reconciler uses existing field manager)
- Update the fake clients to require using one of the approved field managers, to help validate usage in unit tests.
- Add a fake.FieldManager for tests to use.
- Update all the unit tests using once of the fake clients.
- Update all Create, Update, and Patch calls to use the appropriate field manager.
- This should help distinguish between fields modified by the reconciler vs the reconciler-manager, and avoid any accidentally removed fields when using server-side apply. This is necessary to allow the reconciler-manager to start adding labels & annotations to RSyncs, whose spec may already be managed by a reconciler using composition.

Caveats:
- The kpt inventory client and cli-utils applier & destroyer do not yet allow configuring the field manager.
- The client-go FakeDynamicClient and generated client-gen clients do not support passing through client options. So the config-sync fake.DynamicClient just uses the fake.FieldManager instead for now.

### DIFF
--- a/pkg/api/configsync/register.go
+++ b/pkg/api/configsync/register.go
@@ -23,7 +23,8 @@ const (
 	// ConfigSyncPrefix is the prefix for all ConfigSync annotations and labels.
 	ConfigSyncPrefix = GroupName + "/"
 
-	// FieldManager is the field manager name for server-side apply.
+	// FieldManager is the field manager name used by the reconciler.
+	// This avoids conflicts between the reconciler and reconciler-manager.
 	FieldManager = GroupName
 
 	// ControllerNamespace is the Namespace used for Nomos controllers
--- a/pkg/syncer/syncertest/fake/storage.go
+++ b/pkg/syncer/syncertest/fake/storage.go
@@ -34,15 +34,32 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/klog/v2"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/util/log"
 	"sigs.k8s.io/cli-utils/pkg/testutil"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// FieldManager is the field manager used by NewClient, when initializing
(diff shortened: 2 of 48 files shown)


pkg/syncer/syncertest/fake/storage.go
```diff
@@ -34,17 +34,34 @@ import (
...
+	// objects on the cluster.
+	// This avoids conflicts between the reconciler and reconciler-manager.
+	FieldManager = configsync.ConfigSyncPrefix + "fake-client"
+)
+
+var (
+	defaultAllowedFieldManagers = sets.New[string](
+		FieldManager,
+		configsync.FieldManager,
+		reconcilermanager.FieldManager,
```
tiffanny29631, line 61: Should the nomos-cli and resource-group-controller field manager be included?

karlkfi: They could be added in the future, but there aren't any unit tests for those components that use the fake clients here.

pkg/api/configsync/register.go
```diff
@@ -23,7 +23,8 @@ const (
 	// ConfigSyncPrefix is the prefix for all ConfigSync annotations and labels.
 	ConfigSyncPrefix = GroupName + "/"
 
-	// FieldManager is the field manager name for server-side apply.
+	// FieldManager is the field manager name used by the reconciler.
```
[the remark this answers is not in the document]
karlkfi: We already have code to ensure controllers don't fight with each other, and it will be even less likely with watch label filters. Plus, there's no way to tell config sync that you only want to patch an existing object. If you tried, it would probably create a race condition that would fail due to missing required fields until created by a reconciler with a fully valid spec. Plus, we would definitely want to test that kind of new feature if we added support for it. So I don't think we need to do future-proofing for it now.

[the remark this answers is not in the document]
karlkfi: TLDR: Using Config Sync to patch existing objects could be a future feature addition, but it would require a lot more thought. And we can always change the field managers later without much problem.

pkg/syncer/syncertest/fake/storage.go
```diff
@@ -68,16 +85,41 @@ type MemoryStorage struct {
...
+// By default, one of the following must be used:
+// - fake.FieldManager
+// - configsync.FieldManager
+// - reconcilermanager.FieldManager
+//
+// WARNING: fake.DynamicClient does not support passing through field manager
+// due to tech debt in the client-go FakeDynamicClient. So the fake.FieldManager
+// is always used instead.
+// TODO: Pass through FieldManager from Create, Update, & Patch (requires client-go & client-gen changes)
+func (ms *MemoryStorage) SetAllowedFieldManagers(fieldManagers sets.Set[string]) {
```
sdowell, line 114: Why do we need to support changing the allowed field managers?

karlkfi: I thought it might be useful for downstream code users who want to use the fake clients.

sdowell (approved):
/lgtm

PR #1242, opened 2024-05-30
test: fix panic in metric validation

sdowell:
This validation would panic if the got array was larger than the want array. This updates the validation logic so that it handles this case without panicking and returns a diff.


tiffanny29631 (approved):
/lgtm

PR #1243, opened 2024-06-03
Mount empty volume for hydration-controller's service account

Camila-B:
b/342237764

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -67,6 +67,9 @@ data:
            volumeMounts:
            - name: repo
              mountPath: /repo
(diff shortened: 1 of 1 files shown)


manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -67,6 +67,9 @@ data:
            volumeMounts:
            - name: repo
              mountPath: /repo
+           - name: service-account
```
nan-yu, line 70: Suggest adding a comment to explain why we need to override the automounted service account info.

Camila-B: Done

nan-yu (approved):
/lgtm

PR #1244, opened 2024-06-03
Rename declared.RootReconciler to RootScope

karlkfi:
- While ":root" does identify the resource manager named "root-reconciler" in the managed-by annotation, this constant is a Scope type, not a Manager type. So putting the right type in the name of the const should make the code a little easier to read.
- Add help funcs for converting between the Scope and other related types, like the reconciler deployment name, sync kind, and sync namespace.


pkg/declared/scope.go
```diff
@@ -41,3 +45,36 @@ func ValidateScope(s string) error {
 	}
 	return nil
 }
+
+// ScopeFromSync returns the scope associated with a specific RootSync or RepoSync.
+func ScopeFromSync(syncKind, syncNamespace string) Scope {
```
nan-yu, line 50: `ScopeFromSync` can be simplified as `ScopeFromSyncNamspace` with only one parameter because RepoSync can't be defined in the c-m-s namespace, while RootSync can only exist in the c-m-s namespace.

karlkfi: I've never heard of this constraint. Do you know where the validation for it is?

karlkfi: Found it https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/reposync_controller.go#L945

karlkfi: Wish I'd known that before. That's useful for all sorts of naming length reductions. You can just use NN, without needing to specify the kind.

nan-yu: Sorry, missed that. 

And https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/reconcilermanager/controllers/rootsync_controller.go#L823 for RootSync.

I think having kind in the labels is more explicit. For naming length reduction, yes, we can strip out the kind part.

karlkfi: Ok. I refactored the helpers a bit and made a couple of them methods on the Scope type.

nan-yu (approved):
/lgtm

karlkfi:
/hold

A few more changes coming.

karlkfi:
/unhold

I added Scope.String() and Scope.Validate() and removed all the string casting I could find. This should make it clearer if the user wants the string value or the namespace value (same thing for RepoSyncs).

karlkfi:
/retest

TestNoSSLVerifyV1Alpha1 seems flaky

PR #1245, opened 2024-06-04
test: adds cleanup logic to remove fake finalizer

sdowell:
The fake finalizer should usually be cleaned up by the test which adds it, but sometimes the finalizer fails to be removed (e.g. crash). If the finalizer fails to be removed by the test, the cluster becomes wedged and requires manual intervention to fix. This adds logic to Clean which removes the finalizer so that manual intervention is not required.

Also removes the early returns so that goroutines are not leaked by this function.

### DIFF
--- a/e2e/nomostest/clean.go
+++ b/e2e/nomostest/clean.go
@@ -505,7 +505,19 @@ func DeleteObjectsAndWait(nt *NT, objs ...client.Object) error {
 		nn := client.ObjectKeyFromObject(obj)
 		gvk, err := kinds.Lookup(obj, nt.Scheme)
 		if err != nil {
-			return err
+			tg.Go(func() error {
+				return err
+			})
+			continue
+		}
+		// Remove fake test finalizers if they are blocking deletion
+		if reflect.DeepEqual(obj.GetFinalizers(), []string{ConfigSyncE2EFinalizer}) {
+			if err := nt.KubeClient.MergePatch(obj, `{"metadata":{"finalizers":[]}}`); err != nil {
+				tg.Go(func() error {
+					return err
+				})
+				continue
+			}
 		}
 		if !obj.GetDeletionTimestamp().IsZero() {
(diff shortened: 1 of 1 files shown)


e2e/nomostest/clean.go
```diff
@@ -505,7 +505,19 @@ func DeleteObjectsAndWait(nt *NT, objs ...client.Object) error {
...
 		gvk, err := kinds.Lookup(obj, nt.Scheme)
 		if err != nil {
-			return err
+			tg.Go(func() error {
+				return err
+			})
+			continue
+		}
+		// Remove fake test finalizers if they are blocking deletion
+		if reflect.DeepEqual(obj.GetFinalizers(), []string{ConfigSyncE2EFinalizer}) {
```
karlkfi, line 514: I see why you added it here, but this really feels like a test specific cleanup. Like, the test should be in charge of registering it. 

Is there any way we could make the tests all register cleanup, even if they aren't running, then run all the cleanup first before any tests...?

sdowell: There may be a way to hack something like that together, but I think it would be a pretty large refactor to make the test scaffolding work in that way.

Currently [the test environments are set up and `Clean` is run before any of the test cases are invoked](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/main_test.go#L131), so there is no knowledge of the test cases themselves.

karlkfi (approved):
/lgtm

karlkfi:
Maybe try rebasing? otherwise we have a bug on main

sdowell:
> Maybe try rebasing? otherwise we have a bug on main

unit tests seem pretty flaky, I think there is a bug on main

karlkfi:
TestRoot_Parse flake should be fixed by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1261

karlkfi (approved):
/lgtm

PR #1246, opened 2024-06-05
test: fix flakiness of TestResourceGroupCustomResource

Camila-B:
This PR adds cleanup logic to multiple other test that use the Anvil CRD in order to deflake TestResourceGroupCustomResource

### DIFF
--- a/e2e/testcases/multiversion_test.go
+++ b/e2e/testcases/multiversion_test.go
@@ -58,22 +58,6 @@ func TestMultipleVersions_CustomResourceV1(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
-	nt.T.Cleanup(func() {
-		// Remove the CRD and CRs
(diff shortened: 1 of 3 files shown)


e2e/testcases/multiversion_test.go
```diff
@@ -58,6 +58,22 @@ func TestMultipleVersions_CustomResourceV1(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
+	nt.T.Cleanup(func() {
```
sdowell, line 61: This cleanup shouldn't be necessary, since everything is managed by the RootRepo. The test scaffolding will handle cleanup for anything managed by the RootRepo in [Reset](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/reset.go#L83).

sdowell (approved):
/lgtm

PR #1254, opened 2024-06-07
[metric] Reset declared_resource to 0 when commit updates with test fix (#1232)

tiffanny29631:
The current OpenCensus library keeps a metric stream that is uniquely identified by commit alive even though no new values are issued.

This makes the Otel Collector metricstransform processor receives all time value when calculating the max and gives false output.

The temporary work around is to reset the stream of previous commit to 0 when parser sees new commit.

The cherry-pick was modified to accommodate the data structure of Resources that v1.18 currently has.

b/321875474


nan-yu (approved):
/lgtm

PR #1255, opened 2024-06-07
Fix TestReconcileRepoSyncLifecycleValidToInvalid

karlkfi:
- Test was flakey because the update retry wasn't getting the latest object version before retrying. Just use mutate.Spec instead.


sdowell (approved):
/lgtm

karlkfi:
/retest

TestBackoffReimportCount is flaky

PR #1261, opened 2024-06-10
fix: deadlock in Updater.Errors()

karlkfi:
Errors() was calling applyErrors() and both were trying to lock the same mutex. Since nothing else is calling applyErrors() any more, we can just remove it.

Flaky deadlock was caused by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1199, so this does not need cherry-picking to v1.18


sdowell (approved):
/lgtm

tiffanny29631 (approved):
/lgtm

karlkfi:
/retest

TestBackoffReimportCount is known flaky, unrelated.

PR #1266, opened 2024-06-11
test: deflake TestBackoffRetryCount

Camila-B:
* This updates TestBackoffRetryCount so that it uses a timeout of 1 second instead of the reimport count when calling the `Run` function

### DIFF
--- a/pkg/parse/run_test.go
+++ b/pkg/parse/run_test.go
@@ -433,7 +433,8 @@ func TestBackoffRetryCount(t *testing.T) {
 
 	t.Logf("start running test at %v", time.Now())
 
-	ctx, _ := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
 	backoff := defaultBackoff()
 	backoff.Duration = time.Nanosecond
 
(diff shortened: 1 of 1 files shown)


pkg/parse/run_test.go
```diff
@@ -427,41 +427,30 @@ func TestRun(t *testing.T) {
...
 	testState := &namespacecontroller.State{}
-	reimportCount := 0
 	retryCount := 0
-	testIsDone := func(reimportCount *int, _ *int) bool {
-		return *reimportCount == 35
-	}
 
 	t.Logf("start running test at %v", time.Now())
 
-	ctx, cancel := context.WithCancel(context.Background())
```
karlkfi, line 440: defer the cancel to make sure the timer is cleaned up at the end of the function, even if it panics, without waiting for the timeout.

karlkfi (commented):
This looks potentially less flaky, but still not deterministic. If the machine is busy it could start the timer and then pause the thread for more than a second before continuing. This is common on CI clusters with many jobs running in parallel. 

Is there a way we can refactor the test to test it in a way that's not time based? One option would be to inject named timer factories instead of named timeouts.

pkg/parse/run_test.go
Camila-B: Done

PR #1267, opened 2024-06-11
ci: omit nomos migrate output from test logs

sdowell:
The output currently breaks the XML parser used by the junit reporting.


tiffanny29631 (approved):
/lgtm

PR #1268, opened 2024-06-11
test: Only check declared_resources metric

tiffanny29631:
The fix https://github.com/GoogleContainerTools/kpt-config-sync/pull/1232 only affects declared_resources from Config Sync.

The metric from RG controller may take longer to be ready. Skip checking for those target values for now.


karlkfi (approved):
/lgtm

karlkfi (approved):
/lgtm

sdowell (approved):
/lgtm

PR #1269, opened 2024-06-11
Dedupe conflict errors with diff resource versions

karlkfi:
Since objects are indexed by ID in the inventory, we don't need to index conflict errors by GVKNN. So this change modifies the conflict handler to index by ID instead. This should have no real effect.

### DIFF
--- a/pkg/remediator/conflict/handler.go
+++ b/pkg/remediator/conflict/handler.go
@@ -20,15 +20,18 @@ import (
 	"github.com/elliotchance/orderedmap/v2"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
-	"kpt.dev/configsync/pkg/remediator/queue"
+	"kpt.dev/configsync/pkg/core"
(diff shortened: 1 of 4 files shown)


tiffanny29631 (approved):
/lgtm

karlkfi:
/retest 

> new.go:480: 2024-06-11 18:26:30.354638367 +0000 UTC ERROR: waiting for ConfigSync Deployments to become available: WatchObject(Deployment config-management-system/reconciler-manager): predicates not satisfied: expected  config-management-system/reconciler-manager to have status "Current", but got "InProgress": context deadline exceeded
> W0611 18:26:30.354767   24447 reflector.go:458] k8s.io/client-go/tools/watch/informerwatcher.go:146: watch of *v1.Pod ended with: an error on the server ("unable to decode an event from the watch stream: context canceled") has prevented the request from succeeding

This seems like a flake. I didn't change the reconciler-manager at all in this PR.

I suspect the apiserver just stopped the watch prematurely and the test watch client doesn't have a retry.

PR #1272, opened 2024-06-13
[WIP] Refactor parse.Run to use an event generator

karlkfi:
The intent of this refactor is to pull out the timers into their own function so that the rest of the run behavior can be tested in a deterministic way.

This refactor is not an ideal solution, but rather a first step towards making the run loop loop more similar to a control loop.

For now, it still has the same timers driving the creation of events and still has seperate event handling for each event. But in the future, we should be able to further refactor the code to look more like a controller-runtime reconciler.

The first step to be extracted out is the delay of the status update timer, which should happen any time the RunFunc was called. This now happens for all events, including the namespace event.


pkg/parse/run_test.go
```diff
@@ -426,42 +432,43 @@ func TestRun(t *testing.T) {
...
 
-func TestBackoffRetryCount(t *testing.T) {
-	parser := newParser(t, FileSource{}, false, 10*time.Microsecond, 150*time.Microsecond)
-	testState := &namespacecontroller.State{}
-	reimportCount := 0
-	retryCount := 0
-	testIsDone := func(reimportCount *int, _ *int) bool {
-		return *reimportCount == 35
-	}
+// func TestRunRetryBackoffMaxAttempts(t *testing.T) {
```
karlkfi, line 435: Need to replace this with an actual test of the TimedEventProducer, since that's where the backoff logic is now.

pkg/parse/events/timed_event_producer.go
```diff
@@ -0,0 +1,126 @@
...
+	resyncTimer := time.NewTimer(t.ResyncPeriod)
+	defer resyncTimer.Stop()
+
+	statusUpdateTimer := time.NewTimer(t.StatusUpdatePeriod)
+	defer statusUpdateTimer.Stop()
+
+	nsEventTimer := time.NewTimer(t.NamespaceControllerPeriod)
+	defer nsEventTimer.Stop()
+	// Disable namespace events when period is zero/default.
+	if t.NamespaceControllerPeriod == 0 {
```
karlkfi, line 53: This could probably be done more cleanly with a no-op timer

pkg/parse/events/timed_event_producer.go
```diff
@@ -0,0 +1,126 @@
...
+
+			// Schedule next namespace event check attempt
+			nsEventTimer.Reset(t.NamespaceControllerPeriod)
+		}
+
+		// If the RunFunc was called, it will likely have updated the status.
+		// So we can delay the next StatusEvent.
+		// If not, the event was probably handled very quickly, so we don't want
+		// to delay the StatusEvent, otherwise it may never trigger.
+		if !result.Skipped {
```
karlkfi, line 121: Might want to pull this out to a different PR, since it's actually a bug fix for the namespace event handling.

pkg/parse/run.go
```diff
@@ -70,77 +70,76 @@ const (
...
-	defer runTimer.Stop()
-
-	resyncTimer := time.NewTimer(opts.ResyncPeriod)
-	defer resyncTimer.Stop()
-
-	retryTimer := time.NewTimer(opts.RetryPeriod)
-	defer retryTimer.Stop()
+	state := &reconcilerState{}
+
+	runOpts.GenerateFunc(ctx, func(event events.Event) events.HandleResult {
```
karlkfi, line 92: Probably want this extracted to a named function.

pkg/parse/events/events.go
```diff
@@ -0,0 +1,90 @@
...
+type HandleResult struct {
+	// DelayStatusUpdate tells GenerateFunc to reset the status update timer.
+	DelayStatusUpdate bool
+	// ResetRetryBackoff tells GenerateFunc to reset the retry backoff.
+	ResetRetryBackoff bool
+}
+
+// Event represents the cause of HandleEventFunc being called.
+// Some event types may have special fields or methods.
+type Event interface {
```
karlkfi, line 40: None of these events have any special fields any more, now that I added the ResultHandler to the EventProducer to allow all producers to inject generic even cleanup behavior.

So the events could probably be replaced with strings, as long as we don't need to add any other event-specific inputs.

pkg/parse/events/execution.go
```diff
@@ -0,0 +1,117 @@
...
+type EventProducer struct {
+	// EventType of the producer.
+	EventType string
+	// EventChan takes a readable channel value.
+	// Reflection is used to allow channels with any payload type.
+	// This avoids needing to wrap channels to change type, which would
+	// introduce buffering, which would prevent the event producer from waiting
+	// until the event was received.
+	// Usage: reflect.ValueOf(ctx.Done())
+	EventChan reflect.Value
```
karlkfi, line 34: We might want to replace EventChan with a Start function that returns an event channel. That would allow starting the producers in the Execute function.

Then again, it's also useful for testing to be able to make Clock adjustments after all the timers have been started.

pkg/parse/events/meta_scheduler_test.go
```diff
@@ -0,0 +1,233 @@
...
+				},
+				{
+					Event: RetrySyncEvent{},
+					Result: HandleResult{
+						ResetRetryBackoff: true,
+					},
+				},
+			},
+		},
+		// TODO: test retry backoff behavior
```
karlkfi, line 178: Testing retry backoff may not be easy with the way this function is only exposing Events and Results. We need to be able to adjust the Step size between each event to trigger the next retry event.

pkg/parse/events/meta_scheduler_test.go
```diff
@@ -0,0 +1,233 @@
...
+		// TODO: test all handlers together (can cause race conditions)
```
karlkfi, line 179: Testing multiple handlers with this current test function is hard, because concurrent timers create race conditions that affect the event order. When multiple channels have pending events, Select picks a random one to execute.

pkg/parse/run.go
```diff
@@ -70,77 +70,78 @@ const (
...
-	defer resyncTimer.Stop()
-
-	retryTimer := opts.Clock.NewTimer(opts.RetryPeriod)
-	defer retryTimer.Stop()
-
-	statusUpdateTimer := opts.Clock.NewTimer(opts.StatusUpdatePeriod)
-	defer statusUpdateTimer.Stop()
+	state := &reconcilerState{}
+
+	events.Execute(ctx, runOpts.Schedule(ctx), func(event events.Event) events.HandleResult {
```
karlkfi, line 92: This makes the Run function a lot smaller/simpler, but there's still room for improvement. We might want to pull out the inline handler func so we can test it independently. That said, the independent behavior doesn't really mean the whole thing works together as intented.

It might be better to move these event-specific behaviors into methods on the MetaScheduler or EventProducers.

pkg/reconciler/reconciler.go
```diff
@@ -271,6 +273,17 @@ func Run(opts Options) {
 		}
 	}
 
+	// Build an event producer to generate events for parser.Run.
+	scheduler := &events.MetaScheduler{
+		Clock:              parseOpts.Clock,
+		SyncPeriod:         opts.PollingPeriod,
+		FullSyncPeriod:     opts.ResyncPeriod,
+		StatusUpdatePeriod: opts.StatusUpdatePeriod,
+		// TODO: Shouldn't this use opts.RetryPeriod as the initial duration?
```
karlkfi, line 282: This seems like an existing bug. The Backoff wasn't being built with the RetryPeriod option.

pkg/parse/run.go
```diff
@@ -153,95 +154,57 @@ func Run(ctx context.Context, p Parser, nsControllerState *namespacecontroller.S
...
-				// Reset the cache partially to make sure all the steps of a parse-apply-watch loop will run.
-				// The cached sourceState will not be reset to avoid reading all the source files unnecessarily.
-				// The cached needToRetry will not be reset to avoid resetting the backoff retries.
-				state.resetPartialCache()
-				runFn(ctx, p, namespaceEvent, state)
+			// If the commit changed, then reset the retry backoff for the new commit.
+			if result.CommitChanged {
+				return events.Result{
+					DelayStatusUpdate: true,
+					ResetRetryBackoff: true,
```
nan-yu, line 171: It only resets retry backoff when source changed. 
The retry backoff also needs to be reset when a run succeeded. Otherwise, the retry backoff continues from the previous failure before the success.

pkg/parse/run.go
```diff
@@ -70,77 +70,78 @@ const (
...
+			}
 
-	for {
-		select {
-		case <-ctx.Done():
-			return
+		case events.NamespaceResyncEventType:
+			// If the namespace controller indicates that an update is needed,
+			// attempt to re-sync.
+			if !nsControllerState.ScheduleSync() {
```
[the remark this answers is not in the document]
karlkfi: I made it so the event is never sent if the period is >= 0, and the caller only sets the period if nsControllerState is non-nil. So we don't need to nil check here.

pkg/parse/events/events.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// EventType is the name of the event
+type EventType string
+
+const (
+	// FullSyncEventType is the EventType for a full sync, including reading and
+	// parsing resource objects from the shared volume.
+	FullSyncEventType EventType = "FullSyncEvent"
+	// SyncEventType is the EventType for an incremental sync, only parsing
```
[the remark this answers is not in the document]
karlkfi: So you mean FullSyncEvent -> ReimportEvent and SyncEvent -> ResyncEvent?

I guess that's ok. Though FullSyncEvent doesn't just re-import, it also syncs.

This is sorta the problem with naming them discretely when they use 99% of the same logic.

pkg/parse/run.go
```diff
@@ -70,77 +70,78 @@ const (
...
+			// happen until the ongoing reconciliation is done.
+			runOpts.Run(ctx, p, triggerReimport, state)
 
-	runFn := runOpts.runFunc
-	state := &reconcilerState{
-		backoff:     runOpts.backoff,
-		retryTimer:  retryTimer,
-		retryPeriod: opts.RetryPeriod,
-	}
+		case events.StatusEventType:
```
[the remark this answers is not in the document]
karlkfi: The goroutine updates the status while syncing. The event updates the status between sync attempts.

pkg/parse/events/sources.go
```diff
@@ -0,0 +1,232 @@
...
+	// Schedule next resync attempt
+	s.timer.Reset(retryDuration)
+	return result
+}
+
+// HandleResult resets the backoff timer if ResetRetryBackoff is true.
+func (s *TimeDelaySourceWithBackoff) HandleResult(result Result) {
+	if result.ResetRetryBackoff {
+		s.currentBackoff = util.CopyBackoff(s.Backoff)
+		s.timer.Reset(s.currentBackoff.Duration)
```
sdowell, line 178: I understand if it's a lot to ask - But it might be nice to have a table of `scenarios x timers` (where each cell indicates whether the scenario causes the timer to be reset) to help visualize which scenarios/events cause which timers to get reset. Would help clarify the behavior

pkg/parse/events/events.go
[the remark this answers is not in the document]
karlkfi: How about SyncFromDiskEvent & SyncFromCacheEvent?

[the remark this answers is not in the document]
karlkfi: Or SyncEvent & SyncWithReimportEvent

pkg/parse/run.go
karlkfi: Added a Success flag to the RunResult for this.

pkg/parse/run.go
```diff
@@ -70,77 +70,78 @@ const (
...
-	backoff wait.Backoff
+	Run     RunFunc
+	Sources []events.Source
+}
+
+// RunResult encapsulates the result of a RunFunc.
+// This simply allows explicitly naming return values in a way that makes the
+// implementation easier to read.
+type RunResult struct {
+	CommitChanged bool
```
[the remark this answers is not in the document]
karlkfi: Done.

pkg/parse/run.go
```diff
@@ -70,77 +70,78 @@ const (
...
+			// The cached sourceState will not be reset to avoid reading all the source files unnecessarily.
+			// The cached needToRetry will not be reset to avoid resetting the backoff retries.
+			state.resetPartialCache()
+			runOpts.Run(ctx, p, triggerResync, state)
 
-	nsEventPeriod := time.Second
-	nsEventTimer := opts.Clock.NewTimer(nsEventPeriod)
-	defer nsEventTimer.Stop()
+		case events.SyncEventType:
+			// Re-import declared resources from the filesystem (from git-sync).
```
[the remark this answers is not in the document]
karlkfi: Done.

pkg/reconciler/reconciler.go
```diff
@@ -271,6 +273,17 @@ func Run(opts Options) {
 		}
 	}
 
+	// Build an event producer to generate events for parser.Run.
+	producer := &events.SourceProducer{
+		Clock:              parseOpts.Clock,
+		SyncPeriod:         opts.PollingPeriod,
```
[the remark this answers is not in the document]
karlkfi: Done

pkg/parse/events/sources.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+// Source triggers events with a channel and defines how to handle them.
+type Source interface {
+	// Type of the producer.
+	Type() EventType
+	// Channel takes a readable channel value.
+	// Reflection is used to allow channels with any payload type.
+	// This avoids needing to wrap channels to change type, which would
+	// introduce buffering, which would prevent the event producer from waiting
```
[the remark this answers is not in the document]
karlkfi: Fixed.

pkg/parse/run.go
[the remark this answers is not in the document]
nan-yu: ACK

pkg/parse/events/sources.go
karlkfi: I added events/doc.go with more details.

pkg/parse/events/funnel_test.go
```diff
@@ -0,0 +1,256 @@
...
+				{
+					Event: Event{Type: RetrySyncEventType},
+					Result: Result{
+						ResetRetryBackoff: true,
+					},
+				},
+			},
+		},
+		// TODO: test retry backoff behavior
+		// TODO: test all handlers together (can cause race conditions)
```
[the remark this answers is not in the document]
karlkfi: It shouldn't. I just meant that with two timers sharing a clock you don't always get deterministicly ordered events, which makes it hard to compare against an expected result. 

It should be possible with the fake clock tho.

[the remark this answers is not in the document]
karlkfi: I figured out how to test more of them together (SyncEvents, StatusEventType, ResyncEvents), but the retry and namespace events are very spammy, so I left them out. They were already spammy before this PR tho. So I think it's fine to leave them untested for now. Hopefully we can "fix" them to be less spammy in the future with the new event system, perhaps by triggering them with a channel, instead of a timer.

pkg/parse/events/publisher.go
```diff
@@ -0,0 +1,230 @@
...
+		s.timer.Stop()
+	}()
+	return reflect.ValueOf(s.timer.C())
+}
+
+// Publish calls the HandleFunc with a new event and resets the delay timer.
+func (s *TimeDelayPublisher) Publish(subscriber Subscriber) Result {
+	result := subscriber.Handle(Event{Type: s.EventType})
+
+	// Schedule next resync attempt
```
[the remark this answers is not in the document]
karlkfi: Fixed

pkg/parse/events/publisher.go
```diff
@@ -0,0 +1,230 @@
...
+		return Result{}
+	}
+
+	retryDuration := s.currentBackoff.Step()
+	retries := s.retryLimit - s.currentBackoff.Steps
+	klog.Infof("a retry is triggered (retries: %v/%v)", retries, s.retryLimit)
+
+	result := subscriber.Handle(Event{Type: s.EventType})
+
+	// Schedule next resync attempt
```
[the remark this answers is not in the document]
karlkfi: Fixed

pkg/parse/run_test.go
karlkfi: Added backoff tests to TestFunnel_Start

pkg/parse/events/funnel_test.go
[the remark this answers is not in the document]
karlkfi: Also added some backoff tests here too.

pkg/parse/events/builder.go
```diff
@@ -0,0 +1,67 @@
...
+		publishers = append(publishers, NewTimeDelayPublisher(SyncWithReimportEventType, t.Clock, t.SyncWithReimportPeriod))
+	}
+	if t.NamespaceControllerPeriod > 0 {
+		publishers = append(publishers, NewTimeDelayPublisher(NamespaceResyncEventType, t.Clock, t.NamespaceControllerPeriod))
+	}
+	if t.RetryBackoff.Duration > 0 {
+		publishers = append(publishers, NewRetrySyncPublisher(t.Clock, t.RetryBackoff))
+	}
+	if t.StatusUpdatePeriod > 0 {
+		publishers = append(publishers, NewResetOnRunAttemptPublisher(StatusEventType, t.Clock, t.StatusUpdatePeriod))
```
nan-yu, line 64: Does StatusEventType need a ResetOnRunAttemptPublisher? `RunAttempt` will always be false for StatusEvent, so it will never get delayed.

pkg/parse/events/doc.go
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+// Package events contains a set of Events sent by Publishers.
+//
+// PublishingGroupBuilder can be used to Build a set of Publishers.
+//
+// Publisher types:
+// - TimeDelayPublisher    - Publishes events with a configurable time delay.
+// - RetryPublisher        - TimeDelayPublisher with resettable backoff (Result.ResetRetryBackoff).
+// - StatusUpdatePublisher - TimeDelayPublisher with delayable timer (Result.DelayStatusUpdate).
```
nan-yu, line 22: The doc needs to be updated based on the new impl.

pkg/parse/events/doc.go
```diff
@@ -0,0 +1,36 @@
...
+//
+// Events and their Publisher:
+// - SyncWithReimportEvent - TimeDelayPublisher (SyncWithReimportPeriod)
+// - SyncEvent             - TimeDelayPublisher (SyncPeriod)
```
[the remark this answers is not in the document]
karlkfi: Fixed.

pkg/parse/events/doc.go
karlkfi: Fixed.

pkg/parse/events/funnel_test.go
```diff
@@ -0,0 +1,536 @@
...
+				700 * time.Millisecond,
+				1000 * time.Millisecond,
+				1300 * time.Millisecond,
+				1400 * time.Millisecond,
+				1700 * time.Millisecond,
+				2000 * time.Millisecond,
+				2100 * time.Millisecond,
+				2400 * time.Millisecond,
+				2700 * time.Millisecond,
+				2900 * time.Millisecond,
```
nan-yu, line 237: This should be fired at 2800ms for the SyncEventType.

pkg/parse/events/builder.go
nan-yu: ACK. It also allows the retry backoff to be reset.

nan-yu (commented):
One nit comment. Otherwise LGTM

pkg/parse/events/funnel_test.go
karlkfi: nice catch! 
I fixed that one and all the subsequent ones up to the 4000ms reimport.

pkg/parse/events/builder.go
karlkfi: retry backoff is reset by a different flag: Result.ResetRetryBackoff

The RunAttempted flag just resets the status update and sync timers.

Should we make the RunAttempted flag ALSO reset the retry backoff?

karlkfi: Nevermind. You just meant that the HandleResult allowed for ResetRetryBackoff to be sent from other events.

nan-yu (approved):
/lgtm

Issue #1273, opened 2024-06-13
BUG: Root reconciler failing in clusters with proper user mapping (OpenShift).


robertosgm:
When running on systems with correct user mapping from RunAsUser (like OpenShift), the root reconciler crashes with this error:

```
$ kubectl logs -n config-management-system deploy/root-reconciler -c reconciler --previous
I0613 16:14:23.856361       1 setup.go:31] Build Version: v1.18.1-rc.1-0-g5d50947e
I0613 16:14:23.858500       1 main.go:216] Starting reconciler for: root
F0613 16:14:23.858592       1 reconciler.go:151] Error creating rest config: failed to build rest config: reading local kubeconfig: loading REST config from "/.kube/config": stat /.kube/config: no such file or directory
```

This was working on v1.16.1 and fails on v1.17.x, v1.18.x
This is because the InClusterConfig from this new code never runs and jumps straight to trying to configure from a kubeconfig file:

```
func NewRestConfig(timeout time.Duration) (*rest.Config, error) {
	var cfg *rest.Config
	// Detect kubectl config file
	path, err := KubeConfigPath()
	if err != nil {
		// Build from k8s downward API
		cfg, err = NewFromInClusterConfig()
		if err != nil {
			return nil, fmt.Errorf("failed to build rest config: kubeconfig not found: reading in-cluster config: %w", err)
		}
	} else {
		// Build from local config file
		cfg, err = NewFromConfigFile(path)
		if err != nil {
			return nil, fmt.Errorf("failed to build rest config: reading local kubeconfig: %w", err)
		}
	}
	// Set timeout, if specified.
	if timeout != 0 {
		cfg.Timeout = timeout
	}
	klog.V(7).Infof("Config: %#v", *cfg)
	return cfg, nil
}
```

The reason is that KubeConfigPath() only fails when user mapping is wrong and user 1000 is not proper, in OpenShift the user mapping is done correctly.

A more correct code should try to do the NewFromInClusterConfig first and if that fails, then try to configure from Kubeconfig. Or retry the NewFromInClusterConfig after the NewFromConfigFile fails.

sdowell:
It looks like this behavior changed with https://github.com/GoogleContainerTools/kpt-config-sync/pull/986.

Another alternative fix: Check whether the file exists at the returned `KubeConfigPath`, and only use `NewFromConfigFile` if the file exists. Otherwise, use `NewFromInClusterConfig`.

PR #1275, opened 2024-06-13
fix: only use local config if file exists

sdowell:
The rest config precedence was changed to prioritize the kubectl config over the in cluster config. This was only checking if KubeConfigPath returns a path rather than whether a file exists at that path.

It has been reported that certain k8s distributions set the user config for the container, in which case KubeConfigPath will return a home directory path for that user. This causes the process to attempt to use a local file config instead of the in cluster config, which fails.

This change adds an additional check that the kube config file exists at the specified path, which is intended to fix the above scenario.

Fixes: https://github.com/GoogleContainerTools/kpt-config-sync/issues/1273

### DIFF
--- a/pkg/client/restconfig/restconfig.go
+++ b/pkg/client/restconfig/restconfig.go
@@ -36,20 +36,39 @@ const DefaultTimeout = 15 * time.Second
 // is enabled or not. If server-side flow control is enabled, then client-side
 // throttling is disabled, vice versa.
 func NewRestConfig(timeout time.Duration) (*rest.Config, error) {
+	builder := restConfigBuilder{
+		newFromConfigFileFn:      NewFromConfigFile,
+		newFromInClusterConfigFn: NewFromInClusterConfig,
+	}
+	return builder.newRestConfig(timeout)
+}
+
+type restConfigBuilder struct {
+	newFromConfigFileFn      func(string) (*rest.Config, error)
+	newFromInClusterConfigFn func() (*rest.Config, error)
+}
+
+func (b *restConfigBuilder) newRestConfig(timeout time.Duration) (*rest.Config, error) {
 	var cfg *rest.Config
 	// Detect kubectl config file
 	path, err := KubeConfigPath()
-	if err != nil {
-		// Build from k8s downward API
-		cfg, err = NewFromInClusterConfig()
-		if err != nil {
-			return nil, fmt.Errorf("failed to build rest config: kubeconfig not found: reading in-cluster config: %w", err)
+	kubeConfigExists := false
+	if err == nil {
+		if _, err := os.Stat(path); err == nil {
+			kubeConfigExists = true
 		}
-	} else {
+	}
+	if kubeConfigExists {
 		// Build from local config file
-		cfg, err = NewFromConfigFile(path)
+		cfg, err = b.newFromConfigFileFn(path)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/client/restconfig/restconfig.go
```diff
@@ -39,18 +39,24 @@ func NewRestConfig(timeout time.Duration) (*rest.Config, error) {
...
 	// Detect kubectl config file
 	path, err := KubeConfigPath()
-	if err != nil {
-		// Build from k8s downward API
-		cfg, err = NewFromInClusterConfig()
-		if err != nil {
-			return nil, fmt.Errorf("failed to build rest config: kubeconfig not found: reading in-cluster config: %w", err)
+	kubeConfigExists := false
+	if err == nil {
+		if _, err := os.Stat(path); err == nil {
```
karlkfi, line 44: I think there are other possible errors from stat. Might want to check if it's a missing file error specifically.

sdowell: This is checking for existence, the err should only be nil if the file exists.

Agreed if we were checking the inverse that we would need to check the error type.

karlkfi: Ah, indeed. So used to seeing the other pattern I misread it.

pkg/client/restconfig/restconfig.go
```diff
@@ -39,18 +39,24 @@ func NewRestConfig(timeout time.Duration) (*rest.Config, error) {
 	var cfg *rest.Config
 	// Detect kubectl config file
 	path, err := KubeConfigPath()
-	if err != nil {
-		// Build from k8s downward API
-		cfg, err = NewFromInClusterConfig()
-		if err != nil {
-			return nil, fmt.Errorf("failed to build rest config: kubeconfig not found: reading in-cluster config: %w", err)
+	kubeConfigExists := false
```
karlkfi, line 42: Is there any way we can test this code? Looks right, but it'd be nice to have this little caveat tested to avoid regression.

sdowell: Added a unit test

karlkfi: Sorry to give you a hard time about this, but I really don't love the pattern of using package private variables for test hooks unless there's no other way to do it. It feels invasive, is harder to read, and doesn't provide for extension.

Is there some way we could do this with dependency injection, and then just make a thin shim for` NewRestConfig ()`that calls like `NewRestConfigWithX(...)` that takes factory functions or a StatFunc or something?

sdowell: refactored

karlkfi: Thanks!
I have some more nits, but I'll approve as-is for now.

- In the future maybe we could consider using a fake filesystem. There's a few available. Maybe k8s has one too. 
- I've also seen fake environments with methods for vars that make it easier to test things that use env vars. Might be useful if we want to test the full behavior and not just the delegation to upstream functions.
- Generally I think I prefer the convention of using -Fn for variables, -Func for types, and drop the suffix for struct fields, as if they were a normal method.
- Since there's no Build method, this isn't really a builder pattern. It's more like a factory pattern. But since we don't expose an interface, it doesn't really matter.

karlkfi (approved):
/lgtm

PR #1276, opened 2024-06-17
Fix sync status lost on reconciler pod replacement

karlkfi:
- Fix reconciler ignoring/erasing status from previous reconciler
  instances by priming the in-memory cache from the RSyncs status.
  Example: status.sync.commit
  Exception: errors are not parsable and may still be lost.

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1283
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1280
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1282
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1290
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1296
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1299

### DIFF
--- a/pkg/parse/namespace.go
+++ b/pkg/parse/namespace.go
@@ -20,6 +20,9 @@ import (
 	"strconv"
 
 	"github.com/google/go-cmp/cmp"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
@@ -54,7 +57,7 @@ func (p *namespace) options() *Options {
 }
 
 // parseSource implements the Parser interface
-func (p *namespace) parseSource(ctx context.Context, state sourceState) ([]ast.FileObject, status.MultiError) {
+func (p *namespace) parseSource(ctx context.Context, state *sourceState) ([]ast.FileObject, status.MultiError) {
 	p.mux.Lock()
 	defer p.mux.Unlock()
 
@@ -132,7 +135,7 @@ func (p *namespace) setSourceStatusWithRetries(ctx context.Context, newStatus *S
 
 	currentRS := rs.DeepCopy()
 
-	setSourceStatusFields(&rs.Status.Source, p, newStatus, denominator)
+	setSourceStatusFields(&rs.Status.Source, newStatus, denominator)
 
 	continueSyncing := (rs.Status.Source.ErrorSummary.TotalCount == 0)
 	var errorSource []v1beta1.ErrorSource
@@ -156,8 +159,8 @@ func (p *namespace) setSourceStatusWithRetries(ctx context.Context, newStatus *S
 	}
 
 	if klog.V(5).Enabled() {
-		klog.Infof("Updating source status for RepoSync %s/%s:\nDiff (- Expected, + Actual):\n%s",
-			rs.Namespace, rs.Name, cmp.Diff(currentRS.Status, rs.Status))
+		klog.V(5).Infof("Updating source status:\nDiff (- Removed, + Added):\n%s",
+			cmp.Diff(currentRS.Status, rs.Status))
 	}
 
 	if err := p.Client.Status().Update(ctx, &rs, client.FieldOwner(configsync.FieldManager)); err != nil {
@@ -209,7 +212,7 @@ func (p *namespace) setRenderingStatusWithRetires(ctx context.Context, newStatus
 
 	currentRS := rs.DeepCopy()
 
-	setRenderingStatusFields(&rs.Status.Rendering, p, newStatus, denominator)
+	setRenderingStatusFields(&rs.Status.Rendering, newStatus, denominator)
 
 	continueSyncing := (rs.Status.Rendering.ErrorSummary.TotalCount == 0)
 	var errorSource []v1beta1.ErrorSource
@@ -233,8 +236,8 @@ func (p *namespace) setRenderingStatusWithRetires(ctx context.Context, newStatus
 	}
 
 	if klog.V(5).Enabled() {
-		klog.Infof("Updating rendering status for RepoSync %s/%s:\nDiff (- Expected, + Actual):\n%s",
-			rs.Namespace, rs.Name, cmp.Diff(currentRS.Status, rs.Status))
+		klog.V(5).Infof("Updating rendering status:\nDiff (- Removed, + Added):\n%s",
+			cmp.Diff(currentRS.Status, rs.Status))
 	}
 
 	if err := p.Client.Status().Update(ctx, &rs, client.FieldOwner(configsync.FieldManager)); err != nil {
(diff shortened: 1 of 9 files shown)


karlkfi:
Looks like the initializing the cache from RSync status has broken `status.rendering.oci.dir` somehow, probably because it's not part of the `renderingStatus` struct in-memory. That broke TestPublicOCI, TestPolicyDirUnset, and TestUpdateRootSyncGitDirectory.

And it looks like there's another potentially related issue with the `status.rendering.commit` not being as expected. That broke TestHydrateKustomizeComponents and TestHydrateHelmOverlay.

karlkfi:
Alright, it took a fair amount of refactoring, but I was able to add the git/oci/helm source spec into the reconcilerStatus in a way that allows it to be considered when testing for quality. Without this change, the status update is skipped if the RSync spec changed without affecting the status (ex: sync dir changed but not commit).

pkg/parse/status.go
```diff
@@ -0,0 +1,245 @@
...
+	Errs       status.MultiError
+	LastUpdate metav1.Time
+}
+
+// Equals returns true if the specified SourceStatus equals this
+// SourceStatus, excluding the LastUpdate timestamp.
+func (gs SourceStatus) Equals(other SourceStatus) bool {
+	return gs.Commit == other.Commit &&
+		status.DeepEqual(gs.Errs, other.Errs) &&
+		(gs.Spec == nil && other.Spec == nil || gs.Spec.Equals(other.Spec))
```
karlkfi, line 199: Occurs to me that these could panic if gs.Spec is nil. What's the easiest way to fix this?

sdowell: I'd suggest perhaps just breaking this up into multiple if statements instead of the oneliner conditional. I think it will help readability as well.

As a oneliner you could do this with something like `(gs.Spec == nil && other.Spec == nil) || (gs.Spec != nil && gs.Spec.Equals(other.Spec))`

karlkfi:
/retest

Huh. Got a CrashLoopBackoff, but I don't see any error in the logs. Maybe just a kind/docker flake issue on CI.

```
          Warning  FailedMount       8m8s                kubelet            MountVolume.SetUp failed for volume "configs" : failed to sync configmap cache: timed out waiting for the condition
          Warning  FailedMount       8m8s                kubelet            MountVolume.SetUp failed for volume "otel-agent-config-vol" : failed to sync configmap cache: timed out waiting for the condition
```

pkg/parse/opts.go
```diff
@@ -89,9 +89,10 @@ type Options struct {
 // Parser represents a parser that can be pointed at and continuously parse a source.
 type Parser interface {
 	parseSource(ctx context.Context, state sourceState) ([]ast.FileObject, status.MultiError)
-	setSourceStatus(ctx context.Context, newStatus sourceStatus) error
-	setRenderingStatus(ctx context.Context, oldStatus, newStatus renderingStatus) error
-	SetSyncStatus(ctx context.Context, newStatus syncStatus) error
+	LoadStatus(ctx context.Context) (*ReconcilerStatus, error)
+	setSourceStatus(ctx context.Context, newStatus SourceStatus) error
+	setRenderingStatus(ctx context.Context, oldStatus, newStatus RenderingStatus) error
```
karlkfi, line 94: It bugs me that some of these are public and some private. It doesn't make any sense to have private methods on an interface. But maybe I'll save that for a different PR.

pkg/parse/root.go
```diff
@@ -334,55 +338,193 @@ func (p *root) setRenderingStatusWithRetires(ctx context.Context, newStatus rend
...
+			}
+			syncingConditionLastUpdate = condition.LastUpdateTime
+			break
+		}
+	}
+
+	return reconcilerStatusFromRSyncStatus(rs.Status.Status, p.options().SourceType, syncing, syncingConditionLastUpdate), nil
+}
+
+func reconcilerStatusFromRSyncStatus(rsyncStatus v1beta1.Status, sourceType v1beta1.SourceType, syncing bool, syncingConditionLastUpdate metav1.Time) *ReconcilerStatus {
```
karlkfi, line 413: This func is particularly repetitive. But the only way I can think to fix it would just be to not use an internal representation at all and just use the v1beta1 structs. However, mixing external and internal structs for the in-memory cache seems messy. Ideally it would just cache the whole RSync internally, or have some internal type that can easily map to both RepoSync & RootSync.

If I was doing this all over again, I would just make RepoSync & RootSync the same CRD. That would save so much headache, tech debt, and code duplication.

pkg/parse/namespace.go
```diff
@@ -248,16 +251,46 @@ func (p *namespace) setRenderingStatusWithRetires(ctx context.Context, newStatus
 	return nil
 }
 
+// SyncStatus implements the Parser interface
+// SyncStatus gets the RootSync sync status.
+func (p *namespace) LoadStatus(ctx context.Context) (*ReconcilerStatus, error) {
```
karlkfi, line 256: I dont love "Load" but I couldnt think of something better. I didn't want to just use "Status" or "GetStatus" since it makes an API call, but maybe that's more of an implementation detail.

The Parser interface method names are kind of a mess...

karlkfi:
/hold

Blocked on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1286 and needing to redo https://github.com/GoogleContainerTools/kpt-config-sync/pull/1283

pkg/parse/status.go
karlkfi: Added an isSpecEqual func to handle the nil case so it can be used inline.

pkg/parse/root.go
```diff
@@ -381,6 +381,140 @@ func setRenderingStatusFields(rendering *v1beta1.RenderingStatus, p Parser, newS
...
+			Spec:    renderSpec,
+			Commit:  rsyncStatus.Rendering.Commit,
+			Message: rsyncStatus.Rendering.Message,
+			// Can't parse errors.
+			// Errors will be reset the next time the reconciler updates the status.
+			Errs:       nil,
+			LastUpdate: rsyncStatus.Rendering.LastUpdate,
+			// Whether there exists a kustomize.yaml in the source is not persisted in the RSync status.
+			// TODO: Move requiresRendering out of reconcilerStatus.
+			RequiresRendering: false,
```
sdowell, line 503: This can be read from the `RequiresRenderingAnnotationKey` annotation

pkg/parse/source_test.go
```diff
@@ -100,6 +100,7 @@ func TestReadConfigFiles(t *testing.T) {
 			}()
 
 			srcState := &sourceState{
+				spec:    nil, // TODO: ???
```
sdowell, line 103: What is the TODO?

pkg/parse/namespace.go
sdowell: Maybe `ReconcilerStatusFromCluster`?

pkg/parse/root.go
karlkfi: Should we do that here though?

It seems like it's populated by the `RENDERING_ENABLED` env var or `--rendering-enabled` cmd flag, then passed into `parse.Options.RenderingEnabled` which affects control flow in RunFunc which calls `setRequiresRendering` to update the annotation value. None of that seems to check the cached status... 

Rendering seems to behave inconsistently to source and sync, which have `needToSetSourceStatus` and `needToSetSyncStatus` that compare against the cached status for optimization.

pkg/parse/source_test.go
karlkfi: Ideally we'd have some tests that validate the code branches when `sourceState.spec` in not nil. But I haven't dug into those private methods enough to know how they will behave, or even if there's any change behavior and not just a pass through. `TestReadHydratedDirWithRetry` and `TestReadConfigFiles` are affected.

karlkfi: Added a more useful TODO comment

pkg/parse/namespace.go
karlkfi: Renamed

pkg/parse/namespace.go
```diff
@@ -248,6 +251,35 @@ func (p *namespace) setRenderingStatusWithRetires(ctx context.Context, newStatus
 	return nil
 }
 
+// ReconcilerStatusFromCluster gets the RepoSync sync status from the cluster.
+func (p *namespace) ReconcilerStatusFromCluster(ctx context.Context) (*ReconcilerStatus, error) {
+	p.mux.Lock()
```
sdowell, line 256: Is the lock needed? I don't see anything that needs to be synchronized

pkg/parse/root.go
sdowell: It looks like it should get set shortly after when files are read from the filesystem, so perhaps it doesn't make a difference.

pkg/parse/namespace.go
karlkfi: Probably copied from SetSyncStatus. Removed.

sdowell (approved):
/lgtm

PR #1277, opened 2024-06-17
test: clean up ConfigManagement in e2e test

sdowell:
If either of the migrate tests fail early, the ConfigManagement object may be left lingering on the cluster. This will cause the Create call to fail for subsequent executions of these tests on the cluster, because the config-management object already exists.

### DIFF
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -1297,6 +1297,37 @@ func TestNomosMigrate(t *testing.T) {
 		}
 	})
 	nt.T.Cleanup(func() {
+		cmObj := &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "configmanagement.gke.io/v1",
+				"kind":       "ConfigManagement",
+				"metadata": map[string]interface{}{
+					"name": "config-management",
+				},
+			},
+		}
+
+		if err := nt.KubeClient.MergePatch(cmObj, `{"metadata":{"finalizers":[]}}`); err != nil {
+			if apierrors.IsNotFound(err) {
+				return // object already deleted, exit early to prevent watch error (GVK not found)
+			}
+			nt.T.Error(err)
+		}
+		// Delete the ConfigManagement CR, in case the test failed early.
+		if err := nt.KubeClient.Delete(cmObj); err != nil {
+			if apierrors.IsNotFound(err) {
+				return // object already deleted, exit early to prevent watch error (GVK not found)
+			}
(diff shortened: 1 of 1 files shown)


e2e/testcases/cli_test.go
```diff
@@ -1331,6 +1331,25 @@ func TestNomosMigrate(t *testing.T) {
 		},
 	}
 
+	nt.T.Cleanup(func() {
```
nan-yu, line 1334: Besides deleting ConfigManagement object in a cleanup block, do we also need to delete it before creating?
I'm thinking of a case when the cleanup block failed to be executed.

sdowell: In that case the `Create` call will fail but the `Cleanup` block will delete the object and the test will pass on the next execution. We could add a duplicate delete call but I think this is sufficient

nan-yu: ACK. The `Clean` function will also delete test data before all tests start.

e2e/testcases/cli_test.go
```diff
@@ -1331,6 +1331,28 @@ func TestNomosMigrate(t *testing.T) {
...
+			Object: map[string]interface{}{
+				"apiVersion": "configmanagement.gke.io/v1",
+				"kind":       "ConfigManagement",
+				"metadata": map[string]interface{}{
+					"name": "config-management",
+				},
+			},
+		}
+		// Delete the ConfigManagement CR, in case the test failed early.
+		if err := nt.KubeClient.Delete(cmObj); err != nil {
```
nan-yu, line 1345: Do we need to remove the finalizer before deleting the CR?

sdowell: In this case my intention was to just allow the finalizer to run, since the test manifests will be re-installed by the subsequent cleanup block

sdowell: If you prefer there's a few alternative approaches:
- Use `Apply` instead of `Create`, so it doesn't error if the object already exists (or other variations of not erroring if the object already exists)
- Change the cleanup ordering to delete the Deployment first, and then delete the ConfigManagement object and manually remove finalizers (otherwise the operator would re-apply the finalizers)

nan-yu: I would prefer removing the Deployment first and then removing the finalizer and deleting the ConfigManagement object. My concern is that operator may fail to garbage collect resources in the cleanup stage.

sdowell: done

PR #1282, opened 2024-06-20
cleanup: remove unused methods from remediator

karlkfi:
The conflict and fight handlers are injected now, so these methods are unused.


karlkfi:
/restest

Looks like TestRoot_SourceReconcilerErrorsMetricValidation is flaky

PR #1283, opened 2024-06-20
fix: remediator error reporting after new commit

karlkfi:
- Fix remediator error updates not being published after parsing has started for a new commit. Normally parsing is quick, but if there was an error, this state would often continue until the error was fixed by the user in the source.
- Modify setSyncStatus to take a commit, so we can avoid reverting the status.sync.commit after parsing has started and modified state.cache.source.commit.
- Add Remediator.Remediating() to avoid updating the sync status before syncing has succeeded or while remediating is paused during a sync attempt.


sdowell (approved):
/lgtm

PR #1286, opened 2024-06-21
Revert "fix: multiple sync status update bugs (#1283)"

sdowell:
This reverts commit 3e0915218d44dc238bb3a7b21e66c9e999e50821.

Reason for revert: This introduced frequent test failures for:
- TestHydrateKustomizeComponentsChanges
- TestHydrateHelmOverlay

### DIFF
--- a/pkg/parse/namespace.go
+++ b/pkg/parse/namespace.go
@@ -321,6 +321,12 @@ func (p *namespace) SyncErrors() status.MultiError {
 	return p.SyncErrorCache.Errors()
 }
 
+// Syncing returns true if the updater is running.
+// SyncErrors implements the Parser interface
+func (p *namespace) Syncing() bool {
+	return p.Updating()
+}
+
 // K8sClient implements the Parser interface
 func (p *namespace) K8sClient() client.Client {
 	return p.Client
--- a/pkg/parse/opts.go
+++ b/pkg/parse/opts.go
@@ -96,6 +96,8 @@ type Parser interface {
 	// SyncErrors returns all the sync errors, including remediator errors,
 	// validation errors, applier errors, and watch update errors.
 	SyncErrors() status.MultiError
+	// Syncing returns true if the updater is running.
+	Syncing() bool
 	// K8sClient returns the Kubernetes client that talks to the API server.
 	K8sClient() client.Client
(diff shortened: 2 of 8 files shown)


karlkfi:
Looks like the failures are coming from Gets after the Watch. The Watch saw the error, but the Get doesn't. So it's losing the status somehow in a short amount of time, but the test doesn't log the full state of the invalid RSync to know what else changed.

For example:

```
  testlogger.go:77: 2024-06-21 16:52:14.565284054 +0000 UTC WatchObject(RootSync config-management-system/root-sync) watched for 13.544000813s
    hydration_test.go:137: 2024-06-21 16:52:14.612161534 +0000 UTC ERROR: Error: rootSync does not match expected. Got: commit: bb7b4c8f5ba29f1a15231fbc9e6dd9cca3649282, status: SYNCED
        Error Summary: <nil>
        Expected: commit: a9d62866db1b069c531ab419c49b0624a43880ab, status: ERROR
        
        Expected error: Kustomization should be missing.
```

Less than a second after the Watch suceeded with the RootSyncHasRenderingError predicate, the validateRootSyncRepoState failed to find the error and the commit had changed.

karlkfi:
Summarizing findings from chat:

It looks like the bug I fixed in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1283 was a load-bearing bug.

`setRenderingStatus` (and `setSourceStatus`) both update the Syncing condition, not just `SetSyncStatus`. But when `setRenderingStatus` is called, it only populates the Syncing condition with rendering errors, which means there may still be remediator errors in `status.sync.errors` from the previous commit, but those errors will no longer be present in the Syncing condition, because the Syncing condition has been updated to reference the newer commit.

So at some point, a check was added to stop calling `SetSyncStatus` after the source or rendering commit had changed, to stop reverting the Syncing condition commit & errors. But this had the side effect of also stopping updates to the `status.sync.errors` from the remediator running in parallel (which is the bug https://github.com/GoogleContainerTools/kpt-config-sync/pull/1283 was supposed to fix).

So I'm going to have to re-think this fix and figure out how to fix or refactor  `SetSyncStatus` so we can still update `status.sync.errors` without reverting the Syncing condition.

karlkfi (approved):
/lgtm

PR #1287, opened 2024-06-24
e2e: cleanup hydration tests

karlkfi:
- Refactor nomos status validation as Predicate
- Move nomos status validation into existing watches to ensure both conditions are true at the same time. This works around a race condition caused by retries removing errors from the Syncing condition.
- Use NT.Must where possible
- Use TaskGroup to parallelize validations

This PR changes the expectations for rendering tests to account for the possibility that the rendering errors may or may not always show up in the Syncing condition after the error. This allows us to later update the code to update the `status.source.commit` & Syncing condition after fetching but before rendering.

### DIFF
--- a/e2e/testcases/hydration_test.go
+++ b/e2e/testcases/hydration_test.go
@@ -15,15 +15,12 @@
 package e2e
 
 import (
-	"fmt"
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
-	nomosstatus "kpt.dev/configsync/cmd/nomos/status"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/kustomizecomponents"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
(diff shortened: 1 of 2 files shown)


e2e/testcases/hydration_test.go
```diff
@@ -278,61 +249,41 @@ func TestHydrateHelmOverlay(t *testing.T) {
...
 
 	nt.T.Log("Make the hydration fail by checking in an invalid kustomization.yaml")
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Copy("../testdata/hydration/resource-duplicate/kustomization.yaml", "./helm-overlay/kustomization.yaml"))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Copy("../testdata/hydration/resource-duplicate/namespace_tenant-a.yaml", "./helm-overlay/namespace_tenant-a.yaml"))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("Update kustomization.yaml with duplicated resources"))
+	latestCommit = nt.RootRepos[configsync.RootSyncName].MustHash(nt.T)
 	nt.Must(nt.Watcher.WatchObject(kinds.RootSyncV1Beta1(), configsync.RootSyncName, configsync.ControllerNamespace,
 		[]testpredicates.Predicate{
-			testpredicates.RootSyncHasRenderingError(status.ActionableHydrationErrorCode, ""),
+			testpredicates.RootSyncHasRenderingError(status.ActionableHydrationErrorCode, "failed to run kustomize build"),
```
sdowell, line 274: Why did the expected error message change here?

sdowell: Seems like it's just making the assertion stricter, which LGTM

sdowell (approved):
/retest
/lgtm

e2e/testcases/hydration_test.go
karlkfi: Similar text was in the validation Fatalf wrapper. So I moved it to the actual assertion so it would print what was expected when it fails.

PR #1288, opened 2024-06-25
test: increase helm-sync memory for TestPublicHelm

sdowell:
This test occasionally causes an OOMKill for the helm-sync container on Autopilot.


e2e/testcases/helm_sync_test.go
```diff
@@ -62,10 +65,66 @@ func TestPublicHelm(t *testing.T) {
...
+  }
+}`),
+				},
+			},
+		},
+	}
+	if nt.IsGKEAutopilot {
+		nt.T.Log("Increasing memory request/limit for helm-sync on Autopilot")
+		rs.Spec.SafeOverride().Resources = []v1beta1.ContainerResourcesSpec{
+			{ // This chart sometimes causes OOMKill on Autopilot with default limit
```
nan-yu, line 119: While the WordPress Helm chart is not resource-intensive, if OOMKill errors occur in our end-to-end tests, we should probably consider bumping the default limit.

sdowell: I agree, although it typically involves more convincing to increase the default limits

nan-yu: Do you know how frequent we are hitting OOMKills and which Autopilot channel it is?

sdowell: It's currently the greatest source of flakiness on autopilot, which is why I decided to take a closer look at this test. See:
- https://testgrid.k8s.io/googleoss-kpt-config-sync-main#autopilot-regular
- https://testgrid.k8s.io/googleoss-kpt-config-sync-main#autopilot-rapid
- https://testgrid.k8s.io/googleoss-kpt-config-sync-main#autopilot-rapid-latest

It appears to be failing roughly 20% of the time or so

nan-yu: I see. Let's go with increasing the limit in e2e tests only. We can increase the default when more test cases are failing.

nan-yu (approved):
/lgtm

PR #1289, opened 2024-06-25
refactor: use SourceType type in RSync spec

sdowell:
This changes the type of the sourceType field from string to the defined type in the RSync spec. This reduces the number of type conversions required and makes the code more self documenting.

The intent is to apply a similar refactor to `SourceFormat` with a subsequent PR.

### DIFF
--- a/pkg/api/configsync/v1beta1/rootsync_types.go
+++ b/pkg/api/configsync/v1beta1/rootsync_types.go
@@ -16,6 +16,7 @@ package v1beta1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"kpt.dev/configsync/pkg/api/configsync"
 )
 
 // +kubebuilder:object:root=true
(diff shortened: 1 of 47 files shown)


pkg/api/configsync/v1beta1/rootsync_types.go
```diff
@@ -58,8 +59,9 @@ type RootSyncSpec struct {
 	// Must be one of git, oci, helm. Optional. Set to git if not specified.
 	// +kubebuilder:validation:Pattern=^(git|oci|helm)$
```
sdowell, line 60: I think it might make sense to change this to an `enum` instead of a `pattern`, but this results in a CRD diff so I think that should be a separate change

nan-yu: Agreed. We can change `sourceFormat` to enum as well.

nan-yu (approved):
/lgtm

PR #1290, opened 2024-06-25
fix: multiple status update bugs 

karlkfi:
- Update status.source after fetch before rendering if the commit
  changed. This allows using status.source.commit to see what
  commit has been fetched even if rendering fails.
  This also causes the Syncing condition to be updated at the same
  time, which tells the user the remediator is working on a new
  commit.
- Update status.sync periodically after a new commit has been
  published to status.source.commit or status.rendering.commit.
  This allows continuing to update remediator errors after a new
  commit has been fetched. Normally this isn't a long time, but it
  can be if there's a source or rendering error requiring change
  by the user.
- Update SetSyncStatus to only update the Syncing condition if
  the reconciler hasn't fetched a new commit. This prevents
  periodic updates from overwriting the Syncing condition for newer
  commits.
- Update SetSyncStatus to only record the last_sync_timestamp metric
  if the reconciler hasn't fetched a new commit. This prevents
  re-recording the metric, flapping between commits.
- Update SetSyncStatus, to only aggregate errors for a specific
  commit, avoiding the Syncing condition from referencing errors from
  old commits.

This replaces https://github.com/GoogleContainerTools/kpt-config-sync/pull/1283, which was reverted by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1286. The test failures should be fixed by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1287, which updated the expectations to check the rendering status and Syncing condition at the same time.

### DIFF
--- a/pkg/parse/namespace.go
+++ b/pkg/parse/namespace.go
@@ -271,14 +271,22 @@ func (p *namespace) setSyncStatusWithRetries(ctx context.Context, newStatus sync
 
 	setSyncStatusFields(&rs.Status.Status, newStatus, denominator)
 
-	errorSources, errorSummary := summarizeErrors(rs.Status.Source, rs.Status.Sync)
-	if newStatus.syncing {
-		reposync.SetSyncing(rs, true, "Sync", "Syncing", rs.Status.Sync.Commit, errorSources, errorSummary, rs.Status.Sync.LastUpdate)
-	} else {
-		if errorSummary.TotalCount == 0 {
-			rs.Status.LastSyncedCommit = rs.Status.Sync.Commit
+	// The Syncing condition should only represent the status and errors for the latest commit.
+	// So only update the Syncing condition here if we haven't fetched a new commit.
+	// Ideally, checking the source commit would be enough, but because fetching and parsing share the source status,
+	// we also have to check the rendering commit, which may be updated first.
+	var lastSyncStatus string
+	if rs.Status.Source.Commit == rs.Status.Sync.Commit && rs.Status.Rendering.Commit == rs.Status.Sync.Commit {
+		errorSources, errorSummary := summarizeErrorsForCommit(rs.Status.Source, rs.Status.Rendering, rs.Status.Sync, rs.Status.Sync.Commit)
+		if newStatus.syncing {
+			reposync.SetSyncing(rs, true, "Sync", "Syncing", rs.Status.Sync.Commit, errorSources, errorSummary, rs.Status.Sync.LastUpdate)
+		} else {
+			if errorSummary.TotalCount == 0 {
+				rs.Status.LastSyncedCommit = rs.Status.Sync.Commit
+			}
+			reposync.SetSyncing(rs, false, "Sync", "Sync Completed", rs.Status.Sync.Commit, errorSources, errorSummary, rs.Status.Sync.LastUpdate)
 		}
-		reposync.SetSyncing(rs, false, "Sync", "Sync Completed", rs.Status.Sync.Commit, errorSources, errorSummary, rs.Status.Sync.LastUpdate)
+		lastSyncStatus = metrics.StatusTagValueFromSummary(errorSummary)
 	}
 
 	// Avoid unnecessary status updates.
@@ -294,8 +302,9 @@ func (p *namespace) setSyncStatusWithRetries(ctx context.Context, newStatus sync
 		klog.Infof("New sync errors for RepoSync %s/%s: %+v",
 			rs.Namespace, rs.Name, csErrs)
 	}
-	if !newStatus.syncing && rs.Status.Sync.Commit != "" {
-		metrics.RecordLastSync(ctx, metrics.StatusTagValueFromSummary(errorSummary), rs.Status.Sync.Commit, rs.Status.Sync.LastUpdate.Time)
+	// Only update the LastSyncTimestamp metric immediately after a sync attempt
+	if !newStatus.syncing && rs.Status.Sync.Commit != "" && lastSyncStatus != "" {
+		metrics.RecordLastSync(ctx, lastSyncStatus, rs.Status.Sync.Commit, rs.Status.Sync.LastUpdate.Time)
 	}
 
 	if klog.V(5).Enabled() {
@@ -319,12 +328,6 @@ func (p *namespace) setSyncStatusWithRetries(ctx context.Context, newStatus sync
 // SyncErrors implements the Parser interface
 func (p *namespace) SyncErrors() status.MultiError {
 	return p.SyncErrorCache.Errors()
-}
-
-// Syncing returns true if the updater is running.
-// SyncErrors implements the Parser interface
-func (p *namespace) Syncing() bool {
-	return p.Updating()
 }
 
 // K8sClient implements the Parser interface
--- a/pkg/parse/root.go
+++ b/pkg/parse/root.go
@@ -396,14 +396,22 @@ func (p *root) setSyncStatusWithRetries(ctx context.Context, newStatus syncStatu
 
 	setSyncStatusFields(&rs.Status.Status, newStatus, denominator)
 
-	errorSources, errorSummary := summarizeErrors(rs.Status.Source, rs.Status.Sync)
-	if newStatus.syncing {
-		rootsync.SetSyncing(rs, true, "Sync", "Syncing", rs.Status.Sync.Commit, errorSources, errorSummary, rs.Status.Sync.LastUpdate)
-	} else {
-		if errorSummary.TotalCount == 0 {
-			rs.Status.LastSyncedCommit = rs.Status.Sync.Commit
+	// The Syncing condition should only represent the status and errors for the latest commit.
+	// So only update the Syncing condition here if we haven't fetched a new commit.
+	// Ideally, checking the source commit would be enough, but because fetching and parsing share the source status,
+	// we also have to check the rendering commit, which may be updated first.
+	var lastSyncStatus string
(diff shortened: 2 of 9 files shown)


karlkfi:
Looks like this causes some issues with periodic sync status updates:

1. The periodic sync status updates will replace the Syncing condition of a newer commit with one from an older commit.
2. The periodic sync status updates trigger `metrics.RecordLastSync` ever time there's a change in remediator errors (new error added, old error removed, or even just a change of order)
3. The periodic sync status updates include source errors (and render errors?) from new commits, even if they don't match the sync commit. This was previously being prevented by requiring the commits to match in the run loop.

There's also another concern that seems to be pre-existing:

- The source/rendering/sync status fields and syncing condition do not include the RSync generation. So it's possible that the status is out of date, even if the commit is the same. For example, the sync dir could have changed.
- This problem is compounded by error aggregation in the Syncing condition,  which could group errors from different RSync specs.
- This was surfaced in go/config-sync-status-improvements (in 2022) but never fixed.

karlkfi:
Alright, I figured out how to fix the Syncing status problems.

- I moved the old commit check from the run loop into the setSyncStatusWithRetries impl. So it can now update `status.sync` fields without updating the Syncing condition, when `status.source.commit` or  `status.rendering.commit` are newer than (don't match)  `status.sync.commit`. This should allow populating remediator errors without overwriting the Syncing condition for new commits.
- I added a check around `metrics.RecordLastSync` so it's only called when the all the commits match. This should stop it from being called after a new commit has been fetched. 
- Just in case, I also changed `summarizeErrors` to `summarizeErrorsForCommit` and made it match the sync commit. So it will not include source and rendering errors from other commits. It shouldn't be getting called, but this makes doubly sure it won't aggregate across commits. It also includes render errors, if any exist, but since there aren't any non-terminal rendering errors, this isn't likely.

The fix for the remediator error updates partially depends on the fix for the source updates before rendering, because they update the `status.source.commit`, which stops the periodic sync updates from updating the Syncing condition. This would otherwise happen a little later, when  `status.rendering.commit`  is updated.

pkg/parse/root.go
```diff
@@ -457,18 +463,30 @@ func setSyncStatusErrors(syncStatus *v1beta1.Status, cse []v1beta1.ConfigSyncErr
...
+// for a specific commit.
+//
+// Since we don't keep errors for old commits after that stage has started
+// working on a new commit, this process is a little bit lossy.
+func summarizeErrorsForCommit(sourceStatus v1beta1.SourceStatus, renderingStatus v1beta1.RenderingStatus, syncStatus v1beta1.SyncStatus, commit string) ([]v1beta1.ErrorSource, *v1beta1.ErrorSummary) {
 	var errorSources []v1beta1.ErrorSource
-	if len(sourceStatus.Errors) > 0 {
+	var summaries []*v1beta1.ErrorSummary
+
+	if sourceStatus.Commit == commit && len(sourceStatus.Errors) > 0 {
```
sdowell, line 475: So this means we might have an error for a previous commit on the `renderingStatus`, but the `Syncing` condition will only contain the `errors`/`errorSummary` for the current commit? That makes sense, given that the `commit` is a condition field.

pkg/parse/run.go
```diff
@@ -236,22 +235,31 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
...
+			if setSourceStatusErr != nil {
+				// If there were errors from git-sync, log those too
+				state.invalidate(status.Append(gs.errs, setSourceStatusErr))
+				return
 			}
+			// Cache the latest source status in memory
+			state.sourceStatus = gs
+			state.syncingConditionLastUpdate = gs.lastUpdate
+		}
+		// If there were errors from git-sync, stop, log them, and retry later
```
sdowell, line 258: nit: could be helm-sync/oci-sync. Maybe call it `*-sync`

pkg/parse/root.go
karlkfi: That's actually how it is today, because the old code here didn't include any rendering errors and is only called from `setSyncStatusWithRetries`. 

I assume that's because it was originally only sync errors, but then source errors were added when we added non-blocking errors to handle the PoCo case that removes CRs without live CRDs from the apply set.

This new `summarizeErrorsForCommit` is more robust because it includes defensive logic to only ever include errors for the specified commit. But I also moved the commit check from the run loop to `setSyncStatusWithRetries`, which means  `summarizeErrorsForCommit` is actually only ever called when the commits all match. So the change here is redundant/defensive, but I think it also makes the code a little easier to understand and might prevent breakage if we add other updates in the future, like for source or rendering in-progress.

Ideally, the update code would look similar for each stage, but it's all grown very organically and drifted quite a bit just to fix bugs as they are discovered. I'd really love to go in and clean it all up, but it never seems to be top priority.

pkg/parse/run.go
karlkfi: done.

pkg/parse/namespace.go
```diff
@@ -271,14 +271,19 @@ func (p *namespace) setSyncStatusWithRetries(ctx context.Context, newStatus sync
 
 	setSyncStatusFields(&rs.Status.Status, newStatus, denominator)
 
-	errorSources, errorSummary := summarizeErrors(rs.Status.Source, rs.Status.Sync)
-	if newStatus.syncing {
-		reposync.SetSyncing(rs, true, "Sync", "Syncing", rs.Status.Sync.Commit, errorSources, errorSummary, rs.Status.Sync.LastUpdate)
-	} else {
-		if errorSummary.TotalCount == 0 {
-			rs.Status.LastSyncedCommit = rs.Status.Sync.Commit
+	// Only update the Syncing condition if we haven't already fetched a new commit.
```
sdowell, line 274: What are the scenarios we're trying to protect against here? Could you add a bit more context to the comment?

karlkfi: If there's been a new commit, it's ok for the new commit fetch status to override the old fetch/parse status. But if it's an old commit, there could be parse errors we don't want to erase. This is necessary because the fetch and parse status both share the status.source fields. 

A better fix long term would be to give the fetch & parse phases different status fields, but that should have been done when rendering was added and wasn't. So now it's API debt. Fixing it would require also changing the CLI & UI. 

@nan-yu may end up needing to do this as a blocker for hooks. But for now the API is left as-is with slightly different behavior.

karlkfi: We also don't need to update the last updated timestamp or syncing condition if the commit hasn't changed. So we can just bypass the whole update. 

It's hypothetically possible that parsing succeeds after previously failing, but I can't think of many cases where this would happen without a new commit. Maybe a disk read error? If that's the case, the error will eventually be removed when parsing succeeds or fails, which is how it works today already.

karlkfi: Updated the comment with more context.

pkg/parse/root.go
```diff
@@ -396,14 +396,22 @@ func (p *root) setSyncStatusWithRetries(ctx context.Context, newStatus syncStatu
...
-		rootsync.SetSyncing(rs, true, "Sync", "Syncing", rs.Status.Sync.Commit, errorSources, errorSummary, rs.Status.Sync.LastUpdate)
-	} else {
-		if errorSummary.TotalCount == 0 {
-			rs.Status.LastSyncedCommit = rs.Status.Sync.Commit
+	// The Syncing condition should only represent the status and errors for the latest commit.
+	// So only update the Syncing condition here if we haven't fetched a new commit.
+	// Ideally, checking the source commit would be enough, but because fetching and parsing share the source status,
+	// we also have to check the rendering commit, which may be updated first.
+	var lastSyncStatus string
+	if rs.Status.Source.Commit == rs.Status.Sync.Commit && rs.Status.Rendering.Commit == rs.Status.Sync.Commit {
```
sdowell, line 404: Correct me if I'm wrong, but I think the intent is that the `Syncing` condition should be set to `true` here whenever the applier is currently running (`newStatus.syncing`).

When the applier is running it's expected that the on-cluster RSync will have updated `status.source` and `status.rendering`, but not yet `status.sync`. The `status.sync` field is set by `setSyncStatusFields` call above, which won't be written to the cluster until the Update below on line 440.

I believe this means the first time the applier calls [SetSyncStatus](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L613), the `rs.Status.Source.Commit` won't yet be updated and we would skip the condition update here. To avoid this I think this comparison needs to compare the commits for `currentRS` instead of `rs`.

TL;DR: `s/rs/currentRS/g`

pkg/parse/namespace.go
sdowell: To clarify - the reason for all of this complexity is because of the remediator status updates, right? It seems like in the absence of the remediator status updates, all of the status updates would be sequential and we wouldn't need to compare commits.

karlkfi: Correct. Having the remediator run in parallel to the rest of the pipeline makes this complicated. And having remediator errors share the same status fields & conditions makes it worse.

pkg/parse/root.go
karlkfi: `rs.Status.Source.Commit` updates are published by `setSourceStatus`, not `SetSyncStatus`. So that's always published to the cluster before `SetSyncStatus` runs.

`rs.Status.Sync.Commit` is updated locally by `setSyncStatusFields`, so it's safe to use it before it's published to the cluster.

sdowell: Ah I mixed up the two, `currentRS` is the unchanged on-cluster RSync and `rs` is the updated one. Disregard

sdowell (approved):
/lgtm

PR #1292, opened 2024-06-25
refactor: use SourceFormat in RSync spec

sdowell:
This changes the type of the sourceFormat field from string to the defined type in the RSync spec. This reduces the number of type conversions required and makes the code more self documenting.


karlkfi (approved):
/lgtm

PR #1294, opened 2024-06-26
test: clean up root reconciler ClusterRoleBindings

sdowell:
If a finalizer is force removed during cleanup, it is possible that root reconciler ClusterRoleBindings may be orphaned on the cluster. This has caused some test failures due to the lingering resources.

Example test failure: https://oss.gprow.dev/view/gcs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-regular/1805843581710110720


nan-yu (approved):
/lgtm

PR #1296, opened 2024-06-27
Refactor reconciler status caching

karlkfi:
Prior to this change, the reconcilerState in the parser package contained both status reported to the RSync and internal state used to keep track of progress.

With this change, the status persisted to the RSync is now under the ReconcilerStatus struct, under reconcilerState. This seperates the two different domains in preperation for adding more stage-specific caching for RSync spec, which is used to populate the RSync status.

ReconcilerStatus uses primarily public fields, because it helps distinguish between fields intended for external manipulation and fields that should only be manipulated by methods on the struct, even if the manipulator is in the same package. Hopefully, we can expand this convention in the future, but for now it's limited in scope to keep the size of this PR reviewable.

Blocks: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1276


sdowell (approved):
/lgtm

PR #1298, opened 2024-06-27
feat: error backoff for oci-sync

sdowell:
This adds exponential backoff to the oci-sync when it encounters repeated errors. This makes the behavior comparable to the behavior for helm-sync. The code is refactored to use a common backoff creation function and a minimum cap is defined for the sync containers.

### DIFF
--- a/cmd/oci-sync/main.go
+++ b/cmd/oci-sync/main.go
@@ -18,10 +18,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math"
 	"os"
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2/textlogger"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/oci"
@@ -49,6 +51,11 @@ var flOneTime = flag.Bool("one-time", util.EnvBool("OCI_SYNC_ONE_TIME", false),
 	"exit after the first sync")
 var flMaxSyncFailures = flag.Int("max-sync-failures", util.EnvInt("OCI_SYNC_MAX_SYNC_FAILURES", 0),
 	"the number of consecutive failures allowed before aborting (the first sync must succeed, -1 will retry forever after the initial sync)")
+
+func errorBackoff() wait.Backoff {
+	durationLimit := math.Max(*flWait, float64(util.MinimumSyncContainerBackoffCap))
+	return util.BackoffWithDurationAndStepLimit(util.WaitTime(durationLimit), math.MaxInt32)
+}
 
 func main() {
 	utillog.Setup()
@@ -82,6 +89,8 @@ func main() {
 
 	initialSync := true
 	failCount := 0
+	backoff := errorBackoff()
+
 	for {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(*flSyncTimeout))
 		if err := oci.FetchPackage(ctx, log, *flAuth, *flImage, *flRoot, *flDest); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/oci-sync/main.go
```diff
@@ -50,6 +52,11 @@ var flOneTime = flag.Bool("one-time", util.EnvBool("OCI_SYNC_ONE_TIME", false),
 var flMaxSyncFailures = flag.Int("max-sync-failures", util.EnvInt("OCI_SYNC_MAX_SYNC_FAILURES", 0),
 	"the number of consecutive failures allowed before aborting (the first sync must succeed, -1 will retry forever after the initial sync)")
 
+func errorBackoff() wait.Backoff {
+	durationLimit := math.Max(*flWait, float64(util.MinimumSyncContainerBackoffCap))
+	return util.BackoffWithDurationLimit(util.WaitTime(durationLimit), math.MaxInt32)
```
nan-yu, line 57: We can use `*flMaxSyncFailures` as the step cap if it is not `-1`.
We'll need to update how flMaxSyncFailures is used.

nan-yu: In L47, the description of the `--wait` flag is not true. Instead of a fixed wait period, it now represents the wait duration cap. We need to update it, and the helm-sync too.

pkg/util/retry.go
```diff
@@ -65,16 +67,16 @@ func IsErrorRetriable(err error) bool {
...
 }
 
-// BackoffWithDurationLimit returns backoff with a duration limit in 10 steps.
+// BackoffWithDurationLimit returns backoff with a duration limit.
 // Here is an example of the duration between steps:
 //
 //	1.055843837s, 2.085359785s, 4.229560375s, 8.324724174s, 16.295984061s,
 //	34.325711987s, 1m5.465642392s, 2m18.625713221s, 4m24.712222056s, 9m18.97652295s.
-func BackoffWithDurationLimit(duration time.Duration) wait.Backoff {
+func BackoffWithDurationLimit(duration time.Duration, steps int) wait.Backoff {
```
nan-yu, line 75: Now it takes both duration and steps, we need to rename the function.

nan-yu, line 75: https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.18.2-rc.2/pkg/parse/state.go#L100 can be consolidated with the same backoff function.

cmd/oci-sync/main.go
sdowell: I intentionally chose not to add a step cap here, I think it just makes the backoff more complex to understand. For this use case I consider it simpler to just define either a duration cap or a step cap, but not both. Considering the process exits when `*flMaxSyncFailures` is reached, it makes no difference if we map this to the step cap.

`--wait` is still the number of seconds between syncs for the happy path. It's only used as the duration cap if the provided value is greater than the default cap of 1 hour. I think this can be treated as an implementation detail.

pkg/util/retry.go
sdowell: done

sdowell: done

nan-yu (approved):
/lgtm

pkg/parse/state.go
```diff
@@ -98,12 +99,7 @@ const retryLimit = 12
 //	16.295984061s, 34.325711987s, 1m5.465642392s, 2m18.625713221s,
 //	4m24.712222056s, 9m18.97652295s, 17m15.344384599s, 35m15.603237976s.
 func defaultBackoff() wait.Backoff {
-	return wait.Backoff{
-		Duration: 1 * time.Second,
-		Factor:   2,
-		Steps:    retryLimit,
-		Jitter:   0.1,
-	}
+	return util.BackoffWithDurationAndStepLimit(0, retryLimit)
```
karlkfi, line 102: Hmm. What happened to RetryPeriod? Isn't the base duration supposed to be configurable?

The reconciler.Options.RetryPeriod get passed to the parser.Options.RetryPeriod, but then never used?

sdowell: Appears it's used in a few places:
- https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L92
- https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L106
- https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L162
- https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/run.go#L329

PR #1299, opened 2024-06-29
Inject Parser.Clock for testing

karlkfi:
- Update TestRun & TestRoot_Parse to validate the RootSync status
- Change SourceStatus, RenderingStatus, and SyncStatus to pointers and add DeepCopy methods, for easier generation of test permutations.
- Remove `cacheForCommit.errs`, which was only being used for testing errors in TestRun. Errors are now tested from the RSync status.


nan-yu (approved):
/lgtm

PR #1239, opened 2024-05-30
[WIP] Add watch filtering by package-id label (obsolete)

karlkfi:
go/config-sync-watch-filter

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1199
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1204
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1241
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1244

TODO:
- Extract SyncStatusUpdater changes
- Extract conflict handler changes
- Finalize design approval (some open questions about label naming)

PR #1301, opened 2024-07-02
Add watch filtering by applyset label

karlkfi:
- Add "applyset.kubernetes.io/part-of" label to managed objects
- Add "applyset.kubernetes.io/id" label to RootSyncs/RepoSyncs
- Add "applyset.kubernetes.io/tooling" annotation to RootSyncs/RepoSyncs
- Error if the RSync already has a tooling annotation for a different tool (unlikely, but required by KEP).
- Generate ApplySet IDs using kubectl code for compatibility.
- Update TestKubectlCreatesManagedNamespaceResourceMultiRepo & TestAddUpdateDeleteLabels with new labels & annotations

### Design

go/config-sync-watch-filter

#### Metadata Prefixes

In the KEP, the labels and annotations begin with `applyset.k8s.io`. However, when implemented in kubectl, this was changed to `applyset.kubernetes.io`. The reason for this is unclear. But we decided compatibility with kubectl was more important, since it's the only other known ApplySet implementation at this time.

#### ApplySet ID Format

In the KEP, the ID format is specified as `base64(sha256(<NAME>.<NAMESPACE>.<KIND>.<RESOURCE_GROUP>))`. However, when implemented in kubectl, this format was wrapped with the prefix `applyset-` and suffix `-v1`, because "Label values must start and end with alphanumeric values" (https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/applyset.go#L175).

If I were designing this again, I would recommend using `hex(fnv32a(<NAME>.<NAMESPACE>.<KIND>.<RESOURCE_GROUP>))`, because it's faster and only includes alphanumeric values. But we decided compatibility with the KEP & kubectl was more important.

### Tooling Format

For consistency with other Config Sync labels and annotations, we decided on `configsync.gke.io` as the name of the tool.

In the KEP, the tooling format is recommended to be `<toolname>/<semver>`, with examples that show a loose definition of semver with a leading `v` prefix and without requiring all three version segments.

In order to facilitate listing ALL the objects of a resource that are managed by Config Sync, and avoid the complexity of needing to inject the latest version into the code before the tag has been created, we opted here to just use `v1` as the version. Specifying a version also conveniently allows us to share more code with kubectl and stick to the guidelines in the KEP for compatibility with other tools that may try to parse the tooling value and not be as lenient as kubectl.

So the tooling value is `configsync.gke.io/v1`, and should continue using this version for the foreseeable future.

### Dependencies

- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1274
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1328
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1335
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1342
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1343
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1344
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1354
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1356

### Follow Ups

- Race condition: Remediate deletion of objects between remediator pasue and unpause, while the applier is running (b/353334354 & b/355532135)
- Distinguish between resource version conflict & management conflict (b/355533809)
- Remove "applyset.kubernetes.io/part-of" label when an object is abandoned. (b/355534413)

### Unplanned

- Add "applyset.k8s.io/additional-namespaces" annotation to RootSyncs/RepoSyncs.
  - This annotation would need to be applied by the reconciler, not the reconciler-manager.
  - The list of namespaces may change frequently for RootSync, whenever a new source commit is seen.
- Add "applyset.k8s.io/contains-group-kinds" annotation to RootSyncs/RepoSyncs.
  - This annotation would need to be applied by the reconciler, not the reconciler-manager.
  - The list of namespaces may change frequently for RootSync, whenever a new source commit is seen.
  - For large apply sets, especially when using may resource types (ex: KCC or Crossplane), the inventory annotation length may dramatically increase the object size, increasing the risk of reaching the max object size in etcd.
- Add "applyset.k8s.io/inventory" annotation to RootSyncs/RepoSyncs.
  - This annotation would need to be applied by the reconciler, not the reconciler-manager.
  - The list of namespaces may change frequently for RootSync, whenever a new source commit is seen.
  - For large apply sets, the inventory annotation length may overflow the max annotation length.
  - For large apply sets, the inventory annotation length may dramatically increase the object size, increasing the risk of reaching the max object size in etcd.
- Error when attempting to apply an object with a non-matching "applyset.kubernetes.io/part-of" label.
  - Look ups before applying were removed from Config Sync (a few years ago) to reduce the number of API calls made and speed up applying large apply sets. Adding them back would be required to perform this kind of validation/protection.

### Notes

This PR replaces https://github.com/GoogleContainerTools/kpt-config-sync/pull/1239 with less refactoring. The refactoring can be done separately.

### DIFF
--- a/e2e/testcases/managed_resources_test.go
+++ b/e2e/testcases/managed_resources_test.go
@@ -22,66 +22,65 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
+	"kpt.dev/configsync/e2e/nomostest/testkubeclient"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/e2e/nomostest/testwatcher"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/applyset"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/core/k8sobjects"
+	"kpt.dev/configsync/pkg/declared"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/metadata"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // This file includes tests for drift correction and drift prevention.
 //
-// The drift correction in the mono-repo mode utilizes the following two annotations:
-//  * configmanagement.gke.io/managed
-//  * configsync.gke.io/resource-id
+// Drift Correction uses the following metadata:
+//  * The "configmanagement.gke.io/managed" annotation must be set to "enabled".
+//  * The "configsync.gke.io/resource-id" annotation must match the object ID.
+//  * The "configsync.gke.io/manager" annotation must exists and match the RSync's ResourceManager name.
+//  * The "applyset.kubernetes.io/part-of" label must match the RSync's ApplySet ID.
 //
-// The drift correction in the multi-repo mode utilizes the following three annotations:
-//  * configmanagement.gke.io/managed
-//  * configsync.gke.io/resource-id
-//  * configsync.gke.io/manager
-//
-// The drift prevention is only supported in the multi-repo mode, and utilizes the following Config Sync metadata:
-//  * the configmanagement.gke.io/managed annotation
-//  * the configsync.gke.io/resource-id annotation
-//  * the configsync.gke.io/declared-version label
+// Drift Prevention uses the following metadata:
+//  * The "configmanagement.gke.io/managed" annotation must be set to "enabled".
+//  * The "configsync.gke.io/resource-id" annotation must match the object ID.
 
-// The reason we have both TestKubectlCreatesManagedNamespaceResourceMonoRepo and
-// TestKubectlCreatesManagedNamespaceResourceMultiRepo is that the mono-repo mode and
-// CSMR handles managed namespaces which are created by other parties differently:
-//   * the mono-repo mode does not remove these namespaces;
-//   * CSMR does remove these namespaces.
-
-// TestKubectlCreatesManagedNamespaceResourceMultiRepo tests the drift correction regarding kubectl
-// tries to create a managed namespace in the the multi-repo mode.
-func TestKubectlCreatesManagedNamespaceResourceMultiRepo(t *testing.T) {
+// TestDriftKubectlApplyClusterScoped tests drift correction after
+// cluster-scoped changes are made with kubectl.
+func TestDriftKubectlApplyClusterScoped(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.DriftControl, ntopts.Unstructured)
+
+	rootSync2Name := "abcdef"
+	rootSync1ApplySetID := applyset.IDFromSync(configsync.RootSyncName, declared.RootScope)
+	rootSync2ApplySetID := applyset.IDFromSync(rootSync2Name, declared.RootScope)
+	rootSync1Manager := declared.ResourceManager(declared.RootScope, configsync.RootSyncName)
+	rootSync2Manager := declared.ResourceManager(declared.RootScope, rootSync2Name)
 
 	namespace := k8sobjects.NamespaceObject("bookstore")
 	nt.Must(nt.RootRepos[configsync.RootSyncName].Add("acme/ns.yaml", namespace))
 	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add a namespace"))
-	if err := nt.WatchForAllSyncs(); err != nil {
-		nt.T.Fatal(err)
-	}
+	nt.Must(nt.WatchForAllSyncs())
 
 	/* A new test */
-	ns := []byte(`
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: test-ns1
-  annotations:
-    configmanagement.gke.io/managed: enabled
-    configsync.gke.io/resource-id: _namespace_test-ns1
-    configsync.gke.io/manager: :root
-`)
+	ns1Obj := &corev1.Namespace{}
+	ns1Obj.SetName("test-ns1")
+	ns1Obj.SetAnnotations(map[string]string{
+		metadata.ResourceManagementKey: metadata.ResourceManagementEnabled,
+		metadata.ResourceIDKey:         "_namespace_test-ns1",
+		metadata.ResourceManagerKey:    rootSync1Manager,
+	})
+	ns1Obj.SetLabels(map[string]string{
+		metadata.ApplySetPartOfLabel: rootSync1ApplySetID,
+	})
 
-	if err := os.WriteFile(filepath.Join(nt.TmpDir, "test-ns1.yaml"), ns, 0644); err != nil {
-		nt.T.Fatalf("failed to create a tmp file %v", err)
+	if err := writeObjectYAMLFile(filepath.Join(nt.TmpDir, "test-ns1.yaml"), ns1Obj, nt.Scheme); err != nil {
+		nt.T.Fatal(err)
 	}
 
 	out, err := nt.Shell.Kubectl("apply", "-f", filepath.Join(nt.TmpDir, "test-ns1.yaml"))
@@ -89,7 +88,7 @@ metadata:
 		nt.T.Fatalf("got `kubectl apply -f test-ns1.yaml` error %v %s, want return nil", err, out)
 	}
 
-	// Remediator should delete `test-ns1`, because it says it's managed by the
+	// Remediator SHOULD delete `test-ns1`, because it says it's managed by the
 	// root-reconciler, but it's not in the source declared resources.
 	err = nt.Watcher.WatchForNotFound(kinds.Namespace(), "test-ns1", "")
 	if err != nil {
@@ -97,19 +96,19 @@ metadata:
 	}
 
 	/* A new test */
-	ns = []byte(`
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: test-ns2
-  annotations:
-    configmanagement.gke.io/managed: enabled
-    configsync.gke.io/resource-id: _namespace_test-ns2
-    configsync.gke.io/manager: :abcdef
-`)
+	ns2Obj := &corev1.Namespace{}
+	ns2Obj.SetName("test-ns2")
+	ns2Obj.SetAnnotations(map[string]string{
+		metadata.ResourceManagementKey: metadata.ResourceManagementEnabled,
+		metadata.ResourceIDKey:         "_namespace_test-ns2",
+		metadata.ResourceManagerKey:    rootSync2Manager,
+	})
+	ns2Obj.SetLabels(map[string]string{
+		metadata.ApplySetPartOfLabel: rootSync2ApplySetID,
+	})
 
-	if err := os.WriteFile(filepath.Join(nt.TmpDir, "test-ns2.yaml"), ns, 0644); err != nil {
-		nt.T.Fatalf("failed to create a tmp file %v", err)
+	if err := writeObjectYAMLFile(filepath.Join(nt.TmpDir, "test-ns2.yaml"), ns2Obj, nt.Scheme); err != nil {
+		nt.T.Fatal(err)
 	}
 
 	out, err = nt.Shell.Kubectl("apply", "-f", filepath.Join(nt.TmpDir, "test-ns2.yaml"))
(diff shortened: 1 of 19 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -320,11 +320,15 @@ func (r *RepoSyncReconciler) upsertManagedObjects(ctx context.Context, reconcile
 }
 
 // setup performs the following steps:
+// - Patch RepoSync to upsert package-id label
```
nan-yu, line 323: applyset-id labels

pkg/metadata/applyset.go
```diff
@@ -0,0 +1,56 @@
...
+	ApplySetToolingName = configsync.GroupName
+
+	// ApplySetToolingVersion is the version used to represent Config Sync in
+	// the ApplySet tooling annotation.
+	//
+	// The ApplySetKEP and kubectl require this to be a semantic version,
+	// implying that it should be the version of the tool. But we're using a
+	// static version instead, to allow listing all objects managed Config Sync,
+	// regardless of version.
+	ApplySetToolingVersion = "v1"
```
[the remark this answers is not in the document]
karlkfi: I considered that, and proposed it to Janet. She wanted to drop the version completely. But I eventually went with "v1" because the kubectl code expects a version, and that's the major version we use for the product and components. 

I put the reasoning in the PR description, but basically, it allows us to query for RSyncs managed by ConfigSync, regardless of minor or patch version. And it won't need changing if we ever decide to increment the resource group version.

e2e/testcases/managed_resources_test.go
```diff
@@ -203,11 +216,28 @@ metadata:
...
+	}
+	if err := os.WriteFile(path, nsBytes, 0644); err != nil {
+		return fmt.Errorf("failed to create a tmp file: %w", err)
+	}
+	return nil
+}
+
 // TestKubectlCreatesManagedConfigMapResource tests the drift correction regarding kubectl
 // tries to create a managed non-namespace resource in both the mono-repo mode and the multi-repo mode.
 func TestKubectlCreatesManagedConfigMapResource(t *testing.T) {
```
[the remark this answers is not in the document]
karlkfi: One is cluster-scoped (Namespace) and the other is namespace-scoped (ConfigMap). I renamed them for clarity.

But it might be nice in the future to change one to test with RepoSync, not just RootSync.

pkg/applier/applier_test.go
```diff
@@ -89,19 +89,20 @@ func TestApply(t *testing.T) {
...
 	})
 	abandonObj.SetLabels(map[string]string{
-		metadata.ManagedByKey:   metadata.ManagedByValue,
-		metadata.SystemLabel:    "anything",
-		metadata.ArchLabel:      "anything",
-		"example-to-not-delete": "anything",
+		metadata.ManagedByKey:        metadata.ManagedByValue,
+		metadata.SystemLabel:         "anything",
+		metadata.ArchLabel:           "anything",
+		metadata.ApplySetPartOfLabel: "anything", // not removed
```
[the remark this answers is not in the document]
karlkfi: added TODO with bug

e2e/testcases/managed_resources_test.go
```diff
@@ -22,33 +22,40 @@ import (
...
 )
 
 // This file includes tests for drift correction and drift prevention.
 //
-// The drift correction in the mono-repo mode utilizes the following two annotations:
-//  * configmanagement.gke.io/managed
-//  * configsync.gke.io/resource-id
+// The drift correction in the mono-repo mode utilizes the following metadata:
+//  * the configmanagement.gke.io/managed annotation
+//  * the configsync.gke.io/resource-id annotation
```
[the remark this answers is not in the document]
karlkfi: removed

pkg/reconcilermanager/controllers/reposync_controller.go
karlkfi: fixed.

e2e/testcases/managed_resources_test.go
```diff
@@ -133,18 +150,16 @@ metadata:
 	}
 
 	/* A new test */
```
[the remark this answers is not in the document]
karlkfi: This is a behavior change with the new filtering that I discussed in the design doc under caveats. 

Previously, if the remediator was watching the resource type, because it matched the resource of an object in the source, it would see when a new object was created with the `configsync.gke.io/manager` annotation matching the existing reconciler, then it would delete it, because it knows it doesn't manage it.

But with watch filtering, this only happens if the object ALSO has a matching `applyset.kubernetes.io/part-of` label.

Personally, I don't think users are ever going to care about or want this behavior, because it relies on too many coincidences. 

Like, if you make a ConfigMap and give it the manager annotation of an existing reconciler and there isn't already a ConfigMap in the source for that reconciler, then it wouldn't be watching ConfigMaps, so it wouldn't have deleted it. 

Realistically, this scenario is only going to happen if you export an existing ConfigMap, synced by a reconciler, and then gave it a new name and re-applied it. It could happen, but it's not clear that the user would really want or expect Config Sync to delete it. It might be better to just error. And if that is was happened, then it would also probably have the ApplySet label too.

I don't know if it really make sense to test all the permutations of these edge cases, unless someone runs into them and complains about the behavior as unexpected. Then we would at least have some signal that the behavior we're testing is actually desired. Otherwise, we're just writing regression tests for the implementation, not validating intended behavior.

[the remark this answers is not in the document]
nan-yu: Noted the behavior change. I think it would be good to have a test to demonstrate the new behavior. When a support ticket is filed, we can point to the test to explain the behavior.

A scenario I can think of is a R*Sync is deleted without deletion propagation before watch filtering is supported. All managed resources are kept without the applyset labels. Then a resource is removed from the source repo. 
Then Config Sync is upgraded with the watch filtering support and the RS*sync is re-created. It should be the user intention that the resource should be pruned.

Noted that this is kind of an edge case, but it is introduced by this PR. It would be god to have a test case to cover the new behavior.

e2e/testcases/managed_resources_test.go
```diff
@@ -22,45 +22,48 @@ import (
...
-//
-// The drift correction in the multi-repo mode utilizes the following three annotations:
-//  * configmanagement.gke.io/managed
-//  * configsync.gke.io/resource-id
-//  * configsync.gke.io/manager
+// The drift correction utilizes the following metadata:
+//  * the configmanagement.gke.io/managed annotation
+//  * the configsync.gke.io/resource-id annotation
+//  * the configsync.gke.io/manager annotation
+//  * the config.k8s.io/owning-inventory label
```
nan-yu, line 48: Doesn't it also require the applyset label for drift correction?

pkg/metadata/applyset.go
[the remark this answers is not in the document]
nan-yu: ACK

pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -269,11 +269,15 @@ func (r *RootSyncReconciler) upsertManagedObjects(ctx context.Context, reconcile
 }
 
 // setup performs the following steps:
+// - Patch RootSync to upsert package-id label
```
nan-yu, line 272: package-id label -> ApplySet metadata

e2e/testcases/managed_resources_test.go
```diff
@@ -69,19 +72,20 @@ func TestKubectlCreatesManagedNamespaceResourceMultiRepo(t *testing.T) {
...
+		metadata.ResourceIDKey:         "_namespace_test-ns1",
+		metadata.ResourceManagerKey:    rootSync1Manager,
+	})
+	ns1Obj.SetLabels(map[string]string{
+		metadata.ApplySetPartOfLabel: rootSync1ApplySetID,
+	})
 
-	if err := os.WriteFile(filepath.Join(nt.TmpDir, "test-ns1.yaml"), ns, 0644); err != nil {
-		nt.T.Fatalf("failed to create a tmp file %v", err)
+	if err := writeObjectYAMLFile(filepath.Join(nt.TmpDir, "test-ns1.yaml"), ns1Obj, nt.Scheme); err != nil {
```
sdowell, line 87: Is there a reason we need to write yaml to the filesystem instead of applying with the KubeClient?

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -711,3 +713,80 @@ func (r *reconcilerBase) isWebhookEnabled(ctx context.Context) (bool, error) {
...
+				"labelValueOld", currentApplySetID, "labelValue", applySetID)
+		}
+	}
+
+	currentApplySetToolingValue := core.GetAnnotation(rs, metadata.ApplySetToolingAnnotation)
+	if currentApplySetToolingValue != applySetToolingValue {
+		updateRequired = true
+
+		// If already set, validate that we can safely adopt the ApplySet
+		if currentApplySetToolingValue != "" {
```
sdowell, line 751: When would this happen? If we bump the `metadata.ApplySetToolingVersion`?

e2e/testcases/managed_resources_test.go
karlkfi: The test was explicitly applying with kubectl already, I think to simulate using another client more fully as an example.

I only did a quick change to make it easier to use the existing constants. 

But no, I don't think we really need to use kubectl. You wanna rewrite the test to simplify? Seemed out of scope for this PR.

pkg/reconcilermanager/controllers/reconciler_base.go
karlkfi: Yeah. I figured I'd add it so old versions wouldn't freak out if we decided to change it in the future and someone had to downgrade.

sdowell: I assume we would need to flesh out the user journey more in that case, right? I figure we don't expect the user to manually go in and mutate the old annotations

e2e/testcases/managed_resources_test.go
sdowell: Yeah that's fine, we can refactor in a follow-up

karlkfi: Looks like all the tests in this file use kubectl... Maybe that's better than using client-go directly because it more accurately reflects the type of change the user would want reverted, as opposed to one from a controller.

kubectl also might add annotations too, tho we don't test that they are preserved or reverted.

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: fixed

e2e/testcases/managed_resources_test.go
karlkfi: fixed

e2e/testcases/managed_resources_test.go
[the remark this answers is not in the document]
karlkfi: I added some tests to cover the changed behavior to show that objects won't be deleted if they dont have the right ApplySet ID in the part-of label value. 

I also added a new test at the bottom to validate that the remediator can catch the case where the part-of label is changed or removed.

pkg/reconcilermanager/controllers/reconciler_base.go
karlkfi: 1. If you upgrade and we've changed the version, it would auto-upgrade all the RSyncs to have the new version.
2. If you downgrade after we've changed the version, it would auto-downgrade all the RSyncs to have the new version.
3. We don't allow multiple versions of Config Sync installed at the same time, so we shouldn't have to deal with any fights or conflicts.
4. If we DID allow multiple versions concurrently, we'd need to rethink how this behavior works to make sure there was some way to delineate which reconciler-manager managed which RSyncs and how to manage upgrades/downgrades without losing access to the RSyncs we owned before. This would probably require a new label with a unique name for each install.

e2e/testcases/managed_resources_test.go
```diff
@@ -812,9 +918,64 @@ func TestDeleteManagedFieldsWithIgnoreMutationAnnotation(t *testing.T) {
...
+		testpredicates.HasLabel(metadata.ApplySetPartOfLabel, rootSync1ApplySetID),
+		testpredicates.HasAnnotation("season", "summer"),
+	}, testwatcher.WatchTimeout(30*time.Second)))
+
+	nt.T.Log("Removing the ApplySet ID label")
+	nt.Must(nt.KubeClient.MergePatch(namespace,
+		fmt.Sprintf(`{"metadata": {"labels": {%q: null}, "annotations": {"season": "winter"}}}`,
+			metadata.ApplySetPartOfLabel)))
+
+	time.Sleep(10 * time.Second)
```
sdowell, line 962: Is the sleep needed?

e2e/testcases/managed_resources_test.go
```diff
@@ -22,94 +22,95 @@ import (
...
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: test-ns1
-  annotations:
-    configmanagement.gke.io/managed: enabled
-    configsync.gke.io/resource-id: _namespace_test-ns1
-    configsync.gke.io/manager: :root
-`)
+	ns1Obj := &unstructured.Unstructured{}
```
sdowell, line 71: nit: can we use `k8sobjects` helpers for these instead of unstructured? It simplifies the setup a bit

sdowell (approved):
LGTM but I'll give @nan-yu a chance to leave final review

e2e/testcases/managed_resources_test.go
karlkfi: Probably. But SerializeObject will convert them back to Unstructured after. As long as it handles adding the GVKs from the scheme it should work though.

e2e/testcases/managed_resources_test.go
karlkfi: nope. removed.

e2e/testcases/managed_resources_test.go
karlkfi: done.

PR #1302, opened 2024-07-03
Update k8s to v1.30

karlkfi:
- k8s.io/kubernetes v1.30.1
- sigs.k8s.io/controller-runtime v0.18.4
- github.com/open-policy-agent/cert-controller v0.10.2+ 
  Using HEAD to pull in compatibility with controller-runtime v0.18.

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1308
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1314
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1315


karlkfi:
I don't really understand why, but go-licenses is now erroring with `Non go modules projects are no longer supported.` errors on a lot of go stdlib packages. AFAIK, that requirement was added in go-licenses v1.1.0. So I don't know what changed in this PR to trigger a new error from the same version of go-licenses.

go.mod
```diff
@@ -1,9 +1,9 @@
 module kpt.dev/configsync
 
-go 1.22
+go 1.22.4
```
sdowell, line 3: Is there a reason this pins to a patch version?

karlkfi: It happens automatically now with Go 1.22 when you run `go mod tidy`. If I remove it, it comes back. Saw it happen in k8s upstream too.

karlkfi: If I run `go mod tidy -go=1.22.0`, I can change it to  1.22.0, but If I try `go mod tidy -go=1.22` I get an error that `github.com/open-policy-agent/cert-controller@v0.10.2-0.20240531181455-2649f121ab97 requires go@1.22.0, but 1.22 is requested`

karlkfi: Seems like if I run `go mod tidy -go=1.22.0` and then `go mod tidy`, it keeps the older version. So maybe that's the right move, to set the minimum requirement as low as possible.

sdowell: Does this dictate which Go patch version is used when binaries are built? If so, we might want this to be the latest patch version. The golang builder image is currently set to `1.22.5`

karlkfi: AFAIK, the version in go.mod is just the minimum version requirement to build with.

karlkfi: > The minimum version of Go required by the current module.

https://go.dev/doc/modules/gomod-ref

karlkfi: For comparison, upstream k8s is also using 1.22.0: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L9

sdowell (approved):
/lgtm

PR #1304, opened 2024-07-03
Run make in buildenv container by default

karlkfi:
- Use "make" to run in docker
- Use "make all-local" to run locally

This should make it easy for us to all get the same result when running tests locally, even if we have subtly different versions of dependencies installed locally or have the repo checked out in different locations.


nan-yu (approved):
/lgtm

PR #1308, opened 2024-07-03
Replace lint-license.sh with make lint-license

karlkfi:
Calling go-licenses from inside the makefile makes it easier to ensure that we are installing and using the vendored version every time.

This also fixes an issue in Go 1.22 that was causing errors related to missing module info.


Makefile
```diff
@@ -355,8 +358,21 @@ lint-bash:
 	@./scripts/lint-bash.sh
 
 .PHONY: lint-license
-lint-license: buildenv-dirs
-	./scripts/lint-license.sh
```
nan-yu, line 359: `lint-license.sh` can be removed.

karlkfi: oops. thought i had done that. fixed.

nan-yu (approved):
/lgtm

PR #1309, opened 2024-07-03
ci: publish all infra images in postsubmit

sdowell:
Prior to this change, only the buildenv image was published on
postsubmit and required manually bumping the buildenv image tag. This
makes two key changes:
- Forms the image tag using the image inputs (dependencies), such that a
  new image will be published whenever one of the inputs changes. For
  example when the Go base image is updated, the postsubmit will publish
  a new image.
- Adds steps to publish both the gke-e2e and vulnerability-scanner
  images during postsubmit, using the same logic as buildenv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,8 @@ GO_DIR := $(OUTPUT_DIR)/go
 
 # Base image used for all golang containers
 # Uses trusted google-built golang image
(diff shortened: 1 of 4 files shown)


nan-yu (approved):
/lgtm

PR #1314, opened 2024-07-08
Add fake.Cache.RemoveInformer

karlkfi:
This new method is required to satisfy the Informers interface in new versions of controller-runtime. It allows removing and stopping an informer from a set of informers. The individual informers are stopped using informer-specific context wrappers with cancel funcs.

This blocks upgrading our controller-runitme dependency.

Ref: https://github.com/kubernetes-sigs/controller-runtime/pull/2285


sdowell (approved):
/lgtm

PR #1319, opened 2024-07-09
Update golangci-lint to v1.59.1

karlkfi:
Fix Fprintf and Fprintln errors not being handled. If we can't print to stderr, just panic.


pkg/applier/status.go
```diff
@@ -118,12 +126,14 @@ func (m ObjectStatusMap) Log(logger infofLogger) {
...
+func writeStatus(w io.Writer, status interface{ String() string }, ids []core.ID) error {
+	var err error
 	if len(ids) == 0 {
-		fmt.Fprintf(w, "%s (%d)", status, len(ids))
+		_, err = fmt.Fprintf(w, "%s (%d)", status, len(ids))
 	} else {
-		fmt.Fprintf(w, "%s (%d): [%s]", status, len(ids), joinIDs(commaSpaceDelimiter, ids...))
+		_, err = fmt.Fprintf(w, "%s (%d): [%s]", status, len(ids), joinIDs(commaSpaceDelimiter, ids...))
 	}
+	return err
```
nan-yu, line 136: nit: can we move the panic inside the function instead of returning an error and then panic outside?

karlkfi: done

pkg/applier/status.go
```diff
@@ -119,10 +119,14 @@ func (m ObjectStatusMap) Log(logger infofLogger) {
 }
 
 func writeStatus(w io.Writer, status interface{ String() string }, ids []core.ID) {
+	var err error
 	if len(ids) == 0 {
-		fmt.Fprintf(w, "%s (%d)", status, len(ids))
```
sdowell, line 123: Should this be using the same logger as the rest of the file?

pkg/util/env.go
```diff
@@ -59,7 +60,7 @@ func EnvInt(key string, def int) int {
 	if env := os.Getenv(key); env != "" {
 		val, err := strconv.ParseInt(env, 0, 0)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "WARNING: invalid env value (%v): using default, key=%s, val=%q, default=%d\n", err, key, env, def)
```
sdowell, line 62: Maybe klog instead?

pkg/applier/status.go
karlkfi: huh? It's writing to the string builder passed into it. the builder gets logged and reset multiple times.

pkg/util/env.go
karlkfi: wasn't clear to me if nomos is using klog or not

karlkfi: i think klog prints timestamps, and that isn't desired for the nomos CLI.

pkg/applier/status.go
sdowell: My mistake, I misread this as logging

sdowell: I think this should probably return an error instead of panic in that case

karlkfi: Writing to a string builder never fails. If it does, it's a programmer error and we'll want to catch it in testing.

```
func (b *Builder) Write(p []byte) (int, error) {
	b.copyCheck()
	b.buf = append(b.buf, p...)
	return len(p), nil
}
```

sdowell: That seems like an implementation detail that's not guaranteed by this function. This function expects an `io.Writer` interface, so it would be an easy mistake to make in the future if someone passed a different struct. We could change the argument to `*strings.Builder` to make that more clear

pkg/util/env.go
sdowell: This is used by container entrypoints as well. I think panicking may be acceptable for the CLI but seems like an undesirable behavior for backend components

pkg/applier/status.go
karlkfi: done

pkg/util/env.go
karlkfi: EntInt and EnvString are only called from package variable flag definitions, before the logger has been configured in `main()`. They can't use the logger, because some of the flags are used to configure the logger. AFAIK, there's no way to parse the flags in stages. Plus, if the service can't write to STDERR, it should probably crash anyway, because it was executed with a broken output stream.

sdowell: Ack

sdowell (approved):
/lgtm

karlkfi:

TestRoot_SourceReconcilerErrorsMetricValidation flake

karlkfi:
/retest

e2e tests seem to be failing with FailedScheduling of deployment pods... not clear why

PR #1321, opened 2024-07-09
test: bump kind and default k8s version

sdowell:
This bumps the kind version to latest and the default k8s version used for kind clusters.


nan-yu (approved):
/lgtm

PR #1322, opened 2024-07-10
feat: allow nomos migrate to continue on error

sdowell:
ACM operator will soon report an error when monorepo mode is enabled. This change allows nomos migrate to still operate for new versions of ACM operator.

Alternatively we could filter by specific error message, however currently the only errors set by ACM operator are related to the git config issues. This solution is simpler and less brittle on account of hard coded string comparison.


karlkfi (approved):
/lgtm

nan-yu (approved):
/lgtm

PR #1323, opened 2024-07-10
cherry-pick: e2e: attempt to uninstall packages in cleanup (#1231)

karlkfi:
If uninstall fails or times out, continue with existing more aggressive cleanup. This may take longer, but should give Config Sync a chance to clean up objects following reverse dependency ordering. Hopefully, this will help avoid race conditions and deadlocks.


nan-yu (approved):
/lgtm

PR #1326, opened 2024-07-10
ci: reduce number of presubmit kind clusters

sdowell:
There has been a noticable reducation in stability of the kind presubmits. This is an attempt to try to improve stability.


nan-yu (approved):
/lgtm

PR #1328, opened 2024-07-10
Remediate watch filter label removal

karlkfi:
This change handles future cases where watch filtering by label causes delete events when the required label is removed.

This blocks watch filtering: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1301

### DIFF
--- a/pkg/remediator/reconcile/worker_test.go
+++ b/pkg/remediator/reconcile/worker_test.go
@@ -44,84 +44,144 @@ import (
 // TestWorker_Run_Remediates verifies that worker.Run remediates declared
 // objects added to the queue.
 func TestWorker_Run_Remediates(t *testing.T) {
-	ctx := context.Background()
-
-	existingObjs := []client.Object{
-		fake.ClusterRoleBindingObject(syncertest.ManagementEnabled),
-		fake.ClusterRoleObject(syncertest.ManagementEnabled),
+	testCases := []struct {
+		name         string
+		existingObjs []client.Object
+		declaredObjs []client.Object
+		changedObjs  []client.Object
+		eventObjs    []client.Object
+		expectedObjs []client.Object
+	}{
+		{
+			name: "revert delete",
+			existingObjs: []client.Object{
+				fake.ClusterRoleBindingObject(syncertest.ManagementEnabled),
+			},
+			declaredObjs: []client.Object{
+				fake.ClusterRoleBindingObject(syncertest.ManagementEnabled),
+			},
+			changedObjs: []client.Object{
+				queue.MarkDeleted(context.Background(), fake.ClusterRoleBindingObject()),
+			},
+			eventObjs: []client.Object{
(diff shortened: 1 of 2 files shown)


pkg/remediator/reconcile/worker_test.go
```diff
@@ -44,84 +44,144 @@ import (
...
-	for _, obj := range changedObjs {
-		q.Add(obj)
-	}
-
-	// Give the worker a few seconds to remediate
-	// TODO: use client.Watch to watch for the desired changes (requires FakeClient to impl Watch).
-	time.Sleep(2 * time.Second)
-	cancel()
+			// Simulate watch events to add the objects to the queue
+			for _, obj := range tc.changedObjs {
```
nan-yu, line 164: changedObjs -> eventObjs

karlkfi: good catch! fixed.

pkg/remediator/reconcile/worker_test.go
```diff
@@ -44,84 +44,144 @@ import (
...
+			declaredObjs: []client.Object{
+				fake.ClusterRoleBindingObject(syncertest.ManagementEnabled,
+					core.Label("example-label", "example-value")),
+			},
+			changedObjs: []client.Object{
+				// Update object to remove label
+				fake.ClusterRoleBindingObject(syncertest.ManagementEnabled),
+			},
+			eventObjs: []client.Object{
+				// Watch server treats update to remove required label as a delete.
```
nan-yu, line 91: Do you have a link to the code? I'm curious about what labels are required.

karlkfi: When you use label selectors in the watch, it sends a delete event when the selected label is removed.

I asked apimachinery if this was a bug and they said it was working as intended. I haven't gone and tracked down the code for it in the apiserver. It's probably buried in the storage layer.

With our current watches, we don't have any label selectors, so this hasn't been an issues. But for https://github.com/GoogleContainerTools/kpt-config-sync/pull/1301 it's causing two of the conflict e2e tests to fail because they not correctly detecting the label removal, so there's no conflict error.

nan-yu (approved):
/lgtm

pkg/remediator/reconcile/worker_test.go
karlkfi: More explanation here: go/config-sync-watch-filter#heading=h.tpnna8sp25vg

karlkfi: I found a test for similar behavior upstream using a field selector: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/testing/watcher_tests.go#L103

karlkfi: Actually the "filtering" test immediately after that shows the same behavior for label selectors.

PR #1329, opened 2024-07-11
feat: filter contexts before creating rest configs

sdowell:
The rest config creation can be expensive since it performs an API call with UpdateQPS. This can take an especially long time if you have a large kubeconfig that contains stale configs (e.g. cluster deleted) since it will wait for API timeout.

### DIFF
--- a/pkg/client/restconfig/config.go
+++ b/pkg/client/restconfig/config.go
@@ -91,7 +91,7 @@ func currentContextNameFromConfig() (string, error) {
 // AllKubectlConfigs creates a config for every context available in the kubeconfig. The configs are
 // mapped by context name. There is no way to detect unhealthy clusters specified by a context, so
 // timeout can be used to prevent calls to those clusters from hanging for long periods of time.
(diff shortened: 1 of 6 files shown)


pkg/client/restconfig/config.go
```diff
@@ -135,3 +139,17 @@ func AllKubectlConfigs(timeout time.Duration) (map[string]*rest.Config, error) {
 	}
 	return configs, nil
 }
+
+func filterContexts(wantContexts []string, contexts map[string]*clientcmdapi.Context) map[string]*clientcmdapi.Context {
+	filteredContexts := make(map[string]*clientcmdapi.Context)
+	if wantContexts == nil {
+		return contexts
+	} else {
```
nan-yu, line 147: nit: the nested else is not necessary.

nan-yu (commented):
one minor comment, otherwise lgtm

pkg/client/restconfig/config.go
sdowell: The linter agrees 🥲  fixed

nan-yu (approved):
/lgtm

PR #1330, opened 2024-07-11
feat: add --version flag to nomos CLI

sdowell:
There currently is not a clean way to programmatically capture the version of the nomos CLI. The nomos version command also fetches the config sync version installed on the cluster from the kube context.

This provides an easy way to simply print the version of the CLI.


nan-yu (approved):
/lgtm

PR #1331, opened 2024-07-11
ci: fix go lint deprecated field

sdowell:
The staticcheck go field is deprecated. It should be set in the run.go field, however this field defaults to the version in go.mod. So no need to define the go version in this file.


karlkfi (approved):
/lgtm

PR #1334, opened 2024-07-12
Permitting Empty Responses for Server Resource Discovery

tiffanny29631:
Custom metric servers often return an empty response to metric queries as a means to conserve resources. This update logs a warning when such responses are received and then proceeds without further action, aligning with the upstream workaround.

Currently, it is not feasible to develop a valid end-to-end test because the cluster version impacts the reproducibility of the issue, which does not occur on clusters version 1.28 and above. Therefore, only unit tests will be added.

### DIFF
--- a/pkg/util/discovery/server_resourcer.go
+++ b/pkg/util/discovery/server_resourcer.go
@@ -16,6 +16,7 @@ package discovery
 
 import (
 	"errors"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -64,6 +65,11 @@ func GetResources(discoveryClient ServerResourcer) ([]*metav1.APIResourceList, s
 			for gv, subErr := range discoErr.Groups {
 				if apierrors.IsNotFound(subErr) {
 					klog.Warningf("Failed to discover APIGroups %s due to not found, moving on: %v", gv, subErr)
+				} else if strings.Contains(subErr.Error(), "received empty response") {
+					// Allow "empty response" as common responses for metrics following
+					// upstream workaround https://github.com/kubernetes/kubernetes/pull/115978
+					// TODO b/353596666 change to type match when emptyResponseError in upstream is made public
+					klog.Warningf("Failed to discover resources for API group %s: assuming zero resources: %v", gv, subErr)
 				} else {
 					// return all the discovery errors if any one of them is a blocking error
 					return nil, status.APIServerError(discoveryErr, "API discovery failed")
--- a/pkg/util/discovery/server_resourcer_test.go
+++ b/pkg/util/discovery/server_resourcer_test.go
@@ -1,0 +1,97 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package discovery
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+)
+
+// Test mock of ServerResourcer
+type CustomServerResourcer struct {
+	ExpectedResources []*v1.APIResourceList
+	ExpectedError     error
+	Called            bool
+}
+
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -139,6 +139,7 @@ require (
 	github.com/spf13/afero v1.6.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/src-d/gcfg v1.4.0 // indirect
+	github.com/stretchr/objx v0.5.2 // indirect
```
karlkfi, line 142: why didn't go.sum change when this was added?

pkg/api/configsync/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -5,7 +5,7 @@
 package v1alpha1
 
 import (
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
karlkfi, line 8: This doesn't seem related to this change. Is this just a change with Go 1.22? I haven't seen this locally.

pkg/util/discovery/server_resourcer.go
```diff
@@ -64,6 +65,11 @@ func GetResources(discoveryClient ServerResourcer) ([]*metav1.APIResourceList, s
 			for gv, subErr := range discoErr.Groups {
 				if apierrors.IsNotFound(subErr) {
 					klog.Warningf("Failed to discover APIGroups %s due to not found, moving on: %v", gv, subErr)
+				} else if strings.Contains(subErr.Error(), "received empty response") {
+					// Allow "empty response" as common responses for metrics
+					// Upstream workaround https://github.com/kubernetes/kubernetes/pull/115978
```
karlkfi, line 70: Can you make an upstream PR to make `emptyResponseError` public, so we can catch it more explicitly in the future, then add a TODO here?

pkg/util/discovery/server_resourcer.go
```diff
@@ -64,6 +65,11 @@ func GetResources(discoveryClient ServerResourcer) ([]*metav1.APIResourceList, s
...
+					// TODO b/349378765 change to type match when emptyResponseError in upstream is made public
+					klog.Warningf("Failed to discover APIGroups %s due to empty response received, moving on: %v", gv, subErr)
```
karlkfi, line 72: I think you can drop the "due to empty response received" here, since you're including the text of the error. How about `Failed to discover resources for API group %s: assuming zero resources: %v`

pkg/util/discovery/server_resourcer_test.go
```diff
@@ -0,0 +1,96 @@
...
+// limitations under the License.
+
+package discovery
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
```
karlkfi, line 22: I know we use testify elsewhere, but I think this is our first use of mock.Mock. Everywhere else we're mostly using custom fakes. I don't want to make this more work, but I also don't really want to start a precedent of using mocks without discussing it with the team first. Using Mocks than get out of hand pretty quick. That said, it does look a little simpler than writing a fake.

go.mod
tiffanny29631: It seems to be already included in go.sum when updating to K8s 1.30, but no worries it's removed when not using mock.Mock

pkg/api/configsync/v1alpha1/zz_generated.deepcopy.go
tiffanny29631: Reverted

pkg/util/discovery/server_resourcer.go
tiffanny29631: Made https://github.com/kubernetes/kubernetes/pull/126137, also created http://b/353596666 as TODO.

pkg/util/discovery/server_resourcer.go
tiffanny29631: Done

pkg/util/discovery/server_resourcer_test.go
tiffanny29631: Changed to a custom mock in the latest patch.

sdowell:
> Currently, it is not feasible to develop a valid end-to-end test because the cluster version impacts the reproducibility of the issue, which does not occur on clusters version 1.28 and above. Therefore, only unit tests will be added.

Can we write a test that only runs on certain k8s versions where the issue is reproducible? Conversely, do we need this if the issue isn't reproducible on any tested/supported k8s versions?

tiffanny29631:
> Can we write a test that only runs on certain k8s versions where the issue is reproducible? Conversely, do we need this if the issue isn't reproducible on any tested/supported k8s versions?

Just tried again this morning. Master version that can recreate the error (1.27.11) is no longer available, node version still exists, we would specify this cluster config but it might not worth that much. I'm fine of aborting the change since Config Sync has updated k8s library and the k8s 1.28+ also has aggregated discovery, meaning this error is likely no longer an issue.

tiffanny29631:
https://github.com/kubernetes/kubernetes/pull/115978 Should be fixing the issue from memory cached discovery client, as Config Sync has upgraded to over K8s 1.28 the discovery client should no longer receive empty response error for custom metric servers which might not have implemented the resource list.

For cx who are seeing the issue, upgrade Config Sync to over 1.18.3 and upgrade cluster to over 1.28 should fix the issue.

PR #1335, opened 2024-07-15
Fix remediator to catch conflicts during apply

karlkfi:
This fix starts new watches before applying, to ensure that conflicts on existing objects are caught by the remediator immediately after the applier mutates an object and before the whole apply set has succeeded and reconciled.

This is only a partial fix. It doesn't work for custom resources which are applid with their CRD in the same ApplySet. To resolve that case, we will need to asynchronously start watches when the resource is available, which we can add later.

### DIFF
--- a/pkg/parse/updater.go
+++ b/pkg/parse/updater.go
@@ -120,6 +120,13 @@ func (u *Updater) update(ctx context.Context, cache *cacheForCommit) status.Mult
 		if err != nil {
 			return err
 		}
+		// Add new resources to the watch list, without removing old ones.
+		// This ensures controller conflicts are caught while the applier is running.
+		declaredGVKs, _ := u.Resources.DeclaredGVKs()
+		err = u.addWatches(ctx, declaredGVKs)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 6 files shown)


pkg/parse/updater.go
```diff
@@ -213,9 +220,23 @@ func (u *Updater) apply(ctx context.Context, objs []client.Object, commit string
 	return nil
 }
 
-// watch updates the Remediator's watches to start new ones and stop old
-// ones.
-func (u *Updater) watch(ctx context.Context, gvks map[schema.GroupVersionKind]struct{}) status.MultiError {
+// addWatches tells the Remediator to watch additional resources without
+// stopping any.
+func (u *Updater) addWatches(ctx context.Context, gvks map[schema.GroupVersionKind]struct{}) status.MultiError {
+	klog.V(1).Info("Remediator watches addeding...")
```
sdowell, line 226: `adding`

karlkfi: fixed

nan-yu (approved):
/lgtm

PR #1336, opened 2024-07-15
ci: sideload dockerhub images into kind

sdowell:
When running many parallel kind clusters, each kind cluster pulls these public dockerhub images independently. This can result in 429 too many request errors in our CI. Loading these images a single time on the host and then sideloading them should minimize the number of pull attempts.


e2e/nomostest/clusters/kind.go
```diff
@@ -202,3 +204,48 @@ apiServer:
...
+	testing.RegistryImage,
+	testing.NginxImage,
+}
+
+// Pull each public image a single time on the host, and then sideload the images
+// into each kind cluster. This reduces the number of times each image is pulled
+// to avoid 429 too many requests errors from public docker registries.
+// Alternatively we may choose to host these images in a registry we control.
+func sideLoadImages(clusterName string) error {
+	if err := pullImages(); err != nil {
```
tiffanny29631, line 219: How about extract pullImages to before creating kind cluster?

sdowell: I don't think it makes much difference if we pull before vs after the kind cluster is created. I changed it to pull in parallel to kind cluster creation as that could be an optimization.

Makefile
```diff
@@ -442,7 +442,7 @@ clean-helm:
 	@rm -rf $(HELM)
 
 "$(GOBIN)/kind":
-	go install sigs.k8s.io/kind@$(KIND_VERSION)
+	CGO_ENABLED=0 go install sigs.k8s.io/kind@$(KIND_VERSION)
```
karlkfi, line 445: TIL CGO_ENABLED=0 is only default on OSes without a C toolchain installed.

What error does this fix here?

e2e/nomostest/clusters/kind.go
```diff
@@ -202,3 +210,45 @@ apiServer:
 	// We failed to create the cluster maxKindTries times, so fail out.
 	return err
 }
+
+var preloadImages = []string{
+	testing.HTTPDImage,
+	testing.RegistryImage,
+	testing.NginxImage,
+	"prom/prometheus:v2.37.6", // this is defined in a prometheus testdata yaml file
```
karlkfi, line 218: Seems like kind of a gotcha to hard code this. Can it be parsed out of the testdata file?

Makefile
sdowell: Without this the `kind` binary cannot be invoked inside the `kind-e2e` container: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/1336/kpt-config-sync-presubmit-e2e-multi-repo-test-group1/1812970304033525760

e2e/nomostest/clusters/kind.go
karlkfi: I wonder if all of them could be parsed out? Or are the others templated?

e2e/nomostest/clusters/kind.go
```diff
@@ -94,13 +96,19 @@ func (c *KindCluster) Create() error {
...
-	} else {
+	tg := taskgroup.New()
+	tg.Go(func() error {
+		return createKindCluster(c.provider, c.Name, c.KubeConfigPath, version)
+	})
+	tg.Go(func() error {
+		return pullImages()
+	})
+	if err := tg.Wait(); err != nil {
 		c.creationSuccessful = false
```
karlkfi, line 107: doesn't it default to false? why do we need to set it?

e2e/nomostest/clusters/kind.go
```diff
@@ -202,3 +210,45 @@ apiServer:
 	// We failed to create the cluster maxKindTries times, so fail out.
 	return err
 }
+
+var preloadImages = []string{
```
karlkfi, line 214: nit: I think this could be hard-coded in the Create() func or generated dynamically from testdata with a findDockerhubImages() func or something and then passed as an array to the pullImages and sideLoadImages functions, just for a bit of readability/flexibility. But it's not a big deal.

Makefile
karlkfi: Can you add a comment. Something like "Build kind with CGO disabled so it can be used in containers without C installed.

sdowell: done

e2e/nomostest/clusters/kind.go
sdowell: The other server components are defined as Go types rather than yaml manifests, which makes them a bit easier to work with in code - especially for static constants like this. Refactored prometheus installation to use Go types.

e2e/nomostest/clusters/kind.go
sdowell: I'd rather this be an explicit singleton shared across threads, since each thread expects the same static images to be sideloaded.

e2e/nomostest/clusters/kind.go
sdowell: removed

karlkfi (approved):
/lgtm

PR #1337, opened 2024-07-16
ci: sideload dockerhub images into kind (#1336)

tiffanny29631:
When running many parallel kind clusters, each kind cluster pulls these public dockerhub images independently. This can result in 429 too many request errors in our CI. Loading these images a single time on the host and then sideloading them should minimize the number of pull attempts.

Cherrypick for b/353389403


sdowell (approved):
/lgtm

PR #1342, opened 2024-07-18
fix: remediator management conflict detection

karlkfi:
- Change the remediator worker to detect management conflicts.
  This fixes a bug when using watch filtering where conflicts
  may be ignored for objects that were "deleted" by having the
  required label removed after being successfully applied by
  the applier but before being remediated.
- Change the remediator worker to clear management conflicts.
  This fixes a bug where the conflict error might be prematurely
  removed from the status and then not actually resolved,
  causing the error to flap.

Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1354

### DIFF
--- a/pkg/remediator/conflict/handler.go
+++ b/pkg/remediator/conflict/handler.go
@@ -37,6 +37,8 @@ type Handler interface {
 	ConflictErrors() []status.ManagementConflictError
 	// HasConflictErrors returns true when there are conflict errors
 	HasConflictErrors() bool
+	// HasConflictError returns true when there is a conflict for the specified object ID.
+	HasConflictError(core.ID) bool
 }
 
 // handler implements Handler.
@@ -73,6 +75,15 @@ func (h *handler) AddConflictError(id core.ID, newErr status.ManagementConflictE
 	}
 
 	h.conflictErrs.Set(id, newErr)
+}
+
+// HasConflictError returns true when there is a conflict for the specified object ID.
+func (h *handler) HasConflictError(id core.ID) bool {
+	h.mux.RLock()
+	defer h.mux.RUnlock()
+
+	_, found := h.conflictErrs.Get(id)
+	return found
 }
 
 func (h *handler) RemoveConflictError(id core.ID) {
--- a/pkg/remediator/reconcile/reconciler.go
+++ b/pkg/remediator/reconcile/reconciler.go
@@ -25,6 +25,7 @@ import (
 	"kpt.dev/configsync/pkg/importer/analyzer/validation/nonhierarchical"
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/metrics"
+	"kpt.dev/configsync/pkg/remediator/conflict"
 	"kpt.dev/configsync/pkg/status"
 	syncerclient "kpt.dev/configsync/pkg/syncer/client"
 	syncerreconcile "kpt.dev/configsync/pkg/syncer/reconcile"
@@ -47,7 +48,8 @@ type reconciler struct {
 	// declared is the threadsafe in-memory representation of declared configuration.
 	declared *declared.Resources
 
-	fightHandler fight.Handler
+	conflictHandler conflict.Handler
+	fightHandler    fight.Handler
 }
 
 // newReconciler instantiates a new reconciler.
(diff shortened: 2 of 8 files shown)


nan-yu:
> Prior to this change, namespace remediators did not report management conflict errors. Instead, they set a flag to tell the run loop to retry. 

Namespace remediators didn't report management conflict errors because the corresponding namespace applier would report the conflict error. Namespace applier uses the [PolicyAdoptIfNoInventory](https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.18.3-rc.4/pkg/applier/applier.go#L168). If management conflict occurs, the applier would never succeed and report an error. So no need for the remediator to report a duplicate of the same error.

> This had multiple issues:
> - The run loop flag to retry doesn't actually trigger a retry. It just ensures that the next retry will attempt an apply.

If a retry was not triggered, then the status would still include the management conflict error, which doesn't seem to be a problem.

> - If the retry backoff had expired (max 12 attempts) or the last attempt was successful, the conflict was not reported until the next full resync or source change.

If the retry backoff had expired, the run loop would not be executed, so no update to the status and conflict should still be reported.

If the last attempt was successful, that means the conflict should have been fixed (otherwise, it would NOT be successful due to the AdoptIfNoInventory policy). It is correct to NOT report conflict any more.

> - The applier conflict error only knows one manager, while the remediator conflict error knows both. So for namespace remediators, the error with less information was always reported.

The applier used to know the other manager. https://github.com/GoogleContainerTools/kpt-config-sync/pull/1344 changed the behavior. `Who detected the conflict is irrelevent to the user` made it less informative to the applier.

karlkfi:
@nan-yu 

> Namespace remediators didn't report management conflict errors because the corresponding namespace applier would report the conflict error. Namespace applier uses the [PolicyAdoptIfNoInventory](https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.18.3-rc.4/pkg/applier/applier.go#L168). If management conflict occurs, the applier would never succeed and report an error. So no need for the remediator to report a duplicate of the same error.

The applier error is inferior to the remediator error, because it only includes the current manager, not the remote manager. So by delegating to the applier, only the inferior error was being reported.

> If a retry was not triggered, then the status would still include the management conflict error, which doesn't seem to be a problem.

This is only true for the root reconciler. For the namespace reconciler, the remediator was setting the flag and exiting early, without recording the status error to the conflict handler or recording the conflict metric. So for namespace reconciler, if it doesn't retry, because it applied successfully prior to the conflict, then it wouldn't report the conflict until the next resync or source change triggered the applier.

Plus, the applier has a different bug: it reports the management conflict status error but doesn't record the conflict metric.

> If the retry backoff had expired, the run loop would not be executed, so no update to the status and conflict should still be reported.

Only in the root reconciler. As mentioned, in the namespace reconciler it was exiting early without reporting the conflict.

> The applier used to know the other manager. https://github.com/GoogleContainerTools/kpt-config-sync/pull/1344 changed the behavior. `Who detected the conflict is irrelevent to the user` made it less informative to the applier.

This is incorrect. Both before and after that PR, the local and remote manager are included in the conflict error from the remediator. Only the applier error does not have the remote manager, because the kpt applier returns the desired object instead of the current object in the error.

- remediator error: `Sprintf("The %q reconciler detected a management conflict with the %q reconciler. Remove the object from one of the sources of truth so that the object is only managed by one reconciler.", newManager, currentManager)`
- applier error: `Sprintf("The %q reconciler detected a management conflict with another reconciler. Remove the object from one of the sources of truth so that the object is only managed by one reconciler.", newManager).`

Prior to this PR, the remediator error was only ever being surfaced by root reconcilers, and the remediator error would replace the applier error (de-duped by the conflict handler).

After this PR, the namespace reconciler now behaves more similarly to the root reconciler.

karlkfi:
FWIW, `Who detected the conflict is irrelevent to the user.` in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1344 was in reference to the ordering of the desired/new and current/old manager in the error message from the remediator. Prior to that PR, the error message was being reversed when writing directly to a remote RootSync status. But after that PR, the message now stays in the initial order, so the manager that detected the error is always first. This fast is irrelevant to the user, so we don't need to hide it by reversing the order. On top of that, the existing remote status update behavior was subtlely broken, which we discovered in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1345. Simplifying the message made it easier to fix and test, because you can tell who reported it and check that it's de-duped correctly.

karlkfi:
I extracted a subset of these changes to a smaller PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1354

nan-yu:
Discussed offline that the old applier didn't carry the remote manager info, only the inverted conflict error had, but with a wrong value. With this change, both the applier and remediator will report conflict error, and the conflict handler will de-dup them.

nan-yu (approved):
/lgtm

PR #1343, opened 2024-07-18
e2e: watch Role during RootToRoot conflict test

karlkfi:
Add a background watch to TestConflictingDefinitions_RootToRoot to make it faster to detect conflict and easier to debug when it's not working, with diffs logged for the Role as well as both RootSyncs.

### DIFF
--- a/e2e/nomostest/testpredicates/predicates.go
+++ b/e2e/nomostest/testpredicates/predicates.go
@@ -633,6 +633,42 @@ func IsManagedBy(scheme *runtime.Scheme, scope declared.Scope, syncName string) 
 	}
 }
 
+// HasBeenManagedBy validates that an object has been managed by all of the
+// specified managers, one at a time.
+func HasBeenManagedBy(scheme *runtime.Scheme, logger *testlogger.TestLogger, managers ...string) Predicate {
+	managersSeen := make(map[string]struct{}, len(managers))
+	return func(obj client.Object) error {
+		if obj == nil {
+			return ErrObjectNotFound
+		}
+		// Make sure GVK is populated (it's usually not for typed structs).
+		gvk := obj.GetObjectKind().GroupVersionKind()
+		if gvk.Empty() {
+			var err error
+			gvk, err = kinds.Lookup(obj, scheme)
+			if err != nil {
+				return err
+			}
+			obj.GetObjectKind().SetGroupVersionKind(gvk)
+		}
+		manager := core.GetAnnotation(obj, metadata.ResourceManagerKey)
+		managersSeen[manager] = struct{}{}
+		logger.Infof("%s %s update: ResourceManager: %s, ResourceVersion: %s",
+			gvk.Kind, core.ObjectNamespacedName(obj), manager, obj.GetResourceVersion())
+		var unseenManagers []string
+		for _, manager := range managers {
+			if _, found := managersSeen[manager]; !found {
+				unseenManagers = append(unseenManagers, manager)
+			}
+		}
+		if len(unseenManagers) > 0 {
+			return fmt.Errorf("waiting for %s %s to be managed by: %s",
+				gvk.Kind, core.ObjectNamespacedName(obj), strings.Join(unseenManagers, ","))
+		}
+		return nil
+	}
+}
+
 // IsNotManaged checks that the object is NOT managed by configsync.
 // Use differ.ManagedByConfigSync if you just need a boolean without errors.
 func IsNotManaged(scheme *runtime.Scheme) Predicate {
--- a/e2e/testcases/multi_sync_test.go
+++ b/e2e/testcases/multi_sync_test.go
@@ -574,15 +574,6 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
 	nt.T.Logf("Stop the admission webhook, to confirm that the remediator still reports the conflicts")
 	nomostest.StopWebhook(nt)
 
-	nt.T.Logf("The Role resource version should be changed because two reconcilers are fighting with each other")
-	err = nt.Watcher.WatchObject(kinds.Role(), "pods", testNs,
-		[]testpredicates.Predicate{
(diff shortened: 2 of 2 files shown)


e2e/testcases/multi_sync_test.go
```diff
@@ -574,14 +577,35 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
...
-	}
+	// Watch Role until the other conditions are met.
+	// Collect data from role updates for validation later
+	// This will log diffs for easier debugging.
+	var roleUpdates []roleUpdate
+	nt.T.Log("Starting Role Watcher")
+	bgCtx, bgCancel := context.WithCancel(nt.Context)
+	defer bgCancel()
+	bgDone := make(chan struct{})
+	go func() {
```
sdowell, line 588: could this be part of the taskgroup below?

karlkfi: The tasks group tasks exit when the predicates are satisfied. This one stays running until cancelled, after the task group is done.

I tried putting this one in the task group, but it requires either two channels to close or an atomic counter to figure out when the other two are done, which felt messy. I don't really know if this is better, but it's at least obvious that it's different.

karlkfi: Oh, the other reason it can't be in the TaskGroup is that the Role watcher needs to be _interrupted_ when the TaskGroup is done, because there may not be another Role update to trigger the predicate evaluation.

e2e/testcases/multi_sync_test.go
```diff
@@ -574,14 +577,34 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
...
+							ResourceVersion: o.GetResourceVersion(),
+							Manager:         core.GetAnnotation(o, metadata.ResourceManagerKey),
+						}
+						roleUpdates = append(roleUpdates, ru)
+						nt.T.Logf("Role update: %#v", ru)
+					}
+					return fmt.Errorf("waiting for other conditions")
+				},
+			}, testwatcher.WatchTimeout(0), testwatcher.WatchContext(bgCtx))
+		if err != nil && !errors.Is(err, context.Canceled) {
```
sdowell, line 604: nit: err should always be != nil. I think it would be simpler/clearer to just write

```
if !errors.Is(err, context.Canceled) {
```

e2e/testcases/multi_sync_test.go
sdowell: Ack, makes sense

e2e/testcases/multi_sync_test.go
```diff
@@ -609,6 +632,16 @@ func TestConflictingDefinitions_RootToRoot(t *testing.T) {
...
 	}
 
+	nt.T.Log("Stopping Role Watcher")
+	// Stop the background watcher
+	bgCancel()
+	// Wait for background watcher to stop
+	<-bgDone
+
+	if !reconcilerConflictObserved(roleUpdates) {
+		nt.T.Fatalf("Role updates do not indicate a management conflict between reconcilers: %+v", roleUpdates)
```
sdowell, line 642: From the latest test run:
```
multi_sync_test.go:642: 2024-07-18 00:45:28.271563589 +0000 UTC ERROR: Role updates do not indicate a management conflict between reconcilers: [{ResourceVersion:2271 Manager::root} {ResourceVersion:2271 Manager::root}]
```
Seems to indicate there might be an issue with this approach

karlkfi: Yeah, This worked with all the other changes from my watch filter PR but isn't working by itself. That means the test as-written doesn't actually do what I thought it did.

I think the problem is that the TaskGroup predicates are being satisfied after just one of the root reconcilers detects the conflict and reports it to both RootSyncs. It doesn't actually wait for both reconcilers to apply the Role and flip it back and forth.

So I think I'm going to have to change the approach here to just use a task that watches until both managers have been seen on the Role.

karlkfi: Alright, I rewrote this to just use a predicate that watches for both managers.

Since the other predicates are satisfied after only one root reconciler sees the conflict, this will wait for both to have actually updated the Role.

Note: When the Role is applied by a reconciler, it will succeed and remove the error from the status.  So the status will still flap back and forth... That's no really ideal, but that's how it works today for RootSync conflicts.

e2e/testcases/multi_sync_test.go
karlkfi: obsolete.

e2e/nomostest/testpredicates/predicates.go
```diff
@@ -633,6 +633,38 @@ func IsManagedBy(scheme *runtime.Scheme, scope declared.Scope, syncName string)
...
+			}
+			obj.GetObjectKind().SetGroupVersionKind(gvk)
+		}
+		manager := core.GetAnnotation(obj, metadata.ResourceManagerKey)
+		managersSeen[manager] = struct{}{}
+		logger.Infof("%s %s update: ResourceManager: %s, ResourceVersion: %s",
+			gvk.Kind, core.ObjectNamespacedName(obj), manager, obj.GetResourceVersion())
+		for _, manager := range managers {
+			if _, found := managersSeen[manager]; !found {
+				return fmt.Errorf("waiting for %s %s to be managed by %q",
```
sdowell, line 660: nit: might be nice to return all of the managers which haven't been found yet instead of just the first one

sdowell: Perhaps rewriting this like something like:
```
wantManagers := make(map[string]struct{}, len(managers))
for _, manager := range managers {
  wantManagers[manager] = struct{}{}
}
return func(obj client.Object) error {
  ...
  delete(wantManagers, manager)
  if len(wantManagers) > 0 {
    return fmt.Errorf(...)
  }
  return nil
```

karlkfi: done

sdowell (approved):
/lgtm

PR #1344, opened 2024-07-18
Simplify management conflict error

karlkfi:
Use the same message for conflicts detected by the current reconciler and a remote reconciler. Who detected the conflict is irrelevent to the user.

Clarify the error message to make sense for non-git sources of truth.

Depends on: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1345

### DIFF
--- a/pkg/parse/root.go
+++ b/pkg/parse/root.go
@@ -729,20 +729,20 @@ func prependRootSyncRemediatorStatus(ctx context.Context, c client.Client, syncN
 		return status.APIServerError(err, "failed to get RootSync: "+syncName)
 	}
 
+	// Skip updating if already reported by this reconciler or the remote reconciler (inverted)
 	var errs []v1beta1.ConfigSyncError
 	for _, conflictErr := range conflictErrs {
-		conflictCSEError := conflictErr.ToCSE()
-		conflictPairCSEError := conflictErr.CurrentManagerError().ToCSE()
+		cse := conflictErr.ToCSE()
+		invertedCSE := conflictErr.Invert().ToCSE()
 		errorFound := false
 		for _, e := range rs.Status.Sync.Errors {
-			// Dedup the same remediator conflict error.
-			if e.Code == status.ManagementConflictErrorCode && (e.ErrorMessage == conflictCSEError.ErrorMessage || e.ErrorMessage == conflictPairCSEError.ErrorMessage) {
+			if e.Code == status.ManagementConflictErrorCode && (e.ErrorMessage == cse.ErrorMessage || e.ErrorMessage == invertedCSE.ErrorMessage) {
 				errorFound = true
 				break
 			}
 		}
 		if !errorFound {
-			errs = append(errs, conflictCSEError)
+			errs = append(errs, cse)
 		}
 	}
 
--- a/pkg/parse/root_test.go
+++ b/pkg/parse/root_test.go
@@ -1793,17 +1793,17 @@ func TestPrependRootSyncRemediatorStatus(t *testing.T) {
 	const rootSyncName = "my-root-sync"
 	const thisManager = "this-manager"
 	const otherManager = "other-manager"
-	conflictingObject := fake.NamespaceObject("foo-ns",
-		core.Annotation(metadata.ResourceManagerKey, otherManager))
-	conflictError := status.ManagementConflictErrorWrap(conflictingObject, thisManager)
-	invertedObject := fake.NamespaceObject("foo-ns",
-		core.Annotation(metadata.ResourceManagerKey, thisManager))
-	invertedConflictError := status.ManagementConflictErrorWrap(invertedObject, otherManager)
-	kptManagementConflictError := applier.KptManagementConflictError(conflictingObject)
-	currentManagerError := conflictError.CurrentManagerError()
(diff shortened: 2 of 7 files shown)


pkg/parse/root.go
```diff
@@ -732,7 +732,7 @@ func prependRootSyncRemediatorStatus(ctx context.Context, c client.Client, syncN
 	var errs []v1beta1.ConfigSyncError
 	for _, conflictErr := range conflictErrs {
 		conflictCSEError := conflictErr.ToCSE()
-		conflictPairCSEError := conflictErr.CurrentManagerError().ToCSE()
+		conflictPairCSEError := conflictErr.ToCSE()
 		errorFound := false
 		for _, e := range rs.Status.Sync.Errors {
 			// Dedup the same remediator conflict error.
```
sdowell, line 738: It looks like the `if` statement here should change since the error message is now always the same

karlkfi: The format is the same, but the inputs are different. The message includes both the old and new manager, which will swap places depending on which reconciler detected the conflict.

sdowell: If that's the intent then it doesn't look like that's what the code here is doing. Both error messages it's comparing in the OR are the same `conflictErr.ToCSE()`

karlkfi: Hmm. You're right. Before this PR, you could inverted the error and simulate an error written by another root reconciler, but now you can't. So this isn't actually de-duping those any more, even if it does now de-dupe the applier and remediator errors.

So I guess we still need to be able to invert the message for deduping.

karlkfi: Fixed.

pkg/status/management_conflict_error.go
```diff
@@ -99,6 +79,13 @@ func (m *managementConflictErrorImpl) Errors() []Error {
 	return []Error{m}
 }
 
+func (m *managementConflictErrorImpl) Invert() ManagementConflictError {
```
sdowell, line 82: Clean abstraction 👍

pkg/parse/root.go
```diff
@@ -729,20 +729,20 @@ func prependRootSyncRemediatorStatus(ctx context.Context, c client.Client, syncN
...
 	for _, conflictErr := range conflictErrs {
-		conflictCSEError := conflictErr.ToCSE()
-		conflictPairCSEError := conflictErr.CurrentManagerError().ToCSE()
+		cse := conflictErr.ToCSE()
+		invertedCSE := conflictErr.Invert().ToCSE()
 		errorFound := false
 		for _, e := range rs.Status.Sync.Errors {
-			// Dedup the same remediator conflict error.
-			if e.Code == status.ManagementConflictErrorCode && (e.ErrorMessage == conflictCSEError.ErrorMessage || e.ErrorMessage == conflictPairCSEError.ErrorMessage) {
+			if e.Code == status.ManagementConflictErrorCode && (e.ErrorMessage == cse.ErrorMessage || e.ErrorMessage == invertedCSE.ErrorMessage) {
```
sdowell, line 739: Are we lacking test coverage for this? I'm surprised the tests were passing with the previous bugged implementation

karlkfi: TestConflictingDefinitions_RootToRoot is the only test covering remote conflict status reporting. The way those task group predicates are written it passes whether the conflict was reported by the local or remote reconciler. That's what was causing the test to fail in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1343

Also, those predicates pass regardless of how many errors are reported. So it doesn't strictly require de-duping either.

karlkfi: I considered changing the predicates to require exactly one conflict error, but that won't work either, because they will see exactly one error before they see two.

I'm not really sure how to prove the negative that it only ever sees one.

Wanna backfill some unit tests for me?

sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1345

karlkfi: Thanks for the tests!
They caught a subtle bug in my Invert method, and I figured out why the message for KptManagementConflictError was weird (left a comment).
All but one of the unexpected test behaviors have been fixed now. The other I will leave for another time.

pkg/parse/root_test.go
```diff
@@ -1793,124 +1793,97 @@ func TestPrependRootSyncRemediatorStatus(t *testing.T) {
...
-				invertedConflictError.ToCSE(),
+				conflictBA.ToCSE(),
 			},
 			expectedErrors: []v1beta1.ConfigSyncError{
-				conflictError.ToCSE(),
-				invertedConflictError.ToCSE(),
+				conflictBA.ToCSE(),
 			},
 		},
 		"prepend ConflictingManagerError": {
```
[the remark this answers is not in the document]
karlkfi: done.

pkg/parse/root.go
sdowell: Glad it helped! I was also surprised by some of the existing behavior when I wrote the test.

It also helps with understanding the before/after behavior with this change. Maybe lets create a ticket for the other unexpected behavior

sdowell (approved):
/lgtm

PR #1345, opened 2024-07-18
test: add tests for prependRootSyncRemediatorStatus

sdowell:


pkg/parse/root_test.go
```diff
@@ -1788,3 +1788,59 @@ func TestSummarizeErrors(t *testing.T) {
...
+			err := prependRootSyncRemediatorStatus(ctx, fakeClient, rootSyncName,
+				[]status.ManagementConflictError{conflictError}, defaultDenominator)
+			if err != nil {
+				t.Fatal(err)
+			}
+			var updatedRootSync v1beta1.RootSync
+			if err := fakeClient.Get(ctx, rootsync.ObjectKey(rootSyncName), &updatedRootSync); err != nil {
+				t.Fatal(err)
+			}
+			if diff := cmp.Diff(tc.expectedErrors, updatedRootSync.Status.Sync.Errors); diff != "" {
```
karlkfi, line 1841: nit: cmp.Diff works, but I like the output from testutil.AssertEqual better. It shows the left and right and then diff. plus it's shorter to write.

pkg/parse/root_test.go
```diff
@@ -1788,3 +1788,59 @@ func TestSummarizeErrors(t *testing.T) {
...
+	}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			rootSync := fake.RootSyncObjectV1Beta1(rootSyncName)
+			rootSync.Status.Sync.Errors = tc.thisSyncErrors
+			fakeClient := syncertest.NewClient(t, core.Scheme, rootSync)
+			ctx := context.Background()
+			err := prependRootSyncRemediatorStatus(ctx, fakeClient, rootSyncName,
+				[]status.ManagementConflictError{conflictError}, defaultDenominator)
+			if err != nil {
```
karlkfi, line 1834: nit: I usually used require.NoError for this kind of thing.

sdowell: done

pkg/parse/root_test.go
sdowell: done

pkg/parse/root_test.go
```diff
@@ -1788,3 +1788,54 @@ func TestSummarizeErrors(t *testing.T) {
...
+	const rootSyncName = "my-root-sync"
+	const thisManager = "this-manager"
+	const otherManager = "other-manager"
+	conflictingObject := fake.NamespaceObject("foo-ns",
+		core.Annotation(metadata.ResourceManagerKey, otherManager))
+	conflictError := status.ManagementConflictErrorWrap(conflictingObject, thisManager)
+	testCases := map[string]struct {
+		thisSyncErrors []v1beta1.ConfigSyncError
+		expectedErrors []v1beta1.ConfigSyncError
+	}{
```
karlkfi, line 1802: possible extra cases:
- prepends when there are existing errors
- de-dupes ManagementConflictErrorWrap with inverted ManagementConflictErrorWrap
- de-dupes ManagementConflictErrorWraps with KptManagementConflictError

sdowell: I already covered the inverted case, but added testcases for the other two scenarios you mentioned. Note that the current logic does not actually de-dupe KptManagementConflictError

pkg/parse/root_test.go
```diff
@@ -1788,3 +1788,145 @@ func TestSummarizeErrors(t *testing.T) {
...
+		},
+		"unchanged CurrentManagerError": {
+			thisSyncErrors: []v1beta1.ConfigSyncError{
+				currentManagerError.ToCSE(),
+			},
+			expectedErrors: []v1beta1.ConfigSyncError{
+				currentManagerError.ToCSE(),
+			},
+		},
+		// TODO: the follow testcases document the current behavior, but probably isn't the desired behavior
```
karlkfi, line 1888: So it doesn't actually prepend them?

sdowell: It does prepend, I assume the intent is to de-dupe - which should be fixed by your change

pkg/parse/root_test.go
```diff
@@ -1788,3 +1788,145 @@ func TestSummarizeErrors(t *testing.T) {
...
+For more information, see https://g.co/cloud/acm-errors#knv1060`
+	const kptManagementConflictMessage = `KNV1060: The "other-manager" reconciler cannot manage resources declared in another repository. Remove the declaration for this resource from either the current repository, or the managed repository.
+
+metadata.name: foo-ns
+group:
+version: v1
+kind: Namespace
+
+For more information, see https://g.co/cloud/acm-errors#knv1060`
+	// TODO: CurrentManagerError currently seems to print an incorrect error message
```
karlkfi, line 1830: Yeah, I noticed this when I was working on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1344

It uses the same manager in both places.

karlkfi (approved):
/lgtm

PR #1346, opened 2024-07-18
refactor: rename fake package

sdowell:
This package is not only used in testing, so the path/naming is a bit misleading. This package primarily contains helper functions to construct k8s objects.

### DIFF
--- a/pkg/core/k8sobjects/cluster_config.go
+++ b/pkg/core/k8sobjects/cluster_config.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package fake
+package k8sobjects
 
 import (
 	v1 "kpt.dev/configsync/pkg/api/configmanagement/v1"
--- a/pkg/core/k8sobjects/namespace_config.go
+++ b/pkg/core/k8sobjects/namespace_config.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Google LLC
+// Copyright 2022 Google LLC
 //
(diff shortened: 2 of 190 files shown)


pkg/core/k8sobjects/cluster_config.go
nan-yu, line 1: ClusterConfig is a CRD used by the monorepo only. We can get rid of it in a separate PR.

pkg/core/k8sobjects/errors.go
nan-yu, line 1: This should go outside of the `k8sobjects` package.

pkg/core/k8sobjects/namespace_config.go
nan-yu, line 1: NamespaceConfig is also monorepo only, and can be removed in a separate PR.

pkg/core/k8sobjects/repo.go
nan-yu, line 1: Another monorepo only CR.

pkg/core/k8sobjects/errors.go
sdowell: Done

pkg/core/k8sobjects/cluster_config.go
sdowell: Ack

pkg/core/k8sobjects/namespace_config.go
sdowell: Ack

pkg/core/k8sobjects/repo.go
sdowell: Ack

nan-yu (approved):
/lgtm

PR #1348, opened 2024-07-19
fix: remediator ignores no resource/kind match

karlkfi:
- Update remediator to record NoMatchErrors as a resource conflict.
  These are reported as metrics and cause retry, but are still not
  reported as sync errors, for now.
- Update comments in TestCRDDeleteBeforeRemoveCustomResourceV1
  to explain why a conflict is expected and which ones might be
  encountered due to race condition.

### DIFF
--- a/e2e/testcases/custom_resources_test.go
+++ b/e2e/testcases/custom_resources_test.go
@@ -99,6 +99,7 @@ func TestCRDDeleteBeforeRemoveCustomResourceV1(t *testing.T) {
 
 	// Remove CRD
 	// This will garbage collect the CR too and block until both are deleted.
+	nt.T.Log("Deleting Anvil CRD")
 	_, err = nt.Shell.Kubectl("delete", "-f", crdFile)
 	if err != nil {
 		nt.T.Fatal(err)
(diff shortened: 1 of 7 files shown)


e2e/testcases/custom_resources_test.go
```diff
@@ -109,16 +109,13 @@ func TestCRDDeleteBeforeRemoveCustomResourceV1(t *testing.T) {
...
+	// Conflict implies either a different reconciler manager or drift between GET (or watch event) & UPDATE.
+	// TODO: distinguish between management conflict (unexpected manager annotation) and resource conflict (resource version change)
+	nt.Must(nomostest.ValidateMetrics(nt,
 		nomostest.ReconcilerErrorMetrics(nt, rootSyncLabels, firstCommitHash, metrics.ErrorSummary{
-			Conflicts: 1,
-		}))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+			Conflicts: 0,
```
sdowell, line 117: Why can this pass with `Conflict: 0` or `Conflict: 1` with no code changes? Seems a bit suspicious

nan-yu, line 117: The intent of the check here is to confirm the remediator detects the drift, but couldn't correct it.

Instead of checking no conflicts, what about adding a new metric for the [remediator error](https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.18.3-rc.4/pkg/remediator/reconcile/reconciler.go#L94), and validating a remediator error is recorded in the metrics? We can still check no conflicts, but it is just no enough.

karlkfi: Looking through the code again, I think this is actually a race condition. But I was also confused because the conflict metric is overloaded. It's being used to mean 4 different things:
1. Remediator got StatusReasonNotFound when performing an Update. This is a race condition, with another controller deleting an object before the Remediator had a chance to perform an update event in response to a previous update event.
2. Remediator got StatusReasonConflict when performing an Update. This just means some object of the same resource type got updated between event received and event handled. It may or may not be the same object at all.
3. Remediator observed a management conflict (manager annotation says some other reconciler is managing the object). This only gets reported by root reconcilers today.

This is problematic because:
- It confuses management conflicts with resource conflicts
- It doesn't record the metric reason #3 at all from namespace reconcilers
- It doesn't record the metric from the applier, even though it can report the same management conflict error

The problem with this particular test is with reason #2, which is a race condition that depends on the timing of whether the garbage collector deletes the object before the remediator has a chance to handle the previous update event from the applier, when the remediator was paused.

karlkfi: Figured out the actual problem. The remediator wasn't reporting NoMatchErrors as conflicts. So it was only reporting a conflict when it got an (Object) NotFound error or Resource Version Conflict BEFORE the custom resource controller deleted the CR.

karlkfi: I added handling of NoMatchErrors as conflicts using the existing metric. This should make it report the error metric consistently in this test.

If we want to add a new metric, it'll need a one-pager, because we also need to separate management conflicts from resource conflict, and list all the cases to make sure they're handled correctly. The management conflict reporting is spotty at best today. It only works for RootSyncs and only sometimes.  I don't really have time to do this now, due to the watch filter project.

sdowell (approved):
/lgtm

PR #1349, opened 2024-07-19
chore: bump k8s patch version

sdowell:


nan-yu (commented):
Can you update the commit to explain why we need to bump the k8s version, CVE fixes, or feature dependency?

sdowell:
> Can you update the commit to explain why we need to bump the k8s version, CVE fixes, or feature dependency?

done

nan-yu (approved):
/lgtm

PR #1352, opened 2024-07-23
feat: remove PoCo from nomos bugreport

sdowell:
PoCo is no longer included with ACM, so there is no longer a need to bundle the PoCo logs with nomos bugreport.


tiffanny29631 (approved):
/lgtm

nan-yu (approved):
/lgtm

PR #1353, opened 2024-07-23
feat: remove kube-system from nomos bugreport

sdowell:
config-management-operator has not been installed in the kube-system namespace since before 1.9.0. 1.9.0 was released in 2021, so it should not be necessary to include kube-system logs in current versions of nomos bugreport.

See: https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/release-notes#September_23_2021


tiffanny29631 (approved):
/lgtm

PR #1354, opened 2024-07-25
fix: RepoSync management conflict reporting

karlkfi:
- Change the run loop retry handler to check the conflict handler for errors, instead of using a flag in the remediator watchers. This fixes a bug where the applier may only be triggered once after a conflict, even if the conflict was not resolved.
- Change the remediator to treat namespace reconcilers the same as root reconcilers when reporting a management conflict error. This fixes a bug which could happen when the applier succeeded and then a new conflict was discovered, which wouldn't report the conflict until the next resync or source change.
- Use a RWMutex in the conflict handler, to allows parallel reads.
- Change the remote reporting (RootSync -> RootSync) of conflict status errors to read the errors from the conflict handler, instead of the applier errors. This just simplifies the code. The reported errors should be identical.

### DIFF
--- a/pkg/remediator/watch/filteredwatcher.go
+++ b/pkg/remediator/watch/filteredwatcher.go
@@ -33,7 +33,6 @@ import (
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/declared"
 	"kpt.dev/configsync/pkg/diff"
-	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/metrics"
 	"kpt.dev/configsync/pkg/remediator/conflict"
 	"kpt.dev/configsync/pkg/remediator/queue"
@@ -73,9 +72,6 @@ const maxWatchRetryFactor = 18
 type Runnable interface {
 	Stop()
 	Run(ctx context.Context) status.Error
-	ManagementConflict() bool
-	SetManagementConflict(object client.Object, commit string)
-	ClearManagementConflicts()
 	ClearManagementConflictsWithKind(gk schema.GroupKind)
 }
 
@@ -101,14 +97,13 @@ type filteredWatcher struct {
 	scope      declared.Scope
 	syncName   string
 	// errorTracker maps an error to the time when the same error happened last time.
-	errorTracker map[string]time.Time
+	errorTracker    map[string]time.Time
+	conflictHandler conflict.Handler
 
 	// The following fields are guarded by the mutex.
-	mux                sync.Mutex
-	base               watch.Interface
-	stopped            bool
-	managementConflict bool
-	conflictHandler    conflict.Handler
+	mux     sync.Mutex
+	base    watch.Interface
+	stopped bool
 }
 
 // filteredWatcher implements the Runnable interface.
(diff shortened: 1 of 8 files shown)


pkg/remediator/watch/filteredwatcher.go
```diff
@@ -185,30 +166,22 @@ func (w *filteredWatcher) SetManagementConflict(object client.Object, commit str
...
 	}
 
 	// The remediator detects conflict between two root reconcilers.
 	// Add the conflict error to the remediator, and the updateStatus goroutine will surface the ManagementConflictError (KNV1060).
-	// It also sets `managementConflict` true to keep retrying the parse-apply-watch loop
-	// so that the error can auto-resolve if the resource is removed from the conflicting manager's repository.
-	w.managementConflict = true
+	// Conflict error also triggers applier retry.
 	newManager := declared.ResourceManager(w.scope, w.syncName)
 	klog.Warningf("The remediator detects a management conflict for object %q between root reconcilers: %q and %q",
```
sdowell, line 179: This no longer is necessarily between two root reconcilers (could be ns-reconcilers). comment above also needs to be updated

karlkfi: Fixed. Rewrote the logs and comments.

I also merged shouldProcess & SetManagementConflict, since it doesn't need to be external.

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -504,14 +450,48 @@ func (w *filteredWatcher) shouldProcess(object client.Object) bool {
...
+			"The namespace reconciler should report the management conflict as an error.",
+			desiredManager, id, currentManager)
 		return false
 	}
-	w.conflictHandler.RemoveConflictError(id)
-	return true
+
+	// The remediator detected a conflict between two reconcilers.
+	klog.Errorf("Management conflict detected. "+
+		"Reconciler %q received a watch event for object %q, which is manage by namespace reconciler %q. ",
```
sdowell, line 489: nit: `which is managed`

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -504,14 +450,48 @@ func (w *filteredWatcher) shouldProcess(object client.Object) bool {
...
+
+	// A root reconciler can override resource managed by namespace reconciler.
+	// It is not a conflict from the root reconciler's perspective.
+	// The namespace reconciler should detect and report the conflict.
+	if w.scope == declared.RootScope && !declared.IsRootManager(currentManager) {
+		klog.V(5).Infof("Reconciler %q received a watch event for object %q, which is managed by namespace reconciler %q. "+
+			"Root reconcilers are allowed to adopt objects managed by namespace reconcilers. "+
+			"The namespace reconciler should report the management conflict as an error.",
+			desiredManager, id, currentManager)
 		return false
```
sdowell, line 484: Shouldn't this return `true`, since the root reconciler is allowed to manage the object?

karlkfi: If it gets here, CanManage is false, so no.

I'm not actually sure if this branch will ever execute, but if it does this prevents it from recording a conflict.

These branches were in SetManagementConflict, which didn't return anything. But if it got that far, shouldProcess was returning false.

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -504,14 +450,48 @@ func (w *filteredWatcher) shouldProcess(object client.Object) bool {
...
+		// Should never happen, since CanManage returned false.
+		klog.Errorf("Reconciler %q received a watch event for object %q, which has no manager annotation, but CanManage returned false.",
+			desiredManager, id)
+		return false
+	}
+
+	// A root reconciler can override resource managed by namespace reconciler.
+	// It is not a conflict from the root reconciler's perspective.
+	// The namespace reconciler should detect and report the conflict.
+	if w.scope == declared.RootScope && !declared.IsRootManager(currentManager) {
```
nan-yu, line 479: Are we going to run into this condition? If the condition is true, `diff.CanManage()` should have already removed the conflict and returned true.

pkg/remediator/watch/filteredwatcher.go
karlkfi: I think SetManagementConflict must have been called by some other code in the past, but it's not any more. Maybe it was from the applier code, which now adds conflicts to the conflict handler directly.

That said, the applier doesn't need either of these checks, and shouldn't actually use them. The object from the applier is the desired object, not the current object. So the currentManager here would be wrong. And the root applier should never report a management conflict event, because it always adopts.

Maybe we should just remove both checks as redundant.

pkg/remediator/watch/filteredwatcher.go
karlkfi: see my other thread with sam

pkg/remediator/watch/filteredwatcher.go
karlkfi: fixed.

pkg/remediator/watch/filteredwatcher.go
karlkfi: Removed

pkg/remediator/watch/filteredwatcher.go
karlkfi: Removed

sdowell (approved):
/lgtm

PR #1355, opened 2024-07-25
test: always override resources for wordpress

sdowell:
There are several tests which use the public wordpress helm chart which are susceptible to OOMKill on autopilot. This updates each of these tests to use a common helper function to consistently set resource overrides for the helm chart.


e2e/testcases/helm_sync_test.go
```diff
@@ -439,7 +327,8 @@ image:
 		testpredicates.HasExactlyImage("wordpress", "bitnami/wordpress", "", "sha256:362cb642db481ebf6f14eb0244fbfb17d531a84ecfe099cd3bba6810db56694e"),
 		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_USERNAME", "test-user-2"),
 		testpredicates.DeploymentHasEnvVar("wordpress", "WORDPRESS_EMAIL", "test-user@example.com"),
-		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_INLINE", "inline"),
+		testpredicates.DeploymentHasEnvVar("wordpress", "TEST_1", "val1"),
```
karlkfi, line 330: what happened to TEST_INLINE?

sdowell: The previous inline manifest defined `TEST_INLINE`, the common manifest declares these `extraEnvVars`. I don't think the name of the variables was that important, it's just asserting that the env vars from `values` persists.

The mutator could be updated to change the name of the env vars, but it seemed unnecessary

e2e/testcases/helm_sync_test.go
```diff
@@ -211,39 +153,13 @@ func TestPublicHelm(t *testing.T) {
...
-			"valuesFileRefs": [
-			  {
-				"name": "helm-watch-config-map"
-			  }
-			]
-		  }
-		}
-	  }`)
+	rs := rootSyncForWordpressHelmChart(nt, func(m map[string]interface{}) {
+		delete(m, "wordpressUsername") // omit username so it can be set with values file
```
nan-yu, line 157: Doesn't the values file overwrite the username?

sdowell: `values` takes precedence over `valuesFileRefs` it seems, which is what this test was verifying previously. I'm just retaining the assertion here. See [this assertion](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/helm_sync_test.go#L352)

karlkfi (approved):
/lgtm

e2e/testcases/helm_sync_test.go
nan-yu: ACK

nan-yu (approved):
/lgtm

mikebz:
thank you for doing this cleanup.  Also I think whether using Wordpress is the right thing in general.  What do we not test if we just have a Helm chart with simple instance of `nginx`?

karlkfi:
Wordpress is pretty common, like 43% to 65% of websites, depending who you ask.

Nginx is simpler and less resource intensive, but that also means it uses fewer helm template features.

I don't know how many of our customers use Wordpress but the reason it was chosen was popularity and complexity. It's a decent middle ground on chart complexity that gets us decent test coverage without writing a bespoke chart that tries to exercise as many helm features as possible.

mikebz:
I think the question is how much of Helm should we test.   If the cost of negligent, great, if the cost starts adding to our KTLO then we should figure out how to minimize it.  We assume that Helm, when it's available as for GA, has done through their regression testing.

karlkfi:
Just because helm works by itself doesn't mean it works when integrated with Config Sync. But yes, that's why we don't do more exhaustive testing, just a couple e2e tests to validate the integration.

In general, investment in testing raises quality, lowers support costs, and reduces turnover.

PR #1356, opened 2024-07-25
fix: applier not recording conflict metric

karlkfi:
Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1342

### DIFF
--- a/pkg/remediator/conflict/handler.go
+++ b/pkg/remediator/conflict/handler.go
@@ -15,12 +15,14 @@
 package conflict
 
 import (
+	"context"
 	"sync"
 
 	"github.com/elliotchance/orderedmap/v2"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
 	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/metrics"
 	"kpt.dev/configsync/pkg/status"
 )
 
@@ -39,6 +41,16 @@ type Handler interface {
 	HasConflictErrors() bool
 	// HasConflictError returns true when there is a conflict for the specified object ID.
 	HasConflictError(core.ID) bool
+}
+
+// Record a management conflict error, including log and metric.
+func Record(ctx context.Context, handler Handler, err status.ManagementConflictError, commit string) {
+	klog.Errorf("Management conflict detected. "+
+		"Reconciler %q received a watch event for object %q, which is manage by namespace reconciler %q. ",
+		err.DesiredManager(), err.ObjectID(), err.CurrentManager())
+	handler.AddConflictError(err.ObjectID(), err)
+	// TODO: Use separate metrics for management conflicts vs resource conflicts
+	metrics.RecordResourceConflict(ctx, commit)
 }
 
 // handler implements Handler.
@@ -66,9 +78,9 @@ func (h *handler) AddConflictError(id core.ID, newErr status.ManagementConflictE
 	// Ignore KptManagementConflictError if a ManagementConflictError was already reported.
 	// KptManagementConflictError don't have a real ConflictingManager recorded.
 	// TODO: Remove if cli-utils supports reporting the conflicting manager in InventoryOverlapError.
-	if newErr.ConflictingManager() == UnknownManager {
+	if newErr.CurrentManager() == UnknownManager {
 		if oldErr, found := h.conflictErrs.Get(id); found {
-			if oldErr.ConflictingManager() != UnknownManager {
+			if oldErr.CurrentManager() != UnknownManager {
 				return
 			}
 		}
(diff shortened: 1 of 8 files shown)


pkg/remediator/conflict/handler.go
```diff
@@ -41,6 +43,16 @@ type Handler interface {
 	HasConflictError(core.ID) bool
 }
 
+// Record a management conflict error, including log and metric.
+func Record(ctx context.Context, handler Handler, err status.ManagementConflictError, commit string) {
```
sdowell, line 47: Perhaps this should be a method of `Handler`?

pkg/status/management_conflict_error.go
```diff
@@ -113,6 +117,6 @@ func (m *managementConflictErrorImpl) Is(target error) bool {
 	return m.underlying.Is(target)
 }
 
-func (m managementConflictErrorImpl) ConflictingObjectID() core.ID {
+func (m managementConflictErrorImpl) ObjectID() core.ID {
 	return core.IDOf(m.resource)
```
nan-yu, line 121: missing the pointer

pkg/remediator/conflict/handler.go
karlkfi: I considered that, but it seemed weird. None of the other methods take a context. AddConflictError doesn't need the commit. And the handler doesn't otherwise "handle" metrics.

It could replace AddConflictError, but handler.Handle and handler.Record seem like generic names that would be weird with 4 arguments.

It's just sort of a helper function, not a method on the handler...

pkg/status/management_conflict_error.go
karlkfi: Should be fine. It doesn't mutate the object, so it doesn't need a pointer. Methods on a struct are accessible to pointers of the struct, so it still satisfies the interface.

Though, we should probably be internally consistent and either use or don't use the pointer on all the methods that don't mutate the struct.

karlkfi: Added a pointer here just to be consistent with all the other methods on this struct.

pkg/remediator/conflict/handler.go
karlkfi: I guess my feeling is that if the handler's reason for existing included logging and metrics it would compose things that help with that, right... like a logger or a metrics client? But we don't really do that in this code base. They all use package functions with global state. So it feels weird to have a handler method that calls both package functions and a handler method.

I can smell the code smell, I just can't put my finger on exactly why it smells.

karlkfi: Maybe this is a better argument: Handler is currently just a type-safe & thread-safe OrderedMap. Adding a Record method that takes context and commit makes it messier and expands the scope and interface of the abstraction into something larger and nebulous. This is why coming up with a good name for that method is hard.

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -461,7 +461,7 @@ func (w *filteredWatcher) shouldProcess(object client.Object) bool {
 
 	desiredManager := declared.ResourceManager(w.scope, w.syncName)
 	conflictErr := status.ManagementConflictErrorWrap(object, desiredManager)
-	currentManager := conflictErr.ConflictingManager()
+	currentManager := conflictErr.CurrentManager()
 
 	// TODO: Move logging & metric recording into the conflict handler
```
tiffanny29631, line 466: Can this be part be consolidated as well?

pkg/status/management_conflict_error.go
```diff
@@ -53,20 +55,22 @@ func ManagementConflictErrorWrap(resource client.Object, newManager string) Mana
 }
 
 type managementConflictErrorImpl struct {
-	underlying Error
-	resource   client.Object
-	// newManager refers to the manager annotation for the current remediator/reconciler.
```
tiffanny29631, line 58: Maybe keep the comments or update them for change of definition.

pkg/remediator/watch/filteredwatcher.go
karlkfi: consolidated?

pkg/status/management_conflict_error.go
karlkfi: The methods have descriptions now. The definition didn't really change, just the names. 

The comments here are actually specific to the -Wrap constructor inputs, not required or used by the impl itself. The impl shouldn't really care about the annotation or where the inputs came from.

That said, I see that technically speaking, the Invert() method does pull the current manager from the object, which it probably shouldn't. it should use the field instead. And we could probably make this slightly more efficient by not storing the whole Object and only storing the ID instead. THat would completely remove the annotation bit from the interface.

karlkfi: Ah, nevermind. We can't replace the object with just a ID, because the Body() formatter calls PrintResource, which reads the source annotation. I guess we could make an object-like thing that includes annotations, but that'd be a larger refactor.

I can fix Invert() though

nan-yu (approved):
/lgtm

pkg/status/management_conflict_error.go
karlkfi: Done.

pkg/remediator/watch/filteredwatcher.go
tiffanny29631: Change the below section to use conflict.Record?

karlkfi: Good catch. Fixed.

tiffanny29631 (approved):
/lgtm

PR #1357, opened 2024-07-25
feat: only output yaml in bugreport

sdowell:
Previously nomos bugreport output duplicate JSON and YAML files for all of the k8s objects. This is redundant, and YAML is generally more readable/useful for the purpose of the bugreport. The JSON increased the bugreport size unnecessarily and is therefore removed.

Also updates the test to perform a stricter file existence check, so that it will report an error for any unexpected files in the bugreport.


nan-yu (approved):
/lgtm

PR #1358, opened 2024-07-26
build: vendor all tools

sdowell:
This updates the build scaffolding to consistently vendor all tools using the tools.go file. This ensures all versions stay consistent when upgrading dependencies in go.mod and keeps builds hermetic due to vendoring. Doing so bumps the clientgen version which requires some changes to the command line arguments of generator binaries.


nan-yu (approved):
/lgtm

PR #1359, opened 2024-07-26
test: cleanup monorepo CRDs after migrate test

sdowell:
These CRDs were not being removed and left lingering on the cluster after the test.

This is causing a side effect for the nomos bugreport test which is outputting files for these CRDs.


tiffanny29631 (approved):
/lgtm

PR #1360, opened 2024-07-29
fix: bugreport logs for resourcegroup and askpass

sdowell:
- The resource-group-controller logs were only collected when the config-management label was on the Namespace. This meant they were only collected when the operator-applied label was on the Namespace. Now it will consistently collect the ResourceGroup logs whenever the configmanagement "system" label is on the Namespace. This fixes a flake that could happen if the bugreport test ran after the nomos migrate test, and also improves the behavior.
- Update test expectations to expect the askpass logs if the gcenode auth type is in use. This fixes a test failure on the CSR tests.


karlkfi (approved):
/lgtm

PR #1361, opened 2024-07-29
test: also expect askpass for gcpserviceaccount

sdowell:
The askpass sidecar is also present for the gcpserviceaccount git auth type, and thus the logs are expected.


nan-yu (approved):
/lgtm

PR #1364, opened 2024-07-30
test: fix logging to print correct line numbers

sdowell:
The Wrapper struct was unnecessary and broke the Helper method of the testing struct. The Helper method inspects the stack trace and adds the caller as a helper function, but this breaks when the Helper call is invoked in a nested helper function.

With this change the Helper calls work as expected and the correct line numbers are logged.

### DIFF
--- a/e2e/nomostest/testing/wrapper.go
+++ b/e2e/nomostest/testing/wrapper.go
@@ -16,6 +16,7 @@ package testing
 
 import (
 	"strings"
+	"testing"
 	"time"
 
 	"kpt.dev/configsync/e2e"
@@ -23,15 +24,15 @@ import (
 
 const errorPrefix = "ERROR:"
 
-// Wrapper implements NTB.
-type Wrapper struct {
-	t NTB
+// wrapper wraps testing.T to implement NTB with some custom behavior
+type wrapper struct {
+	*testing.T // embed struct for parent methods (e.g. Helper)
 }
 
 // NewShared creates a new instance of the Testing Wrapper that provides additional
 // functionality beyond the standard testing package.
 // It doesn't require the testFeature param.
-func NewShared(t NTB) *Wrapper {
+func NewShared(t *FakeNTB) NTB {
 	t.Helper()
 
 	testFeaturesFlag := *e2e.TestFeatures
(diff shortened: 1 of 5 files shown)


e2e/nomostest/testing/wrapper.go
```diff
@@ -16,22 +16,15 @@ package testing
 
 import (
 	"strings"
-	"time"
+	"testing"
 
 	"kpt.dev/configsync/e2e"
 )
 
-const errorPrefix = "ERROR:"
```
nan-yu, line 24: Do we still need the `ERROR:` prefix? It was added to make test failures more obvious. It was broken when timestamp was added. We could make it work again.

sdowell: I don't mind having the error prefix, but where do you propose injecting it?

sdowell: Updated this to keep the wrapper abstraction with a slightly different implementation so that we can keep the injected prefixes

e2e/nomostest/testing/wrapper.go
```diff
@@ -45,15 +46,12 @@ func NewShared(t NTB) *Wrapper {
 		}
 	}
 
-	tw := &Wrapper{
-		t: t,
-	}
-	return tw
+	return t
```
nan-yu, line 49: No need to return the original input argument given that we're not wrapping it anymore.

e2e/nomostest/testing/wrapper.go
karlkfi: I use "ERROR: "a lot to search for failures in the log.

karlkfi (approved):

e2e/nomostest/testing/wrapper.go
nan-yu: nt.Must -> wrapper -> testing.T. Isn't it also a two tiers of Helper function?

sdowell: With embedding, a call to `wrapper.Helper` now invokes the real `Helper` method - so when it inspects the stack trace it adds the correct function name

PR #1365, opened 2024-07-30
fix: syncer client not handling resource not found

karlkfi:
- Refactor the syncer Client and clientApplier to handle errors more consistently.
- Handle both NotFound and NoMatchFound errors everywhere they can be returned by the controller-runtime client.


sdowell (approved):
/lgtm

PR #1366, opened 2024-07-30
fix: CRD deletion race not reporting conflict

karlkfi:
- Now that UpdateWatches checks for the resource to exist before starting the watch, it can prevent remediator unpause. This is intentional, but it caused TestCRDDeleteBeforeRemoveCustomResourceV1 to become flaky, due to the resource conflict metric not always being recorded, after deleting a non-managed CRD required for a managed CR.
- This fix adds a new resource conflict error type for watches and records a conflict metric if it's returned by UpdateWatches.

### DIFF
--- a/pkg/remediator/watch/manager.go
+++ b/pkg/remediator/watch/manager.go
@@ -30,6 +30,7 @@ import (
 	"kpt.dev/configsync/pkg/remediator/conflict"
 	"kpt.dev/configsync/pkg/remediator/queue"
 	"kpt.dev/configsync/pkg/status"
+	syncerclient "kpt.dev/configsync/pkg/syncer/client"
 )
 
 // Manager accepts new resource lists that are parsed from Git and then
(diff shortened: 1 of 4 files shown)


pkg/remediator/watch/manager.go
```diff
@@ -221,6 +222,10 @@ func (m *Manager) UpdateWatches(ctx context.Context, gvkMap map[schema.GroupVers
 		// Only start watcher if the resource exists.
 		// All resources should exist and be established by now. If not, error.
 		if _, err := m.mapper.RESTMapping(gvk.GroupKind(), gvk.Version); err != nil {
+			if meta.IsNoMatchError(err) {
```
sdowell, line 225: style nit: switch on error and then `continue` after the switch

karlkfi: ya know, i think i tried that on the AddWatches last time and it popped me a linter error telling me to just use an if statement.

karlkfi: but maybe i'm misremembering and two cases are ok but one case is bad.

Either way, if this were a return, you'd expect it like this. It just looks redundant with continue .

sdowell: This seems like good style IMO:
```golang
switch {
case meta.IsNoMatchError(err):
  errs = status.Append(errs, syncerclient.ConflictWatchResourceDoesNotExist(err, gvk))
default: 
  errs = status.Append(errs, status.APIServerErrorWrap(err))
}
continue
```

karlkfi: done

sdowell (approved):
/lgtm

PR #1368, opened 2024-07-31
fix: remediator deleting objects after adoption

karlkfi:
- Root reconciler's remediator should not delete objects that have been adopted by another manager. This allows a namespace reconciler to adopt objects previously managed by a root reconciler without deleting the object first.
- This fixes a race condition that was causing TestConflictingDefinitions_RootToNamespace to be flaky.

Release Note:
- Fixed a bug where RootSync reconcilers could delete objects managed by RepoSync reconcilers. This simplifies the adoption process of handing off objects between reconcilers.


sdowell (approved):
/lgtm

PR #922, opened 2023-10-05
Update otelcontribcol image from 0.54.0 to 0.86.0 (#916)

victorpras:
The currently used version 0.54.0 has two fixable vulnerabilities, which requires to update jaeger to 1.47.0.
The openTelemetry-collector-contrib versions that are earlier than 0.85.0 use jaeger 1.41.0 or older.
Hence, this commit updates the image to the latest version, which unfortunately introduces breaking changes.

This commit includes the following changes:
- removed `--feature-gates=-exporter.googlecloud.OTLPDirect` because feature gate "exporter.googlecloud.OTLPDirect" is stable, can not be disabled.
- removed retry_on_failure because it was from the googlecloud exporter. The exporter itself handles retries, and retrying can cause issues. (#57233)
- added `--feature-gates=pkg.translator.prometheus.NormalizeName` to enable metric suffix trimming because the prometheus transformer appends `_ratio` to gauge metrics. link: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.86.0/pkg/translator/prometheus/normalize_name.go#L149
- updated the expected error message because otel-collector no longer has the message `failed to export time series to GCM`. See the old log[1] and the new log[2].

[1] old log:
```
[...]

PR #1369, opened 2024-07-31
Update otelcontribcol to v0.106.0-gke.2

tiffanny29631:
Upgrading to latest for vulnerability fix.

To upgrade to otelcontribcol 0.106.0 we need to mitigate two breaking changes in https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.104.0.

The opencensus receiver depends on 0.0.0.0 so disabling feature gate component.UseLocalHostAsDefaultHost.

With style `$FOO` deprecation, all env vars are changed to match the style. This brings a problem in the otel-agent container start when reconciler-manager and resource-group-controller do not initialize the following env vars:

- `configsync_sync_kind`
- `configsync_sync_name`
- `configsync_sync_namespace`

So splitting the otel-agent config between the reconciler and controllers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ HELM_STAGING_DIR := $(OUTPUT_DIR)/third_party/helm
 GIT_SYNC_VERSION := v4.2.3-gke.5__linux_amd64
 GIT_SYNC_IMAGE_NAME := gcr.io/config-management-release/git-sync:$(GIT_SYNC_VERSION)
 
-OTELCONTRIBCOL_VERSION := v0.102.0-gke.6
+OTELCONTRIBCOL_VERSION := v0.106.0-gke.2
 OTELCONTRIBCOL_IMAGE_NAME := gcr.io/config-management-release/otelcontribcol:$(OTELCONTRIBCOL_VERSION)
 
 # Keep KIND_VERSION in sync with the version defined in go.mod
--- a/manifests/otel-agent-reconciler-cm.yaml
+++ b/manifests/otel-agent-reconciler-cm.yaml
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -15,7 +15,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: otel-agent
+  name: otel-agent-reconciler
   namespace: config-management-system
   labels:
     app: opentelemetry
@@ -23,7 +23,7 @@ metadata:
     configmanagement.gke.io/system: "true"
     configmanagement.gke.io/arch: "csmr"
 data:
-  otel-agent-config.yaml: |
+  otel-agent-reconciler-config.yaml: |
     receivers:
       opencensus:
     exporters:
@@ -32,6 +32,24 @@ data:
         tls:
           insecure: true
     processors:
+      # Attributes processor adds custom configsync metric labels to applicable
+      # metrics to identify the sync object used to configure this deployment.
+      #
+      # Note: configsync.sync.generation is explicitly excluded here, because it
+      # is high cardinality. So we don't want to send it as a label, only as a
+      # resource attribute. That way it's only propagated to Prometheus, and not
+      # Monarch or Cloud Monitoring, which ignore custom resource attributes.
+      attributes:
+        actions:
+          - key: configsync.sync.kind
+            action: upsert
+            value: ${CONFIGSYNC_SYNC_KIND}
+          - key: configsync.sync.name
+            action: upsert
+            value: ${CONFIGSYNC_SYNC_NAME}
+          - key: configsync.sync.namespace
+            action: upsert
+            value: ${CONFIGSYNC_SYNC_NAMESPACE}
       batch:
       # Populate resource attributes from OTEL_RESOURCE_ATTRIBUTES env var and
       # the GCE metadata service, if available.
@@ -44,7 +62,7 @@ data:
       pipelines:
         metrics:
           receivers: [opencensus]
-          processors: [batch, resourcedetection]
+          processors: [batch, resourcedetection, attributes]
           exporters: [opencensus]
       telemetry:
         logs:
--- a/manifests/templates/resourcegroup-manifest.yaml
+++ b/manifests/templates/resourcegroup-manifest.yaml
@@ -231,6 +231,7 @@ spec:
             - ALL
       - args:
         - --config=/conf/otel-agent-config.yaml
+        - --feature-gates=-component.UseLocalHostAsDefaultHost
         command:
         - /otelcontribcol
         env:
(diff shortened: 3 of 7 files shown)


Makefile
sdowell, line 1: nit: typo in commit message (`veresion ` -> `version`)

tiffanny29631: Fixed

sdowell (approved):
/lgtm

nan-yu (approved):
/lgtm

tiffanny29631:
The failure is due to the two `very important breaking change` in the https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.104.0. 

For the first one we could disable the feature gate as an equivalent of using 0.0.0.0.

The second one can be mitigated with a temporary feature gate disable, but still requires change in ConfigMap to match the style when the previous format is deprecated.

manifests/otel-agent-reconciler-cm.yaml
```diff
@@ -0,0 +1,69 @@
+# Copyright 2022 Google LLC
```
sdowell, line 1: 2024

manifests/otel-agent-reconciler-cm.yaml
```diff
@@ -0,0 +1,69 @@
...
+    configmanagement.gke.io/arch: "csmr"
+data:
+  otel-agent-reconciler-config.yaml: |
+    receivers:
+      opencensus:
+    exporters:
+      opencensus:
+        endpoint: otel-collector.config-management-monitoring:55678
+        tls:
+          insecure: true
```
sdowell, line 33: What is the justification for disabling TLS?

tiffanny29631: This exporter in otel-agent sends data to the otel-collector receivers that runs on the same local host, so TLS is likely not necessary.

manifests/otel-agent-reconciler-cm.yaml
tiffanny29631: Done

manifests/otel-agent-reconciler-cm.yaml
sdowell: I'm not super familiar with these configurations - but it looks like the endpoint referenced in this block is for another Service running in the cluster, which may not be on the same local host. Am I missing something?

karlkfi: This looks like the sidecar agent config for the exporter sending to the otel-collector service/deployment.

That said, does the otel-collector have a TLS cert generated and injected or is it just using http?

karlkfi: If it's supposed to be the opencensus config for the reconciler writing to the sidecar agent, then the endpoint it wrong. It should be localhost. Otherwise, you're bypassing the agent.

tiffanny29631: Ah yeah sorry, it's sending to a different pod. In this case
- our metrics are custom metric data which are not critical to the healthiness of k8s or our product
- encryption may introduce extra overhead, and the cert configuration is an extra step in UX (or the certs need to be shipped as well)
- (this part I'm not sure) e2e tests are verifying metrics for every sync operation, the encryption may add work to the verification

karlkfi: FWIW, it looks like this section is just a copy of the existing agent config, which is already exporting metrics with TLS disabled.

IMO, we should probably enable TLS, but that would be a different feature/project.

So what we have here is fine for now.

tiffanny29631: A copy it is indeed :]

manifests/templates/resourcegroup-manifest.yaml
```diff
@@ -231,6 +231,8 @@ spec:
             - ALL
       - args:
         - --config=/conf/otel-agent-config.yaml
+        - --feature-gates=-component.UseLocalHostAsDefaultHost
+        - --feature-gates=-pkg.translator.prometheus.NormalizeName
```
karlkfi, line 235: What's pkg.translator.prometheus.NormalizeName for?

manifests/otel-agent-cm.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2024 Google LLC
```
karlkfi, line 1: nit: google policy on copyright headers is to set it when the file is created and leave it. So it doesn't need to be updated here.

Makefile
```diff
@@ -69,14 +69,14 @@ KUSTOMIZE_VERSION := v5.4.2-gke.0
 KUSTOMIZE := $(BIN_DIR)/kustomize
 KUSTOMIZE_STAGING_DIR := $(OUTPUT_DIR)/third_party/kustomize
 
-HELM_VERSION := v3.14.4-gke.2
+HELM_VERSION := v3.15.3-gke.0
```
karlkfi, line 72: Can you move the helm update to a different PR?
Best to keep these seperate in case we need to revert.

manifests/templates/resourcegroup-manifest.yaml
tiffanny29631: This is for dropping the feature gate AKA disabling it, to revert to a simple fashion of name sanitizing, I realized metrics from reconilers and reconciler-manager are using this but not the resource-group-controller so adding it here. 

See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheus/README.md

Makefile
tiffanny29631: Done, modified commit and title.

manifests/otel-agent-cm.yaml
tiffanny29631: ack

manifests/templates/resourcegroup-manifest.yaml
karlkfi: Can you make this specific change in a followup PR as well? Seems like a bug fix. Need to explain what it actually changes and what was broken.

tiffanny29631: Just removed the flag. Took a closer look - resource group controller only has gauge and distribution metrics, and the names are only composed with letters and underscores, so even with the normalizer on, no conversion is happening. Plus the test cases are not checking the r-g-c metrics, so no observations if there's a mismatch in name somehow.

We could add the flag just to be safe and align with other deployments.

karlkfi: Feel free to add it in a new PR just for consistency.

karlkfi (approved):
/lgtm

manifests/templates/resourcegroup-manifest.yaml
tiffanny29631: Ohhhh, I see, change was made a while back to actually enable this feature, but adding a `-` disabled it instead. https://github.com/GoogleContainerTools/kpt-config-sync/pull/922/commits/f3e2a5e1b371855afa83f89f009e2befd988bcdd

To re-enable it we could just remove the feature-gate flag as the feature is default on. But the trim & append will start to work and e2e test would fail (just ran a quick test).

From my reading of the doc it'd be better to have the normalizer running to convert the names for the favor of the Prometheus exporter/importer, which would break our e2e by a bit as it explicitly looks for the metric names defined and registered upon creation.

I lean keeping it as is and ignoring the naming, since any user with Prometheus operator would probably already hooked with the current naming.

Or we have a thorough & proper enablement of the feature.

WDYT @karlkfi

PR #1370, opened 2024-08-01
fix: treat IsNotFound as conflict on Create

sdowell:
If the CRD does not exist, the create/apply call may return IsNotFound error. This should be treated as a conflict so that the conflict metric is recorded.

Follow up to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1365

### DIFF
--- a/pkg/syncer/client/client.go
+++ b/pkg/syncer/client/client.go
@@ -75,7 +75,7 @@ func (c *Client) Create(ctx context.Context, obj client.Object, opts ...client.C
 		switch {
 		case apierrors.IsAlreadyExists(err):
 			return ConflictCreateAlreadyExists(err, obj)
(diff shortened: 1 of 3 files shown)


karlkfi (approved):
/lgtm

Digging through the code, I'm beginning to think that neither the client nor the server ever throw the NoMatchFound errors. Those only get thrown by the RESTMapper. 

So there may not actually be any way for us to distinguish between the two error states without asking the RESTMapper first. Thankfully, asking the RESTMapper doesn't always require an API call now, but it still leaves the possibility of race condition between asking the mapper and making the CRUD call.

I guess in this particular case, it doesn't really matter, since both errors are surfaced as a conflict error & metric, but it means we may not be able to deterministically know which error we're going to get in e2e tests when the resource is missing.

PR #1371, opened 2024-08-01
refactor: WaitForRootSyncStalledError ns arg

sdowell:
RootSyncs are always in the c-m-s namespace, so the namespace argument is redundant. This makes the function consistent with the other WaitForRootSync* functions.


nan-yu (approved):
/lgtm

nan-yu (approved):
/lgtm

PR #1375, opened 2024-08-02
fix: conflict metric on watched resource not found

sdowell:
This is intended to fix a scenario where watched CRD is deleted and the watch ends and attempts to restart. In this scenario a conflict metric may not have been reported if the watch fails before receiving a deleted event.

Follow up to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1366

### DIFF
--- a/pkg/parse/updater.go
+++ b/pkg/parse/updater.go
@@ -30,7 +30,6 @@ import (
 	"kpt.dev/configsync/pkg/remediator"
 	"kpt.dev/configsync/pkg/remediator/conflict"
 	"kpt.dev/configsync/pkg/status"
-	syncerclient "kpt.dev/configsync/pkg/syncer/client"
 	"kpt.dev/configsync/pkg/util/clusterconfig"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -133,7 +132,7 @@ func (u *Updater) update(ctx context.Context, cache *cacheForCommit) status.Mult
 		// Add new resources to the watch list, without removing old ones.
 		// This ensures controller conflicts are caught while the applier is running.
 		declaredGVKs, _ := u.Resources.DeclaredGVKs()
-		err = u.addWatches(ctx, declaredGVKs)
+		err = u.addWatches(ctx, declaredGVKs, cache.source.commit)
 		if err != nil {
 			return err
 		}
@@ -232,9 +231,9 @@ func (u *Updater) apply(ctx context.Context, objs []client.Object, commit string
 
 // addWatches tells the Remediator to watch additional resources without
 // stopping any.
-func (u *Updater) addWatches(ctx context.Context, gvks map[schema.GroupVersionKind]struct{}) status.MultiError {
+func (u *Updater) addWatches(ctx context.Context, gvks map[schema.GroupVersionKind]struct{}, commit string) status.MultiError {
 	klog.V(1).Info("Remediator watches adding...")
-	watchErrs := u.Remediator.AddWatches(ctx, gvks)
+	watchErrs := u.Remediator.AddWatches(ctx, gvks, commit)
 	u.SyncErrorCache.SetWatchErrs(watchErrs)
 	if watchErrs != nil {
 		klog.Warningf("Failed to add resource watches: %v", watchErrs)
@@ -248,19 +247,10 @@ func (u *Updater) addWatches(ctx context.Context, gvks map[schema.GroupVersionKi
 // watching any old resources that were removed from the apply set.
 func (u *Updater) updateWatches(ctx context.Context, gvks map[schema.GroupVersionKind]struct{}, commit string) status.MultiError {
 	klog.V(1).Info("Remediator watches updating...")
-	watchErrs := u.Remediator.UpdateWatches(ctx, gvks)
+	watchErrs := u.Remediator.UpdateWatches(ctx, gvks, commit)
 	u.SyncErrorCache.SetWatchErrs(watchErrs)
 	if watchErrs != nil {
 		klog.Warningf("Failed to update resource watches: %v", watchErrs)
-		// Since errors here prevent remediator unpause, we need to record
-		// resource conflict metrics here, when a required resource doesn't
-		// exist. This usually only occurs if a required CRD was deleted
-		// after all the objects that needed it but before the applier finished.
-		for _, err := range watchErrs.Errors() {
-			if err.Code() == syncerclient.ResourceConflictCode {
-				metrics.RecordResourceConflict(ctx, commit)
-			}
-		}
 		return watchErrs
 	}
 	klog.V(3).Info("Remediator watches updated")
--- a/pkg/remediator/watch/filteredwatcher.go
+++ b/pkg/remediator/watch/filteredwatcher.go
@@ -37,6 +37,7 @@ import (
 	"kpt.dev/configsync/pkg/remediator/conflict"
 	"kpt.dev/configsync/pkg/remediator/queue"
 	"kpt.dev/configsync/pkg/status"
+	syncerclient "kpt.dev/configsync/pkg/syncer/client"
 	"kpt.dev/configsync/pkg/util/log"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -73,6 +74,7 @@ type Runnable interface {
 	Stop()
 	Run(ctx context.Context) status.Error
 	ClearManagementConflictsWithKind(gk schema.GroupKind)
+	SetLatestCommit(string)
 }
 
 const (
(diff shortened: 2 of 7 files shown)


pkg/remediator/watch/filteredwatcher.go
```diff
@@ -155,6 +159,14 @@ func (w *filteredWatcher) ClearManagementConflictsWithKind(gk schema.GroupKind)
 	w.conflictHandler.ClearConflictErrorsWithKind(gk)
 }
 
+func (w *filteredWatcher) SetLatestCommitForGVK(commit string) {
```
karlkfi, line 162: did you mean to store these in a map by gvk?

Seems like we kinda have to, since we may have GVKs from multiple commits being watched simultaneously.

sdowell: Each `filteredWatcher` is scoped to a single GVK, so by storing the commit on the `filteredWatcher` we are already mapping GVK to commit.

And yes the reason I went this route is because you might have commit A -> commit B where a certain GVK that was watched on commit A is no longer being watched on commit B.

karlkfi: Ah, i see. The method name is just confusing because it doesn't take a gvk. Could use a comment, i guess.

pkg/remediator/watch/manager_test.go
```diff
@@ -35,14 +35,15 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/testutil"
 )
 
-func fakeRunnable() Runnable {
+func fakeRunnable(commit string) Runnable {
```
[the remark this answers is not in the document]
karlkfi: What if instead of re-constructing the watch, we just have the commit as a locked field that we can update on the Runnable after it's been started, and then only read it when there's an error? Sounds less clean, but more efficient in terms of starting and stopping watches.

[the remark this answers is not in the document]
karlkfi: The other option is a locked error handler callback.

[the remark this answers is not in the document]
sdowell: I didn't add any functionality to start/stop watches which wasn't already there before (also this line being commented on is just a fake unit test constructor in case that was unclear).

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -73,6 +73,8 @@ type Runnable interface {
 	Stop()
 	Run(ctx context.Context) status.Error
 	ClearManagementConflictsWithKind(gk schema.GroupKind)
+	SetLatestCommitForGVK(string)
+	latestCommitForGVK() string
```
karlkfi, line 77: If the setter's public, might as well make the getter public too.  Private methods on interfaces are rarely a good idea, especially public interfaces.

pkg/remediator/watch/filteredwatcher.go
karlkfi: Given that the Runnable interface doesn't mention GVK anywhere else, I suggest just dropping ForGVK from the method names on the interface and method them in the method comment for the filteredWatcher methods specifically.

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -329,6 +341,8 @@ func (w *filteredWatcher) start(ctx context.Context, resourceVersion string) (bo
 	if err != nil {
 		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 			return false, status.InternalWrapf(err, "failed to start watch for %s", w.gvk)
+		} else if apierrors.IsNotFound(err) {
+			metrics.RecordResourceConflict(ctx, w.latestCommit)
```
karlkfi, line 345: should proabably use the getter here, and put a lock on the get/set methods to aoid race condition between the start and set methods.

karlkfi: Does the start method block until the watch exits? If so, it can't use the same lock as the one protecting latestCommit.

This is partly why I don't love this approach. Multiple locks on a struct is a smell.

sdowell: Can you describe the race condition that you are concerned about?

pkg/remediator/watch/filteredwatcher.go
sdowell: Done.

pkg/remediator/watch/filteredwatcher.go
sdowell: The interface is only used within the package, so the whole interface could be private TBH. Do you have a style reference on why private methods on interfaces are a bad idea?

Updated to only use public methods.

pkg/remediator/watch/filteredwatcher.go
karlkfi: 1. AddWatches for commit A starts a new filteredWatcher
2. AddWatches for commit B sets the latestCommit on the existing filteredWatcher at the same time the watcher errors and terminates the start method, causing a race between setting and getting the latestCommit field.

karlkfi: I agree that stopping the watch for every commit seems wasteful, but `go test -race` will fail if the `latestCommit` field is read from and written to the same time from two different threads. So it needs locking around the field to allow concurrent use, even if it doesn't solve a specific failure scenario.

sdowell: Good point, done

pkg/remediator/watch/filteredwatcher.go
karlkfi: If the interface is only used in the package, then the interface can be private. If the interface is private, then there's no extra value from making the interface methods private. Making them private is meaningless. They might as well just all be public.

If the interface is public, then all the interface methods should be public, or else it can't be implemented or used by anything outside of the package.

So AFAICT, there's no reason to make private interface methods at all.

karlkfi (approved):
/lgtm

pkg/parse/updater.go
```diff
@@ -248,7 +248,7 @@ func (u *Updater) addWatches(ctx context.Context, gvks map[schema.GroupVersionKi
 // watching any old resources that were removed from the apply set.
 func (u *Updater) updateWatches(ctx context.Context, gvks map[schema.GroupVersionKind]struct{}, commit string) status.MultiError {
 	klog.V(1).Info("Remediator watches updating...")
-	watchErrs := u.Remediator.UpdateWatches(ctx, gvks)
+	watchErrs := u.Remediator.UpdateWatches(ctx, gvks, commit)
```
karlkfi, line 251: Can the error handing below for conflicts be removed now? Are they being handled by UpdateWatches?

karlkfi: Looks like no, but you could move the conflict handling down into the UpdateWatches where it checks the RESTMapper.

sdowell: Done

pkg/remediator/watch/manager.go
```diff
@@ -217,22 +218,21 @@ func (m *Manager) UpdateWatches(ctx context.Context, gvkMap map[schema.GroupVers
...
 		// Only start watcher if the resource exists.
 		// All resources should exist and be established by now. If not, error.
 		if _, err := m.mapper.RESTMapping(gvk.GroupKind(), gvk.Version); err != nil {
-			switch {
-			case meta.IsNoMatchError(err):
-				errs = status.Append(errs, syncerclient.ConflictWatchResourceDoesNotExist(err, gvk))
-			default:
-				errs = status.Append(errs, status.APIServerErrorWrap(err))
+			if meta.IsNoMatchError(err) {
+				metrics.RecordResourceConflict(ctx, commit)
```
karlkfi, line 229: Probably still want to wrap this as a ConflictWatchResourceDoesNotExist so it gets the more descriptive error message. We can just also immediately record the metric.

pkg/remediator/watch/filteredwatcher.go
```diff
@@ -329,6 +348,8 @@ func (w *filteredWatcher) start(ctx context.Context, resourceVersion string) (bo
 	if err != nil {
 		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 			return false, status.InternalWrapf(err, "failed to start watch for %s", w.gvk)
+		} else if apierrors.IsNotFound(err) {
+			metrics.RecordResourceConflict(ctx, w.getLatestCommit())
```
karlkfi, line 352: Should we also wrap the err here with ConflictWatchResourceDoesNotExist?

sdowell: Done

pkg/remediator/watch/manager.go
sdowell: Done

karlkfi (approved):
/lgtm

PR #1379, opened 2024-08-07
test: update test expectations for autopilot bursting

sdowell:
New versions of GKE autopilot support burstable Pods, which invalidates some of the previous test expectations for container resources on autopilot. This change adds a check for autopilot resource bursting and updates test expectations accodirngly.

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -115,6 +115,9 @@ type NT struct {
 	// ClusterVersion is the parsed version of the target test cluster
 	ClusterVersion *clusterversion.ClusterVersion
 
+	// ClusterSupportsBursting indicates if the cluster supports bursting.
+	ClusterSupportsBursting bool
+
 	// ClusterHash is a 64 character long unique identifier formed in hex used to identify a GKE cluster.
 	ClusterHash string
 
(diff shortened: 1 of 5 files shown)


e2e/nomostest/nt.go
```diff
@@ -888,6 +891,47 @@ func (nt *NT) detectGKEAutopilot(skipAutopilot bool) {
 	}
 }
 
+// Prerequisites for autopilot bursting:
+// - You originally created the cluster with GKE version 1.26 or later
+// - The cluster is running GKE version 1.30.2-gke.1394000 or later
+// See: https://cloud.google.com/kubernetes-engine/docs/how-to/pod-bursting-gke#availability-in-gke
+func (nt *NT) autopilotClusterSupportsBursting() (bool, error) {
+	args := []string{"container", "clusters", "describe", nt.ClusterName,
+		"--location", *e2e.GCPRegion, "--format", `value(initialClusterVersion)`}
```
nan-yu, line 900: The projectID also needs to be provided.

e2e/nomostest/nt.go
```diff
@@ -888,6 +891,47 @@ func (nt *NT) detectGKEAutopilot(skipAutopilot bool) {
...
+
+func (nt *NT) detectClusterSupportsBursting() {
+	if nt.IsGKEAutopilot {
+		var err error
+		nt.ClusterSupportsBursting, err = nt.autopilotClusterSupportsBursting()
+		if err != nil {
+			nt.T.Fatal(err)
+		}
+	} else {
+		nt.ClusterSupportsBursting = true
```
nan-yu, line 931: Does Kind cluster support Pod bursting?

sdowell: Yes

e2e/nomostest/nt.go
sdowell: Good catch, done

nan-yu (commented):
With Autopilot bursting, I think we don't need to set resource limits.

sdowell:
> With Autopilot bursting, I think we don't need to set resource limits.

Yes, this was discussed offline. We will probably want to revisit the autopilot adjustment logic in `reconciler-manager` to make it aware of whether the cluster supports bursting. I don't think we want to rush in such a change right before the release though. This change is merely aimed at adjusting the test expectations to make them accurate to the new autopilot behavior.

nan-yu (approved):
/lgtm

PR #1381, opened 2024-08-07
Align otel-agent feature gate in resource group controller

tiffanny29631:
This is to align the feature behavior with reconciler & reconciler-manager.

Currently we see no difference without the flag as

- there are no test coverage on resourcegroup controller metrics
- the naming convention obeys Prometheus rules so no trimming is happening


karlkfi (approved):
/lgtm

PR #1382, opened 2024-08-08
test: fix autopilot bursting check

sdowell:
detectGKEAutopilot is called later, so the IsGKEAutopilot variable is not set. This switches to use the command line flag instead.


tiffanny29631 (approved):
/lgtm

PR #1383, opened 2024-08-08
test: fix reset enabling deletion propagation after update

sdowell:
If the RSync is modified between the List call and the Update call, then the Update fails. Using MergePatch to inject the annotation makes the reset more robust.

Example failure: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-autopilot-rapid/1821388389237133312

### DIFF
--- a/e2e/nomostest/reset.go
+++ b/e2e/nomostest/reset.go
@@ -169,7 +169,10 @@ func ResetRootSyncs(nt *NT, rsList []v1beta1.RootSync) error {
 		// Enable deletion propagation, if not enabled
 		if EnableDeletionPropagation(rs) {
 			nt.T.Logf("[RESET] Enabling deletion propagation on RootSync %s", rsNN)
-			if err := nt.KubeClient.Update(rs); err != nil {
+			// Use MergePatch instead of Update in case modified since list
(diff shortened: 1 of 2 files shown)


e2e/nomostest/reset.go
```diff
@@ -169,7 +169,10 @@ func ResetRootSyncs(nt *NT, rsList []v1beta1.RootSync) error {
 		// Enable deletion propagation, if not enabled
 		if EnableDeletionPropagation(rs) {
 			nt.T.Logf("[RESET] Enabling deletion propagation on RootSync %s", rsNN)
-			if err := nt.KubeClient.Update(rs); err != nil {
+			// Use MergePatch instead of Update in case modified since list
+			patch := fmt.Sprintf(`{"metadata": {"annotations": {"%s": "%s"}}}`,
```
tiffanny29631, line 173: Apply the same for resetRepoSyncs?

sdowell: I did, it's right below on line 250

tiffanny29631: Ack

tiffanny29631 (approved):
/lgtm

PR #1385, opened 2024-08-09
e2e: refactor validateNumberOfObjectsEquals

karlkfi:
Return an error so the caller can decide whether to error or fail the test.


sdowell (approved):
/lgtm

PR #1386, opened 2024-08-12
test: fix cluster version comparison

sdowell:
The previous logic was not fully correct, for example it would consider 1.31.0 to be less than 1.30.2. This makes the logic a proper semver comparison, which should prevent breakage when newer cluster versions are tested in the future.

### DIFF
--- a/e2e/nomostest/clusterversion/cluster_version.go
+++ b/e2e/nomostest/clusterversion/cluster_version.go
@@ -53,6 +53,27 @@ func (cv ClusterVersion) GKESuffix() (int, error) {
 	return intVal, nil
 }
 
+// IsAtLeast returns whether the ClusterVersion is at least the provided version.
+func (cv ClusterVersion) IsAtLeast(other ClusterVersion) bool {
+	if cv.Major != other.Major {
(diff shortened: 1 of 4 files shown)


e2e/nomostest/clusterversion/cluster_version.go
```diff
@@ -53,6 +53,29 @@ func (cv ClusterVersion) GKESuffix() (int, error) {
...
+		return cv.Major > other.Major
+	}
+	if cv.Minor != other.Minor {
+		return cv.Minor > other.Minor
+	}
+	if cv.Patch != other.Patch {
+		return cv.Patch > other.Patch
+	}
+	// Compare -gke suffix, if they exist
+	if myGKESuffix, err := cv.GKESuffix(); err == nil {
```
karlkfi, line 68: If one has a gke suffix and then other doesn't, the gke suffix should probably be treated as newer, even if this is unlikely.

sdowell: Updated the behavior

karlkfi (approved):
/lgtm

PR #1388, opened 2024-08-12
Unvendor golangci-lint

karlkfi:
- golangci-lint uses the GPL license. We only use it in CI, but had vendored it to enable automatic updates by dependabot. Unfortunately, this makes it hard to build the aggregate licenses file included in the released container images, and makes it difficult to prove that Config Sync does not use re-release GPL code. So this change unvendors golangci-lint and installs it as-needed, built from source.


sdowell (approved):
/lgtm

tiffanny29631 (approved):
/lgtm

PR #1389, opened 2024-08-12
ci: make license check presubmit stricter

sdowell:
This updates the lint-license check to validate all dependencies in the vendor directory and use a specific allowlist of licenses. This will make it more apparent when new licenses are added in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -367,11 +367,16 @@ lint-go: "$(GOLANGCI_LINT)"
 lint-bash:
 	@./scripts/lint-bash.sh
 
+# List of currently permitted licenses to be included for go modules.
+# This may need to be updated as new dependencies are added.
+# See https://github.com/google/licenseclassifier/blob/e6a9bb99b5a6f71d5a34336b8245e305f5430f99/license_type.go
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -367,11 +367,16 @@ lint-go: "$(GOLANGCI_LINT)"
 lint-bash:
 	@./scripts/lint-bash.sh
 
+# List of currently permitted licenses to be included for go modules.
+# This may need to be updated as new dependencies are added.
+# See https://github.com/google/licenseclassifier/blob/e6a9bb99b5a6f71d5a34336b8245e305f5430f99/license_type.go
+ALLOWED_LICENSES := "MIT,ISC,Apache-2.0,BSD-2-Clause,BSD-3-Clause"
```
tiffanny29631, line 373: Also found MPL license in the list, from github.com/harshicorp

sdowell: That dependency is not currently included on main. The presubmit with this configuration would fail if it did

tiffanny29631: Ack

tiffanny29631 (approved):
/lgtm

PR #1390, opened 2024-08-13
e2e: Cleanup helm tests

karlkfi:
- Add NT.RootSyncHelmCharts & NT.RepoSyncHelmCharts to track expected RootSyncs & RepoSyncs with the expected Helm chart.
- Update NT.WatchForAllSyncs to validate RSyncs with Helm charts.
- Use NT.Must where applicable to simplify expectations.
- Use Validate after WatchForAllSyncs


e2e/nomostest/config_sync.go
```diff
@@ -692,6 +695,9 @@ func RootSyncObjectV1Alpha1FromRootRepo(nt *NT, name string) *v1alpha1.RootSync
 	} else if rs.Spec.Override != nil {
 		rs.Spec.Override.ReconcileTimeout = nil
 	}
+	// Unregister the Helm repo for syncing, if any
+	rsKey := client.ObjectKeyFromObject(rs)
+	delete(nt.RootSyncHelmCharts, rsKey) // TODO: delete repo?
```
nan-yu, line 700: Should we delete the helm chart from nt.RootSyncHelmCharts here?

If the helm chart is only added to `nt.RootSyncHelmCharts`, then we don't need the delete here.

In the case when a helm chart from `nt.RootSyncHelmCharts` is converted to a Git repo, we should delete it from `nt.RootSyncHelmCharts` with the conversion.

e2e/nomostest/config_sync.go
```diff
@@ -942,23 +957,37 @@ func (nt *NT) RootSyncObjectHelm(name string, chart *registryproviders.HelmPacka
...
+		rs.Spec.Override.ReconcileTimeout = nil
+	}
+	// Enable automatic deletion of managed objects by default.
+	// This helps ensure that test artifacts are cleaned up.
+	EnableDeletionPropagation(rs)
+	// Register the Helm chart for syncing
+	rsKey := client.ObjectKeyFromObject(rs)
+	nt.RootSyncHelmCharts[rsKey] = chartID
+	// Unregister the Git repo for syncing, if any
+	delete(nt.RootRepos, rsKey.Name) // TODO: delete repo?
```
nan-yu, line 973: Same here. We should delete from `nt.RootRepos` when a helm RootSync is converted from a Git RootSync.

e2e/nomostest/nt.go
```diff
@@ -139,6 +139,12 @@ type NT struct {
 	// The key is the namespace and name of the RepoSync object, the value points to the corresponding Repository object.
 	NonRootRepos map[types.NamespacedName]*gitproviders.Repository
 
+	// RootSyncHelmCharts is a map of RootSyncs that sync Helm charts
+	RootSyncHelmCharts map[types.NamespacedName]registryproviders.HelmChartID
```
nan-yu, line 143: With the new `RootSyncHelmCharts`, should we remove `RootRepos` to `RootSyncGitRepos`?

What about OCI repos? Do we want to store them separately?

nan-yu: What's the purpose of separating Helm charts from other repos? From this PR, it simplifies the validation a bit. What are other purposes?

My concern is that if we end up having 6 variables for Git/OCI/Helm * 2  (RootSync and RepoSync), it may be simpler to just pass the `rootSha1Fn` and `syncDirectoryMap` to the validation.

e2e/nomostest/config_sync.go
karlkfi: My thinking was that if we make a helm rsync, we should delete it from the map of git rsyncs, and vice versa. But RootRepos/NonRootRepos is acting as multiple different things: a list of RSyncs, and a list of Repositories, a map from RSync to Repo.

The TODO here is about whether we need to delete the underlying Repository when we remove the record of the RSync that uses it. However, I think the Repository is cleaned up elsewhere. So I don't think we need to do any cleanup here.

In the future, it might be cleaner to combine these maps, and make some generic Source interface with impls for Helm, OCI, & Git. But I didn't really want to do all the work now to make that happen. This effort as just a way to make the tests less flaky and kind of got a little out of hand.

e2e/nomostest/nt.go
karlkfi: The old validation method (passing rootSha1Fn and syncDirectoryMap to WatchForAllSyncs) only worked when RootRepos/NonRootRepos contained a record of the RootSync/RepoSync. Otherwise, the RSync would be skipped by WatchForAllSyncs.

I found this hard to read and it made refactoring difficult. So I added a dedicated way to ensure WatchForAllSyncs would validate RSyncs using Helm sources.

I agree that having a single map for all sources would be superior. I just didn't want to also refactor all the OCI tests too in the same PR. That could be a follow up simplification.

e2e/nomostest/config_sync.go
nan-yu: > My thinking was that if we make a helm rsync, we should delete it from the map of git rsyncs, and vice versa.

There are two ways of making a helm rsync.
- If it is a brand new helm rsync, we don't need to delete it from the map.
- If the helm rsync is converted from a Git rsync, then we should delete from the map when the conversion happens (a.k.a. where `RootSyncObjectHelm` is invoked).

karlkfi: Right, but delete is just a no-op if the key doesn't exist in the map. So we don't need to check that it exists before deleting it.

nan-yu: It is okay that we only have `nt.RootSyncHelmCharts` and `nt.RootRepos` for now by deleting the other one. It may be a bit messier if we're going to have `nt.RootSyncOCIRepos` in the future.

Looks good to me now, but just a headsup when we need to refactor a gain later.

nan-yu (approved):
/lgtm

e2e/nomostest/config_sync.go
karlkfi: I'll take another stab this week at some incremental improvements.

PR #1391, opened 2024-08-13
Update otelcontribcol to 0.118.0-gke.3

tiffanny29631:
Updating otelcontribcol to latest to include fix of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35006 and other essential security updates in components like Prometheus exporter.

The upgrade covers a breaking change in otelcontribcol of defaulting to local host instead of 0.0.0.0, the feature is now stable and [feature gate](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30702) has been [removed](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36589), Config Sync should explicitly set endpoint to host 0.0.0.0 to expose port for external pod access. This applies to the metric receiver `opencensus`, and the exporter `prometheus`.

This update also separates the ConfigMap for the otel-agent, distinguishing the one used by the reconcilers from the ConfigMap used by the reconciler manager and the resource group controller. As a result, only the ConfigMap for the reconcilers will include the Config Sync-related resource tags.

go/cs-upgrade-otelcontribcol-from-103

### DIFF
--- a/e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
+++ b/e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
@@ -17,9 +17,10 @@ data:
   otel-collector-config.yaml: |-
     receivers:
       opencensus:
+        endpoint: 0.0.0.0:55678
     exporters:
       prometheus:
-        endpoint: :8675
+        endpoint: 0.0.0.0:8675
         namespace: config_sync
         resource_to_telemetry_conversion:
           enabled: true
@@ -303,6 +304,7 @@ data:
                 aggregation_type: max
     extensions:
       health_check:
+        endpoint: 0.0.0.0:13133
     service:
       extensions: [health_check]
       pipelines:
--- a/manifests/otel-agent-reconciler-cm.yaml
+++ b/manifests/otel-agent-reconciler-cm.yaml
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This ConfigMap is used in the reconciler manager pod and the resource group controller pod.
-# It contains the OpenTelemetry (OTEL) agent configuration for metrics and resource detection.
+# This ConfigMap is used in the reconciler pod.
+# It contains the OpenTelemetry (OTEL) agent configuration for metrics and resource detection, with sync specific attributes.
 
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: otel-agent
+  name: otel-agent-reconciler
   namespace: config-management-system
   labels:
     app: opentelemetry
@@ -26,7 +26,7 @@ metadata:
     configmanagement.gke.io/system: "true"
     configmanagement.gke.io/arch: "csmr"
 data:
-  otel-agent-config.yaml: |
+  otel-agent-reconciler-config.yaml: |
     receivers:
       opencensus:
         endpoint: 0.0.0.0:55678
@@ -36,7 +36,27 @@ data:
         tls:
           insecure: true
     processors:
+      # Attributes processor adds custom configsync metric labels to applicable
+      # metrics to identify the sync object used to configure this deployment.
+      #
+      # Note: configsync.sync.generation is explicitly excluded here, because it
+      # is high cardinality. So we don't want to send it as a label, only as a
+      # resource attribute. That way it's only propagated to Prometheus, and not
+      # Monarch or Cloud Monitoring, which ignore custom resource attributes.
+      attributes:
+        actions:
+          - key: configsync.sync.kind
+            action: upsert
+            value: ${CONFIGSYNC_SYNC_KIND}
+          - key: configsync.sync.name
+            action: upsert
+            value: ${CONFIGSYNC_SYNC_NAME}
+          - key: configsync.sync.namespace
+            action: upsert
+            value: ${CONFIGSYNC_SYNC_NAMESPACE}
       batch:
+      # Populate resource attributes from OTEL_RESOURCE_ATTRIBUTES env var and
+      # the GCE metadata service, if available.
       resourcedetection:
         detectors: [env, gcp]
     extensions:
@@ -47,7 +67,7 @@ data:
       pipelines:
         metrics:
           receivers: [opencensus]
-          processors: [batch, resourcedetection]
+          processors: [batch, resourcedetection, attributes]
           exporters: [opencensus]
       telemetry:
         logs:
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -175,7 +175,7 @@ data:
            command:
            - /otelcontribcol
            args:
-           - "--config=/conf/otel-agent-config.yaml"
+           - "--config=/conf/otel-agent-reconciler-config.yaml"
            # The prometheus transformer appends `_ratio` to gauge metrics: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.86.0/pkg/translator/prometheus/normalize_name.go#L149
            # Add the feature gate to enable metric suffix trimming.
            - "--feature-gates=-pkg.translator.prometheus.NormalizeName"
@@ -190,8 +190,10 @@ data:
              protocol: TCP
            - containerPort: 8888  # Metrics.
              protocol: TCP
+           - containerPort: 13133 # Health check
+             protocol: TCP
            volumeMounts:
-           - name: otel-agent-config-vol
+           - name: otel-agent-config-reconciler-vol
              mountPath: /conf
            readinessProbe:
              httpGet:
@@ -282,9 +284,9 @@ data:
            secret:
              secretName: git-creds
              defaultMode: 288
-         - name: otel-agent-config-vol
+         - name: otel-agent-config-reconciler-vol
            configMap:
-             name: otel-agent
+             name: otel-agent-reconciler
              defaultMode: 420
          - name: service-account
            emptyDir: {}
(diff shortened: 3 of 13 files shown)


karlkfi:
Why id this safe to re-apply now?

tiffanny29631:
> Why id this safe to re-apply now?

Are you referring to the upstream fix or our release schedule?

karlkfi:
I mean, if we reverted it before, why is it now safe to reapply? What changed?

tiffanny29631:
> I mean, if we reverted it before, why is it now safe to reapply? What changed?

To ensure correct behavior, we've included a no_op_label in the metricstransform processor's label_set filter (explained in the description). This is a temporary solution until the upstream fix is available, but we need to update to the latest otelcontribcol regardless.

sdowell:
> > I mean, if we reverted it before, why is it now safe to reapply? What changed?
> 
> To ensure correct behavior, we've included a no_op_label in the metricstransform processor's label_set filter (explained in the description). This is a temporary solution until the upstream fix is available, but we need to update to the latest otelcontribcol regardless.

This isn't really a revert then, since new changes were added to the original commit. Suggest just creating a new commit instead of reapplying.

> we need to update to the latest otelcontribcol regardless.

why?

tiffanny29631:
Re @sdowell While not currently anticipated, there's a possibility that future vulnerability fixes need dependency upgrades that are incompatible with current version we use (0.103). In such cases, the version upgrade process will be similar.  Otelcontribcol versions beyond 0.104.0 will introduce breaking changes.

I could start a new PR.

tiffanny29631:
Recreated the commit and message.

e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
```diff
@@ -17,9 +17,10 @@ data:
   otel-collector-config.yaml: |-
     receivers:
       opencensus:
+        endpoint: 0.0.0.0:55678
```
karlkfi, line 20: Do we know why upstream switch from 0.0.0.0 to localhost? Why doesn't that work for us? Was it a security fix?

pkg/metrics/otel.go
```diff
@@ -184,7 +185,9 @@ processors:
         new_name: current_declared_resources
         operations:
           - action: aggregate_labels
-            label_set: []
+            # Using a no_op_label to get around issue in the upstream
+            # https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34430
+            label_set: [no_op_label]
```
karlkfi, line 190: Does setting label_set to nil or removing it work?

manifests/otel-agent-cm.yaml
```diff
@@ -26,43 +26,27 @@ data:
...
     receivers:
       opencensus:
+        endpoint: 0.0.0.0:55678
     exporters:
       opencensus:
         endpoint: otel-collector.config-management-monitoring:55678
         tls:
           insecure: true
     processors:
-      # Attributes processor adds custom configsync metric labels to applicable
```
karlkfi, line 35: Why did these attributes get removed from the agent config?

manifests/templates/reconciler-manager.yaml
```diff
@@ -81,6 +81,7 @@ spec:
         ports:
         - containerPort: 55678 # Default OpenCensus receiver port.
         - containerPort: 8888  # Metrics.
+        - containerPort: 13133 # Health check
```
karlkfi, line 84: Is this the otel-agent readiness probe port?

pkg/metrics/otel.go
karlkfi: Looks like it. Maybe clarify the comment?

tiffanny29631:
/blocked  https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34844

Generated files (go.mod etc.) were removed, which would break AutoVC integration. No affect to the continuous build of the internal mirror so far. 

A CI can be added to restore the generated files to a new branch, along with base image update for CVE scan purpose.

e2e/testdata/otel-collector/otel-cm-full-gcm.yaml
janetkuo: Yes it's for CVE https://github.com/open-telemetry/opentelemetry-collector/issues/6151

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -282,9 +284,9 @@ data:
            secret:
              secretName: git-creds
              defaultMode: 288
-         - name: otel-agent-config-vol
+         - name: otel-agent-config-reconciler-vol
            configMap:
-             name: otel-agent
```
janetkuo, line 287: Is this ConfigMap still used anywhere?

manifests/otel-agent-cm.yaml
janetkuo: It seems to be replaced by `manifests/otel-agent-reconciler-cm.yaml`

janetkuo:
@code-review-assist review

manifests/otel-agent-cm.yaml
tiffanny29631: The otel-agent CM is now used by resource group controller and reconciler manager only, which do not have these sync specific labels. The new one that contains these labels is otel-agent-reconciler-cm.yaml.

manifests/templates/reconciler-manager.yaml
tiffanny29631: It is. The readiness check would fail and container will be marked as notReady by k8s if this port is not specified. It has something to do with the otelcontribcol switching from 0.0.0.0 to default using localhost, even though the healthcheck configuration has specified to listen to 0.0.0.0:13133 the port still need to be specified in the manifest.

manifests/otel-agent-cm.yaml
tiffanny29631: Added comment to the file.

pkg/metrics/otel.go
tiffanny29631: Reverted back to use empty array as bug was fixed from upstream.

test/kustomization/expected.yaml
```diff
@@ -5650,6 +5652,47 @@ data:
...
+    service:
+      extensions: [health_check]
+      pipelines:
+        metrics:
+          receivers: [opencensus]
+          processors: [batch, resourcedetection]
+          exporters: [opencensus]
+      telemetry:
+        logs:
+          level: "INFO"
```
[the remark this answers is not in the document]
tiffanny29631: Added in comment for each file.

manifests/otel-agent-cm.yaml
```diff
@@ -26,43 +26,27 @@ data:
...
-          - key: configsync.sync.name
-            action: upsert
-            value: $CONFIGSYNC_SYNC_NAME
-          - key: configsync.sync.namespace
-            action: upsert
-            value: $CONFIGSYNC_SYNC_NAMESPACE
       batch:
       # Populate resource attributes from OTEL_RESOURCE_ATTRIBUTES env var and
       # the GCE metadata service, if available.
       resourcedetection:
```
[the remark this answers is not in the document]
tiffanny29631: Done

sdowell:
Please clean up the commit messages and history

manifests/templates/reconciler-manager.yaml
tiffanny29631: The updated image should already have this port open, but still worth to specify it here in case of any change.

tiffanny29631:
> Please clean up the commit messages and history

Done, commits squashed and fixed message and title.

karlkfi:
@code-review-assist review

test/kustomization/expected.yaml
```diff
@@ -5650,6 +5652,45 @@ data:
   otel-agent-config.yaml: |
     receivers:
       opencensus:
+        endpoint: 0.0.0.0:55678
+    exporters:
+      opencensus:
+        endpoint: otel-collector.config-management-monitoring:55678
+        tls:
+          insecure: true
```
karlkfi, line 5660: this probably isn't new, but do we care that our metrics are being sent unencrypted between the agents and otel-collector?

/cc @janetkuo

janetkuo: Agreed that we shouldn't use insecure, but given that it's unrelated to this PR, let's track that fix separately.

tiffanny29631: For this topic I created b/402998344 to followup.

karlkfi (approved):
/lgtm

manifests/otel-agent-reconciler-cm.yaml
```diff
@@ -0,0 +1,74 @@
...
+      #
+      # Note: configsync.sync.generation is explicitly excluded here, because it
+      # is high cardinality. So we don't want to send it as a label, only as a
+      # resource attribute. That way it's only propagated to Prometheus, and not
+      # Monarch or Cloud Monitoring, which ignore custom resource attributes.
+      attributes:
+        actions:
+          - key: configsync.sync.kind
+            action: upsert
+            value: ${CONFIGSYNC_SYNC_KIND}
```
karlkfi, line 50: Did you verify these are getting replaced correctly?

IIRC, they were previously being injected using the downward API from reconciler Deployment metadata values, which are set by the reconciler-manager.

karlkfi: I might be thinking of these ones: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/manifests/templates/reconciler-manager-configmap.yaml#L242-L256

How do the variables here get replaced?

karlkfi: Oh, I see. This ConfigMap is mounted on the otel-agent. So otel-collector must detect that they're environment variables and render them from the otel-agent container's environment variables, which are set on the reconciler Deployment I linked.

karlkfi: Looks like we don't have test coverage for this, but in chat you said you verified manually. So it doesn't need to block. Please add tests for these metrics fields in another PR.

PR #1392, opened 2024-08-13
test: delete GAR repositories on teardown

sdowell:
Some side effects have been noticed in the periodic jobs, such as an image never getting deleted if the cleanup fails on a prior execution. As an example this can cause TestHelmLatestVersion to become flaky, since the 3.0.0 helm chart already exists before the test starts.

Tearing down the entire GAR repository allows the periodic jobs to self heal side effects after a failed cleanup.

Example failure: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-regular/1823419274467741696


tiffanny29631 (approved):
/lgtm

PR #1393, opened 2024-08-16
test: Add e2e tests to cover additional scenarios where external files are rendered

Camila-B:
* Adds tests that cover rendering outside of the root directory: a local valuesFile, a local file that is referenced in a `configMapGenerator`, and a local file referenced in the `resources` field
* b/355278028


e2e/nomostest/testpredicates/predicates.go
```diff
@@ -1518,3 +1518,20 @@ func RootSyncHasNomosStatus(expectedCommit, expectedStatus string) Predicate {
...
+func ConfigMapHasData(key string, value string) Predicate {
+	return func(o client.Object) error {
+		if o == nil {
+			return ErrObjectNotFound
+		}
+		cm, ok := o.(*corev1.ConfigMap)
+		if !ok {
+			return WrongTypeErr(o, cm)
+		}
+		if cm.Data[key] == value {
```
sdowell, line 1532: Suggest checking that the value exists as well. Currently this does not differentiate between the key missing and the key containing an empty string.

```golang
if gotVal, ok := cm.Data[key]; !ok {
  return fmt.Errorf(...)
} else if gotVal != value {
  return fmt.Errorf(...)
}
return nil
```

e2e/testcases/hydration_test.go
```diff
@@ -170,6 +170,42 @@ func TestHydrateKustomizeComponents(t *testing.T) {
...
+	nt.Must(nt.RootRepos[configsync.RootSyncName].CommitAndPush("add DRY configs to the repository"))
+
+	nt.T.Log("Update RootSync to sync from the external-files directory")
+	rs := k8sobjects.RootSyncObjectV1Beta1(configsync.RootSyncName)
+	nt.MustMergePatch(rs, `{"spec": {"git": {"dir": "external-files"}}}`)
+	nt.Must(nt.WatchForAllSyncs(nomostest.WithSyncDirectoryMap(syncDirMap)))
+
+	// Validate nomos status
+	latestCommit := nt.RootRepos[configsync.RootSyncName].MustHash(nt.T)
+	nt.Must(nt.Validate(configsync.RootSyncName, configsync.ControllerNamespace, &v1beta1.RootSync{},
```
sdowell, line 194: I think this check is redundant with the `WatchForAllSyncs` check above. Is there something specific you're trying to assert here?

Camila-B: No I was just copying the general structure of the other tests which have a similar check. I went ahead and removed it.

e2e/nomostest/testpredicates/predicates.go
Camila-B: Thanks, done

e2e/testdata/hydration/external-files/namespace.yaml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 Google LLC
```
sdowell, line 1: 2024

Camila-B: Thanks for catching that! Done

sdowell (approved):
/lgtm

PR #1395, opened 2024-08-20
e2e: Simplify tracking of RSync sources

karlkfi:
Merge the 4 maps tracking git and helm sources into a map of SyncSources.

For now, the OCISyncSource is a placeholder, since this PR is already a mega huge refactor.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -31,12 +31,14 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e"
 	"kpt.dev/configsync/e2e/nomostest/gitproviders"
 	"kpt.dev/configsync/e2e/nomostest/metrics"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/e2e/nomostest/registryproviders"
+	"kpt.dev/configsync/e2e/nomostest/syncsource"
 	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/e2e/nomostest/testwatcher"
@@ -87,7 +89,58 @@ var (
 	reconcilerPollingPeriod time.Duration
 	// hydrationPollingPeriod specifies hydration-controller polling period as time.Duration
 	hydrationPollingPeriod time.Duration
+
+	// DefaultRootReconcilerName is the root-reconciler name of the default RootSync object: "root-sync".
+	DefaultRootReconcilerName = core.RootReconcilerName(configsync.RootSyncName)
+	// DefaultRootRepoNamespacedName is the NamespacedName of the default RootSync object.
+	DefaultRootRepoNamespacedName = RootSyncNN(configsync.RootSyncName)
+	// DefaultRootSyncID is the ID of the default RootSync object.
+	DefaultRootSyncID = RootSyncID(configsync.RootSyncName)
 )
+
+// RootSyncNN returns the NamespacedName of the RootSync object.
+func RootSyncNN(name string) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: configsync.ControllerNamespace,
+		Name:      name,
+	}
+}
+
+// RepoSyncNN returns the NamespacedName of the RepoSync object.
+func RepoSyncNN(ns, name string) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: ns,
+		Name:      name,
+	}
+}
+
+// RootSyncID returns an ID for the specified RootSync object.
+func RootSyncID(name string) core.ID {
+	return core.ID{
+		GroupKind: schema.GroupKind{
+			Group: v1beta1.SchemeGroupVersion.Group,
+			Kind:  configsync.RootSyncKind,
+		},
+		ObjectKey: client.ObjectKey{
+			Name:      name,
+			Namespace: configsync.ControllerNamespace,
+		},
+	}
+}
+
+// RepoSyncID returns an ID for the specified RepoSync object.
+func RepoSyncID(name, namespace string) core.ID {
+	return core.ID{
+		GroupKind: schema.GroupKind{
+			Group: v1beta1.SchemeGroupVersion.Group,
+			Kind:  configsync.RepoSyncKind,
+		},
+		ObjectKey: client.ObjectKey{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+}
 
 // IsReconcilerTemplateConfigMap returns true if passed obj is the
 // reconciler-manager ConfigMap reconciler-manager-cm in config-management namespace.
@@ -360,8 +413,19 @@ func multiRepoObjects(objects []client.Object, opts ...func(obj client.Object) e
 // RootSync root-sync will be re-created, if not found.
 func ValidateMultiRepoDeployments(nt *NT) error {
 	// Create root-sync RootSync, if registered.
-	if _, found := nt.RootRepos[configsync.RootSyncName]; found {
-		rs := RootSyncObjectV1Beta1FromRootRepo(nt, configsync.RootSyncName)
+	if source, found := nt.SyncSources[DefaultRootSyncID]; found {
+		id := DefaultRootSyncID
+		var rs *v1beta1.RootSync
+		switch tSource := source.(type) {
+		case *syncsource.GitSyncSource:
+			rs = RootSyncObjectV1Beta1FromRootRepo(nt, id.Name)
+		case *syncsource.HelmSyncSource:
+			rs = nt.RootSyncObjectHelm(id.Name, tSource.ChartID)
+		// case *syncsource.OCISyncSource:
+		// TODO: setup OCI RootSyncs
+		default:
+			nt.T.Fatalf("Invalid RootSync source %T: %s", source, id.Name)
+		}
 		if err := nt.KubeClient.Create(rs); err != nil {
 			if !apierrors.IsAlreadyExists(err) {
 				nt.T.Fatal(err)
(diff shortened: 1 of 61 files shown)


e2e/nomostest/config_sync.go
```diff
@@ -87,8 +89,59 @@ var (
...
+		},
+		ObjectKey: client.ObjectKey{
+			Name:      name,
+			Namespace: configsync.ControllerNamespace,
+		},
+	}
+}
+
+// RepoSyncID returns an ID for the specified RepoSync object.
+func RepoSyncID(name, namespace string) core.ID {
```
sdowell, line 132: nit: types.NamespacedName is a bit more user friendly as a parameter. No need to remember the argument ordering of name vs namespaces

e2e/nomostest/config_sync.go
```diff
@@ -87,8 +89,59 @@ var (
...
+// RootSyncNN returns the NamespacedName of the RootSync object.
+func RootSyncNN(name string) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: configsync.ControllerNamespace,
+		Name:      name,
+	}
+}
+
+// RepoSyncNN returns the NamespacedName of the RepoSync object.
+func RepoSyncNN(ns, name string) types.NamespacedName {
```
sdowell, line 110: nit: Is this function really needed?

e2e/nomostest/config_sync.go
```diff
@@ -618,31 +682,53 @@ func setupDelegatedControl(nt *NT) {
...
-		if rsName == configsync.RootSyncName {
+		if id.Name == configsync.RootSyncName {
 			continue
 		}
-		rs := RootSyncObjectV1Beta1FromRootRepo(nt, rsName)
+		var rs *v1beta1.RootSync
+		switch tSource := source.(type) {
+		case *syncsource.GitSyncSource:
+			rs = RootSyncObjectV1Beta1FromRootRepo(nt, id.Name)
+		case *syncsource.HelmSyncSource:
```
sdowell, line 695: Is this needed yet? It looks like you can still only specify Git `RootRepos`/`NamespaceRepos` when creating initializing NT.

karlkfi: I haven't added any NT.New options for helm or oci, but they could be added later. I just put in some placeholder code for now. It also makes it reasonable to use a switch and not just an if statement, which should make for less change later.

e2e/nomostest/config_sync.go
karlkfi: This already existed, I just moved it to the configsync file.

e2e/nomostest/config_sync.go
karlkfi: While true, it would actually require a fair amount of refactoring to ensure all the callers have an NN available. In this refactor, most of the uses have name and namespace as some combination of constants and hard-coded strings. Then the NN is derived from the id.ObjectKey, rather than the other way around.

I agree NN is better on usability from an ordering perspective, but if it turns all the one-liners into two-liners, I'm not sure it's all that much better. 

That said, a lot of the existing functions take name and namespace args, and they're not always in the same order. Converting them all to NN might make sense as a follow up PR, if you're interested.

sdowell: Yeah the main thing that bothers me is the inconsistent ordering, which can be a bit of a gotcha. It would be a pretty big refactor to replace with NamespacedName across the board

e2e/nomostest/config_sync.go
karlkfi: FWIW, this trades off the ordering concern for being less to type and more consistently discoverable and named when together with RootSyncNN. Seems like a wash to me.

e2e/nomostest/config_sync.go
sdowell: I feel it's a bit misleading to have this code path here even though it's currently dead/unreachable. I'd be fine with leaving this as a switch (for easy future expansion) but leave the other types (helm/oci) unimplemented

e2e/nomostest/config_sync.go
karlkfi: I hear you. Mostly I let the IDE tell me what the order is, which makes it less of an issue. But if you think you can make it better, go for it.

e2e/nomostest/config_sync.go
karlkfi: I basically just defensively ensured all the switches check git and helm and had a default failure.

There's only 7 of these switches in the code:
- 2 in setupDelegatedControl
- 2 in setupCentralizedControl
- 2 in WatchForAllSyncs
- 1 in ValidateMultiRepoDeployments (called by WatchForAllSyncs)

I think the bottom 3 are required. But I could remove the helm case in the top 4, if you think it would avoid confusion between now and whenever we decide to add helm setup in NT.New options.

Maybe using Helm in setup doesn't makes sense. The RootSyncs all share the same RBAC, which is why most of the tests still work that switch root-sync to use Helm. But for RepoSyncs, the RBAC and Namespace are defined in the root-sync, so having the root-sync define Helm RepoSyncs should just work if we decide to add it. I'm not sure how many tests that would affect, but I didn't want to have inconsistent code that only implements the helm case for RepoSyncs...

karlkfi: Looks like we have 2 tests that validate Helm RepoSyncs: TestHelmNamespaceRepo & TestHelmConfigMapNamespaceRepo. They both use central control and could be migrated to define the helm RepoSync in the NT setup. But in order for setup to sync successfully, we have to also add something like BuildAndPushHelmPackage as an NT.New option.

karlkfi: I think I'm gonna hold off on that refactor till a follow up PR. It would require changing how opts.NamespaceRepos & opts.RootRepos work, to add helm and reduce them to a single map.

sdowell: Yeah agreed the bottom 3 are required but not the top 4. I'd say let's just remove the top 4 with a comment that cases must be added for helm/oci if we want to support instantiating with NT.New options.

I don't think we have enough use cases currently to warrant the added complexity, but we could always add it later.

karlkfi: A compromise: I left the helm case in the but commented it out for now.

e2e/testcases/acme_test.go
```diff
@@ -45,15 +44,18 @@ func configSyncManagementLabels(namespace, folder string) map[string]string {
 func TestAcmeCorpRepo(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.Reconciliation1)
 
+	rootSyncGitRepo := nt.SyncSourceGitRepository(nomostest.DefaultRootSyncID)
```
sdowell, line 47: nit: If we wanted to break this into two smaller PRs we could potentially break out all of the `nt.SyncSourceGitRepository` refactors to a separate change (the first change just being a wrapper for `nt.RootRepos`/`nt.NamespaceRepos`). Might make this a bit easier to review.

e2e/testcases/helm_sync_test.go
```diff
@@ -387,32 +389,35 @@ func TestHelmLatestVersion(t *testing.T) {
 
 	// helm-sync automatically detects and updates to the new helm chart version
 	newVersion = "2.5.9"
-	chart, err = nt.BuildAndPushHelmPackage(nomostest.RootSyncNN(configsync.RootSyncName),
+	chart, err = nt.BuildAndPushHelmPackage(rootSyncID.ObjectKey,
```
sdowell, line 392: Should `nt.BuildAndPushHelmPackage` just update the ChartID, so that it doesn't need to be manually updated below?

e2e/testcases/invalid_auth_test.go
```diff
@@ -55,7 +55,9 @@ func TestInvalidAuth(t *testing.T) {
 	if err != nil {
 		nt.T.Fatal(err)
 	}
-	commitHash := nt.RootRepos[configsync.RootSyncName].MustHash(nt.T)
+	// TODO: What's the commit in the rsync status if the repo doesn't exist and/or auth fails?
```
sdowell, line 58: This is a good catch, do we have an issue to track this?

e2e/testcases/helm_sync_test.go
karlkfi: Could do that. Tho realistically the NN passed to BuildAndPushHelmPackage is only used for temp file location and artifacts path. It might make more sense to move those to be based on the helm chart name instead, so more than one RSync can use the same chart, if needed.

Updating the expectation for helm charts could use a helper function though, to make it less verbose.

e2e/testcases/invalid_auth_test.go
karlkfi: Filed b/361182373

e2e/testcases/helm_sync_test.go
karlkfi: Think I'm gonna save that optimization for a follow up PR.

e2e/testcases/invalid_auth_test.go
karlkfi: Updated the TODOs

e2e/testcases/acme_test.go
karlkfi: Separating those out across a few hundred tests might take a couple hours... because the old code doesn't use IDs or SyncSource. So let me know if you think it's actually gonna save you a few hours reviewing.

sdowell: I was thinking it could use the same ID argument to return a Repository, but just use RootRepos/NamespaceRepos under the hood - that way the follow up just needs to replace the `SyncSourceGitRepository` implementation to use the new SyncSources. I'm almost done reviewing though so no biggie

e2e/nomostest/wait_for_sync.go
```diff
@@ -171,50 +172,54 @@ func (nt *NT) WatchForAllSyncs(options ...WatchForAllSyncsOptions) error {
...
 	if opts.syncRootRepos {
-		for name := range nt.RootRepos {
-			if _, ok := opts.skipRootRepos[name]; ok {
+		for id, source := range nt.SyncSources.RootSyncs() {
+			if _, ok := opts.skipRootRepos[id.Name]; ok {
 				continue
 			}
-			nn := RootSyncNN(name)
-			syncDir := syncDirectory(opts.syncDirectoryMap, nn)
+			// TODO: Move SyncPath logic into a SyncSource method
```
sdowell, line 179: Can this be done now?

karlkfi: Would rather do it in a followup PR. I';m not sure what all it requires changing and it'll be hard to see the diff if added to the same commit.

e2e/testcases/helm_sync_test.go
karlkfi: unresolving to keep a record of follow up PRs TODO.

sdowell (approved):
/lgtm

e2e/nomostest/wait_for_sync.go
karlkfi: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1406

e2e/testcases/helm_sync_test.go
karlkfi: Moving this into the Helm RSync constructors as part of https://github.com/GoogleContainerTools/kpt-config-sync/pull/1406

PR #1396, opened 2024-08-21
e2e: simplify RSync test setup

karlkfi:
- Combine RootRepos and NonRootRepos to use syncsource.Set.
- Rename NT.New options to be more clear about what they do.
- Move RootSyncID and RepoSyncID to the core package, to allow use in ntopts without import cycle.


sdowell (approved):
/lgtm

PR #1397, opened 2024-08-21
e2e: simplify test sync setup more

karlkfi:
- Combine RootSyncWithGitSource & RepoSyncWithGitSource into SyncWithGitSource that takes an ID. This helps avoid ambiguity about name & namespace argument order.

### DIFF
--- a/e2e/nomostest/ntopts/multi_repo.go
+++ b/e2e/nomostest/ntopts/multi_repo.go
@@ -59,22 +59,11 @@ type MultiRepo struct {
 	RepoSyncPermissions []rbacv1.PolicyRule
 }
 
-// RepoSyncWithGitSource tells the test case that a RepoSync should be applied
-// that points to an empty Git Repository.
+// SyncWithGitSource tells the test case that a RootSync or RepoSync should be
+// applied that points to an empty Git Repository.
 // TODO: Add another option that allows specifying an existing Repository
-func RepoSyncWithGitSource(ns, name string) func(opt *New) {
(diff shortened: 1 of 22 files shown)


e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -59,22 +59,11 @@ type MultiRepo struct {
 	RepoSyncPermissions []rbacv1.PolicyRule
 }
 
-// RepoSyncWithGitSource tells the test case that a RepoSync should be applied
-// that points to an empty Git Repository.
+// SyncWithGitSource tells the test case that a RootSync or RepoSync should be
+// applied that points to an empty Git Repository.
 // TODO: Add another option that allows specifying an existing Repository
-func RepoSyncWithGitSource(ns, name string) func(opt *New) {
+func SyncWithGitSource(id core.ID) func(opt *New) {
```
sdowell, line 65: Hmm this has me feeling like maybe `core.ID` is too flexible - it seems strange that someone could pass in a ConfigMap GK and it would be considered valid here. Should the type for the SyncSources map be stricter (say, using a string enum type for GK)?

karlkfi: :shrug: if it's not valid it'll get ignored and the test would fail. I could add some validation to check it though. I don't feel like making a new tuple type with a bunch of helper functions though. ID already has helper functions and is easily convertible.

sdowell (approved):
/lgtm

PR #1398, opened 2024-08-21
e2e: add options to SyncWithGitSource

karlkfi:
- Replace MultiRepo.SourceFormat with GitSyncSource.SourceFormat.
- Use ntopts.Unstructured on individual SyncWithGitSource now, instead of on the whole parent NT.New. This should make it more obvious that the setting only applies to RootSyncs, and allow more granular configurations.
- Move default SourceFormats for e2e tests into ntopts/multi_repo.go, instead of spread across multiple files.
- Add GitSourceOptions for SyncWithGitSource to further configure individual RSync Git source configs.
- Add RSync Kind validation to setupTestCase, to enforce RootSync and RepoSync.
- Change Repository to use SyncKind, to simplify usage, instead of its own enum type.

### DIFF
--- a/e2e/nomostest/new.go
+++ b/e2e/nomostest/new.go
@@ -67,12 +67,15 @@ func newOptStruct(testName, tmpDir string, ntOptions ...ntopts.Opt) *ntopts.New 
 			SyncSources: syncsource.Set{
 				// All tests default to having a RootSync named root-sync using
 				// a Git repository, unless otherwise specified.
-				DefaultRootSyncID: &syncsource.GitSyncSource{},
+				DefaultRootSyncID: &syncsource.GitSyncSource{
+					// RootSync root-sync defaults to using hierarchy.
+					// Use ntopts.SyncWithGitSource with ntopts.Unstructured to change.
+					SourceFormat: configsync.SourceFormatHierarchy,
+				},
 			},
 			// Default to 1m to keep tests fast.
 			// To override, use WithReconcileTimeout.
 			ReconcileTimeout: ptr.To(1 * time.Minute),
-			SourceFormat:     configsync.SourceFormatHierarchy,
 		},
 	}
 	for _, opt := range ntOptions {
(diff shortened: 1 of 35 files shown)


e2e/nomostest/new.go
```diff
@@ -67,12 +67,15 @@ func newOptStruct(testName, tmpDir string, ntOptions ...ntopts.Opt) *ntopts.New
 			SyncSources: syncsource.Set{
 				// All tests default to having a RootSync named root-sync using
 				// a Git repository, unless otherwise specified.
-				DefaultRootSyncID: &syncsource.GitSyncSource{},
+				DefaultRootSyncID: &syncsource.GitSyncSource{
+					// RootSync root-sync defaults to using hierarchy.
+					// Use ntopts.SyncWithGitSource with ntopts.Unstructured to change.
+					SourceFormat: configsync.SourceFormatHierarchy,
```
sdowell, line 73: Doesn't need to be part of this refactor, but it would be nice to switch the default to unstructured. 99.9% of the time when I'm writing a new test I want to use unstructured mode

karlkfi: Tempting. 

I think it was written this way to reflect the RootSync/RepoSync defaults, which are annoyingly different for reverse compatibility reasons.

It should probably just not specify the format by default, then it would use the actual server-side default. But at least this way it's more obvious to test writers that may not know the server-side defaults well.

e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -59,12 +57,35 @@ type MultiRepo struct {
...
+// Unstructured mutates a GitSyncSource to use the unstructured SourceFormat.
+func Unstructured(source *syncsource.GitSyncSource) {
+	source.SourceFormat = configsync.SourceFormatUnstructured
+}
+
 // SyncWithGitSource tells the test case that a RootSync or RepoSync should be
 // applied that points to an empty Git Repository.
 // TODO: Add another option that allows specifying an existing Repository
-func SyncWithGitSource(id core.ID) func(opt *New) {
+func SyncWithGitSource(id core.ID, gitOpts ...GitSourceOption) func(opt *New) {
```
sdowell, line 71: Do we need the granularity of specifying source format for each git source? I could see it potentially being useful, but most of the time it seems a bit less user friendly to enumerate the option for every RSync used by the test

sdowell (approved):
/lgtm

PR #1399, opened 2024-08-22
update clientgen to use helper script

sdowell:
This updates the clientgen script to use the upstream helper functions from code-generator. This approach is the updated guidance from the upstream repo.

Several APIs were missing the appropriate code generation comments, which are added. This change makes the clientgen result reproducible in CI, so is added as a presubmit check.


scripts/generate-clientset.sh
```diff
@@ -32,91 +32,25 @@
 # Alternatively, make a symlink from the GOPATH package source directory to the
 # repo directory.
 
-set -euox pipefail
+set -euo pipefail
 
-GOPATH="${GOPATH:-$(go env GOPATH)}"
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${SCRIPT_ROOT}"; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
```
nan-yu, line 38: Do we have `code-generator` in the root directory?

scripts/generate-clientset.sh
```diff
@@ -32,91 +32,25 @@
...
-echo "Generation Completed!"
+GOMOD_NAME="$(grep "^module" "${SCRIPT_ROOT}/go.mod" | cut -d' ' -f2)"
+
+kube::codegen::gen_helpers \
+    --boilerplate "${SCRIPT_ROOT}/hack/boilerplate.txt" \
+    "${SCRIPT_ROOT}/pkg/api"
+
+kube::codegen::gen_client \
+    --with-watch \
+    --output-dir "${SCRIPT_ROOT}/pkg/generated" \
```
nan-yu, line 53: What about keeping `output-dir` and `output-pkg` the same as before? The diff will be much smaller.

pkg/generated/clientset/versioned/clientset.go
```diff
@@ -9,57 +9,25 @@ import (
...
-	kptv1alpha1 "kpt.dev/configsync/clientgen/apis/typed/kpt.dev/v1alpha1"
+	configmanagementv1 "kpt.dev/configsync/pkg/generated/clientset/versioned/typed/configmanagement/v1"
 )
 
 type Interface interface {
 	Discovery() discovery.DiscoveryInterface
 	ConfigmanagementV1() configmanagementv1.ConfigmanagementV1Interface
-	ConfigsyncV1alpha1() configsyncv1alpha1.ConfigsyncV1alpha1Interface
-	ConfigsyncV1beta1() configsyncv1beta1.ConfigsyncV1beta1Interface
-	HubV1() hubv1.HubV1Interface
```
nan-yu, line 24: Do you know why are there APIs deleted?

scripts/generate-clientset.sh
sdowell: It's [vendored[(https://github.com/GoogleContainerTools/kpt-config-sync/tree/main/vendor/k8s.io/code-generator)

nan-yu: I meant the fallback path `echo ../code-generator`, which seems to reference a `code-generator` path under the root directory.

scripts/generate-clientset.sh
sdowell: My intent was to go with the default conventions defined by the code-generator library rather than hacking it to stick to our old conventions. The `gen_client` function assumes a subdirectory convention which is different than what we are currently using.

We have
```
output_dir/clientset
├── informer
└── listers

```
Whereas they assume:
```
output_dir
├── clientset
├── informers
└── listers
```

It's mostly just moving files so the diff isn't that large IMO

pkg/generated/clientset/versioned/clientset.go
sdowell: Good catch, looks like it's because those packages were missing the `+genclient` comment

Makefile
```diff
@@ -295,7 +295,7 @@ all: buildenv-dirs
 
 .PHONY: all-local
 # Run tests, cleanup dependencies, and generate CRDs locally
-all-local: test deps configsync-crds
+all-local: test deps clientgen configsync-crds
```
nan-yu, line 298: `clientgen` or `clientgen-in-docker`? Same for `configsync-crds`?

sdowell: `clientgen`. `all-local` is wrapped by `test-presubmit` to run all of these in a docker container.

nan-yu (approved):
/lgtm

PR #1400, opened 2024-08-22
Explicitly set the manifests' license header year

Camila-B:
* This makes sure that the same year is always used when the license header for the manifests is generated, regardless of the current year
* Follow up to #1092 
* b/318417064

### DIFF
--- a/Makefile.reconcilermanager
+++ b/Makefile.reconcilermanager
@@ -42,7 +42,7 @@ configsync-crds: "$(CONTROLLER_GEN)" "$(KUSTOMIZE)" "$(ADDLICENSE)"
 	&& rm ./manifests/configmanagement.gke.io_namespaceconfigs.yaml \
 	&& rm ./manifests/configmanagement.gke.io_repoes.yaml \
 	&& rm ./manifests/configmanagement.gke.io_syncs.yaml \
-	&& "$(ADDLICENSE)" ./manifests
+	&& "$(ADDLICENSE)" -y 2024 ./manifests
 
 "$(CONTROLLER_GEN)": buildenv-dirs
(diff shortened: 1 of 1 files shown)


Makefile.reconcilermanager
```diff
@@ -42,7 +42,7 @@ configsync-crds: "$(CONTROLLER_GEN)" "$(KUSTOMIZE)" "$(ADDLICENSE)"
 	&& rm ./manifests/configmanagement.gke.io_namespaceconfigs.yaml \
 	&& rm ./manifests/configmanagement.gke.io_repoes.yaml \
 	&& rm ./manifests/configmanagement.gke.io_syncs.yaml \
-	&& "$(ADDLICENSE)" ./manifests
+	&& "$(ADDLICENSE)" -y 2024 ./manifests
```
sdowell, line 45: I don't think this really addresses the issue as described - this will just set all CRDs to always have a 2024 license.

- If a new CRD is added in 2025 (for example), it should use 2025 as the year
- I assume we also want previously created CRDs to match the year they were introduced

karlkfi: Agreed. The add-license code always picks the latest year, which is correct for new files. The problem is that they're only marked as new because we're deleting and re-generating them.

We might need a more advanced solution that "fixes" the copyright year after generation if the file already existed beforehand, maybe by using a `git diff --name-only --exit-code path/to/file` in a bash script or something.

karlkfi:
I think we're just gonna have to update the manifest copyright each year, for now.

Alternatively, someone could modify the script to preserve the old generated files and copy the existing license header into the newly generated files to preserve the year.

PR #1401, opened 2024-08-23
test: improve runtime of TestDependencyWithReconciliation

sdowell:
This test is currently one of the longest running tests, with a typical runtime of ~14m30s. The majority of this time is spent on the scenario where Pods should not reconcile, because the reconcileTimeout is very long for the test. Shortening the reconcile timeout for this specific scenario drastically reduces the runtime of the test.

This reduces the runtime from the order of ~14m30s to ~5m.


nan-yu (approved):
/lgtm

PR #1402, opened 2024-08-23
test: Set fake parser FileObjects output to nil

tiffanny29631:
See sample failed run https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/1364/kpt-config-sync-presubmit/1818393399972073472 where an unknown scoped object was returned from Parse and triggers unknown kind error.

Explicitly setting the output to nil and see if it helps.

### DIFF
--- a/pkg/parse/root_test.go
+++ b/pkg/parse/root_test.go
@@ -1267,7 +1267,10 @@ func TestRoot_SourceReconcilerErrorsMetricValidation(t *testing.T) {
 			fakeConfigParser := &fsfake.ConfigParser{
 				Outputs: []fsfake.ParserOutputs{
 					// One Parse call, with errors
-					{Errors: tc.parseErrors},
+					{
+						Errors:      tc.parseErrors,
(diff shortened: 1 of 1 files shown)


pkg/parse/root_test.go
```diff
@@ -1267,7 +1267,10 @@ func TestRoot_SourceReconcilerErrorsMetricValidation(t *testing.T) {
 			fakeConfigParser := &fsfake.ConfigParser{
 				Outputs: []fsfake.ParserOutputs{
 					// One Parse call, with errors
-					{Errors: tc.parseErrors},
+					{
+						Errors:      tc.parseErrors,
+						FileObjects: nil,
```
karlkfi, line 1272: Isn't nil already the default value for a pointer?

mikebz:
Q: is this still relevant or abandoned?

tiffanny29631:
No longer relevant, closing.

PR #1406, opened 2024-08-23
e2e: Add SyncSource.Path

karlkfi:
- Add Path method that wraps Git & OCI dir & Helm chart name.
- Add SetExpectedSyncPath to update the directory expectation.
- Add SetExpectedGitSource, SetExpectedHelmSource, &
  SetExpectedOCISource to update source expectations and replace
  WithSyncDirectoryMap.
- Move all Git repo test defaults to setupTestCase
- Remove obsolete HelmChartVersionShaFn
- Add kinds.LookupID to replace usace of core.IDof using the schema.
- Add OCIImageID to simplify address/url generation
- Add RootSyncObjectGit and RepoSyncObjectGit to handle setting
  defaults that were in multiple places before.
- Add RootSyncObjectOCI & RepoSyncObjectOCI for building RSyncs with
  OCI sources.
- Add RootSyncObjectHelm & RepoSyncObjectHelm for building RSyncs
  with Helm sources.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -393,11 +393,11 @@ func ValidateMultiRepoDeployments(nt *NT) error {
 		var rs *v1beta1.RootSync
 		switch tSource := source.(type) {
 		case *syncsource.GitSyncSource:
-			rs = RootSyncObjectV1Beta1FromRootRepo(nt, id.Name)
+			rs = nt.RootSyncObjectGit(id.Name, tSource.Repository, tSource.Directory, tSource.SourceFormat)
 		case *syncsource.HelmSyncSource:
 			rs = nt.RootSyncObjectHelm(id.Name, tSource.ChartID)
-		// case *syncsource.OCISyncSource:
-		// TODO: setup OCI RootSyncs
+		case *syncsource.OCISyncSource:
+			rs = nt.RootSyncObjectOCI(id.Name, tSource.ImageID, tSource.Directory, tSource.Digest)
 		default:
 			nt.T.Fatalf("Invalid RootSync source %T: %s", source, id.Name)
 		}
@@ -710,8 +710,8 @@ func setupDelegatedControl(nt *NT) {
 	}
 }
 
-// RootSyncObjectV1Alpha1 returns the default RootSync object.
-func RootSyncObjectV1Alpha1(name, repoURL string, sourceFormat configsync.SourceFormat) *v1alpha1.RootSync {
+// RootSyncObjectV1Alpha1Git returns the default RootSync object.
+func RootSyncObjectV1Alpha1Git(name, repoURL string, sourceFormat configsync.SourceFormat) *v1alpha1.RootSync {
 	rs := k8sobjects.RootSyncObjectV1Alpha1(name)
 	rs.Spec.SourceFormat = sourceFormat
 	rs.Spec.SourceType = configsync.GitSource
@@ -757,7 +757,7 @@ func RootSyncObjectV1Alpha1FromRootRepo(nt *NT, name string) *v1alpha1.RootSync 
 	}
 	repoURL := nt.GitProvider.SyncURL(gitSource.Repository.RemoteRepoName)
 	sourceFormat := gitSource.Repository.Format
-	rs := RootSyncObjectV1Alpha1(name, repoURL, sourceFormat)
+	rs := RootSyncObjectV1Alpha1Git(name, repoURL, sourceFormat)
 	if nt.DefaultReconcileTimeout != nil {
 		rs.Spec.SafeOverride().ReconcileTimeout = toMetav1Duration(*nt.DefaultReconcileTimeout)
 	} else if rs.Spec.Override != nil {
@@ -766,8 +766,8 @@ func RootSyncObjectV1Alpha1FromRootRepo(nt *NT, name string) *v1alpha1.RootSync 
 	return rs
 }
 
-// RootSyncObjectV1Beta1 returns the default RootSync object with version v1beta1.
-func RootSyncObjectV1Beta1(name, repoURL string, sourceFormat configsync.SourceFormat) *v1beta1.RootSync {
+// RootSyncObjectV1Beta1Git returns the default RootSync object with version v1beta1.
+func RootSyncObjectV1Beta1Git(name, repoURL string, sourceFormat configsync.SourceFormat) *v1beta1.RootSync {
 	rs := k8sobjects.RootSyncObjectV1Beta1(name)
 	rs.Spec.SourceFormat = sourceFormat
 	rs.Spec.SourceType = configsync.GitSource
@@ -800,8 +800,8 @@ func RootSyncObjectV1Beta1(name, repoURL string, sourceFormat configsync.SourceF
 
 // RootSyncObjectV1Beta1FromRootRepo returns a v1beta1 RootSync object which
 // uses a RootSync repo from nt.SyncSources.
+// Use nt.RootSyncObjectGit is you want to change the source expectation.
 func RootSyncObjectV1Beta1FromRootRepo(nt *NT, name string) *v1beta1.RootSync {
-	// TODO: Allow supplying the Repository, instead of requiring it to already exist
 	id := core.RootSyncID(name)
 	source, found := nt.SyncSources[id]
 	if !found {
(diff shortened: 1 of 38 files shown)


e2e/testcases/workload_identity_test.go
```diff
@@ -598,8 +583,27 @@ func updateRSyncWithGitSourceConfig(nt *nomostest.NT, rs client.Object, repo *gi
...
-	return rsyncValidateMeta{rsRef: rsRef, sha1Func: sc.commitFn, syncDir: sc.dir}
+	id, err := lookupObjectID(rs, nt.Scheme)
+	if err != nil {
+		nt.T.Fatalf("invalid object ID: %v", err)
+	}
+	return rsyncValidateMeta{rsID: id, sha1Func: sc.commitFn, syncDir: sc.dir}
+}
+
+// TODO: promote to helper function, or find existing alternative
+func lookupObjectID(obj client.Object, scheme *runtime.Scheme) (core.ID, error) {
```
nan-yu, line 594: What about using `core.IDOf`? It doesn't verify the scheme, but should work for RSync objects.

e2e/testcases/policy_dir_test.go
```diff
@@ -50,30 +50,24 @@ func TestPolicyDirUnset(t *testing.T) {
...
-	}
+	nt.Must(nt.WatchForAllSyncs())
 
-	nomostest.SetPolicyDir(nt, configsync.RootSyncName, "")
-	err := nt.WatchForAllSyncs(nomostest.WithSyncDirectoryMap(map[types.NamespacedName]string{nomostest.DefaultRootRepoNamespacedName: "."}))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+	nomostest.SetRootSyncGitDir(nt, rootSyncID.Name, "")
+	nomostest.SetExpectedSyncPath(nt, rootSyncID, ".") // TODO: OCI default sync path?
```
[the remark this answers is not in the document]
karlkfi: TIL `controllers.DefaultSyncDir` and `gitproviders.DefaultSyncDir` are different, confusingly so.

e2e/testcases/private_cert_secret_test.go
```diff
@@ -421,20 +421,16 @@ func TestOCICACertSecretRefRootRepo(t *testing.T) {
 	}
 
 	nt.T.Log("Set the RootSync to sync the OCI image without providing a CA cert")
+	rs := k8sobjects.RootSyncObjectV1Beta1(rootSyncID.Name)
 	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"sourceType": "%s", "oci": {"image": "%s", "auth": "none"}, "git": null}}`,
 		configsync.OciSource, imageURL))
-	nt.WaitForRootSyncSourceError(configsync.RootSyncName, status.SourceErrorCode, "tls: failed to verify certificate: x509: certificate signed by unknown authority")
+	nomostest.SetExpectedSyncPath(nt, rootSyncID, ".") // TODO: OCI default sync path?
```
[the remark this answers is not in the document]
karlkfi: IDOf only works if the GK is set on the object, which isn't true of all objects. That convention is unique to Config Sync code and the k8sobjects libs, but it's not shared by upstream k8s code or client-go.

I found a similar lookup function in my fake client code, so I'm going to move it into the core package next to IDof, so it's easy to discover.

At some point in the future, we should require everything to check the scheme unless the object is Unstructured, but that would be a big refactor to handle the possible errors.

e2e/testcases/workload_identity_test.go
karlkfi: IDOf only works if the GK is set on the object, which isn't true of all objects. That convention is unique to Config Sync code and the k8sobjects libs, but it's not shared by upstream k8s code or client-go.

I found a similar lookup function in my fake client code, so I'm going to move it into the core package next to IDof, so it's easy to discover.

At some point in the future, we should require everything to check the scheme unless the object is Unstructured, but that would be a big refactor to handle the possible errors.

e2e/testcases/policy_dir_test.go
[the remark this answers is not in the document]
nan-yu: `controllers.DefaultSyncDir` is used by the reconciler in prod, while `gitproviders.DefaultSyncDir` is only used for testing.

e2e/testcases/workload_identity_test.go
karlkfi: Had to put it in kinds.LookupID to avoid a package dependency cycle

e2e/testcases/policy_dir_test.go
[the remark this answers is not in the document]
karlkfi: Replaced most of these with two in the OCI RSync constructors, tho those only work (for now) if the source was Git or OCI before. I'll add OCISourceSync construction in a followup PR.

This on specifically is an exception, since it's changed after construction, but I changed it to use the constant.

nan-yu (approved):
/lgtm

karlkfi:
last fixes:
- Updated TestOCICACertSecretRefNamespaceRepo to store the Git source and restore it after switching to OCI and back
- Updated TestHelmCACertSecretRefNamespaceRepo to store the Git source and restore it after switching to Helm and back
- Fixed TestMultiSyncs_Unstructured_MixedControl by fixing ResetExpectedGitSync to work with RepoSyncs

karlkfi:
I had planned to save this for another commit, but I had to add a SetExpectedOCIImage to fix TestOCICACertSecretRefNamespaceRepo, which was flipping between Git and OCI and back. There's still some more work to do to use OCISyncSources everywhere we sync from OCI and replace the use of WithRootSha1Func, but I'll save that for a followup PR.

e2e/nomostest/config_sync.go
```diff
@@ -947,17 +931,66 @@ func RepoSyncObjectV1Beta1(nn types.NamespacedName, repoURL string, sourceFormat
...
+func (nt *NT) RootSyncObjectOCI(name string, imageID registryproviders.OCIImageID, syncPath, expectedImageDigest string) *v1beta1.RootSync {
+	id := core.RootSyncID(name)
+	// Register the OCI image for syncing
+	expectedSyncPath := syncPath
+	if syncPath == "" {
+		expectedSyncPath = controllers.DefaultSyncDir
+	}
+	SetExpectedOCISource(nt, id, imageID, expectedImageDigest, expectedSyncPath)
+	rs := k8sobjects.RootSyncObjectV1Beta1(name)
+	rs.Spec.SourceFormat = configsync.SourceFormatUnstructured // TODO: is SourceFormat necessary with OCI?
```
nan-yu, line 989: Technically, OCI source configs can use the hierarchical format. We don't have test coverage for that scenario. Given that unstructured is the recommended format, so we can skip testing OCI with hierarchical format.

e2e/nomostest/config_sync.go
```diff
@@ -974,20 +1007,34 @@ func (nt *NT) RootSyncObjectOCI(name string, image *registryproviders.OCIImage)
...
+	// Register the OCI image for syncing
+	expectedSyncPath := syncPath
+	if syncPath == "" {
+		expectedSyncPath = controllers.DefaultSyncDir
 	}
-
-	rs := RepoSyncObjectV1Beta1FromNonRootRepo(nt, nn)
+	SetExpectedOCISource(nt, id, imageID, expectedImageDigest, expectedSyncPath)
+	rs := k8sobjects.RepoSyncObjectV1Beta1(nn.Namespace, nn.Name)
+	rs.Spec.SourceFormat = configsync.SourceFormatUnstructured // TODO: is SourceFormat necessary with OCI?
```
nan-yu, line 1033: RepoSync always uses unstructured format.

nan-yu (approved):
/lgtm

e2e/nomostest/config_sync.go
karlkfi: We don't need to specify it for RepoSyncs, right?

nan-yu: Correct. The default value is unstructured for RepoSyncs.

e2e/nomostest/config_sync.go
karlkfi: Replaced TODO with a comment in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1410

PR #1407, opened 2024-08-26
chore: rebalance e2e test groups

sdowell:
After recent optimizations to the e2e tests, the runtimes between test groups has become increasingly unbalanced. This change is intended to reduce the variance.


karlkfi (approved):
/lgtm

PR #1409, opened 2024-08-28
test: reduce size of pause deployment

sdowell:
New autopilot clusters with bursting enabled support lower minimum resources requests (50m CPU, 52 MiB memory).

TestStressManyDeployments creates 1000 copies of this deployment, which required an autopilot cluster to scale up to a minimum of 250 vCPU and 500 GiB memory. This requires a large amount of compute resources and can cause quota starvation for other jobs running in the same project. This change reduces the minimum cluster size to 50 vCPU / 50 GiB memory.


e2e/testcases/stress_test.go
```diff
@@ -846,37 +847,60 @@ func validateNumberOfObjectsEquals(nt *nomostest.NT, gvk schema.GroupVersionKind
...
-            memory: 256Mi
-          requests:
-            cpu: 250m
-            memory: 256Mi
-`, name, namespace, name, name, name)
+func pauseDeploymentObject(nt *nomostest.NT, name, namespace string) *appsv1.Deployment {
+	labels := map[string]string{"app": name}
+	deployment := k8sobjects.DeploymentObject(
+		core.Name(name), core.Namespace(namespace), core.Labels(labels))
+	replicas := int32(1)
```
karlkfi, line 854: The pointer library makes this a one liner.

sdowell: done

karlkfi (approved):
/lgtm

PR #1412, opened 2024-08-30
test: fix logic to create ca certs when git=csr

sdowell:
It is possible to run the tests as a combination of git-provider=csr and oci-provider=local. In this case the CA certs for the registry server need to be created in the RepoSync namespace.


nan-yu (approved):
/lgtm

PR #1413, opened 2024-08-30
e2e: Add ReadOnlyRepository

karlkfi:
- ReadOnlyRepository can be used as a substitute for
  ReadWriteRepository when using a public repo that we don't need
  to modify.
- Rename Repository to ReadWriteRepository.
- Add a Repository interface with a SyncURL method for use as a
  Git source.
- Rename SyncSourceGitRepository to SyncSourceGitReadWriteRepository

This PR does not yet use ReadOnlyRepository, but does some of the preliminary refactoring to make it easier to review.

Follow Up: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1410


sdowell (approved):
/lgtm

PR #1414, opened 2024-08-30
e2e: Add SetRSyncTestDefaults

karlkfi:
Move the test defaults into one function, as much as possible.

### DIFF
--- a/e2e/nomostest/config_sync.go
+++ b/e2e/nomostest/config_sync.go
@@ -710,8 +710,40 @@ func setupDelegatedControl(nt *NT) {
 	}
 }
 
-// RootSyncObjectV1Alpha1Git returns the default RootSync object.
-func RootSyncObjectV1Alpha1Git(name, repoURL string, sourceFormat configsync.SourceFormat) *v1alpha1.RootSync {
+// SetRSyncTestDefaults populates defaults for test RSyncs.
+func SetRSyncTestDefaults(nt *NT, obj client.Object) {
+	switch rs := obj.(type) {
+	case *v1alpha1.RootSync:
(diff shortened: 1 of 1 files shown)


e2e/nomostest/config_sync.go
```diff
@@ -710,8 +710,47 @@ func setupDelegatedControl(nt *NT) {
...
 
-// RootSyncObjectV1Alpha1Git returns the default RootSync object.
-func RootSyncObjectV1Alpha1Git(name, repoURL string, sourceFormat configsync.SourceFormat) *v1alpha1.RootSync {
+// SetRSyncTestDefaults populates defaults for test RSyncs.
+func SetRSyncTestDefaults(nt *NT, obj client.Object) {
+	isRepoSync := false
+	// Set type-specific defaults
+	switch rs := obj.(type) {
+	case *v1alpha1.RootSync:
+		if nt.DefaultReconcileTimeout != nil && (rs.Spec.Override == nil || rs.Spec.Override.ReconcileTimeout != nil) {
```
sdowell, line 719: Did you intend to change the behavior with this refactor? It doesn't look like this conditional is equivalent the the previous one.

e.g. if `nt.DefaultReconcileTimeout is nil` and `rs.Spec.Override is not nil`

karlkfi: I did. The old behavior didn't look right. So I updated the behavior to match the name. 

It shouldn't matter though. This func is only ever used on newly constructed objects, so the other case is never being executed.

If someone decides to use it in the future, it will now act as a default, rather than an override.

karlkfi: Oops. Typo on that last bit. Fixed now.

It should set the timeout as a default now only if either Override is nil or Override.ReconcileTimeout is nil.

sdowell (approved):
/lgtm

PR #1418, opened 2024-09-04
ci: remove GCP_PROJECT arg from test-e2e-kind

sdowell:
This argument is already unused when running the tests in kind clusters, but this makes the intent more clear. The e2e tests should/do not require a GCP project to run with kind.


nan-yu (approved):
/lgtm

PR #1419, opened 2024-09-04
test: fix sha1Func for oci workload identity test

sdowell:
This line was removed in a previous change, which leads to a segfault because the sha1Func is nil when accessed.

Fixes breaking change from https://github.com/GoogleContainerTools/kpt-config-sync/pull/1410

For example: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-rapid/1831339095763193856


karlkfi (approved):
/lgtm

PR #1420, opened 2024-09-04
chore: clean up unused e2e-kind make targets

sdowell:
These legacy aliases are no longer in use and can be removed


Camila-B (approved):

nan-yu (approved):
/lgtm

PR #1421, opened 2024-09-04
fix: set expected oci digest for wi tests

sdowell:
This was broken by a previous change which refactored the logic for setting the expected git commit and image digest.

Fixes a breaking change introduced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1410

Example failure: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-regular/1831391944727597056


karlkfi (approved):
/lgtm

PR #1429, opened 2024-09-06
e2e: Update WatchForSync to take a SyncSource

karlkfi:
- Update WatchForSync to take a SyncSource, instead of Sha1Func,
  sha1 Predicate, and SyncPathPredicatePair. These are replaced by
  SyncSource.Path() and SyncSource.Commit() which handle defaulting
  and using the spec value if an expectation was not explicitly set.
- Update SyncSource expectation functions to be test helpers
- Refactor TestComposition to update expecations and use
  WatchForAllSyncs instead of a custom watch wrapper.
- Refactor a few tests that were using WatchForSync directly to
  pass in a SyncSource with expectations updated.

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1430
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1435


nan-yu (approved):
/lgtm

PR #1433, opened 2024-09-17
Support pre-sync annotations

tiffanny29631:


pkg/parse/run.go
```diff
@@ -247,6 +247,14 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
 	// pull the source commit and directory with retries within 5 minutes.
 	gs.Commit, syncDir, gs.Errs = hydrate.SourceCommitAndDirWithRetry(util.SourceRetryBackoff, opts.SourceType, opts.SourceDir, opts.SyncDir, opts.ReconcilerName)
 
+	// Add pre-sync annotations to the object.
+	// If updating the object fails, it's likely due to a signature verification error
+	// from the webhook. In this case, add the error as a source error.
+	err := p.setPreSyncAnnotations(ctx, gs.Commit)
```
nan-yu, line 253: We only need to set the annotations when gs.errs is nil. If there're source errors, it is most likely that `gs.Commit` is empty.

nan-yu, line 253: Suggest renaming to `setSourceAnnotations`. These annotations are used for pre-sync validations for now, but are not limited to pre-sync.

pkg/parse/root.go
```diff
@@ -261,6 +261,18 @@ func setSourceStatusFields(source *v1beta1.SourceStatus, newStatus *SourceStatus
 	source.LastUpdate = newStatus.LastUpdate
 }
 
+func (p *root) setPreSyncAnnotations(ctx context.Context, commit string) status.Error {
+	rs := &v1beta1.RootSync{}
+	if err := p.Client.Get(ctx, rootsync.ObjectKey(p.SyncName), rs); err != nil {
+		return status.APIServerError(err, "failed to get RootSync for parser")
+	}
+	existing := rs.DeepCopy()
+	core.SetAnnotation(rs, metadata.SourceCommitAnnotationKey, commit)
```
nan-yu, line 270: Suggest checking the existing annotations to avoid unnecessary updates if they're the same.

pkg/parse/root.go
```diff
@@ -261,6 +261,18 @@ func setSourceStatusFields(source *v1beta1.SourceStatus, newStatus *SourceStatus
...
+	core.SetAnnotation(rs, metadata.ImageURLAnnotationKey, p.Options.SourceRepo)
+	err := p.Client.Patch(ctx, rs, client.MergeFrom(existing), client.FieldOwner(configsync.FieldManager))
+	return status.SourceError.Wrap(err).Build()
```
nan-yu, line 273: Suggest returning the patch error here and wrap it as a source error where it is called.

The intent is to wrap both the patch error and the get error in line 267 as a source error for consistency.

pkg/parse/run.go
tiffanny29631: Added check

pkg/parse/root.go
tiffanny29631: Added check. Also added similar check in resetSourceAnnotations when removing the annotations when sourceType switched or set to Git.

pkg/parse/root.go

pkg/parse/root.go
```diff
@@ -261,6 +261,38 @@ func setSourceStatusFields(source *v1beta1.SourceStatus, newStatus *SourceStatus
 	source.LastUpdate = newStatus.LastUpdate
 }
 
+func (p *root) setSourceAnnotations(ctx context.Context, commit string) error {
+	rs := &v1beta1.RootSync{}
+	if err := p.Client.Get(ctx, rootsync.ObjectKey(p.SyncName), rs); err != nil {
```
sdowell, line 266: Is the get necessary? Seems like this patch does not require a get

pkg/parse/run.go
```diff
@@ -247,6 +247,21 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
...
+	// from the webhook. In this case, add the error as a source error.
+	if gs.Errs == nil {
+		var err error
+		if opts.SourceType == "git" {
+			err = p.resetSourceAnnotations(ctx)
+		} else {
+			err = p.setSourceAnnotations(ctx, gs.Commit)
+		}
+		if err != nil {
+			gs.Errs = status.Append(gs.Errs, status.SourceError.Wrap(err).Build())
```
sdowell, line 261: There are no test cases which cover this scenario

sdowell (commented):
Please add context to the commit message and PR description

pkg/parse/run.go
tiffanny29631: For now, setting or removing annotations should always succeed since the end-to-end test involving the webhook check has not been implemented yet. In the future, when the webhook service that verifies image signatures is in place, we will be able to simulate error scenarios where patching (setting annotations) fails, and then validate the source errors accordingly.

pkg/parse/root.go
tiffanny29631: It compares the annotations with the new values to avoid unnecessary patching.

pkg/metadata/annotations.go
```diff
@@ -144,6 +144,16 @@ const (
...
 	DynamicNSSelectorEnabledAnnotationKey = configsync.ConfigSyncPrefix + "dynamic-ns-selector-enabled"
+
+	// ImageURLAnnotationKey is the annotation key applied to R*Sync objects
+	// when pre-sync is either disabled or successfully completed.
+	// This annotation stores the repository or image URL.
+	ImageURLAnnotationKey = configsync.ConfigSyncPrefix + "image-url"
+
+	// ImageURLAnnotationKey is the annotation key applied to R*Sync objects
+	// when pre-sync is either disabled or successfully completed.
+	// This annotation stores the Git commit hash or image digest.
```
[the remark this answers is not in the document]
tiffanny29631: Just modified the comment to align with behavior.

pkg/parse/root.go
nan-yu: Alternatively, instead of merging from existing object, we can patch directly with the raw patch data, `client.RawPatch`. With that, we don't need to GET and check for changes.

pkg/parse/run.go
sdowell: IMO that functionality is fundamental to this feature and the feature should not be submitted without test coverage for it. What's the motivation for submitting this feature before validating the webhook user journey?

pkg/parse/run.go
```diff
@@ -247,6 +247,21 @@ func run(ctx context.Context, p Parser, trigger string, state *reconcilerState)
 	// pull the source commit and directory with retries within 5 minutes.
 	gs.Commit, syncDir, gs.Errs = hydrate.SourceCommitAndDirWithRetry(util.SourceRetryBackoff, opts.SourceType, opts.SourceDir, opts.SyncDir, opts.ReconcilerName)
 
+	// Add pre-sync annotations to the object.
+	// If updating the object fails, it's likely due to a signature verification error
+	// from the webhook. In this case, add the error as a source error.
+	if gs.Errs == nil {
+		var err error
+		if opts.SourceType == "git" {
+			err = p.resetSourceAnnotations(ctx)
```
nan-yu, line 256: Suggest merging `resetSourceAnnotations` and `setSourceAnnotations`.

The consolidated `setSourceAnnotations` function should
1. always set the `source-commit` annotation.
2. Set the `image-url` to empty for Git sources, and set to `p.Options.SourceRepo` for OCI and Helm.

nan-yu: For future-proofing, let's rename `image-url` to `source-url`, but still keep it unset for the Git source.

pkg/parse/run.go
tiffanny29631: Fair, I could keep this open until the commit for the full user journey is ready. The original intention was to keep the steps small and make this change only adds the annotations, maybe I can leave the part of returning the SourceError out from this PR for now?

pkg/metadata/annotations.go
```diff
@@ -144,6 +144,16 @@ const (
...
 	// create the reconciler with the Namespace controller in the reconciler container.
 	DynamicNSSelectorEnabledAnnotationKey = configsync.ConfigSyncPrefix + "dynamic-ns-selector-enabled"
+
+	// ImageURLAnnotationKey is the annotation key applied to R*Sync objects
+	// when pre-sync is either disabled or successfully completed.
+	// This annotation stores the source image URL.
+	ImageURLAnnotationKey = configsync.ConfigSyncPrefix + "image-url"
+
+	// SourceCommitAnnotationKey is the annotation key applied to R*Sync objects
+	// when pre-sync is either disabled or successfully completed.
```
nan-yu, line 154: The comment is still confusing. It interprets like the annotation is applied only when pre-sync is disabled or successfully completed.

What about the following?
```
// SourceCommitAnnotationKey stores the Git source commit fetched by git-sync, the OCI image digest fetched by oci-sync, or the Helm chart version fetched by helm-sync.
// It can be used for OCI signature verification.
```

```

File: pkg/metadata/annotations.go [hunk 9]
```diff
@@ -144,6 +144,16 @@ const (
 	// reconciler-manager reads it. If set to true, the reconciler-manager will
 	// create the reconciler with the Namespace controller in the reconciler container.
 	DynamicNSSelectorEnabledAnnotationKey = configsync.ConfigSyncPrefix + "dynamic-ns-selector-enabled"
+
+	// ImageURLAnnotationKey is the annotation key applied to R*Sync objects
```
nan-yu, line 148: Similarly,

```
// SourceURLAnnotationKey stores the source URL where the source commit is fetched from.
// It is only set for OCI and Helm source, and unset for the Git source.
// It can be used for OCI signature verification.
```

pkg/parse/run.go
tiffanny29631: Done with the latest patch, the source-commit is no longer removed when switching back to git. Modified the tests to reflect that.

pkg/metadata/annotations.go
tiffanny29631: Done

pkg/metadata/annotations.go
tiffanny29631: Done

tiffanny29631:
Done in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1452

PR #1435, opened 2024-09-18
e2e: Add watch predicate options

karlkfi:
- Replace predicate argument in Watcher.WatchObject with a new WatchPredicates option. This reduces the number of arguments and avoids callers needing to build an array. They can use varargs.
- Add WithPredicates option for WatchForAllSyncs. This allows customizing the predicates used, instead of needing to skip certain rsyncs with different expectations and validate them manually.


sdowell (approved):
/lgtm

PR #1439, opened 2024-09-25
Only watch ResourceGroups with the correct owner label

### LINKED ISSUE #1297: Do not reconcile unmanaged ResourceGroups
Hi 👋 

We use Kpt and ResourceGroup inventories in addition to Config Sync. They serve different use cases for our clusters. We don't have a need for status to be populated in ResourceGroups and are therefore disabling it using overrides on each of our *Sync resources. ResourceGroups that are created outside of Config Sync, however, are reconciled by default, so we need to explicitly annotate each of them with `configsync.gke.io/status: disabled`. This includes inventories created by Kpt elsewhere, which ends up with a situation where Kpt inventories are reconciled by Config Sync, and that cannot be easily remedied since Kpt doesn't propagate locally defined annotations to the in-cluster inventory: https://github.com/kptdev/kpt/pull/4165.

Would it be reasonable to limit the scope of the ResourceGroup controller to only look at resources labeled with `app.kubernetes.io/managed-by: configmanagement.gke.io`? Alternatively, to provide a way to opt out of the ResourceGroup reconciler altogether?

tomasaschan:
This fixes #1297.

I didn't see any tests for the full NewController function, and decided setting one up was overkill for this change; we should be able to trust here that controller-runtime can properly add and apply a provided predicate, so as long as that predicate works as expected, so whould the watch configuration.

The new tests exhaustively specify the behavior of the new predicate, so that for every type of event, it matches if (and only if) the label `app.kubernetes.io/managed-by` is set to `configmanagement.gke.io`; this is what Config Sync already does to all objects it creates.

### DIFF
--- a/pkg/resourcegroup/controllers/root/root_controller.go
+++ b/pkg/resourcegroup/controllers/root/root_controller.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
 	"kpt.dev/configsync/pkg/api/kpt.dev/v1alpha1"
+	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/resourcegroup/controllers/handler"
 	"kpt.dev/configsync/pkg/resourcegroup/controllers/resourcegroup"
 	"kpt.dev/configsync/pkg/resourcegroup/controllers/resourcemap"
@@ -211,6 +212,8 @@ func NewController(mgr manager.Manager, channel chan event.GenericEvent,
 		WithEventFilter(ResourceGroupPredicate{}).
 		// skip the Generic events
 		WithEventFilter(NoGenericEventPredicate{}).
+		// only reconcile resource groups owned by Config Sync
+		WithEventFilter(OwnedByConfigSyncPredicate{}).
 		Watches(&apiextensionsv1.CustomResourceDefinition{}, &handler.CRDEventHandler{
 			Mapping: resMap,
 			Channel: channel,
@@ -295,4 +298,25 @@ func isConditionTrue(cType v1alpha1.ConditionType, conditions []v1alpha1.Conditi
 	}
 
 	return false
+}
+
+type OwnedByConfigSyncPredicate struct{}
+
+func (OwnedByConfigSyncPredicate) Create(e event.CreateEvent) bool {
+	return isOwnedByConfigSync(e.Object)
+}
+func (OwnedByConfigSyncPredicate) Delete(e event.DeleteEvent) bool {
+	return isOwnedByConfigSync(e.Object)
+}
+func (OwnedByConfigSyncPredicate) Update(e event.UpdateEvent) bool {
+	return isOwnedByConfigSync(e.ObjectOld)
+}
+func (OwnedByConfigSyncPredicate) Generic(e event.GenericEvent) bool {
+	return isOwnedByConfigSync(e.Object)
+}
+
+func isOwnedByConfigSync(o client.Object) bool {
+	labels := o.GetLabels()
+	owner, ok := labels[metadata.ManagedByKey]
+	return ok && owner == metadata.ManagedByValue
 }
--- a/pkg/resourcegroup/controllers/root/root_controller.go
+++ b/pkg/resourcegroup/controllers/root/root_controller.go
@@ -300,17 +300,25 @@ func isConditionTrue(cType v1alpha1.ConditionType, conditions []v1alpha1.Conditi
 	return false
 }
 
+// OwnedByConfigSyncPredicate filters events for objects that have the label "app.kubernetes.io/managed-by=configmanagement.gke.io"
 type OwnedByConfigSyncPredicate struct{}
 
+// Create implements predicate.Predicate
 func (OwnedByConfigSyncPredicate) Create(e event.CreateEvent) bool {
 	return isOwnedByConfigSync(e.Object)
 }
+
+// Delete implements predicate.Predicate
 func (OwnedByConfigSyncPredicate) Delete(e event.DeleteEvent) bool {
 	return isOwnedByConfigSync(e.Object)
 }
+
+// Update implements predicate.Predicate
 func (OwnedByConfigSyncPredicate) Update(e event.UpdateEvent) bool {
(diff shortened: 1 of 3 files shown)


pkg/resourcegroup/controllers/root/root_controller.go
```diff
@@ -296,3 +299,24 @@ func isConditionTrue(cType v1alpha1.ConditionType, conditions []v1alpha1.Conditi
...
+type OwnedByConfigSyncPredicate struct{}
+
+func (OwnedByConfigSyncPredicate) Create(e event.CreateEvent) bool {
+	return isOwnedByConfigSync(e.Object)
+}
+func (OwnedByConfigSyncPredicate) Delete(e event.DeleteEvent) bool {
+	return isOwnedByConfigSync(e.Object)
+}
+func (OwnedByConfigSyncPredicate) Update(e event.UpdateEvent) bool {
+	return isOwnedByConfigSync(e.ObjectOld)
```
nan-yu, line 312: We would also need to check if the new object is owned by Config Sync.
`return isOwnedByConfigSync(e.ObjectOld) || isOwnedByConfigSync(e.ObjectNew)`

pkg/resourcegroup/controllers/root/root_controller.go
```diff
@@ -211,6 +212,8 @@ func NewController(mgr manager.Manager, channel chan event.GenericEvent,
 		WithEventFilter(ResourceGroupPredicate{}).
 		// skip the Generic events
 		WithEventFilter(NoGenericEventPredicate{}).
+		// only reconcile resource groups owned by Config Sync
+		WithEventFilter(OwnedByConfigSyncPredicate{}).
```
sdowell, line 216: Another option that may be worth looking into is setting up a label filter on the `controller-runtime` client.

The `EventFilter` approach solves the functional requirement of ignoring resource groups not owned by Config Sync, but I assume isn't as efficient compared to filtering resources in the client.

pkg/resourcegroup/controllers/root/root_controller.go
tomasaschan: Thanks! I was unsure about what the best policy was here; fixed in 60bde30.

pkg/resourcegroup/controllers/root/root_controller.go
tomasaschan: IIUC, the client will be configured so that the watch is filtered with the predicates specified like this, as well as any predicates specified to `For` (as functional options); see https://github.com/kubernetes-sigs/controller-runtime/blob/v0.19.0/pkg/builder/controller.go#L321-L323

I don't see a way to specify label filters here without significantly reconfiguring how the controller is initialized. Are you suggesting adding a cache configuration to filter `ResourceGroup`s by labels here? https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/resourcegroup/controllers/runner/run.go#L92-L102 I'm not sure if that's a good idea - I worry that is pretty untested code, so if it's done wrong (now or in the future) it'd risk silently breaking stuff for users...

tomasaschan:
/retest

I can't reproduce this error locally.

tomasaschan:
I'm flailing here. Giving up for today... 😩

pkg/resourcegroup/controllers/root/root_controller_test.go
```diff
@@ -196,3 +196,93 @@ func TestRootReconciler(t *testing.T) {
 	time.Sleep(time.Second)
 	assert.Equal(t, "group0", e.Object.GetName())
 }
+
+func TestOnwedByConfigSyncPredicate(t *testing.T) {
```
sdowell, line 200: nit: typo `TestOnwedByConfigSyncPredicate` -> `TestOwnedByConfigSyncPredicate`

sdowell:
> I'm flailing here. Giving up for today... 😩

It looks like the new unit tests you added are failing, I was able to reproduce the failures locally on my workstation as well

tomasaschan:
@sdowell Looks like I made it work! Care to review again?

pkg/resourcegroup/controllers/root/root_controller_test.go
```diff
@@ -196,3 +196,98 @@ func TestRootReconciler(t *testing.T) {
...
+		)
+	}
+
+	generic := func(o client.Object) bool {
+		return OwnedByConfigSyncPredicate{}.Generic(
+			event.TypedGenericEvent[client.Object]{Object: o},
+		)
+	}
+
+	testCases := []testCase{
```
sdowell, line 261: Could also use some test cases for GVK's that are not resource group, since that is part of the predicate's checks

pkg/resourcegroup/controllers/root/root_controller_test.go
```diff
@@ -196,3 +196,98 @@ func TestRootReconciler(t *testing.T) {
...
+		)
+	}
+	update := func(o client.Object, n client.Object) bool {
+		return OwnedByConfigSyncPredicate{}.Update(
+			event.TypedUpdateEvent[client.Object]{ObjectOld: o, ObjectNew: n},
+		)
+	}
+
+	// not allowed to redefine delete, so we use a different name
+	yeet := func(o client.Object) bool {
```
sdowell, line 249: Love this :)

pkg/resourcegroup/controllers/root/root_controller.go
sdowell: Yes that is what I was suggesting, potentially using label selectors on the caching client to reduce the scope and memory usage of the cache. Agreed that it is risky and prone to gotchas though, so we can punt on it for now. This is a good iterative improvement.

pkg/resourcegroup/controllers/root/root_controller_test.go
tomasaschan: That's implicitly tested elsewhere (which is how I realized that the condition is needed) but added some in dc3459d0.

tomasaschan:

It seems like there might be a timing issue in one of the tests; anyway, the failure is a timeout and I don't understand how it's related, so I'm going to rerun while I do some local refactoring.

sdowell (approved):

sdowell:
Thank you for your contribution!

PR #1440, opened 2024-09-25
fix: Remove ApplySet part-of label on abandon

karlkfi:
This change removes the `applyset.kubernetes.io/part-of` label from managed objects when they are abandoned.

Supported Use Cases:
- Unmanage object
  - `configmanagement.gke.io/managed: disabled`
  - `client.lifecycle.config.k8s.io/deletion: detach`
  - `cli-utils.sigs.k8s.io/on-remove: keep`
  - The Config Sync Applier/Destroyer calls RemoveConfigSyncMetadata if object HasConfigSyncMetadata when a Delete Skipped event is received (deletion detach or keep on remove).
  - The Config Sync Remediator calls RemoveConfigSyncMetadata when an Abandon operation is requested (managed disabled or deletion detach).
- Uninstall package
  - `configsync.gke.io/deletion-propagation-policy: Foreground`
  - This use case invokes the Destroyer, which handles the unmanage annotations the same way the Applier does.

Unsupported use cases:
- Orphan Package
  - `configsync.gke.io/deletion-propagation-policy: Orphan`
  - Unsupported because we don't have a finalizer handling this case today. Maybe in the future.

Fixes: b/355534413

This change also fixes (b/256043590) the expectations in `TestReconcilerFinalizer_MultiLevelMixed` to expect removal of metadata on abandoned objects, which was fixed in an applier refactor a few months ago.

### DIFF
--- a/e2e/testcases/lifecycle_directives_test.go
+++ b/e2e/testcases/lifecycle_directives_test.go
@@ -26,9 +26,11 @@ import (
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/core/k8sobjects"
 	"kpt.dev/configsync/pkg/kinds"
+	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/syncer/differ"
 	"kpt.dev/configsync/pkg/util"
 	"sigs.k8s.io/cli-utils/pkg/common"
@@ -42,10 +44,7 @@ func TestPreventDeletionNamespace(t *testing.T) {
 	rootSyncGitRepo := nt.SyncSourceGitReadWriteRepository(nomostest.DefaultRootSyncID)
 
 	// Ensure the Namespace doesn't already exist.
-	err := nt.ValidateNotFound("shipping", "", &corev1.Namespace{})
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+	nt.Must(nt.ValidateNotFound("shipping", "", &corev1.Namespace{}))
 
 	role := k8sobjects.RoleObject(core.Name("shipping-admin"))
 	role.Rules = []rbacv1.PolicyRule{{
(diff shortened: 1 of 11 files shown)


e2e.log
sdowell, line 1: Did you mean to include this file?

e2e/testcases/lifecycle_directives_test.go
```diff
@@ -61,23 +60,22 @@ func TestPreventDeletionNamespace(t *testing.T) {
...
+	applySetID := core.GetLabel(rsObj, metadata.ApplySetParentIDLabel)
+
+	nt.Must(nt.Validate("shipping", "", &corev1.Namespace{},
 		testpredicates.NotPendingDeletion,
 		testpredicates.HasAnnotation(common.LifecycleDeleteAnnotation, common.PreventDeletion),
-		testpredicates.HasAllNomosMetadata())
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+		testpredicates.HasAllNomosMetadata(),
```
sdowell, line 70: Perhaps `HasAllNomosMetadata` should include a check for the apply set label, and `NoConfigSyncMetadata` should include a check that the label is missing.

Is there a case where an object should have other CS metadata but not the apply set label?

karlkfi: The ApplySet label isn't technically "Config Sync metadata", and I didn't want to have to pass the ApplySetID to the Config Sync metadata methods, so I left it out and gave it its own method in the impl and the tests.

One critical difference is that HasConfigSyncMetadata does not detect the ApplySet label, partly because in order to do so it would need to check the label value, not just the key. So the ApplySet label is only removed if other CS metadata is present AND the ApplySet ID matches.

karlkfi: I considered adding a HasApplySetPartOfLabel predicate, but it seemed pretty superfluous given that the HasLabel predicate is the same number of code lines.

e2e.log
karlkfi: lol no. oops. removed.

e2e/testcases/lifecycle_directives_test.go
```diff
@@ -61,23 +60,22 @@ func TestPreventDeletionNamespace(t *testing.T) {
 	nt.Must(rootSyncGitRepo.CommitAndPush("declare Namespace with prevent deletion lifecycle annotation"))
 	nt.Must(nt.WatchForAllSyncs())
 
-	err = nt.Validate("shipping", "", &corev1.Namespace{},
+	rsObj := &v1beta1.RootSync{}
+	nt.Must(nt.KubeClient.Get(rootSyncNN.Name, rootSyncNN.Namespace, rsObj))
+	applySetID := core.GetLabel(rsObj, metadata.ApplySetParentIDLabel)
```
nan-yu, line 65: Instead of getting applySetID from a GET request, can we use `applyset.IDFromSync`?

karlkfi: Yes, but getting it from the RSync is slightly better for a few reasons:
- It validates that the label is set on the RSync
- It validates the ID matches on the RSync and managed objects
- It's more like what customers would do if they try to build automation around ApplySet labels.

The extra API call shouldn't significantly slow down testing.

nan-yu (approved):
/lgtm

PR #1441, opened 2024-09-26
fix: remediator missing custom resource events

karlkfi:
Prior to this change, the remediator watches were only being started
for new custom resources after the apply attempt had fully completed.
This left some time after the object was applied that the remediator
could miss events made by third-parties. Normally, this would be fine,
because the remediator would revert any change after the watch was
started. But if a DELETE event was missed, the object wouldn't be
recreated until the next apply attempt.

This change adds a CRD Controller to the remediator that watches CRDs
and executes any registered handlers when the CRD is established,
unestablished, or deleted. The remediator now registers CRD handlers
for each resource type it watches, starting watchers as soon as
possible, without waiting for the next apply attempt.

This change also adds a ClusterRole and ClusterRoleBinding specifically
for RepoSync reconcilers, to allow watching of CRDs. RootSync
reconcilers now also watch CRDs, but they already have a CR & CRD.

Fixes: b/355532135

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1445
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1446
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1447

### DIFF
--- a/pkg/reconcilermanager/controllers/crd_controller.go
+++ b/pkg/reconcilermanager/controllers/crd_controller.go
@@ -21,125 +21,154 @@ import (
 
 	"github.com/go-logr/logr"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	controllerruntime "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-// CRDHandler is called by the CRDReconciler to handle establishment of a CRD.
-type CRDHandler func() error
+// CRDReconcileFunc is called by the CRDMetaController to handle CRD updates.
+type CRDReconcileFunc func(context.Context, *apiextensionsv1.CustomResourceDefinition) error
 
-var _ reconcile.Reconciler = &CRDReconciler{}
-
-// CRDReconciler watches CRDs and calls handlers once they are established.
-type CRDReconciler struct {
-	loggingController
-
-	registerLock sync.Mutex
-	handlers     map[string]CRDHandler
-	handledCRDs  map[string]struct{}
+// CRDController keeps track of CRDReconcileFuncs and calls them when the
+// CRD changes. Only one reconciler is allowed per GroupKind.
+type CRDController struct {
+	lock        sync.RWMutex
+	reconcilers map[schema.GroupKind]CRDReconcileFunc
 }
 
-// NewCRDReconciler constructs a new CRDReconciler.
-func NewCRDReconciler(log logr.Logger) *CRDReconciler {
-	return &CRDReconciler{
+// SetReconciler sets the reconciler for the specified CRD.
+// The reconciler will be called when the CRD becomes established.
+// If the reconciler errors, it will be retried with backoff until success.
+// A new reconciler will replace any old reconciler set with the same GroupKind.
+func (s *CRDController) SetReconciler(gk schema.GroupKind, crdHandler CRDReconcileFunc) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if s.reconcilers == nil {
+		s.reconcilers = make(map[schema.GroupKind]CRDReconcileFunc)
+	}
+	s.reconcilers[gk] = crdHandler
+}
+
+// DeleteReconciler removes the reconciler for the specified CRD.
+func (s *CRDController) DeleteReconciler(gk schema.GroupKind) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if s.reconcilers != nil {
+		delete(s.reconcilers, gk)
+	}
+}
+
+// Reconcile calls the CRDReconcileFunc registered for this CRD by GroupKind.
+func (s *CRDController) Reconcile(ctx context.Context, gk schema.GroupKind, crd *apiextensionsv1.CustomResourceDefinition) error {
+	// Let getReconciler handle locking to prevent deadlock in case the
+	// reconciler calls SetReconciler/DeleteReconciler.
+	reconciler := s.getReconciler(gk)
+	if reconciler == nil {
+		// No reconciler for this CRD
+		return nil
+	}
+	if err := reconciler(ctx, crd); err != nil {
+		return fmt.Errorf("reconciling CRD: %s: %w", gk, err)
+	}
+	return nil
+}
+
+func (s *CRDController) getReconciler(gk schema.GroupKind) CRDReconcileFunc {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+
+	reconciler, found := s.reconcilers[gk]
+	if !found {
+		return nil
+	}
+	return reconciler
+}
+
+// CRDMetaController watches CRDs and delegates reconciliation to a CRDControllerManager.
+type CRDMetaController struct {
+	loggingController
+	cache             cache.Cache
+	delegate          *CRDController
+	observedResources map[schema.GroupResource]schema.GroupKind
+}
+
+var _ reconcile.Reconciler = &CRDMetaController{}
+
+// NewCRDMetaController constructs a new CRDMetaController.
+func NewCRDMetaController(delegate *CRDController, cache cache.Cache, log logr.Logger) *CRDMetaController {
+	return &CRDMetaController{
 		loggingController: loggingController{
 			log: log,
 		},
-		handlers:    make(map[string]CRDHandler),
-		handledCRDs: make(map[string]struct{}),
+		cache:             cache,
+		delegate:          delegate,
+		observedResources: make(map[schema.GroupResource]schema.GroupKind),
 	}
 }
 
-// SetCRDHandler adds an handler for the specified CRD.
-// The handler will be called when the CRD becomes established.
-// If the handler errors, it will be retried with backoff until success.
-// One the handler succeeds, it will not be called again, unless SetCRDHandler
-// is called again.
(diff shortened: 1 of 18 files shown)


pkg/reconcilermanager/controllers/rootsync_controller.go
```diff
@@ -416,6 +418,14 @@ func (r *RootSyncReconciler) deleteManagedObjects(ctx context.Context, reconcile
 
 // Register RootSync controller with reconciler-manager.
 func (r *RootSyncReconciler) Register(mgr controllerruntime.Manager, watchFleetMembership bool) error {
+	r.lock.Lock()
+	defer r.lock.Unlock()
```
nan-yu, line 422: Do we need the lock? The `Register` function is only invoked once and won't be re-registered.

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -21,125 +21,153 @@ import (
...
+	handler := s.getHandler(gk)
+	if handler == nil {
 		// No handler for this CRD
-		return reconcile.Result{}, nil
+		return nil
 	}
-	if _, handled := r.handledCRDs[crdName]; handled {
-		// Already handled
-		return reconcile.Result{}, nil
+	// Handler must not be called inside the lock or it could deadlock.
```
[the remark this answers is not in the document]
karlkfi: The comment is why the handler is being gotten with a getter.

pkg/reconcilermanager/controllers/rootsync_controller.go
karlkfi: The lock protects the controller from read/write race condition. But the way we're calling Register today it probably won't ever be called in parallel.

karlkfi: We don't technically need it today, it's just good defensive practice to protect variables that could be used in parallel. 

Locking is cheap. Lock contention is what's expensive.

pkg/reconcilermanager/controllers/crd_controller.go
```diff
@@ -21,125 +21,153 @@ import (
...
-			log: log,
-		},
-		handlers:    make(map[string]CRDHandler),
-		handledCRDs: make(map[string]struct{}),
-	}
-}
+// SetReconciler sets the reconciler for the specified CRD.
+// The reconciler will be called when the CRD becomes established.
+// If the reconciler errors, it will be retried with backoff until success.
+// New reconcilers with replace old reconciler for the same GroupKind.
```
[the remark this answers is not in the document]
karlkfi: fixed.

pkg/reconcilermanager/controllers/crd_controller.go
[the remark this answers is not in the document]
karlkfi: Replaced the comment with one that makes the pattern a little clearer.

pkg/remediator/watch/manager.go
```diff
@@ -327,3 +340,23 @@ func (m *Manager) stopWatcher(gvk schema.GroupVersionKind) {
...
 	delete(m.watcherMap, gvk)
 }
+
+func (m *Manager) startWatchingCRD(gvk schema.GroupVersionKind, commit string) {
+	m.crdController.SetReconciler(gvk.GroupKind(), func(ctx context.Context, crd *apiextensionsv1.CustomResourceDefinition) error {
+		gvkMap := map[schema.GroupVersionKind]struct{}{
+			gvk: {},
+		}
+		if customresource.IsEstablished(crd) {
+			return m.AddWatches(ctx, gvkMap, commit)
```
nan-yu, line 350: When the CRD controller detects a change for a specific CRD, instead of calling`AddWatches` to iterate through all gvks, should we just start the watch for that gvk if it hasn't been started yet?

karlkfi: Occurs to me that the Mapper probably needs to be reset though. And it looks to me like the mapper used by the remediator isn't the same as the one used by the reconciler. It gets constructed from DynamicListerWatcherFactoryFromConfig.

I don't actually know if the new discovery API handles evicting stale resource records from the mapper client or if we need to reset it manually, like we do for the reconciler mapper.

karlkfi: Actually, it looks like the NewDynamicRESTMapper used by the remediator...
1. Doesn't have a Reset method
2. Automatically calls discoveryClient.ServerResourcesForGroupVersion when a NoMatchError is received, updating to include newly established resources automatically.
3. Lazily discovers resources on-demand

However, there are two caveats:
1. There's no removal of resources from the cache. So RESTMapping won't error after a CRD is deleted.
2. It doesn't use the new aggregated discovery API,  because it was only added upstream 2 weeks ago: https://github.com/kubernetes-sigs/controller-runtime/pull/2901.

So...
- We don't need to reset the mapper when a CRD is newly established.
- There's no obvious way to reset the mapper when a CRD becomes unestablished or deleted. So instead of skipping the next watch start after recreation, the watch start may fail after the RESTMapper succeeds, causing unnecessary errors in the CRD Controller and triggering retry with backoff, until the CRD is recreated (maybe never).

nan-yu: I see. The gvkMap is reset to only have one GVK in the crdHandler.

> There's no obvious way to reset the mapper when a CRD becomes unestablished or deleted. So instead of skipping the next watch start after recreation, the watch start may fail after the RESTMapper succeeds, causing unnecessary errors in the CRD Controller and triggering retry with backoff, until the CRD is recreated (maybe never).

I don't quite follow this part. When a CRD becomes unestablished or deleted, won't the handler skips calling `m.AddWatches`? If so, there won't be any errors in the CRD controller. 

Per `RESTMapping won't error after a CRD is deleted`, it sounds like started watch will continue to work without errors.

karlkfi: > I don't quite follow this part. When a CRD becomes unestablished or deleted, won't the handler skips calling m.AddWatches? If so, there won't be any errors in the CRD controller.

You're right. The delete/unestablished event doesn't do anything. It used to, while I was prototyping, but I had to disable it because it was causing CR delete events to be missed, because the watch was being stopped prematurely. And it won't error on an established event, because even though the RESTMapper wasn't reset, the resources is back now, so the watch should succeed.

The problem is really just if UpdateWatches or AddWatches is called by the Updater/Applier after a CRD has been deleted. The AddWatches/UpdateWatches will check the RESTMapper, which thinks the resource still exists, and then the startWatch will fail because it actually doesn't. Then that will trigger the applier run loop retry with backoff.

So we need to somehow reset the RESTMapper otherwise the Updater might fail before it has a chance to reapply a managed CRD that was deleted by a third party.

nan-yu: If the deleted CRD is managed by RSync, the remediator should be able to re-create the CRD between Applier retries. That will eventually fix the watch start error when the CRD is established.

Alternatively, we can GET the CRD before checking the RESTMapper. If the CRD is unestablished or deleted, we can skip starting the watcher.
This GET call can be omitted if the CRD controller stores the CRD's status in memory.

karlkfi: > If the deleted CRD is managed by RSync, the remediator should be able to re-create the CRD between Applier retries. That will eventually fix the watch start error when the CRD is established.

The main edge case this PR is trying to solve is missed delete events for newly created CRs before the Applier succeeds. Eventually starting the watch isn't good enough, because if the CR has already been deleted before the watch starts, the remediator won't know to recreate it. Worst case, if the applier succeeded, but the CR was deleted by a 3rd party before UpdateWatches is called, the CR wouldn't be recreated until the next hourly re-sync.

Plus, the remediator is paused between Applier retries. It only unpauses after the Applier succeeds, because the remediator doesn't follow dependency ordering and could break things worse by trying to remediate after an unsuccessful apply.

> Alternatively, we can GET the CRD before checking the RESTMapper. If the CRD is unestablished or deleted, we can skip starting the watcher. This GET call can be omitted if the CRD controller stores the CRD's status in memory.

The CRDController already gets the CRD from the cache used by the controller-manager informer. And the watch.Manager already starts watches if the CRD is established. The gotcha is that if the RESTMapper cache is not invalidated after a CRD is deleted, then it may contain invalid mappings. Then when the CRD is recreated, the mappings may have changed and the RESTMapper won't know, because it didn't get a NoMatchFound so it didn't refresh from the server. So it needs to auto-invalidate for auto-discovery to be reliable.

So I've added a new ReplaceOnResetRESTMapper that can be used to replace the RESTMapper when Reset is called. Then I added some code to the watch.Manager to handle calling Reset on the mapper when a CRD is established or unestablished, if the mapper still knows about the deleted resource or doesn't know about a new resource.

karlkfi:
For posterity, since inline comment threads can get lost...

Unfortunately, neither the DynamicRESTMapper in controller-runtime nor the DeferredDiscoveryRESTMapper/CachedDiscoveryClient in client-go implement auto-invalidation of resources, unless aggregated discovery is enabled on the server. While aggregated discovery is beta in 1.27+ and thus enabled by default, that doesn't guarantee that it will be enabled on non-GKE clusters.

1. DynamicRESTMapper only recently added support for aggregated discovery, but it's in a newer controller-runtime than we're using today. While it does handle auto-discovery when aggregated discovery is disabled, it does not handle auto-invalidation.
2. CachedDiscoveryClient does support with aggregated discovery, but when disabled it doesn't handle auto-discovery or auto-invalidation. It just has a cache TTL. 

So I've added a new ReplaceOnResetRESTMapper that can be used to replace the RESTMapper when Reset is called. Then I added some code to the watch.Manager to handle calling Reset on the mapper when a CRD is established or unestablished, if the mapper still knows about the deleted resource or doesn't know about a new resource. This handles auto-discovery and auto-invalidation of resources in the RESTMapper, but it's a relatively inefficient.

Hopefully at some point in the future we can make aggregated discovery a requirement and use a simplified DiscoveryClient to handle both auto-discovery and auto-invalidation.

pkg/remediator/watch/manager.go
```diff
@@ -327,3 +341,65 @@ func (m *Manager) stopWatcher(gvk schema.GroupVersionKind) {
 	w.Stop()
 	delete(m.watcherMap, gvk)
 }
+
+func (m *Manager) startWatchingCRD(gvk schema.GroupVersionKind, commit string) {
+	m.crdController.SetReconciler(gvk.GroupKind(), func(ctx context.Context, crd *apiextensionsv1.CustomResourceDefinition) error {
+		if customresource.IsEstablished(crd) {
+			if err := discoverResourceForKind(m.mapper, gvk); err != nil {
+				// Trigger retry by controller-manager
```
nan-yu, line 349: We should log the error

karlkfi: The error is returned to the CRDController which returns it from the Reconcile call, which triggers the ControllerManager to log it and schedule a retry (with backoff I think). So we don't need to log it here.

nan-yu (approved):
/lgtm

PR #1446, opened 2024-10-07
chore: cleanup some logs

karlkfi:
- Simplify & fix applier/destroyer logs on success & failure
- Add the component in front of a few logs to make it clear where in the code the log came from, to simplify debugging.
- Log resource conflicts as warnings with explainations of how remediation will retry.
- Only log the keys from the GVK maps in the remediator manager.
- Fix remediator shutdown log to go to info instead of warning.

### DIFF
--- a/pkg/parse/updater.go
+++ b/pkg/parse/updater.go
@@ -209,23 +209,20 @@ func (u *Updater) apply(ctx context.Context, objs []client.Object, commit string
 			}
 		}
 	}
-	klog.V(1).Info("Applier starting...")
+	klog.Info("Applier starting...")
 	start := time.Now()
 	u.SyncErrorCache.ResetApplyErrors()
 	objStatusMap, syncStats := u.Applier.Apply(ctx, eventHandler, objs)
-	if syncStats.Empty() {
-		klog.V(4).Info("Applier made no new progress")
-	} else {
+	if !syncStats.Empty() {
 		klog.Infof("Applier made new progress: %s", syncStats.String())
 		objStatusMap.Log(klog.V(0))
 	}
 	metrics.RecordApplyDuration(ctx, metrics.StatusTagKey(err), commit, start)
 	if err != nil {
-		klog.Warningf("Failed to apply declared resources: %v", err)
+		klog.Warningf("Applier failed: %v", err)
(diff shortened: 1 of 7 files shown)


pkg/parse/updater.go
```diff
@@ -213,19 +213,16 @@ func (u *Updater) apply(ctx context.Context, objs []client.Object, commit string
...
 	}
 	metrics.RecordApplyDuration(ctx, metrics.StatusTagKey(err), commit, start)
 	if err != nil {
-		klog.Warningf("Failed to apply declared resources: %v", err)
+		klog.Warningf("Applier failed: %v", err)
 		return err
 	}
-	klog.V(4).Info("Apply completed without error: all resources are up to date.")
-	klog.V(3).Info("Applier stopped")
+	klog.V(1).Info("Applier succeeded")
```
nan-yu, line 225: We can probably log at the default level, 0.

pkg/reconciler/finalizer/base_finalizer.go
```diff
@@ -42,20 +42,17 @@ func (bf *baseFinalizer) destroy(ctx context.Context) status.MultiError {
...
 	// We don't have the commit here, so we can't send the apply metric.
 	// metrics.RecordApplyDuration(ctx, metrics.StatusTagKey(errs), commit, start)
 	if err != nil {
-		klog.Warningf("Failed to destroy declared resources: %v", err)
+		klog.Warningf("Destroyer failed: %v", err)
 		return err
 	}
-	klog.V(4).Info("Destroyer completed without error: all resources are deleted.")
-	klog.V(3).Info("Applier stopped")
+	klog.V(3).Info("Destroyer succeeded")
```
nan-yu, line 56: It should be logged at the same level as `Applier succeeded`.

pkg/parse/updater.go
karlkfi: done

pkg/reconciler/finalizer/base_finalizer.go
karlkfi: done

nan-yu (approved):
/lgtm

PR #1447, opened 2024-10-07
chore: remove ClearManagementConflictsWithKind

karlkfi:
The watch.Runnable interface doesn't need to know how to clear management conflicts, because the watch.Manager already has access to the conflict.Handler. So just clear the conflicts in AddWatches, where it also deals with other conflicts. AddWatches also knows that the watch is no longer needed, whereas stopWatch only knows that it needs to be stopped.


nan-yu (approved):
/lgtm

PR #1450, opened 2024-10-09
chore: share mapper between controllers

karlkfi:
Use a common ResettableRESTMapper, shared between controllers and reset automatically by the CRDMetaController when a CRD is updated and the new state is different than what known by the RESTMapper.

Followup after https://github.com/GoogleContainerTools/kpt-config-sync/pull/1441


nan-yu (approved):
/lgtm

PR #1451, opened 2024-10-14
core: fix the retry timer

nan-yu:
Currently, the RetrySyncPublisher publishes retry events with backoff regardless if retry needs to be executed.

If retry is not needed, the RetrySyncPublisher should publish events periodically (every 1 second) without backoff to make sure retry events can be fired in the future.

### DIFF
--- a/pkg/parse/events/publisher.go
+++ b/pkg/parse/events/publisher.go
@@ -129,6 +129,8 @@ type RetrySyncPublisher struct {
 	currentBackoff wait.Backoff
 	retryLimit     int
 	timer          clock.Timer
+
+	runAttempted bool
 }
 
 // Type of events produced by this publisher.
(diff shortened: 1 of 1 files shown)


karlkfi:
What problem is this solving exactly?

I hope we don't actually need a new flag. I think the ResetRetryBackoff flag should be sufficient. I suspect there might just be a scenario where it's not being set to true at the end of the event handler.

This is when it's set today, based on the result of the run func.
```
	// If the run succeeded or source changed, reset the retry backoff.
	if runResult.Success || runResult.SourceChanged {
		eventResult.ResetRetryBackoff = true
	}
```

It's possible the RunResult needs additional flags so we can check them to decide when to set ResetRetryBackoff to true. But I'm not sure what the error cases are that you were seeing.

Also, the event producer logs the event even if the event handler then skips calling the run func. So the new log line doesn't mean the run func was called. You might want to add a new log at the top of the DefaultRunFunc so you can tell the difference in the logs. Or alternatively log when the run func is skipped instead. That might help understand the code flow better.

It's confusing right now because we're using both channels and flags. Hopefully in the future we can refactor it again to just use channels so it's cleaner, but doing it this way first was easier to prove it was performing similarly with the existing tests.

nan-yu:
The problem happens in the case of `No RunFunc call`: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/event_handler.go#L134.

When the RunFunc is no-op, we should also reset the retry backoff. This can be fixed by changing [if runResult.Success || runResult.SourceChanged](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/event_handler.go#L148C5-L148C14) to 
```
if runResult.Success || runResult.SourceChanged || !eventResult.RunAttempted {
  eventResult.ResetRetryBackoff = true
}
```

However, it will still reset the retry backoff and keep generating a confusing message `a retry is triggered (retries: 1/12)` every second in the reconciler logs. 

The fix in this PR is to avoid resetting the backoff and generating the retry log when RunFunc is skipped, instead it just set the retry timer to 1s.

The idea is to distinguish between the two following cases:
- reset the backoff when source is changed or a run attempt succeeded
- skip the backoff when retry is never triggered.

karlkfi:
> The fix in this PR is to avoid resetting the backoff and generating the retry log when RunFunc is skipped, instead it just set the retry timer to 1s.

Isn't that what `ResetRetryBackoff` does? Or is the main difference just that `TriggerRetryBackoff` doesn't emit a new log message?

nan-yu:
> Isn't that what `ResetRetryBackoff` does? Or is the main difference just that `TriggerRetryBackoff` doesn't emit a new log message?

If we fix the condition by adding `|| !eventResult.RunAttempted`, it works, but the difference includes both emitting a new log message, and resetting the backoff by copying the initial backoff.

By introducing the new `TriggerRetryBackoff` flag, we don't have to reset the backoff and fix the log. We only reset the timer every second.

nan-yu:
replaced by #1472

PR #1452, opened 2024-10-15
Feature: support OCI signature verification by setting the `configsync.gke.io/image-to-sync` annotation in RSync

tiffanny29631:


e2e/testcases/presync_test.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	nt.T.Log("Create RepoSync with first OCI image")
+	repoSyncOCI := nt.RepoSyncObjectOCI(repoSyncKey, image.OCIImageID().WithoutDigest(), "", image.Digest)
+	nt.Must(rootSyncGitRepo.Add(nomostest.StructuredNSPath(repoSyncID.Namespace, repoSyncID.Name), repoSyncOCI))
+	nt.Must(rootSyncGitRepo.CommitAndPush("Set the RepoSync to sync from OCI"))
+	nt.Must(nt.Watcher.WatchObject(kinds.RepoSyncV1Beta1(), repoSyncID.Name, namespaceRepo,
+		testwatcher.WatchPredicates(testpredicates.RepoSyncHasSourceError(status.SourceErrorCode, "no signatures found"))))
```
nan-yu, line 65: We can create the first OCI image first, and then setup the oci-signature-verification-server to verify unverified image and be verified when the webhook-server is on.

e2e/testcases/presync_test.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		nt.T.Fatal("Failed to get first image remote URL %s", err)
+	}
+	nt.T.Logf("Signing first test image %s", imageURL)
+	_, err = nt.Shell.ExecWithDebug("cosign", "sign", imageURL, "--key", nomostest.CosignPrivateKeyPath(nt), "--yes")
+	if err != nil {
+		nt.T.Fatalf("Failed to sign first test image %s", err)
+	}
+
+	nt.T.Log("Checking no signature error still exists for second image.")
```
nan-yu, line 84: It's hard to tell if the `no signature error` is from the first image, or the second one. 

What about the following test order:
1. create and push an unsigned image ==> RSync is synced successfully
2. set up the webhook server ==> RSync reports a `no signature error`. This is to ensure we can verify unverified image even though the digest is the same.
3. sign the first image ==> RSync is synced successfully again. This is to ensure the verification works.
4. push another unassigned image with the same `latest` tag ==> RSync reports a `no signature error` for the new image. This is to ensure we can verify different images with floating tag.
5. other checks

tiffanny29631: Just tried -

After step 1 the source annotations are already added so there would be no webhook request sent for the same image, thus no verification is performed. This is one of the catch: webhook can only verify later images after setup, not the current syncing one.

e2e/testcases/presync_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"kpt.dev/configsync/pkg/reposync"
+	"kpt.dev/configsync/pkg/rootsync"
+	"kpt.dev/configsync/pkg/status"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestAddPreSyncAnnotationRepoSync(t *testing.T) {
+	repoSyncID := core.RepoSyncID(configsync.RepoSyncName, namespaceRepo)
+	nt := nomostest.New(t, nomostesting.SyncSource,
+		ntopts.RequireOCIProvider,
```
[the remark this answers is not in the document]
tiffanny29631: Not with the current test setup, I was able to run it with just RequireOCIProvider.

build/buildenv/Dockerfile
```diff
@@ -58,6 +58,10 @@ ENV CGO_ENABLED=0
 # but not any more so we have to install it ourselves.
 RUN go install golang.org/x/tools/cmd/goimports@v0.7.0
 
+RUN go install github.com/sigstore/cosign/cmd/cosign@latest
+
+ENV COSIGN_PASSWORD=""
```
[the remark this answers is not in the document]
tiffanny29631: Cosign requires input of password when generating key pairs and signing images, we might be able to get rid of it when changing the authentication type.

e2e/testcases/presync_test.go
nan-yu: I raised the issue in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1452#discussion_r1802160959. We need to update the webhook server to be able to verify this scenario.

nan-yu:
My browser crashed a few times after trying to leave inline comments. 

One comment about the `getGoogleAccessToken` function in `test/docker/presync-webhook-server/oci-signature-verification-server.go`: it uses the go IAM credentials API library, which is specifically for impersonating service accounts. That requires explicitly specifying the service account you want to impersonate, `e2e-test-ar-reader@peip-monitor.iam.gserviceaccount.com` and using its credentials to generate an access token. This is not flexible enough, especially with the support of BYOID in OCI, which doesn't need to impersonate a GSA.

A more common and recommended approach is to use `google.FindDefaultCredentials`, which implements Google's Application Default Credentials (ADC) strategy. It automatically searches for credentials in a predefined order:
- Environment variable (GOOGLE_APPLICATION_CREDENTIALS)
- Google Cloud CLI credentials
- Metadata server (when running on Google Cloud)

This makes the code more portable and allows it to run in different environements.

```
ts, err := googauth.DefaultTokenSource(context.Background(), cloudPlatformScope)
	if err != nil {
		return nil, err
	}

	token, err := ts.Token()
	if err != nil {
		return nil, err
	}

	return &tokenSourceAuth{oauth2.ReuseTokenSource(token, ts)}, nil
```

The  `go-containerregistry` library wraps it with a `NewEnvAuthenticator` function. You can use it in your webhook implementation by replacing `remote.WithAuthFromKeychain(keychain)` with `remote.WithAuthFromKeychain(google.NewEnvAuthenticator())`.

test/docker/presync-webhook-server/main.go
```diff
@@ -0,0 +1,170 @@
...
+		return URLWithSha, nil
+	}
+
+	// image URL has tag
+	URLWithSha := fmt.Sprintf("%s@sha256:%s", imageWithoutTag, commitSHA)
+	return URLWithSha, nil
+}
+
+// auth authenticates Cosign with the Docker registry using gcloud credentials.
+func auth() error {
```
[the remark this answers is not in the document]
tiffanny29631: Changed to using Cosign library in the image verification server instead of using CLI.

test/docker/presync-webhook-server/main.go
```diff
@@ -0,0 +1,170 @@
...
+
+// replaceTagWithDigest replaces the tag in an image URL with the given digest SHA.
+func replaceTagWithDigest(imageURL, commitSHA string) (string, error) {
+	if !strings.Contains(imageURL, ":") {
+		return "", fmt.Errorf("invalid image URL format: no tag or digest found")
+	}
+	imageWithoutTag := strings.Split(imageURL, ":")[0]
+
+	// image URL has digeset
+	if strings.Contains(imageURL, "@sha256:") {
```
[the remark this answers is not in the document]
tiffanny29631: Added check for digest if presents in the image URL.

e2e/testcases/presync_test.go
```diff
@@ -0,0 +1,283 @@
...
+		ntopts.SyncWithGitSource(repoSyncID),
+		ntopts.RepoSyncPermissions(policy.RBACAdmin(), policy.CoreAdmin()))
+	if err := nomostest.SetupPreSync(nt); err != nil {
```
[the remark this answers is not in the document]
tiffanny29631: Added cleanup that removes all signature verification server and webhook related objects at test done.

pkg/parse/root.go
```diff
@@ -261,6 +262,71 @@ func setSourceStatusFields(source *v1beta1.SourceStatus, newStatus *SourceStatus
...
+
+	if len(patchOperations) == 0 {
+		return nil
+	}
+	patchData, err := json.Marshal(patchOperations)
+	if err != nil {
+		return fmt.Errorf("failed to marshal patch operations: %v", err)
+	}
+
+	return p.Client.Patch(ctx, rs, client.RawPatch(types.JSONPatchType, patchData), client.FieldOwner(configsync.FieldManager))
```
nan-yu, line 327: It is simpler to use `types.MergePatchType`.

```
patch := `{"metadata":{"annotations":{"image-to-sync":"url:digest"}}}`
cl.Patch(ctx, rs,
  client.RawPatch(types.MergePatchType, []byte(patch)),
  client.FieldOwner(configsync.FieldManager)
```

test/docker/presync-webhook-server/main.go
```diff
@@ -0,0 +1,170 @@
...
+			response.Allowed = false
+			response.Result = &metav1.Status{
+				Message: fmt.Sprintf("Image validation failed: %v", err),
+			}
+		} else {
+			log.Printf("Image validation successful for %s", newAnnotations[SourceURL])
+			response.Allowed = true
+		}
+	} else {
+		log.Printf("No annotation changes detected")
```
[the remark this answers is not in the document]
tiffanny29631: Modified the error message when verifying the image a bit so that it reflects which one had failed. Now the test would check for failure message for image#2 specifically when image#1 is already signed, to rule out the false positives and negatives.

Since we've agreed that the webhook would not try to verify the current syncing image on-init since there won't be any admission review request sent if the image is already synced, we don't need a local cache in the webhook server.

e2e/testcases/presync_test.go
tiffanny29631: See https://github.com/GoogleContainerTools/kpt-config-sync/pull/1452#discussion_r1813308352

e2e/testcases/presync_test.go
tiffanny29631: See https://github.com/GoogleContainerTools/kpt-config-sync/pull/1452#discussion_r1813308352

pkg/parse/root.go

tiffanny29631:
re @nan-yu google authentication is simplified with one of the recent commits.

build/buildenv/Dockerfile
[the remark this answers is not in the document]
tiffanny29631: Mmm the presubmit e2e is still failing for some reason, my local runs with GKE clusters are fine though. I'll be focus on resolving that today.

pkg/parse/namespace.go
```diff
@@ -174,6 +177,51 @@ func (p *namespace) setSourceStatusWithRetries(ctx context.Context, newStatus *S
 	return nil
 }
 
+func (p *namespace) setSourceAnnotations(ctx context.Context, commit string) error {
```
nan-yu, line 180: This still has a GET call. We can go with a single PATCH call as below:

```
func (p *namespace) setSourceAnnotations(ctx context.Context, commit string) error {
	rs := &v1beta1.RepoSync{}
	rs.Namespace = string(p.Scope)
	rs.Name = p.SyncName

	var patchType types.PatchType
	var patch string
	if p.Options.SourceType == configsync.OciSource ||
		(p.Options.SourceType == configsync.HelmSource && strings.HasPrefix(p.Options.SourceRepo, "oci://")) {
		patchType = types.MergePatchType
		patch = fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"}}}`,
			metadata.ImageToSyncAnnotationKey,
			fmt.Sprintf("%s@sha256:%s", p.Options.SourceRepo, commit),
		)
	} else {
		patchType = types.JSONPatchType
		patch = fmt.Sprintf(`[{"op":"remove", "path":"/metadata/annotations/%s"}]`,
			// TODO: `/` in the annotation key needs to be encoded with `~1`,
			// e.g. "/metadata/annotations/configsync.gke.io~1image-to-sync"
			metadata.ImageToSyncAnnotationKey)
	}
	return p.Client.Patch(ctx, rs,
		client.RawPatch(patchType, []byte(patch)),
		client.FieldOwner(configsync.FieldManager))
}
```

pkg/parse/root.go
```diff
@@ -261,6 +263,51 @@ func setSourceStatusFields(source *v1beta1.SourceStatus, newStatus *SourceStatus
 	source.LastUpdate = newStatus.LastUpdate
 }
 
+func (p *root) setSourceAnnotations(ctx context.Context, commit string) error {
```
nan-yu, line 266: Similarly for RootSync:

```
func (p *root) setSourceAnnotations(ctx context.Context, commit string) error {
	rs := &v1beta1.RootSync{}
	rs.Namespace = configsync.ControllerNamespace
	rs.Name = p.SyncName

	var patchType types.PatchType
	var patch string
	if p.Options.SourceType == configsync.OciSource ||
			(p.Options.SourceType == configsync.HelmSource && strings.HasPrefix(p.Options.SourceRepo, "oci://")) {
		patchType = types.MergePatchType
		patch = fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"}}}`,
			metadata.ImageToSyncAnnotationKey,
			fmt.Sprintf("%s@sha256:%s", p.Options.SourceRepo, commit),
		)
	} else {
		patchType = types.JSONPatchType
		patch = fmt.Sprintf(`[{"op":"remove", "path":"/metadata/annotations/%s"}]`,
			// TODO: `/` in the annotation key needs to be encoded with `~1`,
			// e.g. "/metadata/annotations/configsync.gke.io~1image-to-sync"
			metadata.ImageToSyncAnnotationKey)
	}
	return p.Client.Patch(ctx, rs,
		client.RawPatch(patchType, []byte(patch)),
		client.FieldOwner(configsync.FieldManager))
}
```

test/docker/presync-webhook-server/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Build stage
+FROM golang:1.23 as builder
+
+WORKDIR /app
+COPY . .
+
+RUN go mod init presync-webhook-server || true && go mod tidy
```
nan-yu, line 21: The go.mod file exists. Do we still need `go mod init`?

test/docker/presync-webhook-server/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+ENV PATH "/go/bin:/usr/local/go/bin:$PATH"
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o webhook-server oci-signature-verification-server.go
+
+# Final stage - runtime image
+FROM debian:bullseye-slim
+
+WORKDIR /
+
+# Install minimal necessary packages
```
nan-yu, line 33: Can we use distroless base image with minimal packages?

test/docker/presync-webhook-server/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  ca-certificates \
+  && rm -rf /var/lib/apt/lists/*
+
+# Copy the built binary from the builder stage
+COPY --from=builder /app/webhook-server /webhook-server
+
+RUN chmod +x /webhook-server
```
nan-yu, line 41: Does `/webhook-server` have execution permission already?
Other images like `/reconciler` don't need to change mode.

e2e/nomostest/nt.go
```diff
@@ -515,7 +518,8 @@ func (nt *NT) testLogs(previousPodLog bool) {
 	// temporarily to see how presubmit responds.
 	nt.PodLogs(configmanagement.ControllerNamespace, reconcilermanager.ManagerName, reconcilermanager.ManagerName, previousPodLog)
 	nt.PodLogs(configmanagement.ControllerNamespace, configuration.ShortName, configuration.ShortName, previousPodLog)
-	nt.PodLogs("resource-group-system", "resource-group-controller-manager", "manager", false)
+	nt.PodLogs(configmanagement.RGControllerNamespace, configmanagement.RGControllerName, "manager", previousPodLog)
+	nt.PodLogs(testOCISignatureVerificationNamespace, testOCISignatureVerificationServerName, "", previousPodLog)
```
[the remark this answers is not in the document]
tiffanny29631: Done, moved to specific test case.

e2e/nomostest/ssh.go
```diff
@@ -292,6 +329,110 @@ func generateSSLKeys(nt *NT, syncSource SyncSource, namespace string, domains []
...
+	err := os.Setenv(cosignPasswordEnvVar, cosignPassword)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err := os.Unsetenv(cosignPasswordEnvVar)
+		if err != nil {
+			nt.T.Errorf("Failed to unset COSIGN_PASSWORD: %v", err)
+		}
+	}()
```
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/nomostest/ssh.go
```diff
@@ -292,6 +329,110 @@ func generateSSLKeys(nt *NT, syncSource SyncSource, namespace string, domains []
...
+
+	err = moveFile("cosign.key", CosignPrivateKeyPath(nt))
+	if err != nil {
+		return fmt.Errorf("failed to move cosign.key: %v", err)
+	}
+
+	err = moveFile("cosign.pub", cosignPublicKeyPath(nt))
+	if err != nil {
+		return fmt.Errorf("failed to move cosign.key: %v", err)
+	}
```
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/nomostest/ssh.go
```diff
@@ -292,6 +329,110 @@ func generateSSLKeys(nt *NT, syncSource SyncSource, namespace string, domains []
...
+		return err
+	}
+
+	nt.T.Logf("Secret %s created.\n", secretName)
+	return nil
+}
+
+// generateTLSKeyPair generates a TLS key pair using OpenSSL and saves the
+// certificate and key to files.
+func generateTLSKeyPair(nt *NT) error {
```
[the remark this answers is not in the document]
tiffanny29631: Done

pkg/parse/namespace.go
tiffanny29631: Done, using mergepatchtype for both, for removing the annotation I'm setting the value to null.

test/docker/presync-webhook-server/Dockerfile
tiffanny29631: Cleaned

test/docker/presync-webhook-server/Dockerfile
tiffanny29631: Cleaned

test/docker/presync-webhook-server/Dockerfile
tiffanny29631: Done

pkg/parse/root.go
tiffanny29631: Done

e2e/nomostest/ssh.go
```diff
@@ -292,6 +329,110 @@ func generateSSLKeys(nt *NT, syncSource SyncSource, namespace string, domains []
...
+
+	err = os.Rename(src, dest)
+	if err != nil {
+		return fmt.Errorf("failed to move file from %s to %s: %v", src, dest, err)
+	}
+	return nil
+}
+
+// createKubernetesSecret creates a Kubernetes secret with the given name, namespace, type, and optional extra arguments.
+func createKubernetesSecret(nt *NT, secretName, namespace, secretType string, args ...string) error {
```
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/nomostest/registryproviders/oci_image.go
```diff
@@ -248,3 +248,11 @@ func (id OCIImageID) WithoutTag() OCIImageID {
 		Digest:     id.Digest,
 	}
 }
+
+// DigestWithoutPrefix returns the image digest without the sha256 prefix
+func (id OCIImageID) DigestWithoutPrefix() string {
```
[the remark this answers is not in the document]
nan-yu: removed

e2e/testcases/presync_test.go
```diff
@@ -0,0 +1,265 @@
...
+		testwatcher.WatchPredicates(testpredicates.RepoSyncHasSourceError(status.SourceErrorCode, "no signatures found"))))
+
+	nt.T.Log("Create second image with latest tag.")
+	bookstoreSA := k8sobjects.ServiceAccountObject("bookstore-sa")
+	image1, err := nt.BuildAndPushOCIImage(repoSyncKey, registryproviders.ImageInputObjects(nt.Scheme, bookInfoRole, bookstoreSA))
+	if err != nil {
+		nt.T.Fatal(err)
+	}
+
+	if err = signImage(nt, image); err != nil {
```
nan-yu, line 79: We don't verify the first image is synced successfully. Do we need to sign it?

e2e/testcases/presync_test.go
```diff
@@ -0,0 +1,265 @@
...
+		nt.T.Fatalf("Failed to sign first test image %s", err)
+	}
+
+	nt.T.Log("Check that verification is up-to-date with floating tag")
+	nt.T.Logf("Checking no signature error exists for second image with digest %s", image1.Digest)
+	nt.Must(
+		nt.Watcher.WatchObject(kinds.RepoSyncV1Beta1(), repoSyncID.Name, namespaceRepo,
+			testwatcher.WatchPredicates(testpredicates.RepoSyncHasSourceError(status.SourceErrorCode, "no signatures found"))),
+		nt.Watcher.WatchObject(kinds.RepoSyncV1Beta1(), repoSyncID.Name, namespaceRepo,
```
nan-yu, line 88: This two predicates can be merged with a single `WatchObject` call.

test/docker/presync-webhook-server/oci-signature-verification-server.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, err
+	}
+	if annotations, ok := metadata["metadata"].(map[string]interface{})["annotations"].(map[string]interface{}); ok {
+		annotationsMap := make(map[string]string)
+		for k, v := range annotations {
+			annotationsMap[k] = fmt.Sprintf("%v", v)
+		}
+		return annotationsMap, nil
+	}
+	return nil, fmt.Errorf("no annotations found")
```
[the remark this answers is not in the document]
tiffanny29631: Changed to log only.

test/docker/presync-webhook-server/oci-signature-verification-server.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func verifyImageSignature(image string) error {
+	if image == "" {
+		return nil
+	}
+
+	ctx := context.Background()
```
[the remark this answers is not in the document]
tiffanny29631: Done

test/docker/presync-webhook-server/oci-signature-verification-server.go
```diff
@@ -0,0 +1,146 @@
...
+			klog.Infof("Image verification successful for %s", newAnnotations[imageToSync])
+			response.Allowed = true
+		}
+	} else {
+		response.Allowed = true
+	}
+
+	admissionReview.Response = response
+	if err := json.NewEncoder(w).Encode(admissionReview); err != nil {
+		klog.Infof("Failed to encode admission response: %v", err)
```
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/testcases/presync_test.go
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/testcases/presync_test.go
tiffanny29631: We want to make sure even though the first image is signed, when the latest tag is with second unsigned image, the server would keep blocking the request - to prevent a false positive.

e2e/testcases/presync_test.go
tiffanny29631: Done

test/docker/presync-webhook-server/oci-signature-verification-server.go
```diff
@@ -0,0 +1,154 @@
...
+	newAnnotations, err := getAnnotations(admissionReview.Request.Object.Raw)
+	if err != nil {
+		klog.Errorf("Failed to extract new annotations: %v", err)
+		response.Result = &metav1.Status{
+			Message: fmt.Sprintf("Failed to extract new annotations: %v", err),
+		}
+		response.Allowed = false
+	}
+
+	if newAnnotations[imageToSync] != oldAnnotations[imageToSync] {
```
nan-yu, line 128: This may lead to a nil pointer if `getAnnotations` returns nil.

We can update `getAnnotations` to `getAnnotation` by the annotation key.

test/docker/presync-webhook-server/oci-signature-verification-server.go
```diff
@@ -0,0 +1,154 @@
...
+
+	admissionReview.Response = response
+	if err := json.NewEncoder(w).Encode(admissionReview); err != nil {
+		klog.Errorf("Failed to encode admission response: %v", err)
+	}
+}
+
+func main() {
+	http.HandleFunc("/validate", handleWebhook)
+	klog.Info("Starting webhook server on port 8443...")
```
nan-yu, line 152: I explained why we need to use `8443` instead of `443` in a commit message. It's better to add an inline comment here.

// Use 8443 as a non-privileged port to avoid root privilege to bind to the port

e2e/testcases/oci_signature_verification_test.go
```diff
@@ -0,0 +1,265 @@
...
+				imageToSyncAnnotation,
+				syncingCommit,
+				syncingImage,
+			)
+		}
+		return nil
+	}
+}
+
+func checkRepoSyncPreSyncAnnotations(image *registryproviders.OCIImage) testpredicates.Predicate {
```
nan-yu, line 226: Do we need these two functions?

We can use `testpredicates.HasAnnotation(metadata.ImageToSyncAnnotationKey, image1.OCIImageID().Address())` to replace them.

e2e/testcases/presync_test.go
nan-yu: ack

test/docker/presync-webhook-server/oci-signature-verification-server.go
tiffanny29631: Done

test/docker/presync-webhook-server/oci-signature-verification-server.go
tiffanny29631: Done

e2e/testcases/oci_signature_verification_test.go
tiffanny29631: Done

test/docker/presync-webhook-server/oci-signature-verification-server.go
```diff
@@ -0,0 +1,161 @@
...
+		klog.Errorf("Failed to unmarshal admission review: %v", err)
+		http.Error(w, "Failed to unmarshal admission review", http.StatusBadRequest)
+		return
+	}
+
+	response := &admissionv1.AdmissionResponse{
+		UID: admissionReview.Request.UID,
+	}
+
+	oldAnnotation, err := getAnnotationByKey(admissionReview.Request.OldObject.Raw, imageToSync)
```
nan-yu, line 114: nit: suggest renaming to `oldImage` and `newImage`

tiffanny29631: Done

nan-yu (approved):
/lgtm

PR #1453, opened 2024-10-16
feat: add bash script for uninstalling ACM

sdowell:
This is an alternative approach to uninstalling Config Management and the operator. This is intended as a simple/portable alternative to using the nomos migrate command.

### DIFF
--- a/installation/uninstall_configmanagement.sh
+++ b/installation/uninstall_configmanagement.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script uninstalls ConfigManagement
+
+set -euox pipefail
+
+hnc_enabled="$(kubectl get configmanagements.configmanagement.gke.io config-management -o=jsonpath="{.spec.hierarchyController.enabled}" --ignore-not-found)"
+
+if [[ "${hnc_enabled}" == "true" ]]; then
+  echo "Hierarchy Controller is enabled on the ConfigManagement object. It must be disabled before migrating."
+  echo "This can be done by unsetting the spec.hierarchyController field on ConfigManagement."
+  exit 1
+fi
+
+kubectl delete deployment -n config-management-system config-management-operator --ignore-not-found --cascade=foreground
+
+if kubectl get configmanagement config-management &> /dev/null ; then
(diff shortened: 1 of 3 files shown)


installation/uninstall_configmanagement.sh
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script uninstalls ConfigManagement
+
+set -euox pipefail
+
+kubectl delete deployment -n config-management-system config-management-operator --ignore-not-found --cascade=foreground
```
nan-yu, line 20: Before deleting the operator Deployment, should we disable HNC first to allow the operator to undeploy HNC resources gracefully?

If not in the uninstallation script, the upgrade instruction should have 3 steps:
1. disable HNC if it is enabled
2. uninstall the current operator
3. upgrade to the new version

sdowell: I'm assuming that the desired behavior when removing ACM operator is to leave subcomponents (HNC, PoCo, CS) orphaned on the cluster. Is there a reason the user would want to uninstall HNC during the procedure?

I don't think it needs to be part of the uninstallation script, but if there's a compelling use case then we could include the upgrade instructions that you suggest.

nan-yu: For CS, we can safely leave it orphaned because we don't want to pause the reconciliation and the new installation will upgrade old reconcilers.
For PoCo, I am not sure about the migration process.
For HNC, users won't be able to disable it after the upgrade if we leave the subcomponent around. They will
a. continue using the orphaned and unsupported subcomponent without migration.
b. encounter controller fights with both orphaned HNC and OSS HNC.
c. manually delete each individual resource (controller, RBAC, service account, and etc) before installing OSS HNC.

To prevent the above issues, we should  include disabling HNC as part of uninstallation or as a prerequisite. Then users are forced to disable HNC  before uninstallation and upgrade.

sdowell: The next PR will include the installation tooling and instructions. Let's continue the discussion on that PR.

nan-yu: The installation instruction may affect the uninstallation logic. We may either disable HNC, or block uninstallation if HNC is enabled.

sdowell: This script is intended to be the same behavior as `nomos migrate`, which does a simple uninstall that orphans all components.

It feels like the HNC specific logic is out of scope here. Is there a design doc for the experience of migrating to OSS HNC? It seems like a more consistent approach would be to orphan everything with the ACM uninstall and then have a separate tool that migrates from the legacy HNC to OSS HNC. I assume not all users want to migrate to OSS HNC and some might want to stay on the legacy HNC.

sdowell: Sent https://github.com/GoogleContainerTools/kpt-config-sync/pull/1454 to first update the behavior of `nomos migrate`.

sdowell: Updated the script to block if HNC is enabled

nan-yu (approved):
/lgtm

PR #1454, opened 2024-10-17
feat: stall nomos migrate if HNC is enabled

sdowell:
This requires for the user to disable HNC before removing Config Management, so that ACM helps with the uninstallation of HNC before it is absent from the cluster. This is intended to streamline the experience of migrating to OSS HNC.

### DIFF
--- a/cmd/nomos/migrate/configmanagement.go
+++ b/cmd/nomos/migrate/configmanagement.go
@@ -40,6 +40,11 @@ func migrateConfigManagement(ctx context.Context, cc *status.ClusterClient, kube
 		printNotice("The cluster is already running as an OSS installation.")
 		return nil
 	}
+	if isHNCEnabled, err := cc.ConfigManagement.IsHNCEnabled(ctx); err != nil {
+		return err
+	} else if isHNCEnabled {
+		return fmt.Errorf("Hierarchy Controller is enabled on the ConfigManagement object. It must be disabled before migrating.")
+	}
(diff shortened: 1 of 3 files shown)


cmd/nomos/migrate/configmanagement.go
```diff
@@ -40,6 +40,11 @@ func migrateConfigManagement(ctx context.Context, cc *status.ClusterClient, kube
 		printNotice("The cluster is already running as an OSS installation.")
 		return nil
 	}
+	if isHNCEnabled, err := cc.ConfigManagement.IsHNCEnabled(ctx); err != nil {
+		return err
+	} else if isHNCEnabled {
+		return fmt.Errorf("Hierarchy Controller is enabled on the ConfigManagement object. It must be disabled before migrating.")
```
nan-yu, line 46: Add the migration instruction in the error message: https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/how-to/migrate-hierarchy-controller.

sdowell: I have some concern including a hardcoded link in the output, as it's possible the link could be removed or renamed in the future. Do you see this as a concern?

nan-yu: That is indeed a concern. Let's keep it as is.

nan-yu (approved):
/lgtm

PR #1457, opened 2024-10-21
ci: add tooling to package kustomization bundle

sdowell:
This adds a make target that package the kustomization bundle into a tarball when the manifests are built. This will be published as a versioned release artifact for each release.


nan-yu (approved):
/lgtm

PR #1458, opened 2024-10-23
set the registry for image URL replacement

sdowell:
Previously the README always assumed gcr.io/config-management-release. With this change, the source registry is updated dynamically based on where the images were actually published.


nan-yu (approved):
/lgtm

PR #1460, opened 2024-10-26
chore: merge sync & full-sync events

karlkfi:
- Simplify the event handler by checking if full-sync is needed
  inside reconciler.Reconcile, based on the lastFullSyncTime.

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1461
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1480
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1494
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1495


pkg/parse/run_test.go
```diff
@@ -236,48 +236,62 @@ func TestReconciler_Reconcile(t *testing.T) {
...
 	}{
 		{
-			name:                  "read retryable error because missing source commit directory",
-			retryCap:              5 * time.Millisecond,
-			srcRootCreateLatency:  10 * time.Millisecond,
-			expectedSourceChanged: false,
-			needRetry:             true,
-			parseOutputs:          nil, // parse should not be called
-			expectedRootSync: func() *v1beta1.RootSync {
+			name:                  "reconcile success with existing source commit directory",
```
karlkfi, line 242: I moved the `read retryable error` test down a bit, which broke the expectations, which were apparently hard-coded based on the test order. So I had to change `expectedRootSync`  to `expectedRootSyncFunc`, so I could pass in the sourcePath. 

I also added a new `reconcile success with existing source commit directory` test happy path test.
That's why this diff is a little jumbled.

karlkfi: I also added a new `reconcile success with full-sync after apply error` test, which needed `reconcilerStateFunc` to initialize the state to simulate a previous apply error on the same commit.

pkg/parse/run.go
```diff
@@ -113,10 +114,17 @@ func (r *reconciler) Reconcile(ctx context.Context, trigger string) ReconcileRes
...
+		trigger = triggerFullSync
+	}
+
+	klog.Infof("Starting sync attempt (trigger: %s)", trigger)
+
 	switch trigger {
 	case triggerFullSync, triggerManagementConflict, triggerNamespaceUpdate:
 		// Force parsing and updating, but skip fetch, render, and read unless required.
-		state.RecordFullSyncStart()
+		state.RecordFullSyncStart(startTime)
```
nan-yu, line 127: The full sync event was triggered 1 hour later after the reconcile loop ends. Now, it is changed to be based on the startTime. Should we change it to endTime?

karlkfi: I had a comment about that but removed it cause it was confusing.

The previous method DID delay from the end of the Reconcile attempt, however it did not account for errors, success after retries, or full syncs triggered by namespace updates or management conflicts. So it didn't really work as intended.

I can't think of any way to know when a sync succeeds that it was first triggered by a full sync, and I'm not sure that would even really be desirable behavior.

AFAIK, the point of the full sync is to re-parse the objects. it doesn't actually re-fetch or re-render them. So the only real benefit is to A) recover from transitive parsing/validation errors, and B) re-sync to correct drift that was missed by the remediator.

For both of those cases, I think it's probably fine and even desirable to do full-syncs at least every hour, rather than the old behavior which was a delay.

In fact, AFAICT, full syncs don't reset the retry backoff unless they succeed. So if retries are exhausted, a full sync is the last attempt until the source changes. So doing them at least every hour means the retry backoff effectively maxes out at an hour between attempts, instead of an hour delay.

But yes, if the sync takes an hour, it would start another 15s later on the next sync attempt.

karlkfi: If we wait until the sync succeeds to set the lastFullSyncTime and a full sync errors, it would effectively disable both full syncs and retries until the next commit change.

We could update lastFullSyncTime after errors and success, only for triggerFullSync, triggerManagementConflict, and triggerNamespaceUpdate, but would that have the desired result?

nan-yu: ACK. I wouldn't think a sync would take an hour long especially with the current ResourceGroup size limit. Hence it may not matter that much if it delays at the start time or end time.

nan-yu (approved):
/lgtm

pkg/parse/run.go
karlkfi: It can... but only with a long dependency chain, since the timeout is for each layer of dependencies, not all of them together. So it's probably relatively unlikely, unless users specify a long reconcile timeout.

PR #1464, opened 2024-10-29
e2e: Only create registry secret when using local registry provider

tiffanny29631:
…ovider

### DIFF
--- a/e2e/nomostest/oci-image-verification-server.go
+++ b/e2e/nomostest/oci-image-verification-server.go
@@ -136,14 +136,17 @@ func auth(nt *NT) error {
 		return err
 	}
 
-	// Create secret for the public ca-cert in the oci-image-verification namespace
-	// for the oci-signature-verification webhook server to verify certificate.
-	sharedTmpDir := filepath.Join(os.TempDir(), NomosE2E, nt.ClusterName)
-	sharedTestRegistrySSLDir := filepath.Join(sharedTmpDir, string(RegistrySyncSource), sslDirName)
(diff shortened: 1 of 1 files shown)


e2e/nomostest/oci-image-verification-server.go
```diff
@@ -141,9 +141,11 @@ func auth(nt *NT) error {
 	sharedTmpDir := filepath.Join(os.TempDir(), NomosE2E, nt.ClusterName)
 	sharedTestRegistrySSLDir := filepath.Join(sharedTmpDir, string(RegistrySyncSource), sslDirName)
 	testRegistryCACertPath := filepath.Join(sharedTestRegistrySSLDir, caCertFile)
-	if err := createSecret(nt, OCISignatureVerificationNamespace, PublicCertSecretName(RegistrySyncSource),
-		fmt.Sprintf("cert=%s", testRegistryCACertPath)); err != nil {
-		return err
+	if e2e.ArtifactRegistry == "local" {
```
nan-yu, line 144: `e2e.ArtifactRegistry` is a constant. It will never be `local`.

I think you meant ` if *e2e.OCIProvider == e2e.Local`?

nan-yu: You can also move the lines above inside the if condition. They are not used when using GAR.

nan-yu: The comment also needs to be updated.

tiffanny29631: Done

nan-yu (approved):
/lgtm

PR #1466, opened 2024-10-30
e2e: Grant signature verification KSA with GKE WI

tiffanny29631:


e2e/testinfra/terraform/common/service_accounts.tf
```diff
@@ -102,3 +102,10 @@ resource "google_project_iam_member" "ns-reconciler-wi-sa-iam" {
   member  = "serviceAccount:${data.google_project.project.project_id}.svc.id.goog[config-management-system/ns-reconciler-test-ns]"
   project = data.google_project.project.id
 }
+
+# Grant artifact registry reader permissions to the oci-image-verification-sa KSA with GKE workload identity.
```
nan-yu, line 106: This is already set in https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testinfra/terraform/modules/service_account/service_account.tf#L75.

tiffanny29631: Applied test infra to both kpt-config-sync-ci-main and kpt-config-sync-ci-release.

PR #1468, opened 2024-10-31
[WIP] fix: retry event delay

karlkfi:
- Delay retry events after a run was attempted due to a different type of event.
- Change most event logs to level 3, to reduce spam for customers.
- Add a log at the beginning of each sync attempt, to explain the trigger.

### DIFF
--- a/pkg/parse/event_handler.go
+++ b/pkg/parse/event_handler.go
@@ -136,7 +136,7 @@ func (s *EventHandler) Handle(event events.Event) events.Result {
 		} else if opts.needToUpdateWatch() {
 			trigger = triggerWatchUpdate
 		} else {
-			// No RunFunc call
+			klog.Info("Skipping retry attempt: not needed")
 			break
 		}
 
--- a/pkg/parse/event_handler.go
+++ b/pkg/parse/event_handler.go
@@ -67,22 +67,21 @@ func (s *EventHandler) Handle(event events.Event) events.Result {
 	var runResult RunResult
 	switch event.Type {
 	case events.SyncWithReimportEventType:
-		// Re-apply even if no changes have been detected.
-		// This case should be checked first since it resets the cache.
-		// If the reconciler is in the process of reconciling a given commit, the resync won't
-		// happen until the ongoing reconciliation is done.
-		klog.Infof("It is time for a force-resync")
-		// Reset the cache partially to make sure all the steps of a parse-apply-watch loop will run.
-		// The cached sourceState will not be reset to avoid reading all the source files unnecessarily.
-		// The cached needToRetry will not be reset to avoid resetting the backoff retries.
-		state.resetPartialCache()
-		runResult = runFn(s.Context, s.Reconciler, triggerResync)
+		// Reimport = Read* + Render* + Parse + Update
+		//
+		// * Read & Render will only happen if there's a new commit, new source
+		//   spec change, or a previous error invalidated the cache.
+		//   Otherwise re-import starts from re-parsing the objects from disk.
+		runResult = runFn(s.Context, s.Reconciler, triggerReimport)
 
 	case events.SyncEventType:
-		// Re-import declared resources from the filesystem (from *-sync).
-		// If the reconciler is in the process of reconciling a given commit, the re-import won't
-		// happen until the ongoing reconciliation is done.
-		runResult = runFn(s.Context, s.Reconciler, triggerReimport)
+		// Resync = Read* + Render* + Parse* + Update
+		//
+		// * Read, Render, and Parse will only happen if there's a new commit,
+		//   new source spec change, or a previous error invalidated the cache.
+		//   Otherwise re-sync skips directly to the Update stage, using the
+		//   previously parsed in-memory object cache.
+		runResult = runFn(s.Context, s.Reconciler, triggerResync)
 
 	case events.StatusEventType:
 		// Publish the sync status periodically to update remediator errors.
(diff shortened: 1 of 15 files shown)


karlkfi:
Added a second commit to try out channel driven event. prolly needs some cleanup.

pkg/parse/run.go
```diff
@@ -78,6 +78,7 @@ type RunFunc func(ctx context.Context, r Reconciler, trigger string) RunResult
 
 // DefaultRunFunc is the default implementation for RunOpts.RunFunc.
 func DefaultRunFunc(ctx context.Context, r Reconciler, trigger string) RunResult {
+	klog.Infof("Starting sync attempt (trigger: %s)", trigger)
```
nan-yu, line 81: After tracing back where `trigger` is set, the trigger would be `reimport` for the periodic reimport event, as well as managementConflicts and watchUpdates events, which makes the log quite confusing. We need to reset the trigger when `opts.HasManagementConflict` or `opts.NeedToUpdateWatch`, or use separate events for managementConflicts and watchUpdates.

pkg/parse/events/publisher.go
```diff
@@ -154,27 +155,29 @@ func (s *RetrySyncPublisher) Start(ctx context.Context) reflect.Value {
...
 
 // HandleResult resets the backoff timer if ResetRetryBackoff is true.
 func (s *RetrySyncPublisher) HandleResult(result Result) {
 	if result.ResetRetryBackoff {
 		s.currentBackoff = util.CopyBackoff(s.Backoff)
+		klog.V(3).Infof("Resetting retry backoff (%v)", s.currentBackoff.Duration)
 		s.timer.Reset(s.currentBackoff.Duration)
+	} else if result.RunAttempted && s.currentBackoff.Steps > 0 {
+		klog.V(3).Infof("Run attempted, delaying next retry event (%v)", s.nextDelay)
+		s.timer.Reset(s.nextDelay)
```
nan-yu, line 178: Is that correct to reset the timer in the `HandleResult` function? The function is invoked after each publisher publishes its event. So when it doesn't need to reset retry backoff, it will keep delaying the retry timer. Before the change, the timer is only reset when the retry event is published.

pkg/parse/events/publisher.go
```diff
@@ -154,27 +155,29 @@ func (s *RetrySyncPublisher) Start(ctx context.Context) reflect.Value {
...
+		klog.Infof("Retry limit has been reached (%v)", s.retryLimit)
 		// Don't reset retryTimer if retry limit has been reached.
 		return Result{}
 	}
 
-	retryDuration := s.currentBackoff.Step()
+	s.nextDelay = s.currentBackoff.Step()
 	retries := s.retryLimit - s.currentBackoff.Steps
-	klog.Infof("a retry is triggered (retries: %v/%v)", retries, s.retryLimit)
+	klog.V(3).Infof("Sending retry event (step: %v/%v)", retries, s.retryLimit)
```
nan-yu, line 165: Should we log this message at the default level?

If no retry attempt is needed, we would only publish the retry event once, so the log will only show up once.

pkg/parse/event_handler.go
```diff
@@ -123,20 +123,12 @@ func (s *EventHandler) Handle(event events.Event) events.Result {
 		runResult = runFn(s.Context, s.Reconciler, namespaceEvent)
 
 	case events.RetrySyncEventType:
-		// Retry if there was an error, conflict, or any watches need to be updated.
+		// Retry if there was an error, or conflict.
```
nan-yu, line 126: Is conflict handled by the retry event? From the code, it sounds like conflict is handled by `events.SyncEventType`.

pkg/parse/events/builder.go
```diff
@@ -63,5 +66,13 @@ func (t *PublishingGroupBuilder) Build() []Publisher {
 	if t.StatusUpdatePeriod > 0 {
 		publishers = append(publishers, NewResetOnRunAttemptPublisher(StatusEventType, t.Clock, t.StatusUpdatePeriod))
 	}
+	if t.WatchUpdateCh != nil {
```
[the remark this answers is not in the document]
karlkfi: They're nil in the tests that aren't testing these timers. It's the same reason the other ones check > 0

pkg/parse/run.go
karlkfi: Yeah, I need to clean the log up here, but I wanted to see if the channels would work first.

pkg/parse/events/publisher.go
karlkfi: This was intentional. Before, the retry delay was independent of other events, but that's not technically what we want. We really want retry to be a delay from the last sync attempt. Otherwise, a retry could trigger right after another event triggered a sync attempt, causing back-to-back sync attempts.

pkg/parse/events/publisher.go
karlkfi: Yeah. This is one of the log changes I added to the first commit, before I added the second commit with the channels. With the channels triggering SyncEvents instead of RetryEvents, we don't actually need to 'mute' this.

pkg/parse/event_handler.go
karlkfi: Right. I changed this comment before adding the conflict channel. Need to go fix the comment.

pkg/parse/events/publisher.go
```diff
@@ -201,6 +204,82 @@ type ResetOnRunAttemptPublisher struct {
...
 	s.TimeDelayPublisher.HandleResult(result)
 	if result.RunAttempted {
+		klog.V(3).Infof("Run attempted, delaying next sync event (%v)", s.Period)
 		s.timer.Reset(s.Period)
 	}
 }
+
+// NewDedupingChannelPublisher constructs an ChannelDrivenPublisher that
+// enqueues a new specified event any time the input channel receives an event.
+func NewDedupingChannelPublisher(eventType EventType, inCh <-chan bool) *ChannelDrivenPublisher {
```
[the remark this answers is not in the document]
karlkfi: It doesn't send a RetrySyncEventType. It sends a SyncEventType, which is passed in to the constructor by the builder.

This does mess up the triggers used for logging a bit, but it should work fine, because the handler does basically the same thing for both events now.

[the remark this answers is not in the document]
karlkfi: My ultimate goal is to condense all the events down to just one type. I have a different PR to combine the Sync and Reimport events. So this one moves the conflict and watch update cases to use the Sync event too.

karlkfi:
Obsolete. Replaced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1472

PR #1472, opened 2024-11-01
fix: ignored conflicts and watch updates

karlkfi:
- Reset the retry backoff when syncing is skipped due to
  prior successful sync and there are no conflicts or watch updates.
  This ensures the retry event will still fire to check for new
  conflicts and watch update requests from the remediator.
- Add logging with more details for debugging, but at level 3, to
  avoid spamming customer logs.


nan-yu (approved):
/lgtm

PR #1474, opened 2024-11-01
[WIP] fix: retry events happening too close together

karlkfi:
- Delay the next retry any time there was a sync attempt, even if it wasn't triggered by a retry event. This fixes retry events triggering too close after a sync, full-sync, or namespace event.

TODO: fix e2e tests

This change exposes 2 bugs:
1. Sync events are not backing off correctly, only retry events are backing off. So the max backoff is the effectively the polling period (15s by default, 5s in e2e tests).
2. When there is a management conflict detected by a sync attempt, and the conflict is later resolved by a 3rd party, the conflict error is never removed, because the remediator is paused, because the sync errored. Then syncs are skipped, because the source didn't change, until there's a full-sync that clears the cache and calls the updater again.

### DIFF
--- a/pkg/parse/events/funnel_test.go
+++ b/pkg/parse/events/funnel_test.go
@@ -128,18 +128,25 @@ func TestFunnel_Start(t *testing.T) {
 			stepSize: time.Second,
 			expectedEvents: []eventResult{
 				{
-					Event:  Event{Type: RetrySyncEventType},
-					Result: Result{},
+					Event: Event{Type: RetrySyncEventType},
+					Result: Result{
+						RunAttempted: true,
+					},
 				},
 				{
-					Event:  Event{Type: RetrySyncEventType},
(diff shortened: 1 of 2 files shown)


karlkfi:
So it turns out this change actually causes those two bugs I discovered. AND the "bug" this PR was trying to fix isn't actually a bug.

It turns out SyncEvents and RetryEvents are safe to happen back to back and interleaved, because the SyncEvents only calls ParseAndUpdate if the source has changed, and ParseAndUpdate is the thing we actually want to delay. SyncEvents still needs to run to do the Read stage that checks the shared volume for new commits/changes/rendering. And RetryEvents need to happen in order to check conflicts and watch updates, even if there's no error to retry.

Follow up with more tests: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1475

PR #1475, opened 2024-11-01
chore: test retry with long and short sync periods

karlkfi:
The e2e tests use a short sync period to speed up tests. So this change adds tests that confirm that the retry backoff still works as intended when its shorter and longer than the sync period.

The SyncEvent only calls ParseAndUpdate if a source change was detected, so it's safe to still send SyncEvents when the retry backoff is longer than the SyncPeriod, because what we really want to delay is the ParseAndupdate calls, which cause a lot of API server requests.

### DIFF
--- a/pkg/parse/events/funnel_test.go
+++ b/pkg/parse/events/funnel_test.go
@@ -173,7 +173,7 @@ func TestFunnel_Start(t *testing.T) {
 			},
 		},
 		{
-			name: "RetryEvents From RetryBackoff with ResetRetryBackoff",
+			name: "RetryEvents and SyncEvents with RetryBackoff shorter than SyncPeriod",
 			builder: &PublishingGroupBuilder{
 				RetryBackoff: wait.Backoff{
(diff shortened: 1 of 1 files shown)


pkg/parse/events/funnel_test.go
```diff
@@ -202,6 +204,7 @@ func TestFunnel_Start(t *testing.T) {
 					Event:  Event{Type: RetrySyncEventType}, // 7s
 					Result: Result{},
 				},
+				// No more retry events until another event sets ResetRetryBackoff=true
```
nan-yu, line 207: We can verify the backoff's steps to validate no more retry.

karlkfi: I added another SyncEvent without ResetRetryBackoff to confirm that a RetryEvent won't be called until the ResetRetryBackoff is true.

nan-yu (approved):
/lgtm

PR #1476, opened 2024-11-05
chore: SyncStatusClient return status.Errors

karlkfi:
Returning status.Error instead of just error simplifies usage of the return value, which is usually appended to a MultiError or returned directly to state.invalidate(err). The caller doesn't need to append, if there's no other error that needs to be reported to the RSync status.

This change also returns most SetSourceAnnotations errors as APIServer errors, unless the error is from a kubernetes webhook, which may be used to implement OCI image signature validation. Those are returned as Source errors instead, as they were before.

Note: We could pull out the webhook detection change to a different PR, if needed. LMK.


pkg/parse/root_sync_client.go
```diff
@@ -197,15 +204,32 @@ func (p *rootSyncStatusClient) SetSourceAnnotations(ctx context.Context, commit
...
+	if err == nil {
+		return false
+	}
+	var statusErr *apierrors.StatusError
+	if !errors.As(err, &statusErr) {
+		return false
+	}
+	// TODO: check for a specific error code to distinguish source errors, like
+	// OCI image verification failure.
+	return requestDeniedByWebhookPattern.MatchString(statusErr.Error())
```
nan-yu, line 229: It is not reliable to determine the error type based on the error message. If the error message is changed in the future, we have to deal with multiple versions of strings.

pkg/parse/root_sync_client.go
```diff
@@ -197,15 +204,32 @@ func (p *rootSyncStatusClient) SetSourceAnnotations(ctx context.Context, commit
...
+// *apierrors.StatusError
+func isRequestDeniedByWebhookStatusError(err error) bool {
+	if err == nil {
+		return false
+	}
+	var statusErr *apierrors.StatusError
+	if !errors.As(err, &statusErr) {
+		return false
+	}
+	// TODO: check for a specific error code to distinguish source errors, like
```
nan-yu, line 227: This requires user's webhook server to implement the same contract and Config Sync would assume all webhook servers follow the contract. If the webhook server doesn't return the specific error code, it would return an API error, which is also confusing.

nan-yu (commented):
@janetkuo Would you also take a look at the change?

pkg/parse/root_sync_client.go
karlkfi: The only way to know for sure that a webhook rejected the error because of a source error is to have an interface schema or contact in place for the webhook to distinguish itself from other webhooks. Thats why I suggested using a specific HTTP status code, like 430, which is otherwise unused by HTTP and could be interpreted as a source error.

Alternatively we could require the webhook to send a specific error message and detect that. 

What I put here is a compromise to avoid needing a more specific interface contract. But as you mentioned, it's not perfect.

karlkfi: Another alternative would be to just treat all these patch errors as APIServer errors, to be consistent with the other status update methods.

It's much more likely that our users will hit a network error here than a webhook rejection due to source, especially since we only have a couple customers who have asked for the signature verification feature.

pkg/parse/root_sync_client.go
karlkfi: If the webhook doesn't follow the interface contract, the user would be confused, yes. But that's something the user or their cluster admin can fix. 

Regardless, I've removed the TODO. If we want to use a custom status error, that can be added in a separate PR. 

This change is a compromise, not as good as a custom http code, but a solid improvement over what we have now, significantly reducing the number of possible misclassified errors.

pkg/parse/root_sync_client.go
nan-yu: I think I'm okay with returning API errors when validation fails in the webhook server.

karlkfi: I'm fine with just returning a 2002 APIServer error for now.

It has the downside that the user won't be able to detect that it was blocked by their webhook without doing error text matching, but I guess no customers have asked for that yet, and if no one's depending on it we can freely change the error type later, if we want.

karlkfi: I updated the PR to remove the regex and source error. It's all APIServer errors now, unless the InternalError case happens, which it shouldn't.

nan-yu: LGTM

e2e/testcases/oci_signature_verification_test.go
```diff
@@ -63,7 +63,7 @@ func TestAddPreSyncAnnotationRepoSync(t *testing.T) {
 	nt.Must(rootSyncGitRepo.Add(nomostest.StructuredNSPath(repoSyncID.Namespace, repoSyncID.Name), repoSyncOCI))
 	nt.Must(rootSyncGitRepo.CommitAndPush("Set the RepoSync to sync from OCI"))
 	nt.Must(nt.Watcher.WatchObject(kinds.RepoSyncV1Beta1(), repoSyncID.Name, namespaceRepo,
-		testwatcher.WatchPredicates(testpredicates.RepoSyncHasSourceError(status.SourceErrorCode, "no signatures found"))))
+		testwatcher.WatchPredicates(testpredicates.RepoSyncHasSourceError(status.APIServerErrorCode, "no signatures found"))))
```
janetkuo, line 66: There's no problem accessing the API server when the webhook validation fails. 
https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/reference/errors#knv2002
If we're going with this change, the error reference doc needs to be updated.

nan-yu (approved):
/lgtm

e2e/testcases/oci_signature_verification_test.go
karlkfi: The way we use APIServerError today, it is also sent for client-side errors, like http timeouts, context cancellations, etc. It's also used for server-side errors like authn/authz, as well as invalid requests, etc.

So realistically, the "APIServerError" name doesn't really describe what the root cause is. But if we want to change that, we would need to do it everywhere (it's used in like 50 places).

But for now, I can update the doc to explain that it's more general and has many possible root causes.

PR #1477, opened 2024-11-05
 fix: the ignore mutation annotation in the applier

Camila-B:
A cache has been added to Resources which stores the cluster state of all objects with the ignore-mutation annotation. The Applier uses the in-memory config with the declared CS metadata applied on top of it instead of only the declared object.

### DIFF
--- a/e2e/testcases/ignore_mutation_test.go
+++ b/e2e/testcases/ignore_mutation_test.go
@@ -1,3 +1,17 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package e2e
 
 import (
--- a/e2e/testcases/ignore_mutation_test.go
+++ b/e2e/testcases/ignore_mutation_test.go
@@ -224,6 +224,7 @@ func TestAnnotationDrift(t *testing.T) {
 		core.Annotation("season", "summer"))
 	nt.Must(rootSyncGitRepo.Add("acme/ns.yaml", updatedNamespace))
 	nt.Must(rootSyncGitRepo.CommitAndPush("update namespace"))
+	firstCommitHash := rootSyncGitRepo.MustHash(nt.T)
 
 	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), "bookstore", "",
 		testwatcher.WatchPredicates(
@@ -237,6 +238,7 @@ func TestAnnotationDrift(t *testing.T) {
 
 	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), nsObj.Name, "",
 		testwatcher.WatchPredicates(
+			testpredicates.HasAnnotation(metadata.SyncTokenAnnotationKey, firstCommitHash),
 			testpredicates.HasAnnotation("season", "winter"),
 		)))
 
@@ -245,9 +247,12 @@ func TestAnnotationDrift(t *testing.T) {
 	nsObj2 := k8sobjects.NamespaceObject("new-ns")
 	nt.Must(rootSyncGitRepo.Add("acme/ns2.yaml", nsObj2))
 	nt.Must(rootSyncGitRepo.CommitAndPush("add another namespace"))
+	nt.Must(nt.WatchForAllSyncs())
+	secondCommitHash := rootSyncGitRepo.MustHash(nt.T)
 
 	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), nsObj.Name, "",
 		testwatcher.WatchPredicates(
+			testpredicates.HasAnnotation(metadata.SyncTokenAnnotationKey, secondCommitHash),
 			testpredicates.HasAnnotation("season", "winter"),
 		)))
 
--- a/e2e/testcases/ignore_mutation_test.go
+++ b/e2e/testcases/ignore_mutation_test.go
@@ -298,13 +298,9 @@ func TestDriftKubectlAnnotateConfigSyncAnnotation(t *testing.T) {
 		nt.T.Fatalf("got `kubectl annotate namespace bookstore --overwrite %s=fall` error %v %s, want return nil", metadata.ResourceManagementKey, err, out)
 	}
 
-	time.Sleep(10 * time.Second)
-
 	// Remediator SHOULD NOT correct it
-	err = nt.Validate("bookstore", "", &corev1.Namespace{}, testpredicates.HasAnnotation(metadata.ResourceManagementKey, "fall"))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), "bookstore", "",
+		testwatcher.WatchPredicates(testpredicates.HasAnnotation(metadata.ResourceManagementKey, "fall"))))
 }
 
 // TestDriftKubectlAnnotateDeleteManagedFieldsWithIgnoreMutationAnnotation
@@ -331,13 +327,9 @@ func TestDriftKubectlAnnotateDeleteManagedFieldsWithIgnoreMutationAnnotation(t *
 		nt.T.Fatalf("got `kubectl annotate namespace bookstore season-` error %v %s, want return nil", err, out)
 	}
 
-	time.Sleep(10 * time.Second)
-
 	// Remediator SHOULD NOT correct it
-	err = nt.Validate("bookstore", "", &corev1.Namespace{}, testpredicates.MissingAnnotation("season"))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), "bookstore", "",
+		testwatcher.WatchPredicates(testpredicates.MissingAnnotation("season"))))
 
 	// The reason we need to stop the webhook here is that the webhook denies a request to modify Config Sync metadata
 	// even if the resource has the `client.lifecycle.config.k8s.io/mutation` annotation.
@@ -369,10 +361,8 @@ func TestDriftKubectlAnnotateDeleteManagedFieldsWithIgnoreMutationAnnotation(t *
 	time.Sleep(10 * time.Second)
 
 	// Remediator SHOULD NOT correct it
-	err = nt.Validate("bookstore", "", &corev1.Namespace{}, testpredicates.MissingAnnotation(metadata.ResourceManagementKey))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), "bookstore", "",
+		testwatcher.WatchPredicates(testpredicates.MissingAnnotation(metadata.ResourceManagementKey))))
 }
 
 // TestAddIgnoreMutationAnnotationDirectly verifies the behavior of the applier when the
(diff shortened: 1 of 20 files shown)


pkg/diff/diff.go
```diff
@@ -121,35 +126,34 @@ func (d Diff) updateType(scope declared.Scope, syncName string) Operation {
...
 		return ManagementConflict
 	case differ.ManagementDisabled(d.Declared) && canManage:
-		if metadata.HasConfigSyncMetadata(d.Actual) {
-			manager := d.Actual.GetAnnotations()[metadata.ResourceManagerKey]
-			if manager == "" || manager == declared.ResourceManager(scope, syncName) {
-				return Abandon
-			}
-		}
-		return NoOp
+		return Abandon
```
nan-yu, line 156: I think the old behavior makes more sense.
- If the actual object doesn't have any CS metadata, we can return a no-op directly. 
- If the actual object has CS metadata, but its manager is different from the current reconciler, we should also return a no-op. Abandon will remove CS metadata, which might get reverted by other manager.

Let's keep the old behavior here and I'll update the design doc accordingly.

pkg/metadata/annotations.go
```diff
@@ -177,7 +177,7 @@ const (
 // This annotation is set by Config Sync on a managed resource.
 const OwningInventoryKey = "config.k8s.io/owning-inventory"
 
-// HNCManagedBy is the annotation that indicates the namespace hierarchy is
+// HNCManagedBy is the annotatiƒon that indicates the namespace hierarchy is
```
nan-yu, line 180: typo

pkg/parse/updater.go
```diff
@@ -125,7 +127,9 @@ func (u *Updater) update(ctx context.Context, cache *cacheForCommit) status.Mult
 	// Apply the declared resources
 	if !cache.applied {
 		declaredObjs, _ := u.Resources.DeclaredObjects()
-		if err := u.apply(ctx, declaredObjs, cache.source.commit); err != nil {
+		klog.Infof("%v objects that were declared: %v", len(declaredObjs), core.GKNNs(declaredObjs))
```
nan-yu, line 130: The applier will log declared objects. Do we need an extra log here?
If it is still needed, I would suggest making it a V(3) level log.

pkg/remediator/reconcile/reconciler.go
```diff
@@ -168,6 +169,29 @@ func (r *reconciler) remediate(ctx context.Context, id core.ID, objDiff diff.Dif
...
+		uObj := &unstructured.Unstructured{}
+
+		uObj.SetGroupVersionKind(objDiff.Declared.GetObjectKind().GroupVersionKind())
+		uObj.SetName(objDiff.Declared.GetName())
+		csAnnotations, csLabels := metadata.GetConfigSyncMetadata(objDiff.Declared)
+
+		uObj.SetAnnotations(objDiff.Actual.GetAnnotations())
+		uObj.SetLabels(objDiff.Actual.GetLabels())
+		metadata.RemoveConfigSyncMetadata(uObj)
+		core.RemoveAnnotations(uObj, metadata.LifecycleMutationAnnotation)
```
nan-yu, line 188: Do we need to remove the LifecycleMutationAnnotation explicitly? csAnnotations should reset it.

Camila-B: If Actual has the LifecycleMutationAnnotation but Declared doesn't, then uObj will have the LifecycleMutationAnnotation. That's why it's needed

pkg/parse/updater.go
Camila-B: No it's not necessary. It's left over from testing. I went ahead and removed it

pkg/metadata/annotations.go
Camila-B: Fixed

Camila-B:
I was able to extract the Remediator changes and will be creating a separate PR for them

pkg/applier/applier.go
```diff
@@ -842,3 +860,38 @@ func (s *supervisor) abandonObject(ctx context.Context, obj client.Object) error
...
+				// Object doesn't exist on the cluster
+				if apierrors.IsNotFound(err) {
+					continue
+				}
+
+				if err != nil {
+					return err
+				}
+
+				declaredResources.UpdateIgnored(uObj)
```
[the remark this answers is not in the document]
Camila-B: Done

pkg/applier/applier.go
```diff
@@ -504,12 +511,19 @@ func (s *supervisor) checkInventoryObjectSize(ctx context.Context, c client.Clie
...
 
 // applyInner triggers a kpt live apply library call to apply a set of resources.
-func (s *supervisor) applyInner(ctx context.Context, eventHandler func(Event), objs []client.Object) (ObjectStatusMap, *stats.SyncStats) {
+func (s *supervisor) applyInner(ctx context.Context, eventHandler func(Event), declaredResources *declared.Resources) (ObjectStatusMap, *stats.SyncStats) {
 	s.checkInventoryObjectSize(ctx, s.clientSet.Client)
 	isDestroy := false
 
 	syncStats := stats.NewSyncStats()
 	objStatusMap := make(ObjectStatusMap)
+	objs, _ := declaredResources.DeclaredObjects()
```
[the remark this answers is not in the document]
Camila-B: Doesn't seem like it. I removed it from DeclaredObjects

pkg/diff/diff.go
Camila-B: Ok I changed it back to the old behaviour. I also created a separate PR for the Remediator changes #1486

pkg/declared/resources.go
```diff
@@ -23,34 +23,114 @@ import (
...
+	if found {
+		oCopy := o.DeepCopyObject().(client.Object)
+		return oCopy, found
+	}
+
+	return o, found
+}
+
+// IgnoredObjects returns a slice with a copy of all ignore-mutation objects in the ignoredObjsMap
+func (r *Resources) IgnoredObjects() []client.Object {
```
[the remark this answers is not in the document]
Camila-B: Does that mean we should only keep 1? i.e. there should only be 1 function that returns the cache **OR** a slice of ignored objects?

pkg/applier/applier.go
```diff
@@ -842,3 +860,38 @@ func (s *supervisor) abandonObject(ctx context.Context, obj client.Object) error
...
+				if err != nil {
+					return err
+				}
+
+				declaredResources.UpdateIgnored(uObj)
+			}
+		}
+	}
+
+	if len(declaredResources.IgnoredObjects()) > 0 {
```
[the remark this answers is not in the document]
Camila-B: Done

pkg/declared/resources.go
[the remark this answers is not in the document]
Camila-B: Actually, I think that I will just keep 1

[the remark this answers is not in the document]
Camila-B: Sure, done. I kept `IgnoredObjects()` and replaced each call to `GetIgnoredObjsCache()` with the `*declared.Resources` object itself

pkg/applier/utils.go
```diff
@@ -46,6 +50,35 @@ func partitionObjs(objs []client.Object) ([]client.Object, []client.Object) {
...
+
+		if found && !deleted {
+			cachedCopy, _ := client.Object.DeepCopyObject(cachedObj).(client.Object)
+
+			// Removes problematic fields
+			csAnnotations, csLabels := metadata.GetConfigSyncMetadata(dObj)
+			core.AddAnnotations(cachedCopy, csAnnotations)
+			core.AddLabels(cachedCopy, csLabels)
+			uObj, _ := reconcile.AsUnstructuredSanitized(cachedCopy)
+			unstructured.RemoveNestedField(uObj.Object, "metadata", "managedFields")
```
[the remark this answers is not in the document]
Camila-B: If it's not removed, I get the following error on apply `KNV2009: failed to apply Namespace, /bookstore: metadata.managedFields must be nil`

pkg/declared/resources.go
```diff
@@ -23,34 +23,114 @@ import (
...
-	// objectMap is a map of object IDs to the unstructured format of those
+	// declaredObjectsMap is a map of object IDs to the unstructured format of those
 	// objects. Note that the pointer to this map is threadsafe but the map itself
 	// is not threadsafe. This map should never be returned from a function
 	// directly. The map should never be written to once it has been assigned to
 	// this reference; it should be treated as read-only from then on.
-	objectMap *orderedmap.OrderedMap[core.ID, *unstructured.Unstructured]
+	declaredObjectsMap *orderedmap.OrderedMap[core.ID, *unstructured.Unstructured]
+
+	// mutationIgnoreObjectsMap is a map of object IDs to the Object of objects
```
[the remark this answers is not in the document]
Camila-B: Done

e2e/testcases/ignore_mutation_test.go
```diff
@@ -0,0 +1,261 @@
...
+func TestDeclareExistingObjectWithIgnoreAnnotation(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.DriftControl, ntopts.SyncWithGitSource(nomostest.DefaultRootSyncID, ntopts.Unstructured))
+	rootSyncGitRepo := nt.SyncSourceGitReadWriteRepository(nomostest.DefaultRootSyncID)
+
+	nt.T.Log("Add an umanaged namespace with the ignore-mutation annotation using kubectl ")
+	nsObj := k8sobjects.NamespaceObject("bookstore",
+		core.Annotation(metadata.LifecycleMutationAnnotation, metadata.IgnoreMutation),
+	)
+
+	nt.Must(nt.KubeClient.Apply(nsObj))
```
Camila-B, line 77: I addressed [this comment](https://github.com/GoogleContainerTools/kpt-config-sync/pull/1486/files/3bb9a0ad08e60bb96d435660685ce6d33c980830#r1838504954) here @nan-yu. However, when I run the test, it now fails with the ignore-mutation not being removed from the namespace on the cluster. It seems like it's because the ignore mutation annotation is adopted by the nomostest field manager on `nt.KubeClient.Apply`. How should the updated applier code handle this case?

e2e/testcases/managed_resources_test.go
```diff
@@ -781,9 +781,9 @@ func TestDriftKubectlAnnotateManagedFieldWithIgnoreMutationAnnotation(t *testing
 			))
 	})
 	tg.Go(func() error {
-		// Note: this proves that the applier does not currently honor the ignore-mutation annotation.
```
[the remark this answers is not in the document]
Camila-B: I merged and added those tests in `ignore_mutation_test.go`

pkg/syncer/reconcile/as_unstructured.go
```diff
@@ -56,5 +57,6 @@ func AsUnstructuredSanitized(o client.Object) (*unstructured.Unstructured, statu
 
 	unstructured.RemoveNestedField(u.Object, "metadata", "creationTimestamp")
 	unstructured.RemoveNestedField(u.Object, "status")
+	unstructured.RemoveNestedField(u.Object, "metadata", "managedFields")
```
nan-yu, line 60: The `managedFields` field is tracked by the Kubernetes API server. Custom controllers should not manage this field.

e2e/testcases/ignore_mutation_test.go
```diff
@@ -0,0 +1,429 @@
...
+		testwatcher.WatchPredicates(
+			testpredicates.HasAnnotation("season", "winter"),
+		)))
+
+	// Push a new commit to trigger a new apply.
+	nt.T.Log("Add another namespace to Git to run the applier")
+	nsObj2 := k8sobjects.NamespaceObject("new-ns")
+	nt.Must(rootSyncGitRepo.Add("acme/ns2.yaml", nsObj2))
+	nt.Must(rootSyncGitRepo.CommitAndPush("add another namespace"))
+
```
nan-yu, line 248: Add `nt.Must(nt.WatchForAllSyncs())` before validating the object to ensure the applier applies the new commit.

e2e/testcases/ignore_mutation_test.go
```diff
@@ -0,0 +1,429 @@
...
+	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), nsObj.Name, "",
+		testwatcher.WatchPredicates(
+			testpredicates.HasAnnotation("season", "winter"),
```
nan-yu, line 251: Besides validating the `season` annotation, we should also verify the value of the `configmanagement.gke.io/token` annotation on the Namespace is changed.

Without this, it is hard to tell if the Namespace is updated or not because Line240 already indicates it has a `season: winter` annotation.

Camila-B: Done

e2e/testcases/ignore_mutation_test.go
Camila-B: Done

nan-yu (approved):
/lgtm

PR #1481, opened 2024-11-07
fix: surface source status update error

karlkfi:
The setSourceStatusErr here was being re-initialized inside an if statement, which means it would always be nil outside the if statement.

This fix allows the source status update error to be returned and logged correctly when the checkpoint is invalidated.

The bug this fixes is relatively minor, because SetSourceStatus has its own internal retries. So it's unlikely that SetSourceStatus will return an error unless something is seriously wrong. However, if it does happen, it needs to be logged because the RSync status won't have been updated.


nan-yu (approved):
/lgtm

PR #1482, opened 2024-11-07
chore: extract reconciler.Render method

karlkfi:
- The new Render method handles checking the done file written by the hydration-controller sidecar.
- The code has been cleaned up a bit to handle errors as soon as possible, following the Return Early Pattern, which reduces cyclomatic complexity and improves readability.
- To avoid the Render method calling state.invalidate(errs) directly, the Render method now returns a TransientError when rendering is still in-progress. This is caught by state.invalidate and handled by logging at the info level, instead of as an error. It still causes a retry
- Added a RenderedCommit method that returns read errors instead of just logging them. This is used by the Render method, which should error if the done file cannot be read, unless it was not found. The original doneCommit now wraps RenderedCommit and log errors when called by the hydration-controller.
- Remove ReconcilerState.reset, which is no longer needed, because returning the TransientError causes state.invalidate to be called, which handles logging and setting needToRetry and lastApplied.

### DIFF
--- a/pkg/parse/run.go
+++ b/pkg/parse/run.go
@@ -16,6 +16,7 @@ package parse
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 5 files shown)


pkg/parse/run.go
```diff
@@ -154,43 +155,10 @@ func DefaultRunFunc(ctx context.Context, r Reconciler, trigger string) RunResult
...
 	// set the rendering status by checking the done file.
 	if opts.RenderingEnabled {
-		doneFilePath := opts.RepoRoot.Join(cmpath.RelativeSlash(hydrate.DoneFile)).OSPath()
-		_, err := os.Stat(doneFilePath)
-		if os.IsNotExist(err) || (err == nil && hydrate.DoneCommit(doneFilePath) != gs.Commit) {
-			rs.Message = RenderingInProgress
-			rs.LastUpdate = nowMeta(opts)
-			klog.V(3).Info("Updating rendering status (before parse)")
-			setRenderingStatusErr := r.SyncStatusClient().SetRenderingStatus(ctx, state.status.RenderingStatus, rs)
-			if setRenderingStatusErr == nil {
```
sdowell, line 174: In the refactored code this code path now returns a transient error, which then calls `state.invalidate`. Is it intentional that `state.invalidate` is called now when it wasn't before? I see the logging behavior is changed for a `TransientError` but it still seems like a behavior change

karlkfi: FWIW, the desired behavior is that when rendering is still in progress, the DefaultRunFunc needs to stop and trigger a retry, with the normal backoff.

Since invalidate triggers a retry and is normally called when there's an error, we can use that pattern instead, and just use the TransientError to avoid the error level log message, while still logging that rendering is in progress at the info level.

That said, I left a TODO in there, because I'm not actually sure the resetCache call is needed. I'm not sure what case that handles.

karlkfi: There are also other cases where I think we can use this TransientError behavior later, like changing how SourceCommitAndDirWithRetry works to just return a TransientError, instead of handling its own retry backoff.

karlkfi: AFAICT, the resetCache call will cause re-read (listing files), re-parse, and re-update. Re-reading sounds desirable, if rendering may have changed the files, but I'm not sure if there's a case where upon getting to this code the cache would already be populated. I would expect that if it's rendering there must be a new commit that hasn't been read, parsed, or updated yet.

Maybe you know?

karlkfi: Made a different PR with just the resetCache call removed, to see if it breaks any e2e tests: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1483

sdowell: At this point I think you could have a cache that is populated from a previous commit that was read/parsed/applied. If we are then observing that a new commit is being rendered, it makes sense to me that we would want to wipe the cache of any state from the previous commit so that the new commit is processed with a clean cache

sdowell (approved):
/lgtm

PR #1483, opened 2024-11-07
[WIP] remove resetCache call

karlkfi:
checking to see if removing this resetCache call causes any tests to fail.


karlkfi:
From reading the code, it appears that this resetCache is probably required, but it doesn't actually fail any tests when removed. 

In fact, it seems like we should resetCache when rendering succeeds after a new commit, not just when in-progress. Otherwise parsing will be skipped, because the parse stage doesn't store the last commit it parsed, just whether parsing succeeded or not.

Both of these are problematic, but without tests that fail, it's hard to know for sure that there's not some other variable or field triggering a parse after a successful render. The rendering tests all pass.

It's possible that the tests themselves are broken or too lenient. Like maybe they only pass after a FullSync or namespace update or management conflict, which all reset the cache.

PR #1484, opened 2024-11-08
chore: clarify syncPath vs syncDir

karlkfi:
- syncDir is the relative path from the root of the source to the
  directory to sync to the cluster. It's specified by spec.git.dir,
  spec.oci.dir, or "." for helm.
- syncPath is the absolute physical path to the syncDir on the
  local filesystem. The physical path has resolved symlinks and
  includes the commit-sha/image-checksum/chart-version.
- sourcePath has two meanings, depending on context:
  - The absolute logical path to where the source was fetched.
  - The relative physical path to the file that defines a specific
    object, used in error messages.
- Remove Files.currentSyncDir, which was only used for logging.
  Move the logging up a level from readConfigFiles to readFromSource,
  which already knows if the syncPath changed.


pkg/hydrate/controller.go
```diff
@@ -75,23 +75,23 @@ func (h *Hydrator) Run(ctx context.Context) {
 	defer runTimer.Stop()
 	rehydrateTimer := time.NewTimer(h.RehydratePeriod)
 	defer rehydrateTimer.Stop()
-	absSourceDir := h.absSourceDir()
+	sourcePath := h.sourcePath()
 	var hydrateErr HydrationError
 	var srcCommit string
-	var syncDir cmpath.Absolute
+	var syncPath cmpath.Absolute
```
sdowell, line 81: Could you document the difference between `sourcePath` and `syncPath`?

karlkfi: done

sdowell (approved):
/lgtm

PR #1485, opened 2024-11-08
chore: ReconcilerState cleanup

karlkfi:
- Rename ReconcilerState methods to reflect the reconcile stages they are called after, rather than the state that they change. This makes it cleared when the methods should be called and documents what the state change does and why.
- Rework logging for the state transitions handled by the ReconcilerState methods.
- Wrap the checkpoint "lastApplied" syncPath in a struct that captures lastUpdateTime and lastTransitionTime, like a kubernetes condition. This makes it possible to tell both when the last sync suceeded or failed and also when it first suceeded for the current commit (as long as it hasn't since failed). For now, this is only visible in the logs, but we may want to expose it via the RSync status in the future.

Ideas for future cleanup:
- Add state transition methods for each stage (fetch, render, read, parse, update)
- Untangle the Render & Read stages. Today the render status isn't updated as successful until after `readFromSource` succeeds, because it handles calling `parseHydrationState` and then looks for kustomization files to enable rendering after calling `readConfigFiles`.
- Untangle the Parse & Update stages. It's not super clear why these share a wrapper func, but it's probably just because the DefaultRunFunc was too big. They can probably be split up.
- Split up the `cacheForCommit`. It hold state for multiple stages. Each should get their own struct with lastUpdateTime, lastTransitionTime, and well named methods to update them.
- Track the syncPath when each stage completes, so we know when to skip a stage, rather that having earlier stages partially/fully reset the cache. For example, `parserResultUpToDate` checks a `hasParserResult` bool, but it should probably store the syncPath and check if it changed. If all the stages did that, then RecordRenderInProgress,  RecordReadSuccess, and RecordReadFailure wouldn't need to reset the cache. Only RecordFullSyncStart would reset the cache, because it's timer based instead of state based.

### DIFF
--- a/pkg/parse/state.go
+++ b/pkg/parse/state.go
@@ -15,15 +15,20 @@
 package parse
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
+	"kpt.dev/configsync/pkg/importer/filesystem/cmpath"
 	"kpt.dev/configsync/pkg/status"
 )
 
+const invalidSyncPath = ""
+
 // ReconcilerState is the current state of the Reconciler, including progress
-// indicators and in-memory cache for each of the reconciler stages:
+// indicators and in-memory cache for each of the reconcile stages:
 // - Fetch
-// - Read
 // - Render/Hydrate
+// - Read
 // - Parse/Validate
 // - Update
 //
@@ -33,8 +38,10 @@ import (
 // TODO: break up cacheForCommit into phase-based caches
 // TODO: move sourceState into ReconcilerState so the RSync spec and status are next to each other
 type ReconcilerState struct {
-	// checkpointedSyncPath is the last checkpointed syncPath.
-	checkpointedSyncPath string
+	// checkpoint tracks the sourcePath the last time reconciling succeeded or
+	// failed. Reconciling includes multiple stages: fetch, render, read, parse,
+	// and update.
+	checkpoint checkpoint
 
 	// status contains fields that map to RSync status fields.
 	status *ReconcilerStatus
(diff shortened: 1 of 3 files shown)


pkg/parse/state.go
```diff
@@ -45,46 +52,102 @@ type ReconcilerState struct {
 	syncErrorCache *SyncErrorCache
 }
 
-func (s *ReconcilerState) checkpoint() {
-	applied := s.cache.source.syncPath.OSPath()
-	if applied == s.checkpointedSyncPath {
-		return
+type checkpoint struct {
+	// sourcePath caches the source patch that was last successfully applied.
```
sdowell, line 56: typo: source patch -> source path

karlkfi: fixed

pkg/parse/state.go
```diff
@@ -45,46 +52,102 @@ type ReconcilerState struct {
...
 
-// invalidate logs the errors, clears the state tracking information.
-// invalidate does not clean up the `s.cache`.
-func (s *ReconcilerState) invalidate(errs status.MultiError) {
+// RecordRenderInProgress is called when waiting for rendering status. It
+// invalidates the checkpoint, requests a retry, and resets the cacheForCommit,
+// which tells the retry to re-parse from source.
+func (s *ReconcilerState) RecordRenderInProgress(c clock.Clock) {
+	klog.Info("Rendering in progress")
+	s.updateCheckpoint(c, invalidSyncPath)
```
nan-yu, line 105: The function only needs to reset cache `s.cache = cacheForCommit`.
The caller returns an error, which will call `RecordFailure` to updateCheckpoint and set s.cache.needToRetry to true.

karlkfi: Good catch. I swear I had it doing that at some point. Musta lost it in one of the copies to a new PR.

sdowell (commented):
LGTM after Nan's comment is addressed
/lgtm

pkg/parse/state.go
karlkfi: fixed

nan-yu (approved):
/lgtm

sdowell (approved):
/lgtm

karlkfi:
Had to fix a lint error, please re-lgtm

PR #1487, opened 2024-11-12
Feature: pause hydration when image verification fails

tiffanny29631:


manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -70,10 +70,14 @@ data:
            - name: service-account # This is needed to prevent any credentials from being exposed to malicious users.
              mountPath: /var/run/secrets/kubernetes.io/serviceaccount
              readOnly: true
+           - name: shared-volume  # Mount the shared volume for reading
+             mountPath: /shared
```
nan-yu, line 74: The volume name is too general. It seems the volume is supposed to be written by the reconciler and read by the hydration-controller. I would suggest renaming to something like `/reconciler-signals` to make it self-explanatory.

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -265,6 +271,8 @@ data:
            emptyDir: {}
          - name: kube
            emptyDir: {}
+         - name: shared-volume
+           emptyDir: {}  # Shared volume for ready-to-render file
```
nan-yu, line 275: `Shared volume for ready-to-render file` -> `A shared volume that allows the reconciler to send signals to the hydration-controller`.

pkg/hydrate/controller.go
```diff
@@ -100,12 +103,25 @@ func (h *Hydrator) Run(ctx context.Context) {
 						srcCommit, err)
 				}
 			} else if doneCommit(h.DonePath.OSPath()) != srcCommit {
-				// If the commit has been processed before, regardless of success or failure,
-				// skip the hydration to avoid repeated execution.
-				// The rehydrate ticker will retry on the failed commit.
-				hydrateErr = h.hydrate(srcCommit, syncPath)
-				if err := h.complete(srcCommit, hydrateErr); err != nil {
-					klog.Errorf("failed to complete the rendering execution for commit %q: %v", srcCommit, err)
+				absShared := cmpath.Absolute("/shared")
```
nan-yu, line 106: Either define the volume path as a constant, or define it as a flag (similar to other root directories).

pkg/hydrate/controller.go
```diff
@@ -100,12 +103,25 @@ func (h *Hydrator) Run(ctx context.Context) {
...
+				readyToRenderFile := absShared.Join(cmpath.RelativeSlash(ReadyToRenderFile)).OSPath()
+				readyToRenderCommit, err := ExtractCommit(readyToRenderFile)
+				if err != nil {
+					klog.Error(err)
+					h.complete(readyToRenderCommit, NewActionableError(err))
```
nan-yu, line 111: If the ready-to-render file doesn't exist, should we mark the hydration complete?
`h.complete` updates the done file and the rendering error file. 
In this case, no rendering is being executed. Moreover, if the commit is not ready to render, the reconciler won't move forward with checking the rendering state.

I think we can simply log a warning in the hydration-controller.
Then this can be simplified as
```
else if readyToRenderFile(h.ReadyToRenderPath.OSPath()) == srcCommit && doneCommit(h.DonePath.OSPath()) != srcCommit {
   .... 
}

Basically, we only run rendering when the commit is ready and the commit is not rendered before.
```

nan-yu (commented):
Is it possible to add unit tests for the blocking feature?

pkg/parse/run.go
```diff
@@ -120,6 +120,13 @@ func DefaultRunFunc(ctx context.Context, r Reconciler, trigger string) RunResult
 	if gs.Errs == nil {
 		if err := r.SyncStatusClient().SetImageToSyncAnnotation(ctx, gs.Commit); err != nil {
```
nan-yu, line 121: A comment that is not specific to the unblocking feature. Now that we have `ReconcilerState.ReconcilerStatus` populated from RSync, we can skip setting the annotation and updating the ready-to-render file when the commit is not changed.

In `type ReconcilerStatus struct {}`, we can add a field `ImageToSyncAnnotation` to cache the annotation, and here we update `if gs.Errs == nil {` to `if gs.Errs == nil && r.ReconcilerState().status.ImageToSyncAnnotation != gs.Commit {`.

This can be a separate PR.

manifests/templates/reconciler-manager-configmap.yaml
tiffanny29631: Done

manifests/templates/reconciler-manager-configmap.yaml
tiffanny29631: Done

pkg/hydrate/controller.go
tiffanny29631: Done, only log a warning if hydration is skipped due to unexpected readyToRenderCommit.

tiffanny29631:
>Is it possible to add unit tests for the blocking feature?

Working on an e2e test now. The check of readyToRenderCommit happens in the controller Run function so is not easy to emulate, we could check for the warning message in hydration-controller log when commit is not as expected, and pass when an image with dry config syncs correctly - meaning hydration is resumed.

tiffanny29631:
@nan-yu e2e test in place, ready for another look.

pkg/hydrate/controller.go
```diff
@@ -99,20 +104,34 @@ func (h *Hydrator) Run(ctx context.Context) {
...
+					// skip the hydration to avoid repeated execution.
+					// The rehydrate ticker will retry on the failed commit.
+					// Only proceed to rendering if the ready-to-render file contains the srcCommit
+					hydrateErr = h.hydrate(srcCommit, syncPath)
+					if err := h.complete(srcCommit, hydrateErr); err != nil {
+						klog.Errorf("failed to complete the rendering execution for commit %q: %v", srcCommit, err)
+					}
+				} else {
+					klog.Warningf("skip hydration as readyToRenderCommit does not match srcCommit, want %s, got %s", srcCommit, readyToRenderCommit)
 				}
```
nan-yu, line 120: Suggest changing to a switch block to make it clear.

```
else {
  doneCommit := extractCommit(h.DonePath.OSPath())
  readyToRenderFile := h.ReconcilerSignalDir.Join(cmpath.RelativeSlash(ReadyToRenderFile)).OSPath()
  readyToRenderCommit := extractCommit(readyToRenderFile)
  switch {
  case doneCommit == srcCommit:
    // no-op
    // If the commit has been processed before, skip the hydration to avoid repeated execution regardless of success or failure.
	// The rehydrate ticker will retry on the failed commit.
  case doneCommit != srcCommit && readyToRenderCommit != srcCommit:
    klog.Warningf("skip hydration as readyToRenderCommit does not match srcCommit, want %s, got %s", srcCommit, readyToRenderCommit)
  case doneCommit != srcCommit && readyToRenderCommit == srcCommit:
    // Only proceed to rendering if the ready-to-render file contains the srcCommit
	hydrateErr = h.hydrate(srcCommit, syncPath)
	if err := h.complete(srcCommit, hydrateErr); err != nil {
	  klog.Errorf("failed to complete the rendering execution for commit %q: %v", srcCommit, err)
	}
  }
}
```

pkg/parse/run.go
```diff
@@ -132,6 +132,18 @@ func (r *reconciler) Reconcile(ctx context.Context, trigger string) ReconcileRes
 		state.RecordFailure(opts.Clock, errs)
 		return result
 	}
+	if newSourceStatus.Errs == nil {
```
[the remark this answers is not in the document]
nan-yu: The code only writes the ready-to-render file. Do we need to remove it when there is source error?

pkg/hydrate/controller.go
tiffanny29631: Done

pkg/parse/run_test.go
```diff
@@ -1135,6 +1146,12 @@ func TestReconciler_Reconcile(t *testing.T) {
 			require.NoError(t, err)
 			testutil.AssertEqual(t, tc.expectedRootSyncFunc(sourceRoot), rs)
 
+			if tc.imageVerified {
+				readyToRenderFilePath := filepath.Join(reconcilerSignalDir, "ready-to-render")
+				_, err = os.Stat(readyToRenderFilePath)
```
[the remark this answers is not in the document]
tiffanny29631: Done

pkg/parse/run.go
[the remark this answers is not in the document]
tiffanny29631: Moved the update of ready-to-render in the fetch action, as it's tightly coupled with the result of setting image-to-sync annotation.

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -64,16 +64,21 @@ data:
...
              mountPath: /var/run/secrets/kubernetes.io/serviceaccount
              readOnly: true
+           - name: reconciler-signals
+             mountPath: /reconciler-signals
+             readOnly: true
            imagePullPolicy: IfNotPresent
            securityContext:
              allowPrivilegeEscalation: false
              readOnlyRootFilesystem: false
+
```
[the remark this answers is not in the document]
tiffanny29631: Done.

e2e/testcases/otel_collector_test.go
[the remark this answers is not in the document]
tiffanny29631: Done

e2e/testcases/oci_signature_verification_test.go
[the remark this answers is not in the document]
tiffanny29631: Changed to use wet repo as hydration is not verified in e2e.

e2e/testdata/hydration/kustomize-components-more/base/kustomization.yaml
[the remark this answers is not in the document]
tiffanny29631: Changed to use wet repo as hydration is not verified in e2e.

e2e/testdata/hydration/namespace-repo-more/base/kustomization.yaml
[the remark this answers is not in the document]
tiffanny29631: Changed to use wet repo as hydration is not verified in e2e.

cmd/reconciler/main.go
```diff
@@ -102,27 +102,31 @@ var (
 
 	dynamicNSSelectorEnabled = flag.Bool("dynamic-ns-selector-enabled", util.EnvBool(reconcilermanager.DynamicNSSelectorEnabled, false), "")
 
-	webhookEnabled = flag.Bool("webhook-enabled", util.EnvBool(reconcilermanager.WebhookEnabled, false), "")
+	webhookEnabled       = flag.Bool("webhook-enabled", util.EnvBool(reconcilermanager.WebhookEnabled, false), "")
+	reconcilerSignalsDir = flag.String(flags.reconcilerSignalDir, "/reconciler-signals",
+		"The absolute path in the container that contains the image digest which has passed signature verification")
```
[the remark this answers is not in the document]
tiffanny29631: Done

pkg/hydrate/controller.go
```diff
@@ -99,20 +104,38 @@ func (h *Hydrator) Run(ctx context.Context) {
...
 			}
 			runTimer.Reset(h.PollingPeriod) // Schedule re-run attempt
 		}
 	}
 }
 
+// isReadyToRender returns true if commit in ready-to-render file is the same as
+// commit acquired from source
+func (h *Hydrator) isReadyToRender(srcCommit string) (string, bool) {
```
nan-yu, line 133: Is this function still needed?

pkg/parse/run.go
```diff
@@ -218,6 +218,13 @@ func (r *reconciler) fetch(ctx context.Context) (*SourceStatus, cmpath.Absolute,
 	if newSourceStatus.Errs == nil {
 		if err := r.syncStatusClient.SetImageToSyncAnnotation(ctx, newSourceStatus.Commit); err != nil {
 			newSourceStatus.Errs = status.Append(newSourceStatus.Errs, err)
+		} else {
+			// write the commit into the read-to-render file in hydrated root
```
nan-yu, line 222: typo: `ready-to-render`

pkg/parse/run.go
```diff
@@ -218,6 +218,13 @@ func (r *reconciler) fetch(ctx context.Context) (*SourceStatus, cmpath.Absolute,
...
+			if opts.RenderingEnabled {
```
nan-yu, line 223: nit: this `if` block can be combined with the `else` block -> `else if opts.RenderingEnabled {`

pkg/parse/run.go
```diff
@@ -218,6 +218,13 @@ func (r *reconciler) fetch(ctx context.Context) (*SourceStatus, cmpath.Absolute,
 	if newSourceStatus.Errs == nil {
```
nan-yu, line 218: Do we need to remove the `ready-to-render` file when there is source error?

pkg/hydrate/controller.go
tiffanny29631: Yes, I split it out so that we can do unit test on it.

pkg/parse/run.go

pkg/parse/run.go
tiffanny29631: Added logic to remove this file when a source error occurs, either before or after image verification. If the removal fails, it logs the error without appending it to the source error or interrupting any loop. Since the hydration controller won’t receive the expected commit in the ready-to-render file when a source error exists, rendering is effectively blocked regardless.

pkg/hydrate/controller.go
nan-yu: It is only used in the unit test. If it is not used by the controller, we don't need to unit test this function.

pkg/parse/run.go
```diff
@@ -615,6 +616,24 @@ func unblockHydration(commit string, r Reconciler) status.Error {
...
+// file will not be updated with the expected commit that is expected by
+// hydration-controller, thus rendering will be blocked.
+func blockHydration(r Reconciler) {
+	readyToRenderFile := r.Options().ReconcilerSignalsDir.Join(cmpath.RelativeSlash(hydrate.ReadyToRenderFile)).OSPath()
+	err := os.Remove(readyToRenderFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			klog.Infof("ready-to-render file %s not found, no action needed", readyToRenderFile)
+		}
+		klog.Errorf("failed to removed ready-to-render file %s: %v", readyToRenderFile, err)
```
nan-yu, line 632: typo: removed -> remove

pkg/parse/run.go
```diff
@@ -615,6 +616,24 @@ func unblockHydration(commit string, r Reconciler) status.Error {
...
+// "file not found". This function removes the ready-to-render file at its
+// best effort, since with source errors presenting, the ready-to-render
+// file will not be updated with the expected commit that is expected by
+// hydration-controller, thus rendering will be blocked.
+func blockHydration(r Reconciler) {
+	readyToRenderFile := r.Options().ReconcilerSignalsDir.Join(cmpath.RelativeSlash(hydrate.ReadyToRenderFile)).OSPath()
+	err := os.Remove(readyToRenderFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			klog.Infof("ready-to-render file %s not found, no action needed", readyToRenderFile)
```
nan-yu, line 630: nit: this log can be set at V(3) level.

pkg/parse/run.go
```diff
@@ -584,6 +592,48 @@ func (r *reconciler) startAsyncStatusUpdates(ctx context.Context) <-chan struct{
...
+}
+
+// blockHydration removes the ready-to-render file from the shared
+// directory and returns any error that occurs during the procedure except
+// "file not found". This function removes the ready-to-render file at its
+// best effort, since with source errors presenting, the ready-to-render
+// file will not be updated with the expected commit that is expected by
+// hydration-controller, thus rendering will be blocked.
+func blockHydration(r Reconciler) {
+	readyToRenderFile := r.Options().ReconcilerSignalsDir.Join(cmpath.RelativeSlash(hydrate.ReadyToRenderFile)).OSPath()
```
nan-yu, line 626: This is repeated in both unblock and block functions. Can it be extracted to a variable?

pkg/parse/run.go
```diff
@@ -584,6 +592,48 @@ func (r *reconciler) startAsyncStatusUpdates(ctx context.Context) <-chan struct{
 	return doneCh
 }
 
+// unblockHydration adds the image digest into the ready-to-render file under the
+// shared directory between reconciler and hydration-controller to inform the
+// hydration-controller to proceed rendering
+func unblockHydration(commit string, r Reconciler) status.Error {
+	readyToRenderFile := r.Options().ReconcilerSignalsDir.Join(cmpath.RelativeSlash(hydrate.ReadyToRenderFile)).OSPath()
+	klog.Infof("writing commit %s to ready to render file %s", commit, readyToRenderFile)
```
nan-yu, line 600: nit: `writing commit %s to ready to render file` -> `signaling the hydration-controller that commit %s is ready to render`

pkg/hydrate/controller.go
tiffanny29631: Ah I forgot the logic in Run has changed. Just updated with cleanup.

nan-yu (approved):
/lgtm

PR #1488, opened 2024-11-12
chore: cleanup reconciler set status method usage

karlkfi:
- Use the standard if err != nil pattern for all calls to set status to return early, making the code easier to read and more consistent.
- Remove unused ReconcilerStatus.SyncingConditionLastUpdate
- Use variable names consistently in run.go: newSourceStatus, newRenderStatus, newSyncStatus, and statusErr.


sdowell (approved):
/lgtm

PR #1489, opened 2024-11-12
chore: git sync update to v4.3.0-gke.3__linux_amd64

mikebz:
bumping up the git-sync even though it's not clear that it's addressing any CVEs


tiffanny29631:
Is there any CVE reported before making this build?

mikebz:
in general yes there aer CVE notifications.

sdowell (approved):
This change is acceptable but it doesn't address the CVE findings on the current `git-sync` image. The findings are related to OS packages and require for the image to be rebuilt

/lgtm

mikebz:
yes, this was a low hanging fruit while I was looking into the git sync updates.

PR #1490, opened 2024-11-12
chore: updated the golang.org/x/oauth2 to v0.24.0

mikebz:
http://b/378536256

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842
 	golang.org/x/mod v0.18.0
 	golang.org/x/net v0.26.0
-	golang.org/x/oauth2 v0.20.0
+	golang.org/x/oauth2 v0.24.0
 	google.golang.org/api v0.152.0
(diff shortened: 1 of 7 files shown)


nan-yu (commented):
With the early expiration in the oauth library, do we still need another early expiration in askpass?

karlkfi:
> With the early expiration in the oauth library, do we still need another early expiration in askpass?

Maybe. Maybe not.

For askpass, there's the added complexity that git-sync doesn't request a new token before every git command, and executes multiple in a row. 

But if I had to guess, we probably only need it in the oauth lib. 

Depending on how short the token expiration is set to, it might actually be problematic to reset it too early.

Does this update cover all the cases in b/378536256?

mikebz:
this is just a library update.

nan-yu:
> Does this update cover all the cases in b/378536256?

Nope. The oauth update only covers the ComputeTokenSource, but not `external-account` for fleet workload identity.

In that regards, it is good to keep it in the askpass.

nan-yu (approved):
/lgtm

PR #1491, opened 2024-11-13
chore: DefaultRunFunc -> reconciler.Reconcile

karlkfi:
- Rename DefaultRunFunc as reconciler.Reconciler.
- Remove methods from the Reconciler interface only used by reconciler.Reconcile.
- Add Reconciler.UpdateSyncStatus to simplify event handling.
- Make reconciler methods private that no longer need to be exposed.

### DIFF
--- a/pkg/parse/reconciler.go
+++ b/pkg/parse/reconciler.go
@@ -16,56 +16,36 @@ package parse
 
 import (
 	"context"
-
-	"kpt.dev/configsync/pkg/importer/filesystem/cmpath"
-	"kpt.dev/configsync/pkg/status"
 )
 
-// Reconciler represents a parser that can be pointed at and continuously parse a source.
+// Reconciler reconciles the cluster with config from the source of truth.
 // TODO: Move to reconciler package; requires unwinding dependency cycles
 type Reconciler interface {
 	// Options returns the ReconcilerOptions used by this reconciler.
 	Options() *ReconcilerOptions
-	// SyncStatusClient returns the SyncStatusClient used by this reconciler.
-	SyncStatusClient() SyncStatusClient
-	// Parser returns the Parser used by this reconciler.
-	Parser() Parser
+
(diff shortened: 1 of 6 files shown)


pkg/parse/reconciler.go
```diff
@@ -83,23 +63,14 @@ func (p *reconciler) Options() *ReconcilerOptions {
...
 // ReconcilerState returns the current state of the reconciler.
 func (p *reconciler) ReconcilerState() *ReconcilerState {
 	return p.reconcilerState
 }
 
-// NewRootRunner creates a new runnable parser for parsing a Root repository.
-func NewRootRunner(recOpts *ReconcilerOptions, parseOpts *RootOptions) Reconciler {
+// NewClusterReconciler creates a new reconciler for reconciling cluster-scoped
+// and namespace-scoped resources configured with a RootSync.
+func NewClusterReconciler(recOpts *ReconcilerOptions, parseOpts *RootOptions) Reconciler {
```
sdowell, line 73: nit: I think `NewRootSyncReconciler` and `NewRepoSyncReconciler` may be more intuitive. `Cluster` and `Namespace` seem not as precise as they could be

karlkfi: I considered that. I'm not against it.

It seems to me that it's primarily reconciling the cluster objects (cluster-scoped and/or namespace-scoped) with the source objects, but as a controller it's also reconciling the RSync status with the spec...

karlkfi: But if that wasn't immediately obvious from the name, then I'm ok with your suggestion, which should be pretty clear.

karlkfi: I wish we could rename the resources themselves to be more clear, but that ship sailed before my time.

karlkfi: Renamed.

sdowell (approved):
/lgtm

PR #1494, opened 2024-11-14
fix: out-of-order status updates

karlkfi:
- Wait for the async status updates to complete before continuing after updating.

### DIFF
--- a/pkg/parse/run.go
+++ b/pkg/parse/run.go
@@ -478,10 +478,8 @@ func (r *reconciler) parseAndUpdate(ctx context.Context, trigger string) status.
 		return sourceErrs
 	}
 
-	// Create a new context with its cancellation function.
-	ctxForUpdateSyncStatus, cancel := context.WithCancel(context.Background())
-
-	go r.updateSyncStatusPeriodically(ctxForUpdateSyncStatus)
+	asyncCtx, asyncCancel := context.WithCancel(ctx)
+	asyncDoneCh := r.startAsyncStatusUpdates(asyncCtx)
 
 	klog.V(3).Info("Updater starting...")
 	start := opts.Clock.Now()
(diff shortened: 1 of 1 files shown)


pkg/parse/run.go
```diff
@@ -533,39 +531,46 @@ func (r *reconciler) setSyncStatus(ctx context.Context, newSyncStatus *SyncStatu
...
-				Spec:       state.status.SourceStatus.Spec,
-				Syncing:    true,
-				Commit:     state.cache.source.commit,
-				Errs:       r.ReconcilerState().SyncErrors(),
-				LastUpdate: nowMeta(opts.Clock),
+// startAsyncStatusUpdates starts a goroutine that updates the sync status
+// periodically until the cancel function is called.
+// Wait until the done channel is closed to confirm the goroutine has exited.
+func (r *reconciler) startAsyncStatusUpdates(ctx context.Context) (context.CancelFunc, <-chan struct{}) {
+	ctx, cancelFn := context.WithCancel(ctx)
```
sdowell, line 538: nit: I think this function could just take a context and return a doneCh, and let the caller handle creating the cancel function

karlkfi: That's what it was doing before. I moved it inside the func to avoid the caller needing to manage two contexts, which seemed confusing.

sdowell: Yeah no biggie. I'm curious whether it's conventional to return the cancel function as the second parameter? It seems like it's normally the second return parameter, although I don't know if I've seen it with a doneCh as the other parameter

karlkfi: The most idiomatic way I can think of would be to have a function that takes a Context and returns a doneCh. That's along the lines of what you were proposing. It just makes the caller do more work.

karlkfi: refactored. used async prefix to distinguish the ctx.

sdowell (approved):
/lgtm

PR #1495, opened 2024-11-15
chore: split parse & update methods

karlkfi:
- Split reconciler.parseAndUpdate into parse and update methods.
- Move handling of non-blocking errors and webhook GVK updates out to the Reconcile method.
- Combine parse and parseSource methods.
- Wrap some of the cache fields in a parseResult, to make it clear that those fields are updated atomicly and make a namespace to simplify method names related to parsing.
- Move GVK set building to a parseResult.GVKs method, instead of passing all the objects to webhook.Update.
- Update some of the unit tests to test just the parse method.

I want to do more to clean up cacheForCommit, to add updateResult, and split sourceResult into fetchResult and renderResult, but this seems like enough for one PR.

### DIFF
--- a/pkg/parse/cache.go
+++ b/pkg/parse/cache.go
@@ -17,6 +17,8 @@ package parse
 import (
 	"sort"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


nan-yu (approved):
/lgtm

PR #1498, opened 2024-11-21
fix: tollerate stale resource group discovery

karlkfi:
- This change allows the resource-group-controller-manager to use cached resource group records when the records are stale due to unavailable APIService backends, like metrics-server, which often has only one replica.
- All other discovery errors will now also cause the TypeResolver controller to retry discovery, with logging and retry backoff handled by the controller manager.
- If discovery fails when the TypeResolver is first created, it will cause the process to error and exit. Then the container runtime should restart the container.
- This should prevent ResourceGroup status from marking all resources as NotFound when discovery fails.

Fixes: b/379943950

Note: This only fixes the problem in the resource-group-controller-manager, and only when Aggregated Discovery is enabled, which is what handles server-side caching of the resource groups. We have other similar but different workarounds in place in the reconciler, which uses slightly methods on the DiscoveryClient and RESTMapper.

### DIFF
--- a/pkg/resourcegroup/controllers/typeresolver/typeresolver.go
+++ b/pkg/resourcegroup/controllers/typeresolver/typeresolver.go
@@ -16,6 +16,8 @@ package typeresolver
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"sync"
 
 	"github.com/go-logr/logr"
@@ -23,10 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // TypeResolver keeps the preferred GroupVersionKind for all the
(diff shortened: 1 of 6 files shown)


pkg/resourcegroup/controllers/typeresolver/typeresolver.go
```diff
@@ -16,39 +16,47 @@ package typeresolver
...
+		if isStaleGroupDiscoveryError(err) {
+			// Log and continue, using the cached APIs.
+			// Stale groups means one or more of the APIService backends are
+			// unhealthy. This is especially common for metrics providers.
+			// If the API provider of a managed resource remains unhealthy, it
+			// will be surfaced as a resource-specific API error later, when
+			// watched or listed, which are retried and won't prevent updating
+			// the status of other managed resources.
+			r.log.Error(err, "Failed to discover API resources; using cached results")
+		} else {
```
tiffanny29631, line 53: Will the `err` contain the stale gv for reference?

pkg/resourcegroup/controllers/typeresolver/typeresolver_test.go
```diff
@@ -51,3 +57,211 @@ func TestResolve(t *testing.T) {
 		})
 	}
 }
+
+func TestRefresh(t *testing.T) {
```
tiffanny29631, line 61: Mostly lgtm, nit missing a case for non-errGroupDiscoveryFailed error?

pkg/resourcegroup/controllers/typeresolver/typeresolver.go
karlkfi: yup https://github.com/kubernetes/kubernetes/blob/release-1.31/staging/src/k8s.io/client-go/discovery/discovery_client.go#L437

pkg/resourcegroup/controllers/typeresolver/typeresolver_test.go
karlkfi: There's a test case in there for "other errors"

karlkfi: oh, i see what you mean. I'll add another

karlkfi: added.

tiffanny29631 (approved):
/lgtm

PR #1499, opened 2024-11-22
Use docker hub for golang image, not google-go.pkg.dev

nresare:
To be able to build container images for both arm64 and amd64 we need a container image for each architecture. This change moves to using the upstream docker hub registry, that unlike google-go.pkg.dev contains images for both architectures.

Obviously, an even better way to resolve this would be to have the currently referenced repository be updated to hold images for both architectures. Feel free to do that :)

I'll kick off the CLA process with my employer after I have pushed this, to have something to reference in my conversation with legal. Hopefully it shouldn't take too long

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ GO_DIR := $(OUTPUT_DIR)/go
 # Base image used for all golang containers
 # Uses trusted google-built golang image
 GOLANG_IMAGE_VERSION := 1.22.5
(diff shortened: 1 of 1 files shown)


nan-yu:
@nresare Thanks for your contribution. However, due to security concerns, we'll continue using the `google-go.pkg.dev/golang` container image. These images are built internally by Google using trusted builders and hosted in Artifact Registry, eliminating reliance on third-party sources and mitigating potential security risks associated with the Go language toolchain.

PR #1500, opened 2024-11-22
Remove explicit GOARCH from build/all/Dockerfile

nresare:
With GOARCH unset, docker buildx multi arch will build the arches specified by the --platform argument

### DIFF
--- a/build/all/Dockerfile
+++ b/build/all/Dockerfile
@@ -26,7 +26,7 @@ COPY . .
 ARG VERSION
 
 # Build all our stuff.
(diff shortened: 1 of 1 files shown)


mikebz:
Good contribution, not sure if it helps us to pin down the architecture.  Better to be flexible if there is no downside.

nan-yu:
> With GOARCH unset, docker buildx multi arch will build the arches specified by the --platform argument

Currently, the base image only supports the `linux/amd64` architecture, requiring us to explicitly set GOARCH. We can re-evaluate this change once the base image supports multiple architectures, including `arm`.

Issue #893, opened 2023-09-22
publish arm64 images


seankhliao:
partly out of my own laziness to build the images myself, could the project publish multi platform images for easy deployment onto arm clusters as well?

seankhliao:
When trying to build it myself for linux/arm64, I also saw that a few images (`git-sync`, `gcenode-askpass-sidecar`, and `resource-group-controller`) didn't have their source in this repo

mikebz:
Hi @seankhliao 

git-sync is located here: https://github.com/kubernetes/git-sync
gcenode-askpass-sidecar just got moved into this repo https://github.com/GoogleContainerTools/kpt-config-sync/tree/main/cmd/gcenode-askpass-sidecar

mikebz:
@karlkfi and @sdowell can comment on whether or not the linux/arm64 

I see:
```
PLATFORMS := linux_amd64 linux_arm64 darwin_amd64 darwin_arm64 windows_amd64
```
so it could be that the right platform gets pulled because we are using dockerx

seankhliao:
I did get this working a few days ago
using resource-group-controller built from https://github.com/GoogleContainerTools/kpt-resource-group

The changes I had to make included somehow switching out the amd64 binaries pulled for helm and kustomize, as well as removing the hardcoded GOARCH=amd64 envs in Dockerfiles

sdowell:
> @karlkfi and @sdowell can comment on whether or not the linux/arm64
> 
> I see:
> 
> ```
> PLATFORMS := linux_amd64 linux_arm64 darwin_amd64 darwin_arm64 windows_amd64
> ```
> 
> so it could be that the right platform gets pulled because we are using dockerx

The `PLATFORMS` you referenced is used for building the `nomos` binary. We cross compile the `nomos` binary, but we currently do not support multi-arch images for the controllers. To support this would require some changes to upstream builds for third party components (opentelemetry, git-sync, resource-group, helm, kustomize), and then we would need to update this repo's build tooling to support multi-arch.

nresare:
I found myself needing an arm64 build of the images, so I had a look at the specifics of what would need to be done. There is what I found:

1. There are three different binary artefacts hosted on gcp that are amd64 only: the golang image and the binaries for kustomize and helm. The golang container image might be pretty straightforward, as the upstream docker hub version is multi-arch.
[...]

PR #1501, opened 2024-11-22
Download helm and kustomize in the target docker image

nresare:
Previously, helm and kustomize for a single architecture was downloaded as part of the top level make invocaiton and copied into the target docker environments.

To facilitate mulitple architectures for the resulting docker images, I have moved the artifact downloads into the docker build, extracting the target architecture from the evnironment

To complicate matters a bit, the google location for those images doesn't hold versions for arm64, so the scripts are downloading from the official release repositories of the respective upstream projects instead

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,11 +75,11 @@ CRANE := $(BIN_DIR)/crane
 GOLANGCI_LINT_VERSION := v1.60.3
 GOLANGCI_LINT := $(BIN_DIR)/golangci-lint
 
-KUSTOMIZE_VERSION := v5.4.2-gke.0
+KUSTOMIZE_VERSION := v5.4.2
 KUSTOMIZE := $(BIN_DIR)/kustomize
 KUSTOMIZE_STAGING_DIR := $(OUTPUT_DIR)/third_party/kustomize
 
-HELM_VERSION := v3.15.3-gke.0
+HELM_VERSION := v3.15.3
 HELM := $(BIN_DIR)/helm
 HELM_STAGING_DIR := $(OUTPUT_DIR)/third_party/helm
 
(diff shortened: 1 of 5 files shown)


nresare:
Please note that this is mainly for inspiration. You have my explicit permission to borrow whatever seems useful from this without attribution.

It is unclear to me whether the current kustomize and helm install scripts are needed with these ones in place. 

obviously it would make the change smaller and less dependent on upstream not pulling releases/replacing binaries etc if the missing binaries could be uploaded to the gcp object store.

karlkfi:
Config Sync itself is not being built for arm64, so there's not much point in having helm and kustomize binaries for arm64 yet.

Plus, for compliance & CVE patching reasons, we're required to ship binaries we built from source. We're also required to include the notices and licenses of dependencies and transitive dependencies.

nresare:
I think I have a pretty good understanding of how these things works, and I am aware of the fact that you currently don't build the config sync container images for arm64.  I currently don't have a lot of pull with the GCP account managers, the reason I'm looking into this is that I had some good experience with configsync with a previous employer and now I'm setting up a little raspberry pi cluster at home and it would be nice to be ablet to use it.

Please see my comment at https://github.com/GoogleContainerTools/kpt-config-sync/issues/893 for my breakdown of what I needed to do to build the images.

mikebz:
@nresare thank you so much for your contribution.  We will see how to incorporate this into a holistic solution here with the angle towards providing ARM support in general.  Are you currently unblocked in how you are building and deploying your fork?

karlkfi:
Sorry, I didn't know there was a previous feature request filed for this.

Didn't mean to shut you down by closing it quickly after you invested time in the PR. I just didn't see any way this particular PR could be modified to be acceptable given our security requirements. And since our internal forks for patching are internal, I can't just ask that you make the required changes there.

This PR seems fine on a fork that you build yourself, using public pre-built helm & kustomize binaries, it just won't satisfy FedRAMP requirements, which we have contracts to uphold.

I'll let @mikebz handle the feature request. He has a more influence than I do over what features gets prioritized.

nresare:
No hard feelings at all about closing this PR. Obviously the right way to go about this would be to fix the forked over at your side. @mikebz I just got the reconciler-manager Pod running, so I think I can consider myself unblocked :)

PR #1505, opened 2024-12-07
fix: refresh auth tokens 3m45s before expiration

karlkfi:
- Migrate from the legacy golang.org/x/oauth2 to the new
  cloud.google.com/go/auth for GCP & federated auth.
- Remove the 5m auth token cache in the askpass sidecar.
  Use the 3m45s EarlyTokenRefresh default provided by
  auth.DefaultCredentials instead.
- Make asskpass, helm-sync, oci-sync, and reconciler-manager all use
  the same credential and token provider code for consistency.
- Add unit tests to confirm that IsCredentialsNotFoundError works.
  It's fragile because it depends on an error string instead of a
  sentinel error.
  https://github.com/googleapis/google-cloud-go/issues/11258
  Unfortunately, the not found error is not reproducable or testable
  on GKE/GCE, so if the error text changes, the test may still pass
  in CI but fail when run locally.
  https://github.com/googleapis/google-cloud-go/issues/4920


cmd/oci-sync/main.go
```diff
@@ -93,9 +95,35 @@ func main() {
 	failCount := 0
 	backoff := errorBackoff()
 
+	var authenticator authn.Authenticator
+	switch configsync.AuthType(*flAuth) {
+	case configsync.AuthNone:
+		authenticator = authn.Anonymous
+	case configsync.AuthGCPServiceAccount, configsync.AuthK8sServiceAccount, configsync.AuthGCENode:
+		// authenticator = google.NewEnvAuthenticator()
+		authenticator = &oci.CredentialAuthenticator{
```
nan-yu, line 104: Keeping the authenticator outside of the for loop makes it impossible to auto-refresh the authenticator. For example, if the IAM policy bindings are changed, the authenticator will stay invalid unless the Pod is restated.

pkg/askpass/credentials.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	// Check the error prefix, because credentials.DetectDefault doesn't use a
+	// custom error type.
+	// https://github.com/googleapis/google-cloud-go/blob/auth/v0.12.0/auth/credentials/detect.go#L108
+	return strings.HasPrefix(err.Error(), "credentials: could not find default credentials.")
+}
+
+// CredentialProvider specifies an interface for anything that can return
+// credentials.
+type CredentialProvider interface {
```
nan-yu, line 48: The CredentialProvider is used by OCI, Helm and Otel-collector. Should it be relocated to a separate package rather than askpass. The askpass is more specific to Git.

cmd/oci-sync/main.go
karlkfi: The authenticator here is just a thin wrapper around the CachingCredentialProvider. So using the same authenticator instance allows caching the Credentials. And the new Credentials library handles caching the token as long as it's valid, with early refresh by default.

If we don't use the same authenticator, then we have to re-detect the credentials and re-request a new token every time, which involves one or two API calls, depending on the config.

If the IAM policy changes, does the client really need a new token? Or does it just check the bindings server-side when it makes a new API call?

pkg/askpass/credentials.go
karlkfi: moved to `pkg/auth`

cmd/oci-sync/main.go
karlkfi: CachedTokenProvider: https://github.com/googleapis/google-cloud-go/blob/auth/v0.12.0/auth/auth.go#L282

`cachedTokenProvider.Token()` returns the cached token if it's non-nil, has a non-empty token value, and has not expired, with early refresh. Otherwise it gets a new token from the wrapped TokenProvider. Errors are not cached.

Using Fleet Workload Identity with GOOGLE_APPLICATION_CREDENTIALS set, `readCredentialsFileJSON` will return a `New3LOTokenProvider`, which builds a `NewCachedTokenProvider` wrapping a `tokenProvider3LO`, which fetches a access token & refresh token from the url configured in the json.

So you're right, it won't fetch a new token every time, but that's a feature, not a bug. That's why the caching is part of the core library now.

karlkfi: I'm having trouble finding a source, but I'm 99% sure that Google IAM policy is not stored in the access token. I'm pretty sure the access token is an encrypted JWT that the server can decrypt with the public key of the auth server, but it should only include the scopes that the client requested, which allows the client to prove it has permission to use those scopes, signed by the auth server, but IAM policy must still be checked for those scopes by the server. So if the IAM policy changes, the client doesn't need a new token to pick up those changes.

The client should only need a new token when the token expires or the client wants to change the scopes it uses. Since clients rarely change scopes (maybe after an upgrade) and scopes rarely change on the server (maybe if services are added), the auth server grants permission to use those scopes for the duration of the token, usually 24hr.

nan-yu: ACK. Thanks for looking it up. I recall seeing the permission denied error as a server response, so I think you're right, the IAM policy bindings should be checked by the server.

pkg/reconcilermanager/controllers/otel_controller.go
```diff
@@ -131,11 +132,16 @@ func (r *OtelReconciler) reconcileConfigMap(ctx context.Context, req reconcile.R
...
-	creds, _ := getDefaultCredentials(ctx)
-	if creds == nil || creds.ProjectID == "" {
-		// No injected credentials
-		return nil, nil
+	// Only configure otel-collector to export metrics to Cloud Monitoring if
+	// GCP credentials are configured on the reconciler-manager.
+	// This assumes the otel-collector will be similarly configured.
+	_, err := r.credentialProvider.Credentials()
+	if err != nil {
+		if auth.IsCredentialsNotFoundError(err) {
```
nan-yu, line 140: Can we continue using `creds == nil || creds.projectID == ""` to check the existence of injected credentials.

Using error message is error-prone when updating the otel-collector version.

karlkfi: There's a couple problems with that...
1. The CredentialProvider interface I added actually returns a TokenProvider, not a full Credentials object. I did this, because Credentials is a struct, not an interface, and we only really needed the Token method anyway. Plus using TokenProvider simplifies the fake for the test.
2. Even if we changed the CredentialProvider to return a Credentials struct pointer, or a new custom interface, the Credentials struct doesn't expose the ProjectID as a public field any more. Instead, it has a `ProjectID(context.Context) (string, error)` method, which uses the `CredentialsPropertyProvider` interface to get the project ID.  https://github.com/googleapis/google-cloud-go/blob/auth/v0.12.0/auth/auth.go#L159

AFAICT, the project ID is generally parsed from the json, but the `CredentialsPropertyProvider` interface allows the project ID to be injected, overriding the `project_id` or `project` in the json. To add extra complexity, it looks like GDCH uses a `project` field, instead of a `project_id` filed. So the `project` field value is used if the `project_id` is missing or empty.
https://github.com/googleapis/google-cloud-go/blob/auth/v0.12.0/auth/internal/internal.go#L121

Then there's the `internal.GetProjectID` method, which allows overriding the project ID with the `GOOGLE_CLOUD_PROJECT` env var. So if that's set, checking the `Credentials.ProjectID` will always return the value of the env var, telling us nothing about whether the credentials json is valid.

So, I could add a new Credentials interface that exposes the ProjectID method, update the fake to impl it, and make the reconciler-manager check that ProjectID doesn't error and is non-empty, but I'm not entirely sure how to handle errors from the ProjectID method. The `CredentialsPropertyProvider.GetProperty` method doesn't seem to have any sort of contract as to what kind of errors it might return. Should they be ignored? Should they be terminal?

karlkfi: Anyway, the impl in this PR will surface all other errors from `Credentials()`, which seems like a good improvement. Previously, when the config was invalid it would just silently skip creating the ConfigMap, defaulting otel-collector to just the Kubernetes+Prometheus exporters.

karlkfi: I can add a test for IsCredentialsNotFoundError tho that will error if the error text is changed upstream.

nan-yu: > I can add a test for IsCredentialsNotFoundError tho that will error if the error text is changed upstream.

That works.

karlkfi: Added. It does make the tests more fragile, but that's better than fragile production code, i guess.

nan-yu (approved):
/lgtm

PR #1507, opened 2024-12-12
fix: ignore gcloud operations list warnings

karlkfi:
- Warnings printed to STDERR were being captured as in-progress operations, causing listOperations to error. This change fixes that issue by disabling human output for that command.


tiffanny29631 (approved):
/lgtm

PR #1508, opened 2024-12-12
fix: mutex unlocking in LoggingTokenProvider

karlkfi:
This was causing oci-sync and helm-sync to deadlock and crash.

Bug introduced by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1505


Camila-B (approved):
/lgtm

PR #1513, opened 2024-12-16
test: update bitnami Helm charts to use OCI package

nan-yu:
Per
https://blog.bitnami.com/2023/04/httpsblog.bitnami.com202304bitnami-helm-charts-now-oci.html, Bitnami Helm charts are now GA as OCI packages. This commit updates its repo URL to use the OCI package, which fixes the `helm pull` error:
```
KNV1068: failed to run kustomize build in /repo/source/.worktrees/b2c6bf5708654b1c2b7fbd898b48ca206de5a47b/helm-components, stdout: : # Warning: 'commonLabels' is deprecated. Please use 'labels' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
                Error: Error: looks like "https://charts.bitnami.com/bitnami" is not a valid chart repository or cannot be reached: Get "https://charts.bitnami.com/bitnami/index.yaml": dial tcp: lookup charts.bitnami.com on 10.66.32.10:53: server misbehaving
                : unable to run: 'helm pull --untar --untardir /repo/source/.worktrees/b2c6bf5708654b1c2b7fbd898b48ca206de5a47b/helm-components/charts/wordpress-15.2.35 --repo https://charts.bitnami.com/bitnami wordpress --version 15.2.35' with env=[HELM_CONFIG_HOME=/tmp/kustomize-helm-1930457181/helm HELM_CACHE_HOME=/tmp/kustomize-helm-1930457181/helm/.cache HELM_DATA_HOME=/tmp/kustomize-helm-1930457181/helm/.data] (is 'helm' installed?): exit status 1
```

Link: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-regular-csr/1868596976711372800

It also replaces the deprecated `commonLabels` with `labels` to suppress the warning.

### DIFF
--- a/Makefile.oss.prow
+++ b/Makefile.oss.prow
@@ -59,6 +59,27 @@ push-test-oci-images-public: "$(CRANE)"
 	cd $(KUSTOMIZE_COMPONENTS_DIR) && crane append -f <(tar -f - -c .) -t $(KUSTOMIZE_COMPONENTS_PUBLIC_GCR_IMAGE)
 	cd $(KUSTOMIZE_COMPONENTS_DIR) && crane append -f <(tar -f - -c .) -t $(KUSTOMIZE_COMPONENTS_PUBLIC_AR_IMAGE)
 
+# The e2e test expects both versions exist as public.
+# See `TestHelmVersionRange` and `rootSyncForWordpressHelmChart` in `e2e/testcases/helm_sync_test.go`.
+WORDPRESS_HELM_CHART_VERSIONS := 15.2.35 15.4.1
+WORDPRESS_HELM_CHART_SOURCE_REPO := oci://registry-1.docker.io/bitnamicharts/wordpress
+WORDPRESS_HELM_CHART_PUBLIC_AR_REPO := oci://$(LOCATION)-docker.pkg.dev/$(TEST_INFRA_PROJECT)/config-sync-test-public
+
+# This target is run as a singleton against the ci-artifacts project, since
+# these require for the registries to be public.
+.PHONY: push-test-helm-charts-public
+push-test-helm-charts-public: "$(HELM)"
+	$(foreach version, $(WORDPRESS_HELM_CHART_VERSIONS), \
+		helm pull $(WORDPRESS_HELM_CHART_PUBLIC_AR_REPO)/wordpress \
+			--version $(version) &> /dev/null || \
+			$(MAKE) __push-test-helm-chart-from-pull VERSION=$(version);)
+
+.PHONY: __push-test-helm-chart-from-pull
+__push-test-helm-chart-from-pull: "$(HELM)"
+	helm pull $(WORDPRESS_HELM_CHART_SOURCE_REPO) --version $(VERSION)
+	@gcloud auth print-access-token | helm registry login -u oauth2accesstoken --password-stdin https://$(LOCATION)-docker.pkg.dev
+	helm push wordpress-$(VERSION).tgz $(WORDPRESS_HELM_CHART_PUBLIC_AR_REPO)
+
 # The following targets are used to provision test resources in a prow environment
 
 # kustomize-components private images (per dev/prow env registry)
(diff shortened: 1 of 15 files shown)


e2e/testcases/helm_sync_test.go
```diff
@@ -907,7 +907,7 @@ func rootSyncForWordpressHelmChart(nt *nomostest.NT, valuesMutator func(map[stri
 	rs.Spec.Helm = &v1beta1.HelmRootSync{
 		Namespace: "wordpress",
 		HelmBase: v1beta1.HelmBase{
-			Repo:        "https://charts.bitnami.com/bitnami",
+			Repo:        "oci://registry-1.docker.io/bitnamicharts",
```
sdowell, line 910: Would it be feasible to switch this to a registry that we own so that we have more control over this?

nan-yu: That's a good idea. To keep minimal changes to the test case, I'll just pull the workpress Helm chart from `oci://registry-1.docker.io/bitnamicharts` and push it to the public AR repo in the CI project.

nan-yu: Yes, I'm writing a make target to pull from the public helm repo, and then push to our public AR repo.

There is a bummer that I hit an issue when trying to run the make target (aka. infra code) to push the same repo again.
This prevents us from re-building the test infra.

If there is no good way to bypass this issue, I would rather use the "real" public repo, `oci://registry-1.docker.io/bitnamicharts`.

sdowell: This PR is a good example of why to host our own infra instead of relying on an upstream repository - at any time the upstream maintainers could make a breaking change that breaks our CI overnight. Yes, the tradeoff is that we need to maintain some infra, but we are already hosting image registries for CI infra and it's a small amount of code to retag/push an image.

What is the error? Is it because the registry has immutable image tags? If so, just ensure the image we are re-pushing has a unique tag.

nan-yu: I disabled the immutable image tag option. After deleting the helm charts and re-pushing with the same version, I got an error from the AR server: `Error: failed commit on ref "manifest-sha256:30edc537f73f3c981f575b2db0387d57f243b62b93224b16fe49f0c76a08cb43": cannot reuse body, request must be retried`.

nan-yu: It might be some specific settings of the public AR repo of the CI project. I tried to re-push the same image with my own project, and it is working well.

nan-yu: I updated the make target to check whether the helm chart version exists first to avoid re-pushing the same image.

Now we can use a public Helm chart from a AR repo that we have full control.

e2e/testdata/hydration/deprecated-GK/kustomization.yaml
```diff
@@ -18,5 +18,6 @@ helmCharts:
   version: 0.2.0
   releaseName: my-prometheus-operator
 
-commonLabels:
-  test-case: hydration
+labels:
+  - pairs:
+      test-case: hydration
```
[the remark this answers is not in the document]
nan-yu: Wrong comment: The suggestion doesn't follow the kustomize syntax.

Makefile.oss.prow
```diff
@@ -59,6 +59,27 @@ push-test-oci-images-public: "$(CRANE)"
 	cd $(KUSTOMIZE_COMPONENTS_DIR) && crane append -f <(tar -f - -c .) -t $(KUSTOMIZE_COMPONENTS_PUBLIC_GCR_IMAGE)
 	cd $(KUSTOMIZE_COMPONENTS_DIR) && crane append -f <(tar -f - -c .) -t $(KUSTOMIZE_COMPONENTS_PUBLIC_AR_IMAGE)
 
+# The e2e test expects both versions exist as public.
+# See `TestHelmVersionRange` and `rootSyncForWordpressHelmChart` in `e2e/testcases/helm_sync_test.go`.
+WORDPRESS_HELM_CHART_VERSIONS := 15.2.35 15.4.1
```
karlkfi, line 64: What's the reasoning being using two different versions?
What's the upgrade strategy over time? Is it the same for both?

nan-yu: The `TestHelmVersionRange` test uses multiple versions from the public AR repo to test the version range feature.
We can rewrite the test to make it more hermetic by building and pushing the images to the test AR repo.

PR #1514, opened 2024-12-18
fix: errorCountAfterTruncation=0 not set

karlkfi:
- Fix a bug where errorCountAfterTruncation was not being set for rendering errors when truncated=false.

b/379720690


Camila-B (approved):
/lgtm

PR #1521, opened 2025-01-03
Update manifest copyright to satisfy linter

karlkfi:
The standard make target rebuilds the generated CRD manifests, which causes the copyright date to be updated to the current year. So this change updates the copyright date to match the expectation.

It would be ideal if the generation script only updated the copyright if the file had otherwise changed, but that's a more complicated change.


sdowell (approved):
/lgtm

PR #1522, opened 2025-01-06
e2e: fix version check failing when zero nodes

karlkfi:
Autopilot clusters with zero nodes were causing the version check to fail with an empty string in the split result. So I added a check to tolerate zero nodes if the control plane is new enough, assuming new nodes will use the control plane version.

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -16,6 +16,7 @@ package nomostest
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


e2e/nomostest/nt.go
```diff
@@ -919,8 +920,14 @@ func (nt *NT) autopilotClusterSupportsBursting() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	versions := strings.Split(strings.TrimSpace(string(out)), " ")
-	for _, nodeVersion := range versions {
+	outStr := strings.TrimSpace(string(out))
+	if outStr == "" {
+		// No nodes (scaled to zero)
```
sdowell, line 925: If there aren't any nodes, can we be certain that the cluster is running on the minimum version?

karlkfi: If it weren't, the cluster version check would have failed, right?

karlkfi: Oh right, we only checks the major version. I guess we need to check the minor too?

karlkfi: Wait. Nah. We check the major and minor of the cluster.

sdowell: We check the initial cluster version above for the first requirement (cluster was created with version 1.26 or later).

This check is to ensure the second requirement (current cluster version is `1.30.2-gke.1394000` or later). If we skip this check I don't think we can be certain that the current cluster version meets the second requirement.

karlkfi: Turns out the gcloud response has all the info we need. So I changed it to just parse the json response and check the 3 version fields available there.

I don't know if the `currentNodeVersion` changes before or after all the nodes have been updated, but it probably shouldn't matter in practice. If it fails in CI we can be more defensive and add the check back in to loop through all the nodes or maybe all the node pools in the json response.

sdowell (approved):
/lgtm

PR #1524, opened 2025-01-07
chore: use scheme to get GVK of typed objects

karlkfi:
Change ObjMetaFromObject to use the Scheme to get the GVK from typed objects, instead of relying on the code to populate the GVK on all objects. This requires handling errors when the type is unknown.

This is a prerequisite for removing the GVK from typed objects.


sdowell (approved):
/lgtm

PR #1527, opened 2025-01-09
chore: update IsManageableSystemNamespace

karlkfi:
The IsManageableSystemNamespace function currently requires typed namespace objects to have their GroupVersionKind set. This update adds checks for the v1.Namespace type to avoid that requirement.

This is part of a larger effort to not require typed objects to have their GVK set.


sdowell (approved):
/lgtm

PR #1528, opened 2025-01-09
chore: improve the InvalidCRDName error message

karlkfi:
The old text mentioned the solution but not the problem. Using backticks instead of quotes also makes the error easier to read because they don't need to be escaped in yaml/json or logs.


tiffanny29631 (approved):
/lgtm

PR #1529, opened 2025-01-11
fix: status update events not emitted

sdowell:
When using a short sync period, such as is used by the tests, it was possible for the sync and status update timers to be constantly reset in lockstep. This caused the status update event to never be emitted.

This change makes it so that the timers are only reset when a new commit is processed, which allows the status update event to be emitted.


pkg/parse/event_handler.go
```diff
@@ -59,7 +59,6 @@ func (s *EventHandler) Handle(event events.Event) events.Result {
 	// This delays status update and sync events.
 	runFn := func(ctx context.Context, trigger string) ReconcileResult {
 		result := s.Reconciler.Reconcile(ctx, trigger)
```
karlkfi, line 61: you can just return this directly, or remove the whole runFn wrapper

pkg/parse/events/builder.go
```diff
@@ -46,7 +46,7 @@ type PublishingGroupBuilder struct {
 func (t *PublishingGroupBuilder) Build() []Publisher {
 	var publishers []Publisher
 	if t.SyncPeriod > 0 {
-		publishers = append(publishers, NewResetOnRunAttemptPublisher(SyncEventType, t.Clock, t.SyncPeriod))
+		publishers = append(publishers, NewTimeDelayPublisher(SyncEventType, t.Clock, t.SyncPeriod))
```
karlkfi, line 49: i think the doc.go needs updating to reflect these changes

pkg/parse/event_handler.go
sdowell: comment is out of date, I realized we need the runFn to reset the sync timer whenever a reconcile is triggered by one of the other event types

pkg/parse/events/builder.go
sdowell: comment out of date, the `NewResetOnRunAttemptPublisher` is retained so doc.go doesn't need to be updated

pkg/parse/events/builder.go
```diff
@@ -55,7 +57,10 @@ func (t *PublishingGroupBuilder) Build() []Publisher {
 		publishers = append(publishers, NewRetrySyncPublisher(t.Clock, t.RetryBackoff))
 	}
 	if t.StatusUpdatePeriod > 0 {
-		publishers = append(publishers, NewResetOnRunAttemptPublisher(StatusUpdateEventType, t.Clock, t.StatusUpdatePeriod))
+		// Status updates are on an independent timer from the sync event, and thus
+		// could run immediately after a sync event. The status updater performs a
+		// diff before making API calls so this should be low cost.
+		publishers = append(publishers, NewTimeDelayPublisher(StatusUpdateEventType, t.Clock, t.StatusUpdatePeriod))
```
karlkfi, line 63: pkg/parse/events/doc.go needs updating to reflect this

sdowell: done

PR #1531, opened 2025-01-21
[WIP] Add max-object-count error to nomos vet

karlkfi:
### Max Object Count

- Max defaults to 5000, but is configurable with --max-object-count
- 5000 objects usually fits within 1.5MiB, as long as the managed object metadata isn't too long. In my tests with CronTab, 5k works, but 6k is too big. YMMV with different objects.

Example usage:
```
$ git clone https://github.com/config-sync-examples/crontab-crs
$ cd crontab-crs
$ wget https://raw.githubusercontent.com/GoogleContainerTools/kpt-config-sync/refs/heads/main/e2e/testdata/customresources/changed_schema_crds/old_schema_crd.yaml -O configs/crontab-crd.yaml
$ cd configs

$ nomos vet --source-format unstructured --max-inventory-size 0
Error: KNV1070: Maximum number of objects exceeded. Found 13064, but expected no more than 5000. Reduce the number of objects being synced to this cluster in your source of truth to prevent your ResourceGroup inventory object from exceeding the etcd object size limit. For instructions on how to break up a repository into multiple repositories, see https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/how-to/breaking-up-repo

For more information, see https://g.co/cloud/acm-errors#knv1070

$ nomos vet --source-format unstructured --max-inventory-size 0 --max-object-count 20000
✅ No validation issues found.

$ nomos vet --source-format unstructured --max-inventory-size 0 --max-object-count 0 # disabled
✅ No validation issues found.
```

### Max Inventory Size in Bytes

- Max defaults to 1572864 bytes (aka 1.5 MiB), but is configurable with --max-inventory-size (e.g. 3MiB = 3145728 bytes)

Example usage:
```
$ git clone https://github.com/karlkfi/crontab-crs
$ cd crontab-crs
$ git checkout karl-test-max-object-count
$ cd configs

$ nomos vet --source-format unstructured --max-object-count 0
Error: KNV1071: Maximum inventory size exceeded. Estimated inventory object size is 1601479 bytes, but the maximum in etcd defaults to 1572864 bytes. To fix, split the resources across multiple repositories.https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/how-to/breaking-up-repo

For more information, see https://g.co/cloud/acm-errors#knv1071

$ nomos vet --source-format unstructured --max-object-count 0 --max-inventory-size 3145728 # Max for GKE
✅ No validation issues found.

$ nomos vet --source-format unstructured --max-object-count 0 --max-inventory-size 0 # disabled
✅ No validation issues found.
```

### TODO
- Vet the interfaces & defaults with customers who frequently have this problem.

### Building nomos

To build your own nomos CLI:
```
$ git clone https://github.com/karlkfi/kpt-config-sync
$ cd kpt-config-sync
$ git checkout karl-nomos-vet-limit
$ make build-cli
```

Builder outputs to `./.output/go/bin/${ARCH}/nomos`

### DIFF
--- a/cmd/nomos/vet/vet.go
+++ b/cmd/nomos/vet/vet.go
@@ -20,11 +20,13 @@ import (
 	"github.com/spf13/cobra"
 	"kpt.dev/configsync/cmd/nomos/flags"
 	"kpt.dev/configsync/pkg/api/configsync"
+	"kpt.dev/configsync/pkg/importer/analyzer/validation/system"
 )
 
 var (
 	namespaceValue string
 	keepOutput     bool
+	maxObjectCount int
 	outPath        string
 )
 
@@ -42,6 +44,9 @@ func init() {
 
 	Cmd.Flags().BoolVar(&keepOutput, "keep-output", false,
 		`If enabled, keep the hydrated output`)
+
+	Cmd.Flags().IntVar(&maxObjectCount, "max-object-count", system.DefaultMaxObjectCount,
+		`If greater than zero, error if any repository contains more than the specified number of objects`)
 
 	Cmd.Flags().StringVar(&outPath, "output", flags.DefaultHydrationOutput,
 		`Location of the hydrated output`)
@@ -64,6 +69,11 @@ returns a non-zero error code if any issues are found.
 		// Don't show usage on error, as argument validation passed.
 		cmd.SilenceUsage = true
 
-		return runVet(cmd.Context(), namespaceValue, configsync.SourceFormat(flags.SourceFormat), flags.APIServerTimeout)
+		return runVet(cmd.Context(), vetOptions{
+			Namespace:        namespaceValue,
+			SourceFormat:     configsync.SourceFormat(flags.SourceFormat),
+			APIServerTimeout: flags.APIServerTimeout,
+			MaxObjectCount:   maxObjectCount,
+		})
 	},
 }
--- a/cmd/nomos/vet/vet.go
+++ b/cmd/nomos/vet/vet.go
@@ -24,10 +24,11 @@ import (
 )
 
 var (
-	namespaceValue string
-	keepOutput     bool
-	maxObjectCount int
-	outPath        string
+	namespaceValue        string
+	keepOutput            bool
+	maxObjectCount        int
+	maxInventorySizeBytes int
+	outPath               string
 )
 
 func init() {
@@ -47,6 +48,9 @@ func init() {
 
 	Cmd.Flags().IntVar(&maxObjectCount, "max-object-count", system.DefaultMaxObjectCount,
 		`If greater than zero, error if any repository contains more than the specified number of objects`)
+
+	Cmd.Flags().IntVar(&maxInventorySizeBytes, "max-inventory-size", system.DefaultMaxInventorySizeBytes,
+		`If greater than zero, error if any repository contains enough objects such that the inventory object would exceed the specified number of bytes when serialized as JSON`)
 
 	Cmd.Flags().StringVar(&outPath, "output", flags.DefaultHydrationOutput,
 		`Location of the hydrated output`)
@@ -70,10 +74,11 @@ returns a non-zero error code if any issues are found.
 		cmd.SilenceUsage = true
 
 		return runVet(cmd.Context(), vetOptions{
-			Namespace:        namespaceValue,
-			SourceFormat:     configsync.SourceFormat(flags.SourceFormat),
-			APIServerTimeout: flags.APIServerTimeout,
-			MaxObjectCount:   maxObjectCount,
+			Namespace:             namespaceValue,
+			SourceFormat:          configsync.SourceFormat(flags.SourceFormat),
+			APIServerTimeout:      flags.APIServerTimeout,
+			MaxObjectCount:        maxObjectCount,
+			MaxInventorySizeBytes: maxInventorySizeBytes,
 		})
 	},
 }
--- a/cmd/nomos/vet/vet_impl.go
+++ b/cmd/nomos/vet/vet_impl.go
@@ -37,6 +37,13 @@ import (
 	"kpt.dev/configsync/pkg/status"
 )
 
+type vetOptions struct {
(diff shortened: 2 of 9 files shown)


mikebz:
Looking at this it seems like the current count is just a number of resources.  I think this helps with a set of problems (esp when the resource count is large), but could we also validate via some heuristic function that the status for all of these objects is going to exceed 1.5MB?

karlkfi:
> could we also validate via some heuristic function that the status for all of these objects is going to exceed 1.5MB?

Maybe.

The problem is that the object size is influenced by a number of factors:
1. The number of objects (x2 for spec + status references)
2. The string lengths of the name, namespace, kind, and API group for each managed object
3. The string lengths of name and namespace of the RSync
4. Any metadata added to the ResourceGroup by other cluster addons, fleet features, and third party controllers.
5. Any possible error messages encountered when syncing or reconciling these objects (when status is enabled)

- 1 + 2 we can calculate in nomos vet from source. 
- 3 we can guess or use a hard-coded max length.
- 4 is difficult to guess, but usually not huge, unless giant annotations are added, but we could also hard-code a guess.
- 5 is unbounded. It's theoretically possible for a single error to be so long it exceeds 1.5MB, but this is very unlikely. However, if already close to the limit, it's likely that it works fine until there's an error or too many errors. Errors have no current max length. Most errors are larger than a single object reference. So total error message bytes size can easily eclipse the size of the object references in the spec & status, even if only a dozen have errors.

So we could make a heuristic, but it would be a guess, and probably not particularly accurate.

As with the max object count, we can tune it over time as needed, to make it more or less likely to trigger a false positive, but the higher we go, the more likely it is to give false negatives.

karlkfi:
Looks like I was wrong about number 5.

While the RG CRD does have fields for conditions on each of the object statuses, the code doesn't populate them. The individual errors from the applier are only reported in the RSync status. So only the following status fields are populated for each object:
- group
- kind
- namespace
- name
- status
- strategy
- actuation
- reconcile

All of those fields have implicit max lengths. So we might be able to make a useful heuristic. 

However, I also discovered that GKE currently reports 3MiB as the new limit, not just 1.5MiB. I'm not sure when that was changed, but I think it's GKE-specific. It probably doesn't apply to OSS K8s or EKS/AKS.

mikebz:
thank you for the detailed breakdown.  My guess is that as long as the disclaimer is there the heuristic function works better than anything a user would come up with.  You already know all the things that go into the length.  I think this is sort of like memory and disk space on a server, you want to probably have a warning when the load gets to X% of the etcd row.  I think right now we don't have any way to warn the user or to even explain what can happen.

Also it might be worthwhile for ResourceGroup to deal with excessive error values.  I assume those are mirrored from the resource statuses, so maybe something over x-many characters can be truncated with "..."

karlkfi:
> Also it might be worthwhile for ResourceGroup to deal with excessive error values. I assume those are mirrored from the resource statuses, so maybe something over x-many characters can be truncated with "..."

I was wrong about the error messages. The ResourceGroup doesn't actually store them like I thought they did. There's fields for it in the CRD, but the inventory client from kpt doesn't populate them. So the inventory size should actually be much easier to predict in nomos vet.

karlkfi:
I added another validator that builds a fake inventory object in memory, serializes to json, and counts the bytes.  It's not perfect, but it might be more accurate. Verifying that it's accurate is challenging though.

mikebz:
@karlkfi is the description of the PR current?  Thinking that instead of a Google doc we might want to just send the PR to customers and let them comment on this.

karlkfi:
> is the description of the PR current?

Yes.

> Thinking that instead of a Google doc we might want to just send the PR to customers and let them comment on this.

 That's fine.

karlkfi:
For the record, I also wrote a bash script that can count objects in a directory, but it doesn't handle hydration or cluster selectors the way `nomos vet` does. So this vet solution is superior when working with "dry" sources, but the script works fine on pre-hydrated "wet" sources. https://gist.github.com/karlkfi/22d3da7603ceac4239c157a180454ea2

janetkuo (commented):
Thanks for the fix! 
Should the default be unlimited, for backward compatibility?

karlkfi:
> Thanks for the fix! Should the default be unlimited, for backward compatibility?

I don't think so. But you may disagree.

I'm not worried about breaking manual interactions with the CLI. And breaking CI scripts would require upgrading the CLI first. So it should only really break after human interaction, which makes it relatively easy to fix and less surprising.

janetkuo:
The flag name can be something like "threshold", and the value could take either a fix number (object count) or % (if we build a heuristic function to estimate RG size). The flag can start with one approach and be extended to support both.

If this new flag introduce a new default limit, it breaks backwards compatibility for sure. So the question is whether we want to introduce this breaking change, or keep it unlimited by default.

mikebz:
How about this plan:
- We just focus on merging in the object count for right now, because it's the most straight forward thing to explain to customers.
- When nothing is provided, we don't error for backwards compatibility
- When just the flag with no number is provided, we default to Karl's suggestion of a "likely safe" object number, w/o a suggestion I would pick 500.  If the customer wants to fine tune their threshold they can provide an integer.
- We revisit additional options later if customers say that the simple option is insufficient.

How reversible are these decisions:
- flag name - irreversible, so `threshold` is gonna be it.
- default behavior - reversible, people can upgrade/downgrade to different CLI version.
- default object count - reversible, we can update it with new releases.
- we can add additional functionality to the flag or create a new flag later.

I don't see customers actively engaging on this outside of the channels we have already explored so I am not sure if we can get additional feedback w/o iterative delivery.

mikebz:
⏰  hellooo.... let's make a decision here EOD.  If there are no objections I would propose we go with my plan 😄  .

janetkuo:
@mikebz the plan lgtm

karlkfi:
https://github.com/GoogleContainerTools/kpt-config-sync/pull/1580

PR #1534, opened 2025-01-23
[cherry-pick] ci: fix vuln scan jq to use location in noteName (#1533)

sdowell:
It appears the API was updated for this json payload, it now includes a location subpath in the noteName value.


Camila-B (approved):
/lgtm

PR #1537, opened 2025-01-29
[WIP] Add Stabilizing status condition

karlkfi:
go/config-sync-stabilizing-status
b/393195894

- Add a new Stabilizing condition to the RSync status that communicates the aggregate reconciliation status of all the managed resource objects in the ResourceGroup inventory.
- Update the inventory client to expose the wrapped Unstructured object and copy metadata from it when applying the ResourceGroup. This allows the reconciler to inject metadata before each apply.
  - Update the inventory client to copy the current inventory state back into the wrapped Unstructured object, to allow reading the latest state after running the applier, without needing to get it from the server again. 
- Update the inventory client to populate the kstatus of each resource, as much as possible given that only the strategy, actuation, and reconcile status are available. Previously, it was always setting the kstatus to Unknown.
- Add a new `configsync.gke.io/source-commit` annotation to the ResourceGroup, so the stabilizing controller knows which commit the inventory was last updated for, so it can put the commit into the Stabilizing condition. This is applied by the ResourceGroup inventory client when the spec is applied. It is deliberately not a spec field, because it is not required by kpt, but that means it doesn't affect the generation and doesn't trigger updating the status observed generations.
- Fix the ResourceGroupController to correctly remove its conditions from the RG status after the status has been disabled, instead of relying on the inventory client to delete them. 

TODO:
- Add more unit tests for the new code
- Add e2e tests to validate the Stabilizing condition behavior (It appears to work according to the e2e logs, but it may not be transitioning True->False as intended yet.)

This unblocks some future enhancements:
- Update `nomos status` to show the strategy, actuation, and reconcile status fields, not just the kstatus. This should make it clearer when a NotFound status is expected or not.
- Update `nomos status` to show the `configsync.gke.io/source-commit` annotation value.
- Update `nomos status` to show the Stabilizing condition status/reason/message.
- Update the cli-utils applier to retain the kstatus too, not just the strategy, actuation, and reconcile status. Then pass it to the inventory client, so it can update whether the object is Terminating, InProgress, or Failed. The change in this PR can only tell if it's Current or NotFound.

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1551
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1553
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1565
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1564
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1563
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1566
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1597
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1596
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1593
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1589
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1588
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1576
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1575
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1574
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1572
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1570

### DIFF
--- a/pkg/parse/repo_sync_client.go
+++ b/pkg/parse/repo_sync_client.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
@@ -256,18 +257,26 @@ func (p *repoSyncStatusClient) GetReconcilerStatus(ctx context.Context) (*Reconc
 func (p *repoSyncStatusClient) SetSyncStatus(ctx context.Context, newStatus *SyncStatus) status.Error {
 	p.mux.Lock()
 	defer p.mux.Unlock()
-	return p.setSyncStatusWithRetries(ctx, newStatus, defaultDenominator)
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		return p.setSyncStatusWithRetries(ctx, newStatus, defaultDenominator)
+	})
+	if err != nil {
+		// setSyncStatusWithRetries only returns API errors
+		return status.APIServerErrorWrap(err)
+	}
+	return nil
 }
 
-func (p *repoSyncStatusClient) setSyncStatusWithRetries(ctx context.Context, newStatus *SyncStatus, denominator int) status.Error {
+func (p *repoSyncStatusClient) setSyncStatusWithRetries(ctx context.Context, newStatus *SyncStatus, denominator int) error {
 	if denominator <= 0 {
-		return status.InternalErrorf("denominator must be positive: %d", denominator)
+		klog.Fatalf("denominator must be positive: %d", denominator)
 	}
 	opts := p.options
 
+	rsKey := reposync.ObjectKey(opts.Scope, opts.SyncName)
 	rs := &v1beta1.RepoSync{}
-	if err := opts.Client.Get(ctx, reposync.ObjectKey(opts.Scope, opts.SyncName), rs); err != nil {
-		return status.APIServerError(err, fmt.Sprintf("failed to get the RepoSync object for the %v namespace", opts.Scope))
+	if err := opts.Client.Get(ctx, rsKey, rs); err != nil {
+		return fmt.Errorf("failed to get RepoSync %s: %w", rsKey, err)
 	}
 
 	currentRS := rs.DeepCopy()
@@ -294,7 +303,7 @@ func (p *repoSyncStatusClient) setSyncStatusWithRetries(ctx context.Context, new
 
 	// Avoid unnecessary status updates.
 	if !currentRS.Status.Sync.LastUpdate.IsZero() && cmp.Equal(currentRS.Status, rs.Status, compare.IgnoreTimestampUpdates) {
-		klog.V(5).Infof("Skipping status update for RepoSync %s/%s", rs.Namespace, rs.Name)
+		klog.V(5).Infof("Skipping status update for RepoSync %s", rsKey)
 		return nil
 	}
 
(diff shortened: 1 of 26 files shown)


pkg/reconciler/stabilizing/stabilizing_controller.go
```diff
@@ -0,0 +1,422 @@
...
+			return SyncCondition{},
+				fmt.Errorf("invalid ResourceGroup field value: status.resourceStatuses[%d].actuation: `%s`",
+					i, objStatus.Actuation)
+		}
+		switch objStatus.Reconcile {
+		case v1alpha1.ReconcileFailed, v1alpha1.ReconcileSkipped, v1alpha1.ReconcilePending, v1alpha1.ReconcileSucceeded:
+			reconcileStatusCounts[objStatus.Reconcile]++
+		default:
+			return SyncCondition{},
+				fmt.Errorf("invalid ResourceGroup field value: status.resourceStatuses[%d].actuation: `%s`",
```
tiffanny29631, line 177: Should be checking for Reconcile

pkg/reconciler/stabilizing/stabilizing_controller.go
```diff
@@ -0,0 +1,422 @@
...
+
+	// Get the source commit from the inventory
+	commit := core.GetAnnotation(rgObj, metadata.SourceCommitAnnotationKey)
+
+	syncCondition := SyncCondition{
+		Type:   SyncStabilizing,
+		Commit: commit,
+	}
+
+	switch {
```
tiffanny29631, line 190: Will it be a case when multiple cases are true at the same time, causing the syncCondition being overwritten?

pkg/resourcegroup/controllers/metrics/record.go
```diff
@@ -40,58 +40,67 @@ const (
 	CMSNamespace = "config-management-system"
 )
 
+func record(ctx context.Context, ms ...stats.Measurement) {
```
karlkfi, line 43: TODO: extract

pkg/parse/repo_sync_client.go
```diff
@@ -256,18 +257,26 @@ func (p *repoSyncStatusClient) GetReconcilerStatus(ctx context.Context) (*Reconc
 func (p *repoSyncStatusClient) SetSyncStatus(ctx context.Context, newStatus *SyncStatus) status.Error {
 	p.mux.Lock()
 	defer p.mux.Unlock()
-	return p.setSyncStatusWithRetries(ctx, newStatus, defaultDenominator)
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
karlkfi, line 260: TODO: extract

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller_test.go
```diff
@@ -133,7 +133,6 @@ func TestReconcile(t *testing.T) {
 	channelKpt <- event.GenericEvent{Object: resgroupKpt}
 
 	// Verify that the reconciliation modifies the ResourceGroupStatus field correctly
-	expectedStatus.ObservedGeneration = 1
```
karlkfi, line 136: TODO: extract

pkg/resourcegroup/controllers/root/root_controller.go
```diff
@@ -99,7 +99,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	// Skip ResourceGroup status updates if the status is disabled and has
 	// already been removed.
-	if resourcegroup.IsStatusDisabled(resgroup) {
+	if resourcegroup.IsStatusDisabled(resgroup) && resgroup.Status.Conditions == nil {
```
karlkfi, line 102: TODO: extract

test/kustomization/expected.yaml
```diff
@@ -2823,11 +2823,10 @@ spec:
               observedGeneration:
                 default: 0
                 description: |-
-                  observedGeneration is the most recent generation observed.
-                  It corresponds to the Object's generation, which is updated on
-                  mutation by the API Server.
-                  Everytime the controller does a successful reconcile, it sets
-                  observedGeneration to match ResourceGroup.metadata.generation.
+                  observedGeneration is updated to the latest metadata.generation every
```
karlkfi, line 2826: TODO: extract

mikebz:
will this be merged by EOW?  If not we should close it or put it in a branch.

karlkfi:
@janetkuo decided not to pursue this for now. So I'll just close it.

PR #1538, opened 2025-02-04
cleanup: Stop updating ManageConflictErrors for a remote RSync

Camila-B:
* Given #1342, this PR removes the requirement to report conflict errors to the remote manager in the reconciler when it's a RootSync
* b/355301118


tiffanny29631:
/lgtm

Karl might have more context on this.

karlkfi:
Awesome. I always hated this "feature".

I think it's safe to remove this code, but please confirm that the e2e tests validate equivalent behavior. I'm surprised they passed.

If the behavior did actually change:

1. Please scan through our public docs and see if there's anywhere that mentions this feature. We might need to change them.
2. We probably want confirmation/approval from @janetkuo

sdowell:
> Awesome. I always hated this "feature".
> 
> I think it's safe to remove this code, but please confirm that the e2e tests validate equivalent behavior. I'm surprised they passed.
> 
> If the behavior did actually change:
> 
> 1. Please scan through our public docs and see if there's anywhere that mentions this feature. We might need to change them.
> 2. We probably want confirmation/approval from @janetkuo

There is e2e test coverage here: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testcases/multi_sync_test.go#L521

My read is that this verifies the behavior, so I don't think the behavior actually changed.

Camila-B:
To add to what Sam said, this change removes redundancy and validates that the existing behaviour hasn't changed.

karlkfi (approved):
/lgtm

PR #1539, opened 2025-02-06
feat: rewrite inventory client using simple CRUD

sdowell:
This bumps cli-utils to the new version which surfaces a simpler inventory client interface. The new interface surfaces basic CRUD methods, and initially we can use the provided UnstructuredClient.

The UnstructuredClient is a simple generic client which still operates on a single KRM (in this case ResourceGroup) object.

This is a building block which will enable implementing a new inventory client in Config Sync which leverages sharded ResourceGroup objects.


pkg/applier/applier.go
```diff
@@ -163,25 +161,29 @@ func NewSupervisor(cs *ClientSet, scope declared.Scope, syncName string, reconci
 // objects in a single namespace.
 func NewNamespaceSupervisor(cs *ClientSet, namespace declared.Scope, syncName string, reconcileTimeout time.Duration) (Supervisor, error) {
 	syncKind := configsync.RepoSyncKind
-	invObj := newInventoryUnstructured(syncKind, syncName, string(namespace), cs.StatusMode)
+	syncNamespace := string(namespace)
```
karlkfi, line 164: nit: scope has a SyncNamespace method, but i guess it's not technically required here. Maybe we should combine these constructors in the future tho. They're not all that different.

pkg/applier/applier.go
```diff
@@ -600,7 +598,7 @@ func (s *supervisor) applyInner(ctx context.Context, eventHandler func(Event), d
 			klog.Info(e.ErrorEvent)
 			err := e.ErrorEvent.Err
 			if util.IsRequestTooLargeError(err) {
-				err = largeResourceGroupError(err, idFromInventory(s.inventory))
+				err = largeResourceGroupError(err, idFromInventoryInfo(s.invInfo))
```
karlkfi, line 601: nit: now that Info has its own ID, it might make sense to rename idFromInventoryInfo to coreIDFromInventoryInfo just to disambiguate.

pkg/applier/applier.go
```diff
@@ -796,29 +788,16 @@ func (s *supervisor) handleDisabledObjects(ctx context.Context, rg *live.Invento
...
 // exists.
-func (s *supervisor) removeFromInventory(rg *live.InventoryResourceGroup, objs []client.Object) error {
-	clusterInv, err := s.clientSet.InvClient.GetClusterInventoryInfo(rg)
-	if err != nil {
-		return err
-	}
-	if clusterInv == nil {
-		// If inventory does not exist, there is nothing to remove
+func (s *supervisor) removeFromInventory(ctx context.Context, objs []client.Object) error {
+	inv, err := s.clientSet.InvClient.Get(ctx, s.invInfo, inventory.GetOptions{})
```
karlkfi, line 792: Should we wrap this in a retry.RetryOnConflict? Could be a follow-up too.

karlkfi: oh, nvm. The CreateOnUpdate should already handle that.

pkg/applier/inventory/inventory_converter.go
```diff
@@ -0,0 +1,201 @@
...
+	"sigs.k8s.io/cli-utils/pkg/apis/actuation"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// UnstructuredInventoryConverter is an interface for converting ResourceGroup
+// objects to/from UnstructuredInventory objects.
+// This provides the translation logic to instantiate a UnstructuredClient which
+// works with ResourceGroup objects.
+type UnstructuredInventoryConverter interface {
```
karlkfi, line 37: Q: Do we need an interface? Seems like unnecessary indirection with only one impl.

pkg/applier/inventory/inventory_converter.go
```diff
@@ -0,0 +1,201 @@
...
+			syncKind:      configsync.RootSyncKind,
+			syncName:      syncName,
+			syncNamespace: configsync.ControllerNamespace,
+			statusMode:    statusMode,
+		}
+	}
+	return &unstructuredInventoryConverter{
+		syncKind:      configsync.RepoSyncKind,
+		syncName:      syncName,
+		syncNamespace: string(scope),
```
karlkfi, line 68: nit: scope.SyncNamespace()

karlkfi: you can actually use the same for the RootScope too

pkg/applier/inventory/inventory_converter.go
```diff
@@ -0,0 +1,201 @@
...
+	core.SetLabel(toObj, metadata.ManagedByKey, metadata.ManagedByValue)
+	core.SetLabel(toObj, metadata.SyncNamespaceLabel, ic.syncNamespace)
+	core.SetLabel(toObj, metadata.SyncNameLabel, ic.syncName)
+	core.SetLabel(toObj, metadata.SyncKindLabel, ic.syncKind)
+	core.SetAnnotation(toObj, metadata.ResourceManagementKey, metadata.ResourceManagementEnabled)
+	core.SetAnnotation(toObj, metadata.StatusModeKey, ic.statusMode)
+
+	if len(objs) == 0 { // Unset resource fields
+		klog.V(4).Infoln("clearing inventory resources")
+		unstructured.RemoveNestedField(toObj.UnstructuredContent(),
```
karlkfi, line 176: Generally dont want to actually use UnstructuredContent() when passing to one of these methods. If it creates a new Object map, it doesn't actually set it on the Unstructure.Object. Just use toObj.Object. If you're worried about it not being set, you have to do `toObj.Object = toObj.UnstructuredContent()` beforehand

pkg/applier/inventory/inventory_converter.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		err = unstructured.SetNestedSlice(toObj.UnstructuredContent(),
+			objStatus, "status", "resourceStatuses")
+		if err != nil {
+			return nil, err
+		}
+	}
+	generation := toObj.GetGeneration() // Update status.observedGeneration
+	klog.V(4).Infof("setting observedGeneration %d: ", generation)
+	err := unstructured.SetNestedField(toObj.UnstructuredContent(),
```
karlkfi, line 195: Is updating status.observedGeneration here actually useful? I know the old code did, but IIUC, it will need to again be updated after the spec update and before the status update, right? Does the UnstructuredInventoryClient handle that?

pkg/applier/inventory/inventory_converter_test.go
```diff
@@ -0,0 +1,647 @@
...
+			inv, err := ic.InventoryFromUnstructured(tc.fromObj)
+			if tc.wantErr {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, invName, inv.GetName())
+			require.Equal(t, invNamespace, inv.GetNamespace())
+			require.Equal(t, invID, inv.GetID().String())
+			require.Equal(t, tc.wantObjRefs, inv.GetObjectRefs())
```
karlkfi, line 212: might want testutil.AssertEqual here so it prints a nice diff when it doesn't match.

pkg/applier/inventory/inventory_converter_test.go
```diff
@@ -0,0 +1,647 @@
...
+	ui.SetObjectStatuses(f.objStatuses)
+	return ui
+}
+
+func TestInventoryToUnstructured(t *testing.T) {
+
+	testCases := map[string]struct {
+		fromObj *unstructured.Unstructured
+		toInv   *inventory.UnstructuredInventory
+		wantErr bool
```
karlkfi, line 237: you can do error comparisons with `testerrors.AssertEqual`

pkg/applier/inventory/inventory_converter_test.go
```diff
@@ -0,0 +1,647 @@
...
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+const (
+	invName       = "inv-name"
+	invNamespace  = configsync.ControllerNamespace
+	invID         = "inv-id"
+	invGeneration = 42
```
karlkfi, line 41: nit: you could just type this here to avoid casting later

pkg/applier/inventory/inventory_converter_test.go
```diff
@@ -0,0 +1,647 @@
...
+			})
+			// Set expected labels
+			tc.wantObj.SetLabels(map[string]string{
+				common.InventoryLabel:       invID,
+				metadata.ManagedByKey:       metadata.ManagedByValue,
+				metadata.SyncNameLabel:      invName,
+				metadata.SyncNamespaceLabel: configsync.ControllerNamespace,
+				metadata.SyncKindLabel:      configsync.RootSyncKind,
+			})
+			if !equality.Semantic.DeepEqual(rg, tc.wantObj) {
```
karlkfi, line 614: You can use testutil.AssertEqual to handle the comparison and diff printing for you. 

I don't think we actually want to use Semantic.DeepEqual in the test anyway. Would be good to validate it's exactly what we expect, just not similar, for catching minor regressions.

pkg/applier/inventory/inventory_converter_test.go
sdowell: done

pkg/applier/inventory/inventory_converter_test.go
sdowell: done

pkg/applier/inventory/inventory_converter_test.go
sdowell: done

pkg/applier/inventory/inventory_converter_test.go
sdowell: done

pkg/applier/inventory/inventory_converter.go
sdowell: This function is called again after the spec update and before the status update, so this is valid. It could be considered redundant because the UnstructuredInventoryClient also updates the observedGeneration, but I considered this as just being defensive.

pkg/applier/inventory/inventory_converter.go
sdowell: Done

pkg/applier/inventory/inventory_converter.go
sdowell: Done

pkg/applier/inventory/inventory_converter.go
sdowell: No we don't strictly need an interface. I thought it might make the intent more clear with the direction that this is going. Removed

pkg/applier/applier.go
```diff
@@ -163,25 +161,29 @@ func NewSupervisor(cs *ClientSet, scope declared.Scope, syncName string, reconci
 // objects in a single namespace.
 func NewNamespaceSupervisor(cs *ClientSet, namespace declared.Scope, syncName string, reconcileTimeout time.Duration) (Supervisor, error) {
 	syncKind := configsync.RepoSyncKind
-	invObj := newInventoryUnstructured(syncKind, syncName, string(namespace), cs.StatusMode)
+	syncNamespace := namespace.SyncNamespace()
+	invObj := newInventoryUnstructured(syncName, syncNamespace)
 	// If the ResourceGroup object exists, annotate the status mode on the
 	// existing object.
 	if err := annotateStatusMode(context.TODO(), cs.Client, invObj, cs.StatusMode); err != nil {
```
karlkfi, line 168: Do we need to annotate the status mode in the constructor? Can we move this to the converter?

pkg/applier/applier.go
```diff
@@ -495,7 +493,7 @@ func handleMetrics(ctx context.Context, operation string, err error) {
 // checkInventoryObjectSize checks the inventory object size limit.
 // If it is close to the size limit 1M, log a warning.
 func (s *supervisor) checkInventoryObjectSize(ctx context.Context, c client.Client) {
-	u := newInventoryUnstructured(s.syncKind, s.syncName, s.syncNamespace, s.clientSet.StatusMode)
+	u := newInventoryUnstructured(s.syncName, s.syncNamespace)
```
karlkfi, line 496: Would it be possible to also move this inventory size checking to the converter?

karlkfi: I guess this gets called before Applier.Run... so maybe doing it later isn't great...

pkg/applier/inventory/inventory_converter_test.go
```diff
@@ -0,0 +1,643 @@
...
+				},
+			}.build(t),
+		},
+	}
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			ic := NewInventoryConverter(declared.RootScope, invName, metadata.StatusEnabled)
+			rg, err := ic.InventoryToUnstructured(tc.fromObj, tc.toInv)
+			if tc.wantErr != nil {
+				require.Equal(t, tc.wantErr, err)
```
karlkfi, line 595: I'm kinda surprised this works. This is why I wrote testerrors.AssertEqual, which doesn't require the error pointer to be exactly the same. But maybe that's only a problem on wrapped errors and not the top one.

pkg/applier/applier.go
sdowell: It's also set in the converter. I didn't remove this because I assumed there may be good reason to set this annotation before the applier runs, e.g. to unwedge a ResourceGroup whose status is too large

pkg/applier/applier.go
sdowell: That seems like a good suggestion - technically we could catch inventory size issues earlier that way (during the InvAddTask before the applier performs actuation).

That being said this seems like something we can treat as a separate enhancement and can address in a separate PR

pkg/applier/applier.go
karlkfi: I don't think we need it here. 

- Applier.Run will create a new one for us with the right annotation the first time. 
- It will never be too big on the cluster, because it would get denied before writing to etcd.
- NewNamespaceSupervisor only gets called once when the reconciler initializes. So it can only really "fix" anything very occasionally, and even so, the applier will run shortly after initialization, doing the same thing.

Plus, removing it reduces uses of  newInventoryUnstructured, which seems like something we shouldn't have outside of the converter code or tests.

karlkfi: If the annotation didn't previously exist, it could be added here, and that would trigger the RGC to remove the status, but that seems unnecessary, because InvAddTask would do the same thing a few seconds later at most.

sdowell: Fair points, I'm on board to remove it.

As a side note: It occurred to me that if a ResourceGroup transitions from statusEnabled -> statusDisabled, the applier won't clear the status (since it skips UpdateStatus). Rather it will set the statusDisabled annotation on the ResourceGroup and then afterwards the resource-group-controller will clear the status. This is probably fine although it may be preferable for the applier to actually clear the status, I'm unsure.

sdowell: All that being said, there's an edge case where if a user has too many managed objects they wouldn't be able to disable the status. This is because the applier would be trying to set the annotation on the ResourceGroup with the same update that it's adding objects to the inventory (exceeding the size limit). So the applier would not be able to set the annotation until the user removed objects from the source repository.

This seems less user friendly than the alternative of setting the annotation with this separate initial update

karlkfi: Can you fix this in a separate PR? Doesn't need to block.

It shouldn't be making API calls in a constructor. But it also doesn't need to run it before every apply. It just needs to run once, perhaps wrapped with RetryOnConflict (Get+Update), or just using a patch.

pkg/applier/applier.go
karlkfi: Agreed. Please follow up in a separate PR.

karlkfi (approved):
/lgtm

pkg/applier/applier.go
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1619

PR #1541, opened 2025-02-07
test: configurable google auth for oci sig server

sdowell:
The OCI signature verification server currently always uses the google authenticator, even when running on KinD clusters. After upgrading the k8s libraries, the webhook response changes to a message reporting the unavailability of the GCE metadata server.

This change makes the authenticator configurable so that google auth only needs to be used when running against GAR. This will enable the test assertions to pass after upgrading the k8s libraries.

After this is submitted, the image will be built/published and the test scaffolding will be updated to set the environment variable.


karlkfi (approved):
/lgtm

PR #1542, opened 2025-02-07
refactor: relocate ResourceGroup GVK definition

sdowell:
This moves all usages of the ResourceGroup GVK variable from the kpt live package to the ResourceGroup api package located in the Config Sync repository.

The variable usage predates the migration of the resourcegroup component into the Config Sync repository. This change consolidates the API definition and minimizes imports of the kpt live package prior to removing the dependency on the kpt inventory client.


karlkfi:
Could we make this GVK a function instead of a var? structs aren't immutable, so they can't be const, but something could easily modify it and break the code.

sdowell:
> Could we make this GVK a function instead of a var? structs aren't immutable, so they can't be const, but something could easily modify it and break the code.

done

karlkfi (approved):
/lgtm

PR #1543, opened 2025-02-07
test: set USE_GOOGLE_AUTHENTICATOR for oci sig server

sdowell:
This sets the USE_GOOGLE_AUTHENTICATOR variable on the test oci signature verification webhook server, such that the google authenticator is only enabled when using GKE and GAR.

### DIFF
--- a/e2e/nomostest/oci-image-verification-server.go
+++ b/e2e/nomostest/oci-image-verification-server.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	admissionv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


e2e/nomostest/oci-image-verification-server.go
```diff
@@ -260,6 +261,10 @@ func testOCISignatureVerificationDeployment() *appsv1.Deployment {
 						VolumeMounts: volumeMounts,
 						Env: []corev1.EnvVar{
 							{Name: reconcilermanager.OciCACert, Value: "/etc/ssl/ca-certs/cert"},
+							{
+								Name:  "USE_GOOGLE_AUTHENTICATOR",
+								Value: strconv.FormatBool(*e2e.OCIProvider == e2e.ArtifactRegistry && *e2e.TestCluster == e2e.GKE),
```
karlkfi, line 266: any idea if the capitalization of true/false matters?

karlkfi (approved):
/lgtm

PR #1544, opened 2025-02-07
chore: bump k8s libraries to 1.32.0

sdowell:


go.mod
```diff
@@ -1,6 +1,6 @@
 module kpt.dev/configsync
 
-go 1.23
+go 1.23.0
```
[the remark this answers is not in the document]
sdowell: It's forcibly set when running `go mod tidy`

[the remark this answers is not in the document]
karlkfi: If you remove the `.0` does it replace it when you run go mod tidy?
If so, do you know why?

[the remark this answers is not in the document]
sdowell: > If you remove the .0 does it replace it when you run go mod tidy?

Yes.

> If so, do you know why?

No. I can look into it but it didn't seem worth the time investment to root cause

pkg/kmetrics/fieldcount.go
```diff
@@ -268,9 +268,6 @@ func k8sMetadataCount(k *types.Kustomization) map[string]int {
 	if k.Namespace != "" {
 		result["Namespace"] = 1
 	}
-	if len(k.CommonLabels) > 0 {
```
karlkfi, line 271: Why was this removed?

pkg/remediator/queue/queue.go
```diff
@@ -73,12 +73,12 @@ type ObjectQueue struct {
 	// also allow any call to yield the lock safely (specifically for Get).
 	cond *sync.Cond
 	// rateLimiter enables the ObjectQueue to support rate-limited retries.
-	rateLimiter workqueue.RateLimiter
+	rateLimiter workqueue.TypedRateLimiter[GVKNN]
```
karlkfi, line 76: Are these generic types available in the previous go version? 

If so, can you move this out to a different PR, just to reduce the size of this one.

If not, no worries.

pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -105,6 +105,10 @@ type reconcilerBase struct {
 
 	// syncKind is the kind of the sync object: RootSync or RepoSync.
 	syncKind string
+
+	// controllerName is used by tests to de-dupe controllers
+	// controller-runtime requires unique names for each controller instantiation
+	controllerName string
```
karlkfi, line 111: I assume the requirement is new?

But I'm pretty sure the name method has been around. So can you pull this out to a different PR, just to reduce the size of this one?

pkg/remediator/queue/queue.go
sdowell: Looks like these types aren't available in the previous k8s version, so it needs to be changed at the same time

pkg/reconcilermanager/controllers/reconciler_base.go
sdowell: Yes the requirement is new, pulled the change out to a separate PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1548

pkg/kmetrics/fieldcount.go
karlkfi: I guess I don't really know what this field is used for or if it's safe to remove it. Is this because we updated Kustomize at the same time as Go? Is it necessary to update both at the same time?

karlkfi: > Deprecated: Use the Labels field instead, which provides a superset of the functionality of CommonLabels. CommonLabels to add to all objects and selectors.

If we're parsing the user's kustomize file and using it to render then we probably need to keep passing both to kustomize until it's removed. 

So add a linter comment to ignore it, and maybe log an error that it's deprecated so the user is made aware.

sdowell: Agreed, done

karlkfi (approved):
/lgtm

PR #1545, opened 2025-02-08
test: add e2e tests to verify the max RSync NN length

sdowell:
The RSync name/namespace lengths are constrained by labels which are added to objects created by Config Sync. These test cases document the current behavior.

### DIFF
--- a/e2e/testcases/basic_test.go
+++ b/e2e/testcases/basic_test.go
@@ -16,17 +16,24 @@ package e2e
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
+	"kpt.dev/configsync/e2e/nomostest/policy"
+	"kpt.dev/configsync/e2e/nomostest/syncsource"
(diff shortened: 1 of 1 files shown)


sdowell:
FYI @janetkuo

e2e/testcases/basic_test.go
```diff
@@ -230,3 +237,109 @@ func TestNamespaceKubePublicCanBeManaged(t *testing.T) {
...
+	// Share the repository but set a separate sync dir for the new RootSync
+	defaultSyncDir := rootSyncTooLong.Spec.Dir
+	rootSyncTooLong.Spec.Git.Dir = rootSyncTooLongName
+	nt.Must(repo.Add(fmt.Sprintf("%s/rs1.yaml", defaultSyncDir), rootSyncTooLong))
+	nt.Must(repo.Add(
+		fmt.Sprintf("%s/ns.yaml", rootSyncTooLongName),
+		k8sobjects.NamespaceObject(rootSyncTooLongName)))
+	nt.Must(repo.CommitAndPush("create RootSync with too long name"))
+	nt.WaitForRootSyncSyncError(rootSyncTooLongName, applier.ApplierErrorCode,
+		"must be no more than 63 characters", nil)
```
karlkfi, line 263: Should we set a max length on the name field in the CRD?

sdowell: I was planning to add explicit validation in the controller after this, with an earlier and clearer error message. I don't think we can add the constraint on the CRD because it may fail to apply on a cluster upgrade with an invalid/failing RSync.

I wanted to add the test first to make it clear that the constraint already exists.

karlkfi (approved):
/lgtm

PR #1546, opened 2025-02-10
[WIP] e2e: wait for cluster to be ready

karlkfi:


e2e/nomostest/client.go
```diff
@@ -139,6 +144,9 @@ func upsertCluster(t testing.NTB, cluster Cluster, opts *ntopts.New) (bool, erro
 	if err := cluster.Create(); err != nil {
 		return true, err
 	}
+	if err := cluster.WaitForReady(); err != nil {
```
sdowell, line 147: Is there evidence that suggests this is needed? We currently wait for all operations on the cluster to complete during the Create call above

e2e/nomostest/clusters/gke.go
```diff
@@ -260,6 +265,33 @@ func createGKECluster(t testing.NTB, name string) error {
 	return listAndWaitForOperations(context.Background(), t, name)
```
sdowell, line 265: The `listAndWaitForOperations` call at the end of `createGKECluster` and `getGKECredentials` is currently intended to serve as a readiness check. I'd suggest either:

1. Instead of surfacing a new `WaitForReady` interface method, replace the `listAndWaitForOperations` calls with `waitForReadyGKECluster`, assuming `listAndWaitForOperations` is insufficient to determine readiness.
2. Keep the new `WaitForReady` interface method but remove the existing readiness checks built in to `createGKECluster`/`getGKECredentials`

e2e/nomostest/client.go
sdowell: Just ran the tests locally on GKE and ran into the same issue where this doesn't wait for the cluster to be ready. This wasn't an issue before, perhaps there was a change to the GKE operations during cluster creation.

mikebz:
will this be merged by EOW?  If not we should close it or put it in a branch.

karlkfi:
I think @sdowell should probably take it over. He had some thoughts on how to change it. The bug this solves is still a minor nuisance when running tests that create new clusters.

sdowell:
Moved to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1663

PR #1548, opened 2025-02-11
test: add controllerName to rsync unit tests

sdowell:
The next controller-runtime version requires each controller instantiation to have a unique name. By default the controller is given the name of the CRD type, e.g. rootsync/reposync.

Since each unit test instantiates a new controller, it needs a unique name.

### DIFF
--- a/pkg/reconcilermanager/controllers/reconciler_base.go
+++ b/pkg/reconcilermanager/controllers/reconciler_base.go
@@ -105,6 +105,9 @@ type reconcilerBase struct {
 
 	// syncKind is the kind of the sync object: RootSync or RepoSync.
 	syncKind string
+
+	// controllerName is used by tests to de-dupe controllers
+	controllerName string
 }
 
 func (r *reconcilerBase) serviceAccountSubject(reconcilerRef types.NamespacedName) rbacv1.Subject {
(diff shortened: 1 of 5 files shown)


pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -105,6 +105,9 @@ type reconcilerBase struct {
 
 	// syncKind is the kind of the sync object: RootSync or RepoSync.
 	syncKind string
+
+	// controllerName is used by tests to de-dupe controllers
+	controllerName string
```
karlkfi, line 110: Looks like this needs to be populated by the constructors so it's non-empty.

sdowell: The default value for a string is empty string (`""`)

karlkfi: Right, but then the name will be empty and cause the requirement to fail, right? I'm not sure how it fails or if it's a test-only requirement, but if we have it here in the struct we should populate it.

sdowell: The controller-runtime requirement is that the name is unique, not that it's non-empty. If you don't explicitly set the name then it defaults to the name of the resource (`rootsync`/`reposync`). The problem with the unit tests is that they always use the default name.

karlkfi: Gotcha. Then we probably don't need to require the caller to set them, but it might make sense to hard-code the name in the constructor so it's always set. Otherwise it looks like we have a test-only field in a non-test struct.

karlkfi: Alternatively, we could set it to the syncKind field value and just remove the controllerName field.

sdowell: Okay, added a line in the constructor to explicitly set the variable. I opted to let controller-runtime pick the default instead of choosing our own default.

karlkfi (approved):
/lgtm

PR #1552, opened 2025-02-11
feat: prevent ResourceGroup conflicts in applier

sdowell:
The applier often encounters resource conflicts when updating the ResourceGroup object due to frequent updates by the resource-group-controller. These become even more frequent in an upcoming update which aims to minimize the number of Get requests performed by the applier.

Resource conflicts are avoided by setting an annotation before and after the applier runs, and pausing updates in the resource-group-controller while the annotation is set.

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


sdowell:
We've come up with an alternative solution to pausing the resource-group-controller while the applier is running.

Instead we only pause the r-g-c while the observedGeneration != generation, which is between the spec update done by `InvAddTask` and the spec/status update done by `DeleteOrUpdateInvTask`.

Additionally, the new inventory client is planned to implement some retry-on-conflict logic for each update operation.

PR #1553, opened 2025-02-12
chore: cleanup ResourceGroup controller tests

karlkfi:
- Add watch test helpers to avoid sleeping to wait for controller changes.
- Add controller test helpers to handle setup and teardown.
- Add klog/logr test setup.
- Stop the test controller-manager after the test and wait for shutdown to avoid handing goroutines.


pkg/testing/testcontroller/testcontroller.go
```diff
@@ -0,0 +1,84 @@
...
+		Verbosity: int(klogLevel),
+	})
+}
+
+// StartTestManager starts the controller-manager in a go routine.
+// Call the returned function in a defer to stop the manager and wait until it exits.
+// This avoids async error logs from the test environment stopping before the
+// controller-manager.
+func StartTestManager(t *testing.T, mgr manager.Manager) func() {
+	ctx := context.Background() // TODO: use t.Context() after Go upgrade
```
sdowell, line 67: Which Go version is required? Can you add that to the comment?

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller_test.go
```diff
@@ -414,68 +464,3 @@ func TestReconcileTimeout(t *testing.T) {
 		})
 	}
 }
-
-func TestActuationStatusToLegacy(t *testing.T) {
```
sdowell, line 418: Why was this test removed?

pkg/resourcegroup/controllers/root/root_controller_test.go
```diff
@@ -139,63 +156,82 @@ func TestRootReconciler(t *testing.T) {
...
-	// There should be one event pushed to the channel.
-	go func() { e = <-reconcilerKpt.channel }()
-	time.Sleep(time.Second)
-	assert.Equal(t, "group0", e.Object.GetName())
+func waitForEvent(t *testing.T, inCh <-chan event.GenericEvent, timeout time.Duration) {
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+	select {
+	case e := <-inCh:
+		require.Equal(t, "group0", e.Object.GetName())
```
sdowell, line 231: If the name (`group0`) is not passed as a parameter, suggest moving it to a constant to make it clear that it's assumed this is the only `ResourceGroup` name used for tests

pkg/testing/testwatch/testwatch.go
```diff
@@ -0,0 +1,145 @@
...
+// This is designed for simplified use in tests and is not likely to be suitable
+// for non-test usage.
+func WatchObject(ctx context.Context, c client.WithWatch, exampleObjList client.ObjectList) (watch.Interface, error) {
+	watcher, err := c.Watch(ctx, exampleObjList)
+	if err != nil {
+		return nil, err
+	}
+	go func() {
+		<-ctx.Done()
+		watcher.Stop()
```
sdowell, line 46: Is this needed given that the context is also passed to the watcher?

karlkfi: Calling watcher.Stop allows for not just a downstream signal from caller to watcher but also allows waiting for an upstream signal from the watcher to the caller to confirm that the watcher is fully stopped.

As for whether the Watch method context actually stops the returned watcher... it depends on the watch impl. And frustratingly, there are a half dozen upstream wrappers that all behave subtly differently, especially fake ones used by tests. I spent months last year trying to get them fixed upstream but kept running into roadblocks from needing to be reverse compatible. So it's better to just be defensive.

karlkfi: FWIW, this method isn't waiting for watcher.Stop to exit, so the return signal isn't actually being waited for here, but the caller should be waiting for the watcher.ResultChan to close, which has a similar effect.

pkg/testing/testcontroller/testcontroller.go
karlkfi: Context was added in go 1.24.0, 6 months ago: https://go-review.git.corp.google.com/c/go/+/603959

Make errors when compiling in the build env: `t.Context undefined (type *"testing".T has no field or method Context, but does have unexported field context)`

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller_test.go
karlkfi: Merge mistake. Re-added. I'm gonna delete it in a subsequent PR.

pkg/resourcegroup/controllers/root/root_controller_test.go
karlkfi: Done.

sdowell (approved):
/lgtm

PR #1554, opened 2025-02-12
feat: validate RSync length in reconciler-manager

sdowell:
This adds validation logic for the maximum RSync name/namespace length in the reconciler-manager. This constraint already existed due to labels which were mapped from the RSync name/namespace. This change just adds explicit validation so that the errors are caught earlier and a clear error message is emitted.

### DIFF
--- a/pkg/reconcilermanager/controllers/reposync_controller.go
+++ b/pkg/reconcilermanager/controllers/reposync_controller.go
@@ -968,9 +968,23 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 	return result
 }
 
+// This check surfaces the RepoSync NN length constraint earlier as a validation error.
+func validateRepoSyncName(rs *v1beta1.RepoSync) error {
+	nnLength := len(rs.Name) + len(rs.Namespace)
+	if nnLength > reconcilermanager.MaxRepoSyncNNLength {
+		return fmt.Errorf("maximum combined length of RepoSync name and namespace is %d, but found %d",
+			reconcilermanager.MaxRepoSyncNNLength, nnLength)
+	}
+	return nil
+}
+
 func (r *RepoSyncReconciler) validateRepoSync(ctx context.Context, rs *v1beta1.RepoSync, reconcilerName string) error {
 	if rs.Namespace == configsync.ControllerNamespace {
 		return fmt.Errorf("RepoSync objects are not allowed in the %s namespace", configsync.ControllerNamespace)
+	}
+
+	if err := validateRepoSyncName(rs); err != nil {
+		return err
 	}
 
 	if err := validation.IsDNS1123Subdomain(reconcilerName); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -968,11 +968,28 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 	return result
 }
 
+// The ns-reconciler deployment label is constructed as "ns-reconciler-<ns>-<name>-<len(name)>"
+// Thus, the max length name + namespace for a RepoSync is 45 characters.
+// This check surfaces this constraint earlier as a validation error.
+func validateRepoSyncName(rs *v1beta1.RepoSync) error {
+	nnLength := len(rs.Name) + len(rs.Namespace)
+	maxRepoSyncLength := 45
```
karlkfi, line 976: perhaps make these constants?

sdowell: done

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -968,11 +968,25 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 	return result
 }
 
+// This check surfaces the RepoSync NN length constraint earlier as a validation error.
+func validateRepoSyncName(rs *v1beta1.RepoSync) error {
+	nnLength := len(rs.Name) + len(rs.Namespace)
+	if nnLength > reconcilermanager.MaxRepoSyncNNLength {
+		return fmt.Errorf("maximum RepoSync name/namespace length of %d exceeded: %d",
```
karlkfi, line 975: this error message is slightly ambiguous if you dont know what the constraint is. Perhaps "name+namespace" would be a little more clear that the limit is not for each independently but both together?

sdowell: done

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -968,11 +968,25 @@ func (r *RepoSyncReconciler) populateContainerEnvs(ctx context.Context, rs *v1be
 	return result
 }
 
+// This check surfaces the RepoSync NN length constraint earlier as a validation error.
+func validateRepoSyncName(rs *v1beta1.RepoSync) error {
+	nnLength := len(rs.Name) + len(rs.Namespace)
+	if nnLength > reconcilermanager.MaxRepoSyncNNLength {
+		return fmt.Errorf("maximum combined length of RepoSync name and namespace is %d, but found %d",
```
tiffanny29631, line 975: Is it helpful to output the problematic reposync/rootsync name as well

karlkfi (approved):
/lgtm

This probably needs a release note, even though it's not technically a new constraint. It might be a good idea to start adding proposed release notes to the PR description, just to streamline the release process.

PR #1562, opened 2025-02-18
test: add e2e tests to verify the max RSync NN length

sdowell:
The RSync name/namespace lengths are constrained by labels which are added to objects created by Config Sync. These test cases document the current behavior.

The service account permissions for CSR were missing from the previous change (https://github.com/GoogleContainerTools/kpt-config-sync/pull/1545)


karlkfi (approved):
/lgtm

PR #1565, opened 2025-02-21
chore: extract otelBaseController

karlkfi:


pkg/reconcilermanager/controllers/otel_base_controller.go
```diff
@@ -0,0 +1,71 @@
...
+	"kpt.dev/configsync/pkg/status"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// otelBaseController implements common functionality for otel controllers.
+type otelBaseController struct {
+	loggingController
+
+	client client.Client
+	scheme *runtime.Scheme
```
sdowell, line 34: `scheme` appears unused

karlkfi: Good call. Removed. The scheme is available from the Client interface anyway, so really no controllers should need it as a field.

sdowell (approved):
/lgtm

PR #1566, opened 2025-02-21
chore: fork kpt into internal/third_party

karlkfi:
kpt is primarily used for its ResourceGroup inventory client (InventoryResourceGroup), which is used by cli-utils' Applier.

This change copies the existing kpt code into the
internal/third_party directory of this repo, allowing us to make local Config Sync specific changes, without needing to depend on kpt as an upstream dependency.

This is a temporary solution to allow modifications until the inventory client is rewritten from scratch with a new cleaner interface.

Using internal/third_party allows preserving the upstream copyright and license headers, at least until the rewrite. To facilitate this, the internal/third_party directory is now ignored by make lint-license, and a rewrite was added to go.mod.

This PR does not change the kpt code itself, except the readme.


karlkfi:
@janetkuo please review. 

This is the internal fork of kpt without any changes, except to the README. 

This will unblock making changes to the fork without upstreaming them. This should be acceptable because the bugs we want to fix do not materially affect kpt, and we plan to replace this fork with a rewrite of the inventory client soon. So there's no need to upstream the fixes or depend on their e2e tests.

go.mod
```diff
@@ -2,6 +2,8 @@ module kpt.dev/configsync
 
 go 1.23.0
 
+replace github.com/GoogleContainerTools/kpt => ./internal/third_party/github.com/GoogleContainerTools/kpt
```
janetkuo, line 5: Suggest adding a comment here to clarify that this is only temporary and we plan to remove the dependency later

janetkuo (approved):
/lgtm

go.mod
karlkfi: Will do in my next PR, since this already merged

PR #1568, opened 2025-02-25
[WIP] fix: ResourceGroup controller bugs

karlkfi:
- Fix e2e tests to correctly update the ResourceGroup status
- Optimize e2e tests to watch, instead of polling
- Fix the ResourceGroup controller to wait for status.observedGeneration updates by the applier inventory client. This fixes a race condition and prevents status update conflicts.
- Fix TestStatusEnabledAndDisabled to allow changes to the status.generation by the applier.
- Fix the ResourceGroup controller to skip condition updates that only change the timestamp. This avoids unnecessary updates.
- Fix the ResourceGroup controller to correctly update the kstatus and the reconcile status, not just the kstatus.
- Change ResourceGroup to skip setting Reconciling=True before imediately setting it to Reconciling=False. This avoids unnecessary updates.
- Fix ResourceGroup controller to let the controller-manager handle retries with shared backoff, instead of using inline retries with independent backoff.
- Fix ResourceGroup controller to log errors encountered when computing the reconcile status.
- Fix ResourceGroup "root" controller to remove unnecessary "optimizations" which could cause updates to be missed. The core controller already handles these skips, and the controller-manager already handles de-duping identical events. This avoids temporarily incorrect status and delayed status updates.

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1570
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1573
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1572
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1574
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1575
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1593
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1591

### DIFF
--- a/e2e/nomostest/testkubeclient/client.go
+++ b/e2e/nomostest/testkubeclient/client.go
@@ -78,12 +78,12 @@ func (tc *KubeClient) Get(name, namespace string, obj client.Object) error {
 	return tc.Client.Get(tc.Context, client.ObjectKey{Name: name, Namespace: namespace}, obj)
 }
 
-// List is identical to List defined for clietc.Client, but without requiring Context.
+// List is identical to List defined for client.Client, but without requiring Context.
 func (tc *KubeClient) List(obj client.ObjectList, opts ...client.ListOption) error {
 	return tc.Client.List(tc.Context, obj, opts...)
 }
 
-// Create is identical to Create defined for clietc.Client, but without requiring Context.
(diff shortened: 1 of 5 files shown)


karlkfi:
This is now obsolete. All the fixes and cleanups have been pulled out to separate PRs.

PR #1569, opened 2025-02-25
test: refactor some unit tests to use exact errors

Camila-B:
- Removes reference to obsolete bug
- Cleans some tests up by validating errors by their exact type and message and not just their error code

### DIFF
--- a/pkg/declared/namespace_safeguard_test.go
+++ b/pkg/declared/namespace_safeguard_test.go
@@ -15,7 +15,6 @@
 package declared
 
 import (
-	"errors"
 	"testing"
 
 	"github.com/elliotchance/orderedmap/v2"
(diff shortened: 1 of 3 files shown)


pkg/reconcilermanager/controllers/gitsync_env.go
```diff
@@ -350,7 +350,6 @@ func gitSyncEnvs(_ context.Context, opts options) []corev1.EnvVar {
 			})
 		}
 	default:
-		// TODO b/168553377 Return error while setting up gitSyncData.
```
karlkfi, line 353: Looks like this "feature request" would have been a relatively minor chore PR to fix, to return the error through a few layers of functions. Can you just expose the error instead in a separate PR from the other changes here?

upsertManagedObjects already returns an error, so you probably only have to modify populateContainerEnvs &  gitSyncEnvs.

Camila-B: Sure, I'll do that in a separate PR. For now, I re-added the comment back.

mikebz:
@karlkfi @janetkuo curious if this is waiting on changes or reviews?

Camila-B:
This is ready for review again

PR #1570, opened 2025-02-26
fix: ResourceGroup unnecessary timestamp updates

karlkfi:
The ResourceGroup controller has been updating status conditions every time it reconciles, even when nothing changed but the LastTransitionTime.

This change fixes that problem by only updating the conditions if another condition field changed. This prevents the pre-existing semantic DeepEquals from always failing.


sdowell (approved):
/lgtm

PR #1571, opened 2025-02-26
fix: ResourceGroup unnecessary condition updates

karlkfi:
- Deprecate the `Reconciling` condition on the ResourceGroup, because it doesn't actually reflect the aggregate reconcile status of the objects synced by Config Sync from the source of truth. This condition will now always have a `status` of  `False` and a `reason` of `FinishReconciling` to avoid breaking any clients that were waiting for that state. This change reduces churn on the ResourceGroup objects, which had been frequently flipping back and forth between a `status` of `True` and `False`. To learn how to wait for all the managed objects to be synced and reconciled, see https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/how-to/monitor-rootsync-reposync#confirm_resources_in_the_commit_are_reconciled
- This change also removes the case that was setting the `Reconciling` condition `reason` to `ExceedTimeout`, which is an implementation detail of the ResourceGroup controller, has no value to the user, and doesn't need to be exposed.
- Docs change: cl/731842416
- Fixes: b/399896123

### DIFF
--- a/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
+++ b/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
@@ -99,16 +99,17 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
-	newStatus := r.startReconcilingStatus(rgObj.Status)
-	if err := r.updateStatusKptGroup(ctx, rgObj, newStatus); err != nil {
-		r.Logger(ctx).Error(err, "failed to update ResourceGroup to start reconciling")
-		return ctrl.Result{Requeue: true}, err
-	}
+	// Since this controller doesn't perform any writes in any external systems,
+	// it doesn't need to set the Reconciling condition status to True before
+	// building the new status, because there are no possible side-effects that
+	// the user might need to know about when reconciling fails.
+	// So skip initializing the Reconciling & Stalled conditions and only set
+	// them when reconciling succeeds or fails.
 
 	id := getInventoryID(rgObj.Labels)
-	newStatus = r.endReconcilingStatus(ctx, id, req.NamespacedName, rgObj.Spec, rgObj.Status, rgObj.Generation)
+	newStatus := r.endReconcilingStatus(ctx, id, req.NamespacedName, rgObj.Spec, rgObj.Status, rgObj.Generation)
 	if err := r.updateStatusKptGroup(ctx, rgObj, newStatus); err != nil {
-		r.Logger(ctx).Error(err, "failed to update ResourceGroup to finish reconciling")
+		r.Logger(ctx).Error(err, "failed to update ResourceGroup")
 		return ctrl.Result{Requeue: true}, err
 	}
 
@@ -169,19 +170,6 @@ func (r *reconciler) updateStatusKptGroup(ctx context.Context, resgroup *v1alpha
 	return r.client.Status().Update(ctx, resgroup, client.FieldOwner(FieldManager))
 }
 
-func (r *reconciler) startReconcilingStatus(status v1alpha1.ResourceGroupStatus) v1alpha1.ResourceGroupStatus {
-	// Preserve existing status and only update fields that need to change.
-	// This avoids repeatedly updating the status just to update a timestamp.
-	// It also prevents fighting with other controllers updating the status.
-	newStatus := *status.DeepCopy()
-	// Update conditions, if they've changed
-	newStatus.Conditions = updateCondition(newStatus.Conditions,
-		newReconcilingCondition(v1alpha1.TrueConditionStatus, StartReconciling, startReconcilingMsg))
-	newStatus.Conditions = updateCondition(newStatus.Conditions,
-		newStalledCondition(v1alpha1.FalseConditionStatus, "", ""))
-	return newStatus
-}
-
 func (r *reconciler) endReconcilingStatus(
 	ctx context.Context,
 	id string,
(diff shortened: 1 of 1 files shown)


pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -201,20 +188,29 @@ func (r *reconciler) endReconcilingStatus(
...
+	// it can only be used to confirm that the applier has completed its update.
+	newStatus.ObservedGeneration = generation
+
+	// Previously, the Reconciling condition toggled between True & False while
+	// this controller was reconciling, but this caused unnecessary churn and
+	// confusion, because it didn't reflect the reconciliation status of the
+	// managed objects in status.resourceStatuses. So now it's always set to
+	// False, for reverse compatibility, in case any clients were waiting it.
+	newStatus.Conditions = updateCondition(newStatus.Conditions,
+		newReconcilingCondition(v1alpha1.FalseConditionStatus, FinishReconciling, finishReconcilingMsg))
```
karlkfi, line 212: Should we set the reason to "Obsolete" or "Deprecated" instead of "FinishReconciling" to avoid confusion?

karlkfi: @janetkuo suggests to keep the "FinishReconciling" reason.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -205,20 +193,29 @@ func (r *reconciler) endReconcilingStatus(
...
 			newStalledCondition(v1alpha1.TrueConditionStatus, ExceedTimeout, exceedTimeoutMsg))
 	}
 
+	// Updating the ObservedGeneration is probably a no-op, because the applier
+	// should have updated it already when it updated the spec & status.
+	// But update it here just for completeness.
+	// It's important to note that because there's only one ObservedGeneration
+	// field, it cannot be used to confirm that this controller has completed.
+	// it can only be used to confirm that the applier has completed its update.
+	newStatus.ObservedGeneration = generation
```
sdowell, line 209: It looks like this is a behavioral change - before it did not set the observedGeneration in the case of `reconcileTimeout` exceeded. Rather, it only set the observedGeneration after successfully computing the status.

Is this an intentional change?

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -205,20 +193,29 @@ func (r *reconciler) endReconcilingStatus(
 	}()
```
sdowell, line 193: Is this goroutine being leaked in the case of the `reconcileTimeout` being exceeded? It doesn't look like we wait for this goroutine to exit in that case.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: Your right. Maybe I should move it out to its own PR. 

Ideally it should be a no-op, assuming that the inventory client already updated the ObservedGeneration when it updated the status. However, I think there's actually a bug in the RG inventory client that updates this to the generation before it updates the spec, causing it to be out of date.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: Yes. It's not being cancelled because it has no context... Unrelated bug.

karlkfi: IMO this needs to be rewritten so that it calculates the resource statuses asynchronously, caches them, and just updates the RG status based on the cache, but that's a bigger effort I haven't gotten around to.

karlkfi: The problem is that if this controller is triggered by resource updates, then the cache would almost always be somewhat stale, depending on whether a managed object update triggered it or a RG update triggered it. So it needs to be sequenced AFTER updating the cache to reflect the status of whatever triggered it. That way it wouldn't need to recompute all the resource statused on every reconcile pass.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: Reverted this change.

karlkfi:
@janetkuo please review.

This is the "Deprecation" of the Reconciling condition on the ResourceGroup.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -99,16 +99,17 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// it doesn't need to set the Reconciling condition status to True before
+	// building the new status, because there are no possible side-effects that
+	// the user might need to know about when reconciling fails.
+	// So skip initializing the Reconciling & Stalled conditions and only set
+	// them when reconciling succeeds or fails.
 
 	id := getInventoryID(rgObj.Labels)
-	newStatus = r.endReconcilingStatus(ctx, id, req.NamespacedName, rgObj.Spec, rgObj.Status, rgObj.Generation)
+	newStatus := r.endReconcilingStatus(ctx, id, req.NamespacedName, rgObj.Spec, rgObj.Status, rgObj.Generation)
 	if err := r.updateStatusKptGroup(ctx, rgObj, newStatus); err != nil {
```
janetkuo, line 111: Do we need to check if status has changed before updating, to avoid unnecessary updates?

janetkuo (approved):

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: There's a DeepEquals check in updateStatusKptGroup that handles this.

karlkfi: FWIW, the k8s server also handles no-ops well in most cases, so even if an an update is posted to the server, if there's no change it won't bump the resourceVersion or produce out watch events. So we're double covered.

PR #1572, opened 2025-02-26
fix: ResourceGroup resource reconcile status

karlkfi:
- Fixed a bug that caused the ResourceGroup
  `status.resourceStatuses[*].reconcile` field to become stale
  between apply attempts and no longer reflect the current
  reconciliation status of the managed objects.

For example, when actuation and reconcile succeed, but then the
object is rescheduled or becomes unhealthy, the reconcile status
should be changed to reflect that by the RGC, but it wasn't, until
the applier updated all the resource statuses in the next apply
attempt.

Dependencies: 
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1573
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1576

### DIFF
--- a/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
+++ b/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
@@ -320,16 +320,26 @@ func (r *reconciler) computeStatus(
 		}
 
 		// Keep the existing object statuses written by cli-utils
-		aStatus, exists := actuationStatuses[resStatus.ObjMetadata]
-		if exists {
-			resStatus.Actuation = aStatus.Actuation
+		if aStatus, exists := actuationStatuses[resStatus.ObjMetadata]; exists {
 			resStatus.Strategy = aStatus.Strategy
+			resStatus.Actuation = aStatus.Actuation
 			resStatus.Reconcile = aStatus.Reconcile
+
+			// Update the reconcile status based on the Strategy & Actuation
+			// from the last apply attempt, and the newly computed kstatus.
+			if reconcile, err := UpdateReconcileStatusToReflectKstatus(resStatus); err != nil {
+				// Keep existing Reconcile status
+				log.Error(err, "Resource object status unknown: failed to compute")
+			} else {
+				resStatus.Reconcile = reconcile
+			}
 
 			// Update the status field to reflect source -> spec -> status,
 			// not just spec -> status.
 			resStatus.Status = UpdateStatusToReflectActuation(resStatus)
 		}
+
+		log.V(5).Info("Resource object status computed", "status", resStatus)
 
 		// add the resource status into resgroup
 		statuses = append(statuses, resStatus)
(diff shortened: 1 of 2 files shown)


karlkfi:
It looks like this might not work as intended.

I had assumed the `status.reconcileStatuses[].status` field was suppossed to match the computed kstatus value, but it looks like it doesn't any more. There's a case in `ActuationStatusToLegacy` that sets the status to `Current` when actuation and reconcile are true, without checking the strategy. That means instead of being `NotFound` after a prune, it becomes `Current`.

Unfortunately, I suspect the UI & `nomos status` depend on this behavior now. So I don't think I can fix it to match kstatus.

If that's the case, I might need to add an extra line here to change it to Current after running `KstatusToReconcileStatus` to compute the new reconcile status from the kstatus.

But it also means the status written to the RG doesn't actually match the kstatus stored in the TypeResolver. And we can't change the TypeResolver, because we need the kstatus to compute the reconcile status.

karlkfi:
Actually, it's even more complicated...
It only sets to Current if not already set to NotFound, which happens when...
- the GVK is unknown to api discovery
- the Get returned NotFound or NoMatchError
- the cached status from the TypeResolver is NotFound

So it's possible that it never actually sets to Current after being deleted and that line is a no-op because of the previous early exit...

I'll need to run some tests.

karlkfi:
Test error is in `TestResourceGroupApplyStatus`: `ERROR: expected ResourceGroup ResourceVersion to not change`.

I think this should be fixed by: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1571 & https://github.com/GoogleContainerTools/kpt-config-sync/pull/1573

karlkfi:
I deleted some previous comments here because I misread the code.

We actually do need to keep the `ActuationStatusToLegacy` because it covers a couple important cases. But it needs to be fixed to be clearer to read and cover a few edge cases. I'll do that in another PR.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -346,23 +348,112 @@ func (r *reconciler) computeStatus(
...
+		case v1alpha1.ActuationSucceeded:
+			switch status.Status {
+			case v1alpha1.Current:
+				return v1alpha1.ReconcileSucceeded, nil // Strategy=Apply + Actuation=Succeeded + kstatus=Current => Reconcile=Succeeded
+			case v1alpha1.InProgress:
+				return v1alpha1.ReconcilePending, nil // Strategy=Apply + Actuation=Succeeded + kstatus=InProgress => Reconcile=Pending
+			case v1alpha1.Failed:
+				return v1alpha1.ReconcileFailed, nil // Strategy=Apply + Actuation=Succeeded + kstatus=Failed => Reconcile=Failed
+			case v1alpha1.Terminating:
+				return v1alpha1.ReconcileFailed, nil // Strategy=Apply + Actuation=Succeeded + kstatus=Terminating => Reconcile=Failed // TODO: Should this be Pending?
```
karlkfi, line 367: TODO: Should this be Pending, instead of Failed?

If we assume Failed is a terminal state that needs user interaction, then this isn't really failed. We could mark it as Pending instead, because the applier needs to re-actuate in order to re-apply the deleted object.

However, if the applier succeeded syncing, it might not be retrying. So it might not re-actuate for another hour.

Plus, if something deleted the object after config sync applied it, there's probably actually a controller fight, or some manual change by a user, which might indicate it needs user interaction to fix anyway.

@janetkuo What do you think?

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -342,6 +352,73 @@ func (r *reconciler) computeStatus(
 	return statuses
 }
 
+// UpdateReconcileStatusToReflectKstatus uses the current Strategy and Actuation
+// status from the applier and the newly computed kstatus to compute the
+// Reconcile status. Returns an error if one of the inputs is invalid.
+func UpdateReconcileStatusToReflectKstatus(status v1alpha1.ResourceStatus) (v1alpha1.Reconcile, error) {
+	switch status.Strategy {
+	case v1alpha1.Apply:
+		switch status.Actuation {
```
sdowell, line 361: Suggest breaking these nested switches into helper functions for readability

karlkfi: I left them all together to ensure it was catching all enums and permutations, but yeah, maybe adding function names would help reduce inline comments.

sdowell: I find that switches get kind of hard to follow when they are 2-3+ layers deep

karlkfi: Extracted funcs for the 3rd layer. I don't think it needs it for the 2nd, but let me know if you feel differently.

sdowell: Looks good, thank you

sdowell (approved):
/lgtm

PR #1575, opened 2025-02-26
fix: ResourceGroup conflicts

karlkfi:
The ResourceGroup controller reconciles after each ResourceGroup
update to spec or status. This was causing the ResourceGroup
controller to try to make status updates after the applier made its
initial spec update, without waiting for the applier to do the
status update immediately afterwards. This caused a conflict between
the concurrent status updates.
    
To avoid this, the applier inventory client now correctly updates the
status.observedGeneration and the ResourceGroup controller now waits
for the status.observedGeneration to match the metadata.generation.

Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1566


karlkfi:
Looks like this will require e2e test changes to simulate the applier setting the observedGeneration in the tests that create the ResourceGroup manually.

e2e/testcases/resource_group_controller_test.go
```diff
@@ -53,53 +52,55 @@ func TestResourceGroupController(t *testing.T) {
...
+	nt.Must(nt.Watcher.WatchObject(kinds.ResourceGroup(), configsync.RootSyncName, configsync.ControllerNamespace,
+		testwatcher.WatchPredicates(
+			testpredicates.AllResourcesReconciled(nt.Scheme),
+		)))
 }
 
 func TestResourceGroupControllerInKptGroup(t *testing.T) {
 	nt := nomostest.New(t, nomostesting.ACMController)
 
+	// increase logging
```
karlkfi, line 65: TODO: remove logging change after tests pass

karlkfi: Meh. I'm just gonna leave in the verbose logging. It only gets written to the test log on test failure anyway, in which case we'll probably want verbose logging.

sdowell (approved):
/lgtm

PR #1576, opened 2025-02-27
fix: ResourceGroup resource status bugs

karlkfi:
- Fixed a bug that was causing the `status` field to remain `Current` after the object had successfully reconciled, even if the object later became `InProgress`, `Terminating`, or `Failed`. This will cause nomos status and the Cloud Console UI to correctly show the current status after initial reconciliation, rather than just the status based on the last time the Applier ran. 
- Fixed a bug that was causing the `status` field to become `Unknown` when the object is actually `Terminating` when the `actuation` is empty/unknown, `pending`, `skipped`, or `failed`. Now, the `Terminating` and `NotFound` statuses should be correctly exposed for previously applied objects, even if the object spec has not been updated to match the source of truth yet.
- Refactored ActuationStatusToLegacy to be easier to read, with extra comments to explain the reasoning.
- Adjusted the test expectations for the new behavior

### DIFF
--- a/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
+++ b/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
@@ -323,16 +323,16 @@ func (r *reconciler) computeStatus(
 			hasErr = true
 		}
 
-		// Update the legacy status field based on the actuation, strategy and reconcile
-		// statuses set by cli-utils. If the actuation is not successful, update the legacy
-		// status field to be of unknown status.
+		// Keep the existing object statuses written by cli-utils
 		aStatus, exists := actuationStatuses[resStatus.ObjMetadata]
 		if exists {
 			resStatus.Actuation = aStatus.Actuation
 			resStatus.Strategy = aStatus.Strategy
 			resStatus.Reconcile = aStatus.Reconcile
 
-			resStatus.Status = ActuationStatusToLegacy(resStatus)
+			// Update the status field to reflect source -> spec -> status,
+			// not just spec -> status.
+			resStatus.Status = UpdateStatusToReflectActuation(resStatus)
 		}
 
 		// add the resource status into resgroup
@@ -346,23 +346,45 @@ func (r *reconciler) computeStatus(
 	return statuses
 }
 
-// ActuationStatusToLegacy contains the logic/rules to convert from the actuation statuses
-// to the legacy status field. If conversion is not needed, the original status field is returned
-// instead.
-func ActuationStatusToLegacy(s v1alpha1.ResourceStatus) v1alpha1.Status {
-	if s.Status == v1alpha1.NotFound {
-		return v1alpha1.NotFound
+// UpdateStatusToReflectActuation updates the latest computed kstatus to reflect
+// the desired state in the source of truth, as much as possible, not just the
+// desired state that has been persisted to the Kubernetes API.
+//
+// This is necessary, because kstatus only reflects whether the object status
+// reflects the object spec. But in Config Sync, the desired state is actually
+// in the source or truth. So if the latest desired state has not yet been
+// successfully actuated (applied or deleted), then the current status is
+// Unknown, because neither Kubernetes nor this ResourceGroup controller knows
+// what the desired state is.
+func UpdateStatusToReflectActuation(s v1alpha1.ResourceStatus) v1alpha1.Status {
+	// When actuation is unknown, return the latest computed status.
+	// This should only ever happen when upgrading from an old version.
+	// It means the applier was last run with code that predates the addition
+	// of the strategy, actuation, and reconcile fields.
+	if s.Actuation == "" {
+		return s.Status
 	}
 
-	if s.Actuation != "" &&
-		s.Actuation != v1alpha1.ActuationSucceeded {
-		return v1alpha1.Unknown
+	// When actuation has succeeded, we know the object spec is up to date with
+	// the latest source, so the latest computed status should be correct.
+	if s.Actuation == v1alpha1.ActuationSucceeded {
+		return s.Status
 	}
 
-	if s.Actuation == v1alpha1.ActuationSucceeded && s.Reconcile == v1alpha1.ReconcileSucceeded {
-		return v1alpha1.Current
+	// When the latest computed status is Terminating or NotFound, keep it,
+	// even if actuation is Pending, Skipped, or Failed, because Terminating and
+	// NotFound are independent of the desired state in the source of truth,
+	// while InProgress, Current, and Failed depend on actuation having
+	// successfully updated the spec on the cluster.
+	if s.Status == v1alpha1.NotFound || s.Status == v1alpha1.Terminating {
+		return s.Status
 	}
-	return s.Status
+
+	// When actuation is not successful (Pending, Skipped, Failed), that means
+	// the object spec has not been updated to the desired spec from source yet,
+	// so the computed status can be ignored and the real status is Unknown,
+	// unless the latest computed status is NotFound or Terminating.
+	return v1alpha1.Unknown
 }
 
 // setResStatus updates a resource status struct using values within the cached status struct.
--- a/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller_test.go
+++ b/pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller_test.go
@@ -474,65 +474,106 @@ func TestReconcileTimeout(t *testing.T) {
 	}
 }
 
-func TestActuationStatusToLegacy(t *testing.T) {
+func TestUpdateStatusToReflectActuation(t *testing.T) {
 	tests := []struct {
 		name      string
 		resStatus v1alpha1.ResourceStatus
 		want      v1alpha1.Status
 	}{
 		{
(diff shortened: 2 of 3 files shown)


karlkfi:
Looks like we can't remove the `statusNeedsUpdate` optimization in the root controller until we stop setting Reconciling status=True on every reconcile loop (https://github.com/GoogleContainerTools/kpt-config-sync/pull/1571). So I reverted that change in this PR. It will be removed as part of the PR that removes the ResourceGroupPredicate (https://github.com/GoogleContainerTools/kpt-config-sync/pull/1574), which depends on a few other PRs.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller_test.go
```diff
@@ -481,51 +481,92 @@ func TestActuationStatusToLegacy(t *testing.T) {
...
+			name: "Return not found when status is not found already - disregard actuation success",
+			resStatus: v1alpha1.ResourceStatus{
 				Status:    v1alpha1.NotFound,
 				Actuation: v1alpha1.ActuationSucceeded,
 			},
-			v1alpha1.NotFound,
+			want: v1alpha1.NotFound,
 		},
 		{
-			"Return Current if both Actuation and Reconcile succeeded",
```
sdowell, line 522: It looks like the behavior changed for this input combination. Is this intentional?

karlkfi: Yes. The test expectation asserts undesirable behavior that was changed.

The reason it's undesirable is that it's too generic. If the Strategy is Apply, then it might be correct, but definitely not when the Strategy is Delete. Plus, the function already had an early exit case that makes this expectation incorrect when the status is NotFound, and I added another case for Terminating. But the real gotcha is that if the Strategy is Apply and the Actuation and Reconcile both succeeded, then the kstatus should already be Current, unless the Reconcile status is stale. And if that's the case, then we actually want to show the latest computed kstatus, not override it as Current. So this case is actual a bug that hides status updates, because until we merge https://github.com/GoogleContainerTools/kpt-config-sync/pull/1572, the Reconcile status is almost always stale, because it's only updated by the Applier.

TLDR: This case is what was causing the status field to remain Current, even if the object became InProgress, Terminating, or Failed after successfully reconciling.

karlkfi: I updated the description to be more like release notes

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -350,19 +350,31 @@ func (r *reconciler) computeStatus(
...
+	if s.Status == v1alpha1.NotFound || s.Status == v1alpha1.Terminating {
+		return s.Status
 	}
-	return s.Status
+
+	// When actuation is not successful (pending, skipped, failed), that means
+	// the object spec has not been updated to the desired spec from source yet,
+	// so the computed status can be ignored and the real status is Unknown,
+	// unless the latest computed status is NotFound or Terminating.
+	return v1alpha1.Unknown
```
sdowell, line 377: So given actuation was not successful:
- `InProgress`, `Current`, or `Failed` is replaced with `Unknown`
- `NotFound` and `Terminating` are retained

What is special about `NotFound`/`Terminating` that makes it honored? And why are each of the other statuses replaced with `Unknown`?

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -350,19 +350,31 @@ func (r *reconciler) computeStatus(
 // to the legacy status field. If conversion is not needed, the original status field is returned
```
sdowell, line 350: Why is it referred to as the legacy status field?

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: The status field here is sorta overloaded. It's not just the current kstatus (spec -> status). It's also trying to express whether the desired state from source has been synced (source -> spec -> status).

So when kstatus computes to InProgress, Current, and Failed, those all depend on the spec being the desired state. But if the desired state is actually in the source, and it hasn't been actuated yet, then the ResourceGroup controller doesn't know what the desired state is and neither does Kubernetes. So those all need to map to Unknown.

But when kstatus computes to NotFound or Terminating, those are independent of the desired state and are still correct and valuable to expose to the user.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: IIRC, the desire was for the strategy, actuation, and reconcile statuses to replace the legacy status field. So that's why the inventory client set it to Unknown and the code calls it the "legacy status" here.

But while kpt was updated to use the new strategy, actuation, and reconcile statuses, Config Sync was not completely migrated, partly because we never got around to phase 2 of the project and partly because we learned that the kstatus actually exposed additional useful information. Terminating, for example. The conditions generated when computing the kstatus are also useful, because they include error details and even information about sub-objects, like Pod errors for Deployments.

Would you like me to update the func name and methods to remove the "legacy" terminology? I don't think we're ever really going to remove it at this point.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: I updated the PR description to clarify the fix.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
sdowell: I'm mostly asking for the sake of my own understanding, because I don't have much context around this code. I don't think it needs to be renamed as part of this change but sounds like we might want to consider changing the naming at some point

karlkfi: I can rename it here to clarify the intend of the function

karlkfi: Done. Should be clearer in the code now.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -346,23 +346,45 @@ func (r *reconciler) computeStatus(
...
+// in the source or truth. So if the latest desired state has not yet been
+// successfully actuated (applied or deleted), then the current status is
+// Unknown, because neither Kubernetes nor this ResourceGroup controller knows
+// what the desired state is.
+func UpdateStatusToReflectActuation(s v1alpha1.ResourceStatus) v1alpha1.Status {
+	// When actuation is unknown, return the latest computed status.
+	// This should only ever happen when upgrading from an old version.
+	// It means the applier was last run with code that predates the addition
+	// of the strategy, actuation, and reconcile fields.
+	if s.Actuation == "" {
```
janetkuo, line 364: nit: Can we combine them given that they all return `s.Status`? Or at least combine the `if s.Actuation` ones

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
janetkuo: Do we keep `NotFound` and `Terminating` because we don't care whether it exists in source (desired state) or not anymore?

karlkfi: Right. 

After a sync attempt succeeds, the applier will remove pruned objects from the inventory, so this edge case doesn't apply. 

This case only applies before syncing has started (before inv-add-task) and after syncing has started (after inv-add-task) but before its succeeded (before inv-set-task).

If the actuation status is Pending, Skipped, or Failed, then Kubernetes does not have the latest object spec yet, so the kstatus is mostly incorrect, relative to the desired state in the source. But NotFound and Terminating are always correct, because they're not dependent on Kubernetes having the latest spec from source.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: I considered that and figured it was valuable to explain the reasoning in the long comments before each case. But if you think it's important, I can go combine them and make a larger single comment block with bullet points.

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
karlkfi: Put another way:
- If the object status is NotFound or Terminating, we should show NotFound or Terminating, whether or not the object exists in the source or has changed since it was last applied.
- If the object status is InProgress, Current, or Failed AND we know that the last attempt to apply or prune it Failed or was Skipped or is Pending, then the latest spec has not been persisted to Kubernetes yet, so the RGC doesn't know if the current spec on the cluster matches the desired spec in source or not, so we set the status to Unknown.

janetkuo (approved):
/lgtm

PR #1577, opened 2025-02-27
test: re-introduce e2e syncing condition validation

sdowell:
This was previously commented out due to flakiness.

Reverts a subset of changes from https://github.com/GoogleContainerTools/kpt-config-sync/pull/275


e2e/nomostest/wait_for_sync.go
```diff
@@ -242,11 +243,8 @@ func (nt *NT) WaitForRootSyncSourceError(rsName, code string, message string, op
 			if err := nt.KubeClient.Get(rs.GetName(), rs.GetNamespace(), rs); err != nil {
 				return err
 			}
-			// Only validate the rendering status, not the Syncing condition
-			// TODO: Remove this hack once async sync status updates are fixed to reflect only the latest commit.
-			return testpredicates.ValidateError(rs.Status.Source.Errors, code, message, nil)
-			// syncingCondition := rootsync.GetCondition(rs.Status.Conditions, v1beta1.RootSyncSyncing)
-			// return validateRootSyncError(rs.Status.Source.Errors, syncingCondition, code, message, []v1beta1.ErrorSource{v1beta1.SourceError})
+			syncingCondition := rootsync.GetCondition(rs.Status.Conditions, v1beta1.RootSyncSyncing)
```
karlkfi, line 246: Looks like these are in the WaitFor methods, not the newer preconditions we use with the WatchFor methods.
- Are these WaitFors still used? A lot? Can we replace them with watches?
- Do the WatchFor methods or their predicates have equivalent expectations?

e2e/nomostest/wait_for_sync.go
```diff
@@ -379,36 +368,34 @@ func (nt *NT) WaitForRepoSyncStalledError(rsNamespace, rsName, reason, message s
...
+	if err := testpredicates.ValidateError(statusErrs, code, message, resources); err != nil {
+		return err
+	}
+	if syncingCondition == nil {
+		return fmt.Errorf("syncingCondition is nil")
+	}
+	if syncingCondition.Status == metav1.ConditionTrue {
+		return fmt.Errorf("status.conditions['Syncing'].status is True, expected false")
+	}
+	if !reflect.DeepEqual(syncingCondition.ErrorSourceRefs, expectedErrorSourceRefs) {
```
karlkfi, line 381: Do we want to use semantic DeepEqual here?

e2e/nomostest/wait_for_sync.go
sdowell: > Are these WaitFors still used? A lot? Can we replace them with watches?

Yes they are still used a lot. We could probably replace Wait with watches.

> Do the WatchFor methods or their predicates have equivalent expectations?

There seem to be `R*SyncHas*Error` predicate functions which have similar functionality, although those also don't verify the Syncing condition

karlkfi: I'm frankly surprised this PR passes e2e tests. I have a feeling that it's still flaky, but we might need to replace them all with watches in order to enable debug diffs to figure out what the problem is. So I'm hesitant to merge this as-is, but I guess flaky failures is at least some signal... and we can always revert it again later.

What do you think?

Do you want to convert them all to watches first and add this condition there? That might give us more confidence.

karlkfi: It occurs to me that if the retry timer is short or the resync timer is short, like we set it in the e2e tests, the Syncing condition may spend more time True than False due to it flipping back and forth before and after each apply attempt. We might need to lengthen the timers to make it less flaky.

e2e/nomostest/wait_for_sync.go
sdowell: It's an array of strings, so I'm not sure whether it matters. Do you have a preference?

e2e/nomostest/wait_for_sync.go
sdowell: Haven't noticed any flakiness so far. Converted them all to use watches

e2e/nomostest/wait_for_sync.go
karlkfi: In general it's good to avoid using reflect in prod code. In test code it's less important, but usually slightly faster. The only real behavioral difference here would be accepting nil as equal to an empty list, but that's probably not important here.

e2e/nomostest/wait_for_sync.go
```diff
@@ -233,86 +233,47 @@ func (nt *NT) WatchForSync(
...
+		testpredicates.RootSyncHasSourceError(code, message),
+		testpredicates.RootSyncHasErrorSourceRefs([]v1beta1.ErrorSource{v1beta1.SourceError}),
+	}
+	return nt.Watcher.WatchObject(kinds.RootSyncV1Beta1(), rsName, configsync.ControllerNamespace,
+		testwatcher.WatchPredicates(predicates...))
 }
 
 // WaitForRootSyncSyncError waits until the given error (code and message) is present on the RootSync resource
-func (nt *NT) WaitForRootSyncSyncError(rsName, code string, message string, resources []v1beta1.ResourceRef, opts ...WaitOption) {
+func (nt *NT) WaitForRootSyncSyncError(rsName, code string, message string, resources []v1beta1.ResourceRef) error {
```
karlkfi, line 247: Since it's using a watch now, can you rename it WatchFor- and move it into the testwatcher package? Trying to reduce NT methods.

e2e/nomostest/wait_for_sync.go
karlkfi: Cool. Thanks

e2e/nomostest/wait_for_sync.go
sdowell: done

e2e/nomostest/wait_for_sync.go
sdowell: done

karlkfi (approved):
/lgtm

Thanks!

PR #1579, opened 2025-02-28
test: add ResourceGroup assertions to WatchForSync

sdowell:
This adds validation for the ResourceGroup status in the WatchForSync function, which is used extensively in the test framework to validate RSyncs and wait for them to finish syncing.

This will cause WatchForSync to now also wait for all synced objects to reconcile and for the ResourceGroup to be current. This is intended to provide test coverage for the resource-group-controller when integrated with Config Sync.

### DIFF
--- a/e2e/nomostest/sync.go
+++ b/e2e/nomostest/sync.go
@@ -21,8 +21,10 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/api/kpt.dev/v1alpha1"
 	"kpt.dev/configsync/pkg/parse"
 	"kpt.dev/configsync/pkg/reposync"
+	"kpt.dev/configsync/pkg/resourcegroup"
 	"kpt.dev/configsync/pkg/rootsync"
 	"kpt.dev/configsync/pkg/util/log"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -260,6 +262,61 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
 		if helmChart := status.Rendering.Helm.Chart; helmChart != syncPath {
 			return fmt.Errorf("status.rendering.helm.chart %q does not match the expected chart %q", helmChart, syncPath)
 		}
+	}
+	return nil
+}
+
+func resourceGroupHasReconciled(sourceHash string) testpredicates.Predicate {
+	return func(o client.Object) error {
+		if o == nil {
+			return testpredicates.ErrObjectNotFound
+		}
+		rg, ok := o.(*v1alpha1.ResourceGroup)
+		if !ok {
+			return testpredicates.WrongTypeErr(o, &v1alpha1.ResourceGroup{})
+		}
+
+		// Status is disabled for this ResourceGroup, skip all status checks
+		if resourcegroup.IsStatusDisabled(rg) {
+			return testpredicates.ResourceGroupHasNoStatus()(o)
+		}
+
+		if len(rg.Spec.Resources) != len(rg.Status.ResourceStatuses) {
+			return fmt.Errorf("length of spec.resources (%d) does not equal length of status.resourceStatuses (%d)",
+				len(rg.Spec.Resources), len(rg.Status.ResourceStatuses))
+		}
+
+		resourceCount := make(map[v1alpha1.ObjMetadata]int)
+		for _, s := range rg.Status.ResourceStatuses {
+			if err := resourceStatusIsCurrent(s, sourceHash); err != nil {
+				return fmt.Errorf("object %s is not current: %w", s.ObjMetadata, err)
+			}
+			resourceCount[s.ObjMetadata]++
+		}
+
+		for _, o := range rg.Spec.Resources {
+			if count, ok := resourceCount[o]; ok && count > 0 {
+				resourceCount[o]--
+			} else {
+				return fmt.Errorf("spec.resources does not equal status.resourceStatuses")
+			}
+		}
+		return nil
+	}
+}
+
+func resourceStatusIsCurrent(rs v1alpha1.ResourceStatus, sourceHash string) error {
+	if rs.Status != v1alpha1.Current {
+		return fmt.Errorf("resourceStatus.status is not %s. Got %s", v1alpha1.Current, rs.Status)
+	}
+	if rs.Actuation != v1alpha1.ActuationSucceeded {
+		return fmt.Errorf("resourceStatus.actuation is not %s. Got %s", v1alpha1.ActuationSucceeded, rs.Actuation)
+	}
+	if rs.Reconcile != v1alpha1.ReconcileSucceeded {
+		return fmt.Errorf("resourceStatus.reconcile is not %s. Got %s", v1alpha1.ReconcileSucceeded, rs.Reconcile)
+	}
+	if rs.SourceHash != resourcegroup.TruncateSourceHash(sourceHash) {
+		return fmt.Errorf("resourceStatus.sourceHash is not %s. Got %s", sourceHash, rs.SourceHash)
 	}
 	return nil
 }
--- a/e2e/nomostest/wait_for_sync.go
+++ b/e2e/nomostest/wait_for_sync.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -29,6 +30,7 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/testwatcher"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/api/kpt.dev/v1alpha1"
 	"kpt.dev/configsync/pkg/core/k8sobjects"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/reposync"
@@ -100,6 +102,14 @@ func SkipReadyCheck() WatchForAllSyncsOptions {
 	}
 }
 
+// SkipAllResourceGroupChecks is an optional parameter which specifies to skip
+// ResourceGroup checks for all RSyncs after they finish syncing.
+func SkipAllResourceGroupChecks() WatchForAllSyncsOptions {
+	return func(options *watchForAllSyncsOptions) {
+		options.watchForSyncOpts = append(options.watchForSyncOpts, SkipResourceGroupCheck())
+	}
+}
+
 // WatchForAllSyncs calls WatchForSync on all Syncs in nt.SyncSources.
 //
 // If you want to validate specific fields of a Sync object, use
@@ -159,7 +169,8 @@ func (nt *NT) WatchForAllSyncs(options ...WatchForAllSyncsOptions) error {
 }
 
 type watchForSyncOptions struct {
-	watchOptions []testwatcher.WatchOption
+	watchOptions           []testwatcher.WatchOption
+	skipResourceGroupCheck bool
 }
 
 // WatchForSyncOption is an optional parameter for WatchForSync.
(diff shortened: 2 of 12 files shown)


e2e/nomostest/sync.go
```diff
@@ -263,3 +266,58 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
...
+		}
+		return nil
+	}
+}
+
+func resourceStatusIsCurrent(rs v1alpha1.ResourceStatus, sourceHash string) error {
+	if rs.Status != v1alpha1.Current {
+		return fmt.Errorf("resourceStatus.status is not %s. Got %s", v1alpha1.Current, rs.Status)
+	}
+	if rs.Reconcile != v1alpha1.ReconcileSucceeded {
```
karlkfi, line 313: FYI, the Reconcile status is not currently being updated by the RGC. So it only flips when the applier inv-set-task runs. Fix is in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1572

e2e/nomostest/sync.go
```diff
@@ -263,3 +266,58 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
 	}
 	return nil
 }
+
+func resourceGroupHasReconciled(sourceHash string) testpredicates.Predicate {
```
karlkfi, line 270: Probably want some condition checks in here for Stalled at least. The Reconciling condition is now deprecated by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1571 but it might be worth adding a quick inline comment about it here to remind us why we're not checking it.

e2e/nomostest/wait_for_sync.go
```diff
@@ -229,6 +229,17 @@ func (nt *NT) WatchForSync(
 		return fmt.Errorf("waiting for sync: %w", err)
 	}
 	nt.T.Logf("%s %s/%s is synced", gvk.Kind, namespace, name)
+	rgPredicates := []testpredicates.Predicate{
+		// Wait until status.observedGeneration matches metadata.generation
+		testpredicates.HasObservedLatestGeneration(nt.Scheme),
```
karlkfi, line 234: I'm not really sure what the expected behavior was intended to be for the observedGeneration today. My guess is that nobody really considered how it would work with two controllers.

Anyway, I'm changing how it works in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1575

Instead of adding controller-specific observedGenerations, I was able to get it to work by having the applier inventory client handle setting it correctly when it updates the status, and then the RGC requires them to match before it will do its status updates, to avoid race conditions between spec & status updates.

This means that there's no status flag we can really use to confirm that the RGC has done its job. But that's sort of ok, for now. 

It does mean that the kstatus fields will still be UNKNOWN until we fix cli-utils to pass it through to the inventory client so it can write them.

Another gotcha worth mentioning is that when the status is disabled, the observedGeneration is not updated. So this check likely needs to be embedded in resourceGroupHasReconciled.

e2e/nomostest/sync.go
karlkfi: Actually, the StatusEquals Current check might handle the Stalled condition.

sdowell: Added an additional check to ensure it has the Stalled condition and it is False.

e2e/nomostest/sync.go
```diff
@@ -263,3 +266,58 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
...
+			return testpredicates.ErrObjectNotFound
+		}
+		rg, ok := o.(*v1alpha1.ResourceGroup)
+		if !ok {
+			return testpredicates.WrongTypeErr(o, &v1alpha1.ResourceGroup{})
+		}
+
+		// Status is disabled for this ResourceGroup, skip all status checks
+		if root.IsStatusDisabled(rg) {
+			return nil
```
karlkfi, line 282: Do we want to test that it's empty?
I just updated those assertions in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1593

e2e/testdata/hydration/helm-components/kustomization.yaml
```diff
@@ -22,7 +22,7 @@ helmCharts:
   releaseName: my-wordpress
   namespace: wordpress
   valuesInline:
-    existingSecret: my-wordpress
+    wordpressPassword: abcdefg
```
sdowell, line 25: FYI this was changed because the helm chart never reconciled before, the previous configuration assumed the Secret was created out of band.

This makes it so that the helm chart reconciles and the ResourceGroup passes its validation

e2e/nomostest/sync.go
```diff
@@ -263,3 +266,58 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
...
+		}
+
+		if len(rg.Spec.Resources) != len(rg.Status.ResourceStatuses) {
```
karlkfi, line 285: Do we want to validate that all the resources match in spec & status? Or is that too much here?

e2e/nomostest/sync.go
```diff
@@ -263,3 +266,58 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
...
+}
+
+func resourceStatusIsCurrent(rs v1alpha1.ResourceStatus, sourceHash string) error {
+	if rs.Status != v1alpha1.Current {
+		return fmt.Errorf("resourceStatus.status is not %s. Got %s", v1alpha1.Current, rs.Status)
+	}
+	if rs.Reconcile != v1alpha1.ReconcileSucceeded {
+		return fmt.Errorf("resourceStatus.reconcile is not %s. Got %s", v1alpha1.ReconcileSucceeded, rs.Reconcile)
+	}
+	if rs.Actuation != v1alpha1.ActuationSucceeded {
```
karlkfi, line 316: Maybe move this above the reconcile check. If actuation failed we don't care about reconcile.

e2e/nomostest/sync.go
```diff
@@ -263,3 +266,58 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
...
+		return fmt.Errorf("resourceStatus.actuation is not %s. Got %s", v1alpha1.ActuationSucceeded, rs.Actuation)
+	}
+	if rs.SourceHash != status.TruncateSourceHash(sourceHash) {
```
karlkfi, line 319: Huh. I didn't even know these had a source hash field. Nifty.

So I guess you can tell if they're all on the same commit pretty easy, you just can't tell what commit it's supposed to be from the spec.

I guess if we have this in the status, maybe we should add a SourceHash to the spec, instead of adding a source-commit annotation. That would make it possible to check that the status matches the spec without external knowledge.

Guess we can do that in a different PR, but we're gonna need approval from @janetkuo

e2e/nomostest/wait_for_sync.go
```diff
@@ -37,13 +39,12 @@ import (
...
 
 type watchForAllSyncsOptions struct {
-	timeout            time.Duration
 	readyCheck         bool
 	syncRootRepos      bool
 	syncNamespaceRepos bool
 	skipRootRepos      map[string]bool
 	skipNonRootRepos   map[types.NamespacedName]bool
-	predicates         []testpredicates.Predicate
+	watchForSyncOpts   []WatchForSyncOption
```
karlkfi, line 47: Maybe pull out this Predicates -> WatchForSyncOptions change to its own PR first to reduce the size of this one a bit.

e2e/nomostest/sync.go
sdowell: That's exactly what the code below does.

e2e/testdata/hydration/compiled-json/helm-components/wordpress/secret_my-wordpress.json
```diff
@@ -0,0 +1,28 @@
+{
```
karlkfi, line 1: What are these new files for?

pkg/resourcegroup/controllers/root/root_controller.go
```diff
@@ -163,7 +163,8 @@ func (r *Reconciler) reconcileDisabledResourceGroup(ctx context.Context, req ctr
 	return r.reconcile(ctx, req.NamespacedName, []v1alpha1.ObjMetadata{}, true)
 }
 
-func isStatusDisabled(resgroup *v1alpha1.ResourceGroup) bool {
+// IsStatusDisabled returns whether the ResourceGroup has disabled status updates.
+func IsStatusDisabled(resgroup *v1alpha1.ResourceGroup) bool {
```
karlkfi, line 167: nit: Is there a better place for this than under the controllers package? Like maybe under pkg/resourcegroup?

pkg/resourcegroup/controllers/status/status.go
```diff
@@ -133,7 +133,11 @@ func GetSourceHash(annotations map[string]string) string {
 	if len(annotations) == 0 {
 		return ""
 	}
-	sourceHash := annotations[SourceHashAnnotationKey]
+	return TruncateSourceHash(annotations[SourceHashAnnotationKey])
+}
+
+// TruncateSourceHash truncates the provided source hash after the first 7 characters.
+func TruncateSourceHash(sourceHash string) string {
```
karlkfi, line 140: nit: Is there a better place for this than under the controllers package? Like maybe under pkg/resourcegroup?

e2e/testdata/hydration/helm-components/kustomization.yaml
karlkfi, line 25: what's this change for?

sdowell: See my other comment.

e2e/testdata/hydration/compiled-json/helm-components/wordpress/secret_my-wordpress.json
sdowell: The helm chart changed, and these files are used by another test to verify the contents of the hydrated helm chart.

e2e/nomostest/sync.go
sdowell: Done

pkg/resourcegroup/controllers/root/root_controller.go
sdowell: Done

pkg/resourcegroup/controllers/status/status.go
sdowell: Done

e2e/testdata/hydration/helm-components/kustomization.yaml
karlkfi: maybe pull this out to its own PR with its own description, along with the new files you added which i assume are for the same reason.

e2e/nomostest/sync.go
sdowell: Done

e2e/nomostest/wait_for_sync.go
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1594

e2e/testdata/hydration/helm-components/kustomization.yaml
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1595

e2e/nomostest/sync.go
karlkfi: Ah, right. I was kinda confused by the `resourceCount` map. 

It looks like you're validating that there's not more than one entry with the same ID, which seems good. But it's kinda hard to read.

I think there's a way to enforce that in the apiserver with schema validation, but it doesn't look like our CRD has those defined today. I know resources can define unique compound list keys. They added it for server-side apply list merging. I guess that's something we can add later.

What about building a fake `[]ResourceStatus` from the `status.resourceStatuses` and then using semantic.DeepEquals to compare it to the real `status.resources`. Then you could error with a yaml diff that would make it easy to see how they compare?

karlkfi: FWIW, this only works if they're sorted, which they might not be. So you might have to DeepCopy `status.resources` and then sort both the actual and expected lists before comparing them.

sdowell: Correct, the lists are not sorted - hence the decision to build a map to determine equality. Also it's not validating that the IDs are unique - this is purely verifying list equality. It's validating that the lists has the same set of IDs, including same number of duplicate IDs if there are duplicates.

karlkfi: Alright. I think there's some readability/correctness changed we could make here, but I don't want to hold up our other PRs, and I don't really think it's all that important for general validation we use in every e2e test. 

I wish we could just check the conditions here, and save all the other fields for validating in RG-specific tests, but that's not today's reality.

We can iterate on this later.

karlkfi (approved):
/lgtm

PR #1580, opened 2025-03-01
feat: Add --threshold flag to nomos vet

karlkfi:
- Added a new `--threshold[=MAX]` option to the `nomos vet` command that allows specifying a maximum number of objects. If set, it will enable validation that will error if the number of objects exceeds the specified value, after rendering and cluster selectors. By default, this validation is disabled. If you pass the option without a value, the default value of 1000 is used.
- The value 1000 was chosen as a compromise between scale and safety. Technically we know form e2e tests that the inventory can actually hold at least 5000 objects without needing to disable the status, however, this is unsafe to do in production because each of those objects could error, which adds error conditions to the inventory status, which can significantly increase the size of the inventory.
- Unlike other options, this option has an optional value and different defaults when specified than when not specified. This requires the flag and value to be sent in the same argument, like `--threshold=1000`, instead of as two different arguments, like `--threshold 1000`.
- If the option is specified with a value of zero or lower, the validation will be disabled, the same as if the option was not specified.

This replaces the PoC PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1531


cmd/nomos/vet/vet.go
```diff
@@ -43,6 +45,13 @@ func init() {
 	Cmd.Flags().BoolVar(&keepOutput, "keep-output", false,
 		`If enabled, keep the hydrated output`)
 
+	Cmd.Flags().IntVar(&threshold, "threshold", system.DefaultMaxObjectCount,
+		`If greater than zero, error if any repository contains more than the specified number of objects `)
+	// Set value to zero when `--threshold` is passed without a value.
+	// This requires the flag and value to be sent in the same argument, like
+	// `--threshold=1000`, instead of `--threshold 1000`.
+	Cmd.Flags().Lookup("threshold").NoOptDefVal = "0"
```
janetkuo, line 53: This is different from the PR description 
> If you pass the option without a value, the default value of 1000 is used.

cmd/nomos/vet/vet.go
```diff
@@ -43,6 +45,13 @@ func init() {
 	Cmd.Flags().BoolVar(&keepOutput, "keep-output", false,
 		`If enabled, keep the hydrated output`)
 
+	Cmd.Flags().IntVar(&threshold, "threshold", system.DefaultMaxObjectCount,
+		`If greater than zero, error if any repository contains more than the specified number of objects `)
+	// Set value to zero when `--threshold` is passed without a value.
+	// This requires the flag and value to be sent in the same argument, like
+	// `--threshold=1000`, instead of `--threshold 1000`.
```
janetkuo, line 52: We should allow both `--threshold=1000` and `--threshold 1000`

karlkfi: That's not possible with the cobra CLI library we're using. I'm not even sure it's possible with the Go flags library it wraps.

The only way to make the value optional is to set the `NoOptDefVal` field. This works by only checking a single argument and optionally allowing an `=` delimiter.

cmd/nomos/vet/vet.go
karlkfi: The PR description is correct.

The behavior Mike suggested is that the flag behaves differently whether it's specified or not:
- When --threshold is specified without a value, the default is 1000 (configured by `DefaultMaxObjectCount` passed to `IntVar`)
- When --threshold is NOT specified, the default is 0 (configured by `NoOptDefVal`)

cmd/nomos/vet/vet.go
karlkfi: Personally, I think this makes for confusing UX, and suggests that having 2 different defaults is unusual and undesirable.

cmd/nomos/vet/vet.go
karlkfi: > Set value to zero when `--threshold` is passed without a value.

Ah, the inline comment here is wrong.

cmd/nomos/vet/vet.go
karlkfi: There's a feature request open for the underlying pflag library: https://github.com/spf13/pflag/issues/321

One of the comments notes:

> Python's argparse module actually implements the behavior you desire, but I have yet to use another option/flag parser that has the same behavior.

cmd/nomos/vet/vet.go
karlkfi: Apparently I got it backwards. `NoOptDefVal` specifies the default to use when the flag is specified with no value. Then the normal default is used when the flag is not specified at all.

Added tests to validate this, and fixed the comments. But the PR description is still correct.

cmd/nomos/vet/vet_test.go
```diff
@@ -129,14 +160,173 @@ func TestVet_MultiCluster(t *testing.T) {
...
+			tmpDirPath, err := os.MkdirTemp("", "nomos-test-")
+			require.NoError(t, err)
+			t.Cleanup(func() {
+				err := os.RemoveAll(tmpDirPath)
+				require.NoError(t, err)
+			})
+
+			// Write objects to temp dir as yaml files
+			for _, fileObj := range tc.objects {
+				// fileName := fmt.Sprintf("%s-%s-%s.yaml", obj.Unstructured.GetKind(), obj.GetNamespace(), obj.GetName())
```
janetkuo, line 303: Remove this

janetkuo (commented):
Just a few nits and lgtm otherwise

cmd/nomos/vet/vet_test.go
karlkfi: done

cmd/nomos/vet/vet.go
```diff
@@ -43,6 +46,17 @@ func init() {
 	Cmd.Flags().BoolVar(&keepOutput, "keep-output", false,
 		`If enabled, keep the hydrated output`)
 
+	// The --threshold flag has three modes:
+	// 1. When `--threshold` is not specified, the validation is disabled.
+	// 2. When `--threshold` is specified with no value, the validation is enabled, using the default value, same as `--threshold=1000`.
+	// 3. When `--threshold=1000` is specified with a value, the validation is enabled, using the specified maximum.
+	Cmd.Flags().IntVar(&threshold, "threshold", 0,
+		`If greater than zero, error if any repository contains more than the specified number of objects `)
```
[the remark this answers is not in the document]
karlkfi: done

janetkuo (approved):
/lgtm

PR #1585, opened 2025-03-05
chore: remove duplicate kpt code

karlkfi:
The pkg/pkg directory was a duplicate of the pkg
directory, accidentally added when forked.


sdowell (approved):
/lgtm

PR #1588, opened 2025-03-06
fix: inventory client deleting unmanaged fields

karlkfi:
The ResourceGroup inventory client currently uses the output of the GetObject method to Update the ResourceGroup, but this makes it fight with the ResourceGroup controller, because it erases the status fields set by the RGC, mainly status.conditions.

This fix isn't quite as good as using server-side apply, but that would require the inventory client and RGC to use different field manager names.

This does not fix removal of conditions under status.resourceStatuses, only removal of status.conditions.

Fixes: b/401304237


sdowell (approved):
/lgtm

PR #1591, opened 2025-03-07
e2e: Increase log level for ResourceGroup tests

karlkfi:


sdowell:
Suggest we submit https://github.com/GoogleContainerTools/kpt-config-sync/pull/1592 first to mitigate the size of the stdout log dump

e2e/testcases/custom_resources_test.go
```diff
@@ -51,12 +51,8 @@ func TestCRDDeleteBeforeRemoveCustomResourceV1(t *testing.T) {
 	nt.Must(nomostest.WaitForCRDs(nt, []string{"anvils.acme.com", "clusteranvils.acme.com"}))
 
 	nt.T.Log("Increase log level of reconciler to help debug failures")
```
tiffanny29631, line 53: Is this flakiness still a case? Does it still require increased verbosity?

karlkfi: This test is pretty complicated. I don't know if it's flaky or not.

It looks like the conflict metric is overloaded to include multiple different things:
- management conflicts (unexpected manager annotation)
- resource conflicts (resource version change)

That makes the metric value unpredictable. It should always be at least 1, but if it's more we can't tell exactly why.

Seems reasonable to just leave it for now. I don't really have time now to propose metrics changes.

karlkfi:
> Suggest we submit #1592 first to mitigate the size of the stdout log dump

Done

e2e/testcases/custom_resources_test.go
tiffanny29631: Ack

tiffanny29631 (approved):
/lgtm

PR #1592, opened 2025-03-07
test: truncate Pod logs for each test case

sdowell:
There are some long lived containers, such as reconciler-manager, which will accumulate a large amount of logs over the duration of the test suite. When a test fails, it dumps the Pod logs and can result in a large quantity of logs which are not relevant to the failed test.

This truncates the logs based on the start time of the test, to ensure only relevant logs are included.

### DIFF
--- a/e2e/nomostest/nt.go
+++ b/e2e/nomostest/nt.go
@@ -202,6 +202,9 @@ type NT struct {
 
 	// repoSyncPermissions will grant a list of PolicyRules to NS reconcilers
 	repoSyncPermissions []rbacv1.PolicyRule
+
+	// testStartTime is a timestamp which records when the current test started
+	testStartTime time.Time
 }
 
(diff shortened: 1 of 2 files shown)


e2e/nomostest/nt.go
```diff
@@ -463,7 +466,11 @@ func (nt *NT) MustKubectl(args ...string) []byte {
 func (nt *NT) PodLogs(namespace, deployment, container string, previousPodLog bool) {
 	nt.T.Helper()
 
-	args := []string{"logs", fmt.Sprintf("deployment/%s", deployment), "-n", namespace}
+	args := []string{"logs", fmt.Sprintf("deployment/%s", deployment),
```
karlkfi, line 469: nit: log entry on new line

tiffanny29631 (approved):
/lgtm

PR #1593, opened 2025-03-07
e2e: Use Semantic.DeepEqual to test empty status

karlkfi:
- reflect.DeepEqual worked, but Semantic.DeepEqual is preferred for Kubernetes resources.
- Error with the YAML diff if the status is still set

### DIFF
--- a/e2e/testcases/status_enablement_test.go
+++ b/e2e/testcases/status_enablement_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"kpt.dev/configsync/e2e/nomostest"
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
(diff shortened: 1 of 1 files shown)


e2e/testcases/status_enablement_test.go
```diff
@@ -101,9 +100,9 @@ func resourceGroupHasStatus(obj client.Object) error {
 	if !ok {
 		return testpredicates.WrongTypeErr(obj, &resourcegroupv1alpha1.ResourceGroup{})
 	}
-	// When status is enabled, the resource statuses are computed and populated.
-	if len(rg.Status.ResourceStatuses) == 0 {
-		return fmt.Errorf("found empty status in %s", core.IDOf(obj))
+	emptyStatus := resourcegroupv1alpha1.ResourceGroupStatus{}
+	if equality.Semantic.DeepEqual(emptyStatus, rg.Status) {
```
sdowell, line 104: This seems like a weak assertion - can the test use `testpredicates.ResourceGroupStatusEquals` instead?

karlkfi: Yes, but the output isn't consistent yet until after some other PRs merge.

karlkfi: - https://github.com/GoogleContainerTools/kpt-config-sync/pull/1570 should have fixed the `status.conditions` timestamps continuously updating (b/399892476)
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1588 should have fixed the `status.conditions` disappearing (b/401304237)
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1571 should have fixed the Reconciling condition flip/flopping (b/399896123)
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1575 should fix `status.observedGeneration`
- TODO: fix `status.resourceStatuses[*].conditions` disappearing (b/401315927)

My preference would just be to keep this weak assertion for now, until all the above fixes have landed, especially because the last one might not get fixed until after we land the inventory client rewrite.

sdowell (approved):
/lgtm

PR #1594, opened 2025-03-07
test: refactor WatchForSync to take opt struct

sdowell:
This refactor adds flexibility to WatchForSync to accept options in addition to WatchOption. This is built on by a subsequent PR.

The usage of ResourceGroup predicates are removed because WatchForAllSyncs watches RSyncs, so these predicates were invalid.

### DIFF
--- a/e2e/nomostest/wait_for_sync.go
+++ b/e2e/nomostest/wait_for_sync.go
@@ -37,13 +37,12 @@ import (
 )
 
 type watchForAllSyncsOptions struct {
-	timeout            time.Duration
 	readyCheck         bool
 	syncRootRepos      bool
 	syncNamespaceRepos bool
 	skipRootRepos      map[string]bool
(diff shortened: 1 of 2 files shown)


e2e/nomostest/wait_for_sync.go
```diff
@@ -133,13 +125,6 @@ func (nt *NT) WatchForAllSyncs(options ...WatchForAllSyncsOptions) error {
 		}
 	}
 
-	watchOptions := []testwatcher.WatchOption{
-		testwatcher.WatchTimeout(opts.timeout),
```
karlkfi, line 137: Did this default timeout get lost?

sdowell: No. The default timeout is also configured directly on the Watcher struct.

karlkfi (approved):
/lgtm

PR #1595, opened 2025-03-07
test: fix helm-components chart so that it reconciles

sdowell:
The helm chart referenced a Secret which is expected to be created out of band, but the Secret did not exist. This caused the package to never reconcile.

Replacing this with an inline wordpressPassword handles creation of the Secret by the helm chart so that it can reconcile.

### DIFF
--- a/e2e/testdata/hydration/compiled-json/helm-components/wordpress/secret_my-wordpress.json
+++ b/e2e/testdata/hydration/compiled-json/helm-components/wordpress/secret_my-wordpress.json
@@ -1,0 +1,28 @@
+{
+	"apiVersion": "v1",
+	"items": [
+		{
+			"apiVersion": "v1",
+			"data": {
+				"wordpress-password": "YWJjZGVmZw=="
+			},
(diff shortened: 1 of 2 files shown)


e2e/testdata/hydration/compiled-json/helm-components/wordpress/secret_my-wordpress.json
```diff
@@ -0,0 +1,28 @@
...
+					"test-case": "hydration"
+				},
+				"name": "my-wordpress",
+				"namespace": "wordpress"
+			},
+			"type": "Opaque"
+		}
+	],
+	"kind": "List"
+}
```
karlkfi, line 28: nit: linebreak

sdowell: This is one of many autogenerated JSON files. The test fails if you change the checksum of the file.

karlkfi (approved):
/lgtm

PR #1596, opened 2025-03-10
test: cleanup ResourceGroup controller e2e tests

karlkfi:
Pulled out most these changed from https://github.com/GoogleContainerTools/kpt-config-sync/pull/1575, which should make it easier to see the coordination changes in that PR if this one merges first.

Dependencies:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1591

### DIFF
--- a/e2e/testcases/resource_group_controller_test.go
+++ b/e2e/testcases/resource_group_controller_test.go
@@ -23,7 +23,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"kpt.dev/configsync/e2e/nomostest"
-	"kpt.dev/configsync/e2e/nomostest/retry"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/e2e/nomostest/testresourcegroup"
@@ -32,12 +31,12 @@ import (
 	"kpt.dev/configsync/pkg/api/configmanagement"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/kpt.dev/v1alpha1"
-	"kpt.dev/configsync/pkg/applier"
 	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/core/k8sobjects"
 	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/reconcilermanager"
 	"kpt.dev/configsync/pkg/resourcegroup"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func TestResourceGroupController(t *testing.T) {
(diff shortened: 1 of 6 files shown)


e2e/testcases/resource_group_controller_test.go
```diff
@@ -123,28 +114,48 @@ func TestResourceGroupControllerInKptGroup(t *testing.T) {
 			},
 		},
 	}
-	nt.Must(testresourcegroup.UpdateResourceGroup(nt.KubeClient, rgNN, func(rg *v1alpha1.ResourceGroup) {
```
karlkfi, line 126: FYI, UpdateResourceGroup works fine, but Get & Update is the same number of lines and more consistent with how other tests do the same thing.

We could probably remove UpdateResourceGroup, but it's used in a couple other files I didn't want to touch in this PR.

e2e/testcases/resource_group_controller_test.go
```diff
@@ -169,33 +180,31 @@ func TestResourceGroupControllerInKptGroup(t *testing.T) {
...
 			testpredicates.ResourceGroupStatusEquals(expectedStatus),
 		), testwatcher.WatchTimeout(time.Minute)))
 
-	if err := testresourcegroup.CreateOrUpdateResources(nt.KubeClient, resources, resourceID); err != nil {
-		nt.T.Fatal(err)
-	}
+	nt.T.Log("Simulate the Reconciler applying the managed resources")
+	nt.Must(testresourcegroup.CreateOrUpdateResources(nt.KubeClient, resources, resourceID))
 
-	expectedStatus.ObservedGeneration = 3
```
sdowell, line 182: Why was the observedGeneration expectation changed?

karlkfi: It didn't. It was already being set above on line 148. This was a no-op.

e2e/testcases/resource_group_controller_test.go
```diff
@@ -239,45 +239,50 @@ func TestResourceGroupCustomResource(t *testing.T) {
...
 	expectedStatus.ObservedGeneration = 1
 	nt.Must(nt.Watcher.WatchObject(kinds.ResourceGroup(), rgNN.Name, rgNN.Namespace,
 		testwatcher.WatchPredicates(
 			testpredicates.ResourceGroupStatusEquals(expectedStatus),
 		)))
-	crdObj := anvilV1CRD()
-	crdGVK := anvilGVK("v1")
+
+	nt.T.Log("Add an Anvil object to the ResourceGroup spec")
+	anvilGVK := anvilGVK("v1")
```
sdowell, line 269: Variable name collides with function name

e2e/testcases/resource_group_controller_test.go
sdowell: I see, that's fine but probably makes it a bit less clear as the reader needs to go reference the previous assignment

e2e/testcases/resource_group_controller_test.go
karlkfi: Fixed in a new commit

sdowell (approved):
/lgtm

PR #1597, opened 2025-03-11
test: fix uses of RetryOnConflict

karlkfi:
To avoid conflicts, you have to get the object from the cluster before each retry, to get the latest resource version. It's also good to rebuild the object, to reset the metadata, like UID and resourceVersion, to make sure you get the latest object from the cluster, not an old version.

Fixes: b/399904244


sdowell (approved):
/lgtm

PR #1598, opened 2025-03-11
test: fix coredns helm chart to reconcile

sdowell:
The ConfigMap and Service were always created in the default namespace, which resulted in the Deployment never being able to reconcile. This change ensures the other objects are created in the same namespace as the Deployment.


Camila-B (approved):
/lgtm

karlkfi (approved):
/lgtm

PR #1599, opened 2025-03-12
test: skip resourcegroup checks in stress tests

sdowell:
These stress tests either take too long to reconcile or do not reconcile, and therefore should skip ResourceGroup checks.


Camila-B (approved):
/lgtm

karlkfi (commented):
This will probably make the tests pass, but there's still an underlying issue to fix.

PR #1600, opened 2025-03-12
test: fix ResourceGroup expectations

karlkfi:
- Conditions will be empty when status is disabled
- status.observedGeneration will be 0 when status is disabled
- Validate status.resourceStatuses with DeepEqual and log the yaml diff on error

Extracted:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1601
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1602
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/1603

### DIFF
--- a/e2e/nomostest/sync.go
+++ b/e2e/nomostest/sync.go
@@ -17,16 +17,21 @@ package nomostest
 import (
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	"kpt.dev/configsync/pkg/api/kpt.dev/v1alpha1"
+	"kpt.dev/configsync/pkg/kinds"
 	"kpt.dev/configsync/pkg/parse"
 	"kpt.dev/configsync/pkg/reposync"
 	"kpt.dev/configsync/pkg/resourcegroup"
 	"kpt.dev/configsync/pkg/rootsync"
 	"kpt.dev/configsync/pkg/util/log"
+	kstatus "sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -266,7 +271,7 @@ func statusHasSyncDirAndNoErrors(status v1beta1.Status, sourceType configsync.So
 	return nil
 }
 
-func resourceGroupHasReconciled(sourceHash string) testpredicates.Predicate {
+func resourceGroupHasReconciled(commit string, scheme *runtime.Scheme) testpredicates.Predicate {
 	return func(o client.Object) error {
 		if o == nil {
 			return testpredicates.ErrObjectNotFound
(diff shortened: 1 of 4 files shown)


e2e/testcases/stress_test.go
```diff
@@ -342,8 +341,7 @@ func TestStressManyDeployments(t *testing.T) {
 
 	// Validate that the resources sync without the reconciler running out of
 	// memory, getting OOMKilled, and crash looping.
-	// Skip ResourceGroup checks because it takes too long for this many Deployments to reconcile
-	nt.Must(nt.WatchForAllSyncs(nomostest.SkipAllResourceGroupChecks()))
+	nt.Must(nt.WatchForAllSyncs())
```
sdowell, line 344: Why is this expected to reconcile now?

karlkfi: It should always have reconciled. The reconcile timeout is set to 5m in the applier, which makes each apply attempt wait the normal amount of time, WatchObject uses the DefaultWaitTimeout of 30m for stress tests, and the WatchObject for the RG starts after the WatchObject for the RSync succeeds. So it has plenty of time for the Deployment Pods to schedule, even on autopilot.

e2e/nomostest/sync.go
```diff
@@ -276,47 +281,49 @@ func resourceGroupHasReconciled(sourceHash string) testpredicates.Predicate {
...
+				Actuation:   v1alpha1.ActuationSucceeded,
+				Reconcile:   v1alpha1.ReconcileSucceeded,
+				SourceHash:  shortCommit,
+				Conditions:  nil,
 			}
 		}
-		return nil
-	}
-}
+		if !equality.Semantic.DeepEqual(found, expected) {
```
karlkfi, line 321: I'm actually pretty surprised this passed all the e2e tests... I thought I was going to have to sort them.

But maybe we don't have any tests that churn in the right way to break the ordering...

e2e/testcases/stress_test.go
sdowell: I have doubts about that, but I suppose we'll see. I don't see changes here that address the existing failures for this test

karlkfi: I'm running it locally. Will know shortly.

karlkfi: Looks like you're right. Only about 350 are being correctly scheduled on the standard stress clusters.

```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2025-03-12T20:53:42Z"
    message: '0/3 nodes are available: 3 Too many pods. preemption: 0/3 nodes are
      available: 3 No preemption victims found for incoming pod.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
  qosClass: Guaranteed
```

Not super clear why the nodes are "not available" tho.

Each of the nodes has a lot of available Memory and CPU too.

The ReplicaSets also seem to be timing out reporting status.

karlkfi: Ah. It's probably because of the node's IP limit, which is 110 by default.

karlkfi: Yup. I got it working with 4 nodes per zone in 3 zones.

karlkfi: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1601

e2e/nomostest/wait_for_sync.go
```diff
@@ -263,17 +261,7 @@ func (nt *NT) WatchForSync(
...
-			testpredicates.HasObservedLatestGeneration(nt.Scheme),
-			// Wait until metadata.deletionTimestamp is missing, and conditions do not include Reconciling=True or Stalled=True
-			testpredicates.StatusEquals(nt.Scheme, kstatus.CurrentStatus),
-			// Make sure the ResourceGroup has the Stalled condition and it is "False"
-			// This ensures the condition exists and wasn't removed by the applier
-			testpredicates.HasConditionStatus(nt.Scheme, string(v1alpha1.Stalled), corev1.ConditionFalse),
-			// Wait until all resourceStatuses are consistent with spec and current/reconciled
-			resourceGroupHasReconciled(expectedCommit),
-		}...),
+		testwatcher.WatchPredicates(resourceGroupHasReconciled(expectedCommit, nt.Scheme)),
```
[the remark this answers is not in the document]
karlkfi: Done

e2e/nomostest/sync.go
karlkfi: Well, I guess if it breaks we can always add sorting later.

sdowell (approved):
/lgtm

PR #1601, opened 2025-03-12
test: Fix TestStressManyDeployments reconciling

karlkfi:
For stress tests, use more smaller nodes:
- 3 -> 12 nodes
- n2-standard-8 -> n2-standard-4
- 50Gb -> 25Gb disks

This works around the 110 pod per node limit in GKE, allowing the test to schedule 1,000 pods. The pods are tiny and just sleep.

### DIFF
--- a/e2e/flags.go
+++ b/e2e/flags.go
@@ -211,11 +211,13 @@ var GKEClusterVersion = flag.String("gke-cluster-version", util.EnvString("GKE_C
 	"GKE cluster version to use when creating GKE clusters. Defaults to GKE_CLUSTER_VERSION env var.")
 
 // GKEMachineType is the GKE machine type to use when creating GKE clusters.
-var GKEMachineType = flag.String("gke-machine-type", util.EnvString("GKE_MACHINE_TYPE", DefaultGKEMachineType),
+// See testcases.setDefaultArgs for dynamic defaults.
+var GKEMachineType = flag.String("gke-machine-type", "",
 	"GKE machine type to use when creating GKE clusters. Defaults to GKE_MACHINE_TYPE env var.")
 
 // GKEDiskSize is the GKE disk size to use when creating GKE clusters.
-var GKEDiskSize = flag.String("gke-disk-size", util.EnvString("GKE_DISK_SIZE", DefaultGKEDiskSize),
+// See testcases.setDefaultArgs for dynamic defaults.
+var GKEDiskSize = flag.String("gke-disk-size", "",
 	"GKE disk size to use when creating GKE clusters. Defaults to GKE_DISK_SIZE env var.")
 
 // GKEDiskType is the GKE disk type to use when creating GKE clusters.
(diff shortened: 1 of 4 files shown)


e2e/flags.go
```diff
@@ -257,12 +257,32 @@ const (
 	DefaultGKEChannel = "regular"
 	// DefaultGKEMachineType is the default GKE machine type to use when creating a cluster
 	DefaultGKEMachineType = "n2-standard-8"
+	// DefaultGKEMachineTypeForStressTests is the default GKE machine type to
```
sdowell, line 260: fyi if you want these to be used by the prow jobs you'll need to update the config: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml#L383

karlkfi: thanks. we should probably just remove the GKE_MACHINE_TYPE there, since we had the same as default anyway

e2e/nomostest/clusters/gke.go
```diff
@@ -222,16 +222,31 @@ func createGKECluster(t testing.NTB, name string) error {
 	// Standard-specific options (cannot specify for autopilot)
 	if !*e2e.GKEAutopilot {
 		if *e2e.GKEMachineType != "" {
-			args = append(args, "--machine-type", *e2e.GKEMachineType)
+			machineType := *e2e.GKEMachineType
+			// Override default value for stress tests
+			if *e2e.Stress && machineType == e2e.DefaultGKEMachineType {
```
sdowell, line 227: This could be a bit awkward if the user explicitly specifies the machine type which happens to be `DefaultGKEMachineType` while running stress tests

karlkfi: I know. Couldn't think of another solution. Got a suggestion?

The --stress flag value isn't available at init time when the --machine-type default value is registered. And it needs to also fall back to the env var if not specified.

I could set the default to the empty string when configuring the flag, but then we have to process all the variables after parsing the flags. Since `go test` owns the command handler, including flag parsing, I don't know where to process the flags after parsing that won't cause race conditions when running tests in parallel. After init, the earliest code we have is nomostest.New, which runs in every test function.

karlkfi: The only other solution I could come up with was to process the flags in every test, somewhere in nomostest.New, with a mutex or sync.Once. Seemed kinda overkill.

karlkfi: I guess I could make a `TestMain(m *testing.M)` and parse flags there...

karlkfi: Oh, hey, we have one already. I'll add it there.

karlkfi: Done

e2e/flags.go
karlkfi: I think we'll want to merge this first, then we can remove the env vars from the prow jobs

karlkfi: Looks like the stress tests inherit from the non-stress config, which makes it hard to remove the defaults for stress tests without removing the standard default. So I might have to make another PR for that.

karlkfi: Plus, it looks like prow already has standard clusters using `n2-standard-4` nodes with 25Gb disk. So The defaults in the code are out of date.

karlkfi: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2449

sdowell (approved):
/lgtm

PR #1602, opened 2025-03-13
test: make KubeClient print object kind and type

karlkfi:
This makes it easier to tell what object was created when using Unstructured objects.


sdowell (approved):
/lgtm

PR #1609, opened 2025-03-14
test: Fix TestKCCResourceGroup

karlkfi:
- Update status.observedGeneration to unblock the ResourceGroup controller.
- Use nt.Must
- Add log messages for each test step
- Replace kubeDeleteIgnoreNotFound with nt.KubeClient.DeleteIfExists
- use nt.T.Cleanup to delete the test ResourceGroups


sdowell (approved):
/lgtm

PR #1611, opened 2025-03-14
feat: Delete the ResourceGroup when the RSync is deleted

Camila-B:
- Makes sure that the ResourceGroup is deleted when the RSync is deleted, regardless of the deletion propagation policy 
- b/382533338

### DIFF
--- a/e2e/testcases/reconciler_finalizer_test.go
+++ b/e2e/testcases/reconciler_finalizer_test.go
@@ -128,6 +128,10 @@ func TestReconcilerFinalizer_Orphan(t *testing.T) {
 				testpredicates.HasLabel(metadata.ApplySetPartOfLabel, applySetID),
 			))
 	})
+	tg.Go(func() error {
+		// ResourceGroup SHOULD have been deleted
+		return nt.Watcher.WatchForNotFound(kinds.ResourceGroup(), rootSync.GetName(), rootSync.GetNamespace())
+	})
 
 	nt.Must(tg.Wait())
 
@@ -348,6 +352,9 @@ func TestReconcilerFinalizer_MultiLevelForeground(t *testing.T) {
 		return nt.Watcher.WatchForNotFound(kinds.Namespace(), namespace1.GetName(), namespace1.GetNamespace())
 	})
 	tg.Go(func() error {
+		return nt.Watcher.WatchForNotFound(kinds.ResourceGroup(), rootSync.GetName(), rootSync.GetNamespace())
+	})
+	tg.Go(func() error {
 		return nt.Watcher.WatchForNotFound(kinds.RootSyncV1Beta1(), rootSync.GetName(), rootSync.GetNamespace())
 	})
 	if err := tg.Wait(); err != nil {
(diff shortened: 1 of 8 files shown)


e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -127,9 +128,20 @@ func TestReconcilerFinalizer_Orphan(t *testing.T) {
 				testpredicates.HasLabel(metadata.ApplySetPartOfLabel, applySetID),
 			))
 	})
-	if err := tg.Wait(); err != nil {
-		nt.T.Fatal(err)
-	}
+
```
karlkfi, line 131: might want to also check here that the RG was deleted, like you added in the Foreground test

e2e/testcases/reconciler_manager_test.go
```diff
@@ -453,6 +466,24 @@ func validateRepoSyncDependencies(nt *nomostest.NT, ns, rsName string) []client.
 	return repoSyncDependencies
 }
 
+func validateRootSyncResourceGroup(nt *nomostest.NT, rsName string) *v1alpha1.ResourceGroup {
```
karlkfi, line 469: generally for helper functions it's best to return an error, if possible, instead of failing inside the helper func. It makes them more re-usable later, because the caller can decide if the error fails the test or not.

pkg/core/scheme.go
```diff
@@ -40,6 +40,7 @@ import (
 	configsyncv1alpha1 "kpt.dev/configsync/pkg/api/configsync/v1alpha1"
 	configsyncv1beta1 "kpt.dev/configsync/pkg/api/configsync/v1beta1"
 	hubv1 "kpt.dev/configsync/pkg/api/hub/v1"
+	kptdevv1alpha1 "kpt.dev/configsync/pkg/api/kpt.dev/v1alpha1"
```
karlkfi, line 43: nit: kptv1alpha1 is probably find. the .dev TLD is inconsequential, but it's not a big deal at all

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -480,6 +480,10 @@ func (r *RepoSyncReconciler) deleteManagedObjects(ctx context.Context, reconcile
 		return fmt.Errorf("deleting service account: %w", err)
 	}
 
+	if err := r.deleteResourceGroup(ctx, rsRef); err != nil {
```
karlkfi, line 483: Am I right in understanding that the reconciler-manager is handling deletion of the RG in the normal Reconciler loop, not in the RootSyncFinalizer/RepoSyncFinalizer?

e2e/testcases/reconciler_finalizer_test.go
Camila-B: Done

e2e/testcases/reconciler_manager_test.go
Camila-B: Sounds good. I refactored the function to return both the ResourceGroup and the error

pkg/core/scheme.go
Camila-B: Done

pkg/reconcilermanager/controllers/reposync_controller.go
Camila-B: Yes that's correct

karlkfi (approved):
/lgtm

Even less complicated than I expected. Nice work.

PR #1613, opened 2025-03-18
chore: update base image, Helm and Otelcontribcol

tiffanny29631:
CVE-2024-40635
CVE-2025-0395

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ GOLANG_IMAGE := google-go.pkg.dev/golang:$(GOLANG_IMAGE_VERSION)
 # Base image used for debian containers
 # When updating you can use this command: 
 # gcloud container images list-tags gcr.io/gke-release/debian-base --filter="tags:bookworm*" 
-DEBIAN_BASE_IMAGE := gcr.io/gke-release/debian-base:bookworm-v1.0.4-gke.4
+DEBIAN_BASE_IMAGE := gcr.io/gke-release/debian-base:bookworm-v1.0.4-gke.5
 # Base image used for gcloud install, primarily for test images.
 # We use -slim for a smaller base image where we can choose which components to install.
 # https://cloud.google.com/sdk/docs/downloads-docker#docker_image_options
@@ -79,7 +79,7 @@ KUSTOMIZE_VERSION := v5.4.2-gke.0
 KUSTOMIZE := $(BIN_DIR)/kustomize
 KUSTOMIZE_STAGING_DIR := $(OUTPUT_DIR)/third_party/kustomize
 
-HELM_VERSION := v3.15.3-gke.4
+HELM_VERSION := v3.15.3-gke.5
 HELM := $(BIN_DIR)/helm
 HELM_STAGING_DIR := $(OUTPUT_DIR)/third_party/helm
 
@@ -89,7 +89,7 @@ COSIGN := $(BIN_DIR)/cosign
 GIT_SYNC_VERSION := v4.3.0-gke.13__linux_amd64
 GIT_SYNC_IMAGE_NAME := gcr.io/config-management-release/git-sync:$(GIT_SYNC_VERSION)
 
-OTELCONTRIBCOL_VERSION := v0.118.0-gke.4
+OTELCONTRIBCOL_VERSION := v0.118.0-gke.5
 OTELCONTRIBCOL_IMAGE_NAME := gcr.io/config-management-release/otelcontribcol:$(OTELCONTRIBCOL_VERSION)
 
 # Directory used for staging Docker contexts.
--- a/pkg/hydrate/tool_util.go
+++ b/pkg/hydrate/tool_util.go
@@ -51,7 +51,7 @@ import (
 
 const (
 	// HelmVersion is the recommended version of Helm for hydration.
-	HelmVersion = "v3.15.3-gke.4"
+	HelmVersion = "v3.15.3-gke.5"
 	// KustomizeVersion is the recommended version of Kustomize for hydration.
 	KustomizeVersion = "v5.4.2-gke.0"
 	// Helm is the binary name of the installed Helm.


e2e/testdata/hydration/helm-overlay/base/kustomization.yaml
```diff
@@ -18,10 +18,12 @@ helmCharts:
...
   namespace: coredns
-  valuesFile: ../base_value_overrides.yaml
+  valuesFile: base_value_overrides.yaml
 
 commonAnnotations:
   hydration-tool: kustomize
 
-commonLabels:
-  team: coredns
+labels:
```
sdowell, line 26: Why do these helm overlays need to change with a security patch bump?

janetkuo (approved):

e2e/testdata/hydration/helm-overlay/base/kustomization.yaml
janetkuo: Ideally this could be a separate PR. This change is because commonLabels is deprecated. The behavior should be the same after this PR.

sdowell: Are you saying the security patch (for the `helm` binary I assume) deprecated/removed `commonLabels`?

janetkuo: I mean kustomize deprecated `commonLabels` and users should migrate to use `labels` instead.

sdowell: I'm trying to understand why is the test breaking now with this change?

tiffanny29631: Running the previous failing test TestNomosHydrateAndVetDryRepos from the head of branch would still fail with the warning from Kustomization about deprecated field -  the warning will end up as a helm error. I found https://github.com/kubernetes-sigs/kustomize/issues/5436 but no obvious documentation on the change of `commonLabels` field, not sure why it would suddenly start failing with a patch update of Helm alone, without change in Kustomize.

sdowell: Seems like we should understand the regression, which is an unexpected behavioral change for a security patch.

Can we at least separate the test change to a separate PR and submit that first? It seems like it should be unrelated (e.g. for the purpose of cherry-pick/revert)

tiffanny29631: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1614

sdowell: Do you intend to determine the root cause for the regression?

tiffanny29631: I've been pulling my hair for an hour, https://github.com/GoogleContainerTools/kpt-config-sync/pull/1607 was having the same error but was overlooked by all of us. First thought was something to do with the fix in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1595 and https://github.com/GoogleContainerTools/kpt-config-sync/pull/1598 when test finally reconciles but checking out the code before them would still fail the test. I would guess it's something environmentally causing the warning to surface more strictly.. But still it's not explaining why it would pass before.

sdowell: Looking back at the job history, I don't think the problem was with the usage of `commonLabels`.

See examples of the failing jobs:
- https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/1613/kpt-config-sync-presubmit-e2e-multi-repo-test-group1/1902108569143087104
- https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/pr-logs/pull/GoogleContainerTools_kpt-config-sync/1613/kpt-config-sync-presubmit-e2e-multi-repo-test-group1/1902154277317513216

There is a warning about `commonLabels`, but that's not what's causing the test to fail. It looks like there's a DNS lookup failure which is causing the test to fail.

e.g.
```
{Failed      new.go:107: 2025-03-18 21:29:12.592265172 +0000 UTC created temporary directory "/tmp/nomos-e2e/test-nomos-hydrate-and-vet-dry-repos--hydrate-error:-duplicate-resources2079960854"
    cli_test.go:811: 2025-03-18 21:29:12.713369492 +0000 UTC ERROR: hydrate error: duplicate resources: expected error 'may not add resource with an already registered id', but got 'Error: unable to render the source configs in /repo/e2e/testdata/hydration/resource-duplicate: failed to run kustomize build in /repo/e2e/testdata/hydration/resource-duplicate, stdout: : # Warning: 'commonLabels' is deprecated. Please use 'labels' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
        Error: accumulating resources: accumulation err='accumulating resources from 'github.com/config-sync-examples/kustomize-components?ref=main': open /repo/e2e/testdata/hydration/resource-duplicate/github.com/config-sync-examples/kustomize-components?ref=main: no such file or directory': failed to run '/usr/bin/git fetch --depth=1 https://github.com/config-sync-examples/kustomize-components main': fatal: unable to access 'https://github.com/config-sync-examples/kustomize-components/': Could not resolve host: github.com
        : exit status 128
        
        '}
```

sdowell (approved):
/lgtm

e2e/testdata/hydration/helm-overlay/base/kustomization.yaml
sdowell: Additionally, the DNS lookup failure appears to be unrelated to this change and is already causing tests to fail on HEAD

sdowell: Sent https://github.com/GoogleContainerTools/kpt-config-sync/pull/1616 as a potential fix.

PR #1614, opened 2025-03-20
chore: fix Kustomize deprecated field commonLabels in helm test

tiffanny29631:
Related upstream change https://github.com/kubernetes-sigs/kustomize/pull/5464/files


sdowell (approved):
This change looks fine, but it still seems unexpected that the security patch in https://github.com/GoogleContainerTools/kpt-config-sync/pull/1613 broke these tests

/lgtm

PR #1616, opened 2025-03-21
test: persist helm charts locally

sdowell:
This changes the helm charts used by the cli tests to be persisted locally instead of fetched from a remote third party helm repository. This is intended to fix some flaky DNS lookup errors that have been observed from the presubmit environment.

This also generally makes the tests more hermetic and minimizes reliance on third party infra.


e2e/testcases/cli_test.go
```diff
@@ -742,13 +742,15 @@ func TestNomosHydrateAndVetDryRepos(t *testing.T) {
...
 			expectedOutPath: "../testdata/hydration/compiled/helm-overlay",
 		},
-		{
-			name:            "hydrate a DRY repo with remote base",
-			path:            "../testdata/hydration/remote-base",
-			outPath:         "remote-base/compiled",
-			sourceFormat:    string(configsync.SourceFormatUnstructured),
-			expectedOutPath: "../testdata/hydration/compiled/remote-base",
-		},
+		// remote-base relies on external public helm chart, which occasionally
```
tiffanny29631, line 745: Will the removal of the remote-based test case result in decreased test coverage? The case seems to be passing fine in the post submit CI.

sdowell: > The case seems to be passing fine in the post submit CI.

It's not passing fine in the periodics. You can see that it's very flaky as it is in the presubmits: https://testgrid.k8s.io/googleoss-kpt-config-sync-main#kind

> Will the removal of the remote-based test case result in decreased test coverage?

Yes this means less test coverage, although this is more-so verifying kustomize functionality than Config Sync functionality.

PR #1617, opened 2025-03-24
test: log more details for unexpected status

karlkfi:
- Include the full kstatus response in the error when the StatusEquals predicate fails, including the message and conditions.
- Update WaitForNamespace to use the StatusEquals predicate.


sdowell (approved):
/lgtm

PR #1618, opened 2025-03-25
test: refine ResourceGroup expectations

karlkfi:
Update resourceGroupHasReconciled to make it easier to debug the errors, by grouping all the status checks together, so there's one big diff when it fails.
Use ResourceGroupStatusEquals because it already handles removing the timestamps before comparison.


sdowell (approved):
/lgtm

PR #1619, opened 2025-03-26
refactor supervisor constructor

sdowell:
* refactor: streamline the supervisor constructor

This refactors the supervisor constructor to consolidate the
RepoSync/RootSync supervisors into a single constructor.

The AnnotateStatusMode step is also moved to a separate interface method
so that the constructor does not make api calls or return an error.

* fix: retry on conflict for annotating RG status mode

This wraps AnnotateStatusMode so that it handles conflict errors
gracefully instead of causing the reconciler to crash.

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/applier/stats"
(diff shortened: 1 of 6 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -146,6 +146,8 @@ type RootOptions struct {
 
 // Run configures and starts the various components of a reconciler process.
 func Run(opts Options) {
+	// Start listening to signals
+	signalCtx := signals.SetupSignalHandler()
```
tiffanny29631, line 150: Mostly lgtm, just curious why the SetupSignalHandler is moving to the front

sdowell: So that the context can be passed to `AnnotateStatusMode`

pkg/applier/applier.go
```diff
@@ -121,6 +122,7 @@ func (e ErrorEvent) Type() EventType {
 type Supervisor interface {
 	Applier
 	Destroyer
+	AnnotateStatusMode(ctx context.Context) error
```
karlkfi, line 125: nit: UpdateStatusMode is probably slightly better, to avoid the caller needing to know it's an annotation. Tho it's a little weird since it doesn't take a new status mode. I guess it could, as long as it used the lock to avoid races.

pkg/applier/applier.go
```diff
@@ -150,77 +152,56 @@ var _ Supervisor = &supervisor{}
...
+// AnnotateStatusMode sets the status mode annotation on the ResourceGroup object
+// if it exists.
+// This is invoked once before the applier runs so that ResourceGroups which are
+// too large can have their status cleared by the resource-group-controller.
+func (s *supervisor) AnnotateStatusMode(ctx context.Context) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		u := newInventoryUnstructured(s.syncName, s.syncNamespace)
+		err := s.clientSet.Client.Get(ctx, client.ObjectKey{Name: u.GetName(), Namespace: u.GetNamespace()}, u)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
```
karlkfi, line 191: nit: add comment: If the ResourceGroup doesn't exist yet, cli-utils will create it with the right status policy when it starts the next apply/destroy.

pkg/applier/applier.go
```diff
@@ -150,77 +152,56 @@ var _ Supervisor = &supervisor{}
...
-		syncName:         syncName,
-		syncNamespace:    syncNamespace,
-		reconcileTimeout: reconcileTimeout,
-	}
-	klog.V(4).Infof("Root Supervisor %s is initialized and synced with the API server", syncName)
-	return a, nil
+// AnnotateStatusMode sets the status mode annotation on the ResourceGroup object
+// if it exists.
+// This is invoked once before the applier runs so that ResourceGroups which are
+// too large can have their status cleared by the resource-group-controller.
```
karlkfi, line 185: nit: "cleared by the resource-group-controller before trying to add any new resources from source."

pkg/applier/applier.go
```diff
@@ -150,77 +152,56 @@ var _ Supervisor = &supervisor{}
...
+			return err
+		}
+		annotations := u.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[metadata.StatusModeKey] = s.clientSet.StatusMode
+		u.SetAnnotations(annotations)
+		klog.Infof("annotate the ResourceGroup object with the status mode %s", s.clientSet.StatusMode)
+		return s.clientSet.Client.Update(ctx, u, client.FieldOwner(configsync.FieldManager))
```
karlkfi, line 203: Does updating the annotations bump the metadata.generation?

If so, we might need to update the status to bump the observedGeneration, otherwise the RGC will be paused until the applier runs.

If we're updating the status here, we might as well also remove it, if the status is disabled, rather than waiting for the RGC to do it. It would save 1 API call.

pkg/applier/applier_test.go
```diff
@@ -374,8 +376,7 @@ func TestApply(t *testing.T) {
 				Mapper:     fakeClient.RESTMapper(),
 				// TODO: Add tests to cover status mode
 			}
-			applier, err := NewNamespaceSupervisor(cs, syncScope, syncName, 5*time.Minute)
-			require.NoError(t, err)
+			applier := NewSupervisor(cs, syncScope, syncName, 5*time.Minute)
```
karlkfi, line 379: do we want to call AnnotateStatusMode in these tests?

karlkfi: If not, we need a test for that method explicitly

sdowell: Added a separate unit test

PR #1620, opened 2025-03-27
refactor: return bool from metadata methods

karlkfi:
Return true when the labels or annotations were changed, to allow skipping update calls when there's no change.

### DIFF
--- a/pkg/core/decorate.go
+++ b/pkg/core/decorate.go
@@ -18,26 +18,32 @@ package core
 // some non-objects (such as PodTemplates) define annotations but are not objects.
 type Annotated interface {
 	GetAnnotations() map[string]string
-	SetAnnotations(annotations map[string]string)
+	SetAnnotations(map[string]string)
 }
 
 // SetAnnotation sets the annotation on the passed annotated object to value.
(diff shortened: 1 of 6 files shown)


pkg/core/decorate.go
```diff
@@ -18,26 +18,32 @@ package core
...
+	SetAnnotations(map[string]string)
 }
 
 // SetAnnotation sets the annotation on the passed annotated object to value.
-func SetAnnotation(obj Annotated, annotation, value string) {
-	as := obj.GetAnnotations()
-	if as == nil {
-		as = make(map[string]string)
+// Returns true if the object was modified.
+func SetAnnotation(obj Annotated, annotation, value string) bool {
```
sdowell, line 26: Can you add unit tests for these functions?

sdowell:
nit: this seems more like a `refactor` than a `chore`

pkg/core/decorate.go
karlkfi: Done.

sdowell (approved):
/lgtm

pkg/core/decorate.go
sdowell: Thanks! looks good

PR #1621, opened 2025-03-27
chore: remove owner refs dead code

karlkfi:


pkg/reconcilermanager/controllers/reconciler_base.go
```diff
@@ -125,7 +125,6 @@ func (r *reconcilerBase) upsertServiceAccount(
 	auth configsync.AuthType,
 	email string,
 	labelMap map[string]string,
-	refs ...metav1.OwnerReference,
```
tiffanny29631, line 128: Should the callers be modified as well as an arg is removed?

karlkfi: they weren't passing any, which is why the code still compiles.

tiffanny29631: Ack

tiffanny29631 (approved):
/lgtm

PR #1622, opened 2025-03-27
refactor: cleanup UpdateStatusMode

karlkfi:
- Return status error to get added context for certain errors
- Use core.SetAnnotation to simplify code
- Log at level 3 like other updates

Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1620

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -187,25 +187,23 @@ func NewSupervisor(cs *ClientSet, scope declared.Scope, syncName string, reconci
 func (s *supervisor) UpdateStatusMode(ctx context.Context) error {
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		u := newInventoryUnstructured(s.syncName, s.syncNamespace)
-		err := s.clientSet.Client.Get(ctx, client.ObjectKey{Name: u.GetName(), Namespace: u.GetNamespace()}, u)
+		key := client.ObjectKeyFromObject(u)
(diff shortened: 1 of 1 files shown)


pkg/applier/applier.go
```diff
@@ -187,25 +187,23 @@ func NewSupervisor(cs *ClientSet, scope declared.Scope, syncName string, reconci
...
-		err := s.clientSet.Client.Get(ctx, client.ObjectKey{Name: u.GetName(), Namespace: u.GetNamespace()}, u)
+		key := client.ObjectKeyFromObject(u)
+		err := s.clientSet.Client.Get(ctx, key, u)
 		if err != nil {
 			// RG doesn't exist, it will be created by applier with appropriate status mode
 			if apierrors.IsNotFound(err) {
 				return nil
 			}
-			return err
+			return status.APIServerErrorf(err, "failed to get %s: %s", u.GetKind(), key)
```
sdowell, line 197: What is the motivation for wrapping this in `APIServerErrorf`?

sdowell (commented):
nit: this is more of a `refactor` than a `chore`

pkg/applier/applier.go
karlkfi: We do it almost everywhere else that uses the k8s api. It adds some special messaging for permission errors, and a KNV error code to help with SEO finding the known issues page.

sdowell (approved):
/lgtm

PR #1625, opened 2025-04-01
e2e: fix OOMKill when testing with --debug

karlkfi:
- Limit diffing to object yaml that is at most 256 KB. This should help keep diff memory use under 2 GB.

### DIFF
--- a/pkg/util/log/stringer.go
+++ b/pkg/util/log/stringer.go
@@ -15,6 +15,7 @@
 package log
 
 import (
+	"encoding/binary"
 	"fmt"
 
 	"github.com/kylelemons/godebug/diff"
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -65,6 +65,8 @@ require (
 // When updating cloud.google.com/go/auth, ensure the auth.IsCredentialsNotFoundError still works.
 require cloud.google.com/go/auth v0.12.0
 
+require github.com/jpillora/sizestr v1.0.0
```
sdowell, line 68: did you intend to put this line on its own?

also nit: would be nice to not require this import since it's only used for a few unit test logs

karlkfi: This is where `go get github.com/jpillora/sizestr` put it. I think the tooling gets confused by the `require cloud.google.com/go/auth` above it.

I've replaced the dependency with some slightly more hacky local functions. But if we need to use this elsewhere, I'd prefer to use a well tested library.

sdowell (approved):
/lgtm

go.mod
sdowell: LGTM but still seems a bit overkill for unit test logging. I think a simple modulus and printing the approximate size in KB would be sufficient

karlkfi: Shoulda said that before i went and implemented it :P

PR #1626, opened 2025-04-02
feat: unmanage orphaned objects during RSync deletion

sdowell:
Previously when a RSync was deleted with the Orphan deletion propagation policy (default), all managed objects were left unchanged. This includes all of the Config Sync specific metadata injected into managed objects.

Some of this lingering metadata can prevent objects from being adopted by another RepoSync, since RepoSyncs use PolicyAdoptIfNoInventory. If the admission-webhook is enabled, the lingering metadata prevents any client from updating/deleting the object and prevents and RSync from adopting the object.

This change ensures that orphaned objects are treated as unmanaged by Config Sync so that they can be updated/deleted or adopted by a new RSync.


e2e/nomostest/nt.go
```diff
@@ -206,6 +207,10 @@ type NT struct {
 
 	// testStartTime is a timestamp which records when the current test started
 	testStartTime time.Time
+
+	// rsyncDeletionPropagation is the deletion propagation policy to use for all
+	// rsyncs. Defaults to Foreground to clean up test resources.
+	rsyncDeletionPropagation metadata.DeletionPropagationPolicy
```
karlkfi, line 213: nit: rename rsyncDeletionPropagation -> DeletionPropagationPolicy

e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -55,6 +56,9 @@ type MultiRepo struct {
 
 	// RepoSyncPermissions will grant a list of PolicyRules to NS reconcilers
 	RepoSyncPermissions []rbacv1.PolicyRule
+
+	// RSyncDeletionPropagation sets the deletion propagation policy for all RSyncs
+	RSyncDeletionPropagation metadata.DeletionPropagationPolicy
```
karlkfi, line 61: nit: rename rsyncDeletionPropagation -> DeletionPropagationPolicy

e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -116,16 +117,14 @@ func TestReconcilerFinalizer_Orphan(t *testing.T) {
 		// Namespace1 should NOT have been deleted, because it was orphaned by the RootSync.
 		return nt.Watcher.WatchObject(kinds.Namespace(), namespace1.GetName(), namespace1.GetNamespace(),
 			testwatcher.WatchPredicates(
-				testpredicates.HasAllNomosMetadata(), // metadata NOT removed when orphaned
-				testpredicates.HasLabel(metadata.ApplySetPartOfLabel, applySetID),
+				testpredicates.IsNotManaged(nt.Scheme),
```
karlkfi, line 120: Does this validate just the managed annotation or all the removed metadata?

e2e/testcases/sync_ordering_test.go
```diff
@@ -662,6 +666,114 @@ func TestDependencyWithReconciliation(t *testing.T) {
...
+	nt.Must(verifyUnmanaged(nt, ns1()))
+	nt.Must(verifyUnmanaged(nt, anvilCRD()))
+	nt.Must(verifyUnmanaged(nt, anvilCR()))
+	nt.Must(verifyUnmanaged(nt, cm1()))
+	nt.Must(verifyUnmanaged(nt, sa1()))
+
+	nt.T.Cleanup(func() {
+		// Remove objects from the original RSync for cleanup
+		repo := nt.SyncSourceGitReadWriteRepository(rootSync)
+		nt.Must(repo.Remove("acme/ns1.yaml"))
```
karlkfi, line 747: Might be nice to have an nt.Should that calls nt.T.Error, instead of nt.T.Fatal. Otherwise the first that fails will fail to clean the rest up.

But that would need a different PR.

e2e/testcases/sync_ordering_test.go
```diff
@@ -662,6 +666,114 @@ func TestDependencyWithReconciliation(t *testing.T) {
...
+	nt.Must(repo.Add("acme/anvil_cr.yaml", anvilCR()))
+	nt.Must(repo.CommitAndPush("Take ownership of anvil in root-sync-2"))
+	repo = nt.SyncSourceGitReadWriteRepository(repoSync1)
+	nt.Must(repo.Add("acme/cm1.yaml", cm1()))
+	nt.Must(repo.CommitAndPush("Take ownership of cm1 in repo-sync-1"))
+	repo = nt.SyncSourceGitReadWriteRepository(repoSync2)
+	nt.Must(repo.Add("acme/sa1.yaml", sa1()))
+	nt.Must(repo.CommitAndPush("Take ownership of sa1 in repo-sync-2"))
+	// Ready check recreates the root-sync, which we don't want
+	nt.Must(nt.WatchForAllSyncs(nomostest.SkipReadyCheck(), nomostest.SkipRootRepos(rootSync.Name)))
```
karlkfi, line 769: Will this do anything? Doesn't SkipRootRepos skip the RootSync check?

pkg/reconciler/finalizer/base_finalizer.go
```diff
@@ -56,3 +67,48 @@ func (bf *baseFinalizer) destroy(ctx context.Context) status.MultiError {
...
 	return nil
 }
+
+func (bf *baseFinalizer) unmanageObjects(ctx context.Context, rsyncRef client.ObjectKey) status.MultiError {
+	var errs status.MultiError
+	rg := &v1alpha1.ResourceGroup{}
+	if err := bf.Client.Get(ctx, rsyncRef, rg); err != nil {
+		if apierrors.IsNotFound(err) {
+			// ResourceGroup doesn't exist. This could be for a variety of reasons, but
+			// there's no work for the finalizer to do.
```
karlkfi, line 77: Not entirely sure if this should succeed silently or not. 
Seems like we should at least log something.

I guess one case is that it hasn't successfully applied yet, so the RG hasn't been created, in which case it's fine to exit.

Are there other use cases where it might not?

karlkfi: Does the reconciler-manager delete the RG after the RSync is NotFound or just after it's got a deletionTimestamp?

pkg/reconciler/finalizer/base_finalizer.go
```diff
@@ -56,3 +67,48 @@ func (bf *baseFinalizer) destroy(ctx context.Context) status.MultiError {
...
+		}
+		if err := bf.Client.Get(ctx, key, obj); err != nil {
+			if apierrors.IsNotFound(err) { // skip orphaning objects that don't exist
+				continue
+			}
+			errs = status.Append(errs, status.APIServerError(err, "failed to get object", obj))
+			continue
+		}
+		existing := obj.DeepCopy()
+		if core.RemoveAnnotations(obj, removeAnnotations...) {
```
karlkfi, line 106: What about labels?

We have something like this in the Supervisor code already i think, for abandoning individual objects. Might want to make the code similar.

pkg/reconciler/finalizer/controller.go
```diff
@@ -150,14 +150,10 @@ func (c *Controller) reconcileFinalizer(ctx context.Context, obj client.Object)
...
+	case metadata.DeletionPropagationPolicyForeground, metadata.DeletionPropagationPolicyOrphan:
 		if _, err := c.Finalizer.AddFinalizer(ctx, obj); err != nil {
 			return err
 		}
-	case metadata.DeletionPropagationPolicyOrphan:
-		if _, err := c.Finalizer.RemoveFinalizer(ctx, obj); err != nil {
-			return err
-		}
 	default:
 		klog.Warningf("%T %s has an invalid value for the annotation %q: %q",
```
karlkfi, line 158: weird that there's a warning here. I guess we dont have any CRD field validation for it... Should we add real validation in a separate PR?

pkg/reconciler/finalizer/reposync_finalizer.go
```diff
@@ -72,10 +70,17 @@ func (f *RepoSyncFinalizer) Finalize(ctx context.Context, syncObj client.Object)
 		return fmt.Errorf("setting Finalizing condition: %w", err)
 	}
 
-	if err := f.deleteManagedObjects(ctx, rs); err != nil {
-		return fmt.Errorf("deleting managed objects: %w", err)
+	deletionPropagation, ok := rs.Annotations[metadata.DeletionPropagationPolicyAnnotationKey]
+	if ok && deletionPropagation == string(metadata.DeletionPropagationPolicyForeground) {
```
karlkfi, line 74: is there a String() method on the enum? if not, we could add one

karlkfi, line 74: This If could be a switch.

pkg/reconciler/finalizer/reposync_finalizer_test.go
```diff
@@ -271,11 +279,176 @@ func TestRepoSyncFinalize(t *testing.T) {
...
+							},
+						},
+					},
+				}
+				return obj
+			}(),
+			expectedRsyncBeforeDestroy: func() client.Object {
+				obj := repoSync1.DeepCopy()
+				// No changes - continue deleting
+				// TODO: Should finalizer re-entry be vissible to the user by toggling a condition?
```
karlkfi, line 351: not sure what this is asking

pkg/syncer/syncertest/fake/storage.go
```diff
@@ -304,6 +304,21 @@ func (ms *MemoryStorage) listObjects(gk schema.GroupKind) []*unstructured.Unstru
 	return result
 }
 
+func lookupGVK(obj client.Object, scheme *runtime.Scheme) (schema.GroupVersionKind, error) {
+	switch obj.(type) {
+	case *metav1.PartialObjectMetadata:
```
karlkfi, line 309: wanna add this case to kinds.Lookup?

e2e/testcases/sync_ordering_test.go
sdowell: It skips based on the list of names that are passed in. It still waits for the other RootSyncs and RepoSyncs to be synced

pkg/reconciler/finalizer/base_finalizer.go
sdowell: > I guess one case is that it hasn't successfully applied yet, so the RG hasn't been created, in which case it's fine to exit. Are there other use cases where it might not?

The only happy path case I could think of is if the applier hasn't successfully run yet. It otherwise would be permanently wedged if some other client deleted the ResourceGroup, although nobody else _should_ delete the ResourceGroup

> Does the reconciler-manager delete the RG after the RSync is NotFound or just after it's got a deletionTimestamp?

It waits for the `reconciler` finalizer to complete first before executing the `reconciler-manager` finalizer.

pkg/reconciler/finalizer/rootsync_finalizer_test.go
```diff
@@ -278,11 +288,177 @@ func TestRootSyncFinalize(t *testing.T) {
...
+				return obj
+			}(),
+			expectedError:   nil,
+			expectedStopped: true,
+			expectedRsyncAfterFinalize: func() client.Object {
+				obj := rootSync1.DeepCopy()
+				// +1 to set ReconcilerFinalizing condition
+				// +1 to remove ReconcilerFinalizing condition
+				// +1 to remove Finalizer
+				// TODO: optimize by combining consecutive updates
```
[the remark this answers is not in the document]
sdowell: yes

pkg/reconciler/finalizer/reposync_finalizer_test.go
sdowell: it's just copying existing boilerplate. I didn't really follow the TODO but I can remove if you prefer

e2e/testcases/reconciler_finalizer_test.go
sdowell: It [validates the managed annotation](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/nomostest/testpredicates/predicates.go#L674)

pkg/reconciler/finalizer/base_finalizer.go
sdowell: The change only removes the management annotations which block the resources from being adopted by another client/RSync (by either the admission-webhook or in the applier logic).

It was intentional to leave the other Config Sync metadata (annotations/labels), but I'm not opposed to removing all of the metadata if you feel strongly

karlkfi: I think removing all the CS metadata would be best, unless there's some use case that needs it.

pkg/reconciler/finalizer/base_finalizer.go
karlkfi: How about we check the RSync `status.lastSyncedCommit`. If it's set, then log an error. If it's not set, then log at info level. 

This might reduce visibility on some cases where the apply started, created the RG, and didn't succeed, but it should increase visibility on most other problematic cases where it's synced at least once before.

pkg/reconciler/finalizer/reposync_finalizer_test.go
karlkfi: Hmm. I guess the question was whether the condition should flip back and forth when attempting and retrying. Based on more recent work in the RGC, I think it's undesirable for conditions to flap back and forth. So the answer is probably NO.

So it should be safe to remove these TODOs.

pkg/syncer/syncertest/fake/storage.go
karlkfi: Should be fine. But you might want to do it in a different PR and let the tests run.

e2e/testcases/reconciler_finalizer_test.go
karlkfi: Looks like both the manager and managed annotations.

If we're gonna remove all the CS meta, we should also validate all of it was removed.

e2e/nomostest/ntopts/multi_repo.go
```diff
@@ -55,6 +56,9 @@ type MultiRepo struct {
 
 	// RepoSyncPermissions will grant a list of PolicyRules to NS reconcilers
 	RepoSyncPermissions []rbacv1.PolicyRule
+
+	// DeletionPropagationPolicy sets the deletion propagation policy for all RSyncs
+	DeletionPropagationPolicy metadata.DeletionPropagationPolicy
```
karlkfi, line 61: Do we want these to be a pointer so you can set it to nil to omit the annotation?

pkg/reconciler/finalizer/base_finalizer.go
```diff
@@ -56,3 +69,60 @@ func (bf *baseFinalizer) destroy(ctx context.Context) status.MultiError {
...
+				client.FieldOwner(configsync.FieldManager)); err != nil {
+				errs = status.Append(errs, status.APIServerError(err, "failed to patch object to remove metadata", obj))
+			}
+		}
+	}
+	numErrs := 0
+	if errs != nil {
+		numErrs = len(errs.Errors())
+	}
+	klog.Infof("Finalizer unmanaged %d objects (%d not found) (%d errors)",
```
karlkfi, line 125: Since this might be re-run multiple times, we might want to show progress (total, successfully unmanaged, not found, and errored), like we do in the Applier with SyncStats. There's even a `stats.DisabledObjStats` you might be able to re-use if you add a field for notfound. Then you can log it like: `klog.Infof("Finalizer made new progress: %s", unmanageStats)`. Might be good to populate an ObjectStatusMap too, so it's even more similar, and we can see exactly which objects fall into each category in the logs, for debugging. https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/pkg/parse/updater.go#L194-L198

e2e/nomostest/ntopts/multi_repo.go
sdowell: done

e2e/nomostest/new.go
```diff
@@ -68,6 +70,8 @@ func newOptStruct(testName, tmpDir string, ntOptions ...ntopts.Opt) *ntopts.New
 			// Default to 1m to keep tests fast.
 			// To override, use WithReconcileTimeout.
 			ReconcileTimeout: ptr.To(1 * time.Minute),
+			// Default to foreground to clean up test resources
+			DeletionPropagationPolicy: &foregroundDeletion,
```
karlkfi, line 74: `ptr.To` should make the indirection easier

pkg/kinds/lookup.go
```diff
@@ -27,11 +28,22 @@ import (
 // Lookup returns the GVK of a object based on the types registered with the
 // provided Scheme.
 func Lookup(obj runtime.Object, scheme *runtime.Scheme) (schema.GroupVersionKind, error) {
-	gvk, err := apiutil.GVKForObject(obj, scheme)
-	if err != nil {
-		return schema.GroupVersionKind{}, fmt.Errorf("failed to lookup object type: %w", err)
+	switch obj.(type) {
+	case *metav1.PartialObjectMetadata:
```
karlkfi, line 32: Maybe I misread this earlier...
In what case is version not required?

pkg/reconciler/finalizer/reposync_finalizer_test.go
```diff
@@ -271,11 +278,181 @@ func TestRepoSyncFinalize(t *testing.T) {
...
+					"a-key":                              "a-value",
+				})
+				wantCM.SetLabels(map[string]string{
+					metadata.ManagedByKey:        metadata.ManagedByValue,
+					metadata.ApplySetPartOfLabel: applySetID,
+					"l-key":                      "l-value",
+				})
+			} else { // When "Orphan" is used, all CS metadata should be removed from the CM
+				wantCM.ResourceVersion = "2"
+				wantCM.SetAnnotations(map[string]string{
```
karlkfi, line 422: Looks like these can be pulled out and put up above the if statement. Then you can call SetConfigSyncMetadata on the wantCM if the policy is foreground.

pkg/reconciler/finalizer/reposync_finalizer_test.go
```diff
@@ -290,16 +467,20 @@ func TestRepoSyncFinalize(t *testing.T) {
 				rsync := &v1beta1.RepoSync{}
 				err := fakeClient.Get(context.Background(), key, rsync)
 				require.NoError(t, err)
+				core.SetAnnotation(tc.expectedRsyncBeforeDestroy,
+					metadata.DeletionPropagationPolicyAnnotationKey,
+					string(tc.deletionPolicy))
```
karlkfi, line 472: nit: `.String()`

pkg/reconciler/finalizer/reposync_finalizer_test.go
```diff
@@ -313,8 +494,11 @@ func TestRepoSyncFinalize(t *testing.T) {
 			testerrors.AssertEqual(t, tc.expectedError, err)
 
 			assert.Equal(t, tc.expectedStopped, stopped)
-			var expectedObjs []client.Object
+			expectedObjs := []client.Object{rg, wantCM}
 			if tc.expectedRsyncAfterFinalize != nil {
+				core.SetAnnotation(tc.expectedRsyncAfterFinalize,
+					metadata.DeletionPropagationPolicyAnnotationKey,
+					string(tc.deletionPolicy))
```
karlkfi, line 501: nit: `.String()`

pkg/metadata/metadata.go
```diff
@@ -111,6 +111,32 @@ func HasConfigSyncMetadata(obj client.Object) bool {
 	return false
 }
 
+// ConfigSyncMetadata contains fields needed to set all Config Sync metadata on
+// a managed resource.
+type ConfigSyncMetadata struct {
+	ApplySetID      string
```
karlkfi, line 117: linter was happy with these public fields not having descriptions?

pkg/reconciler/finalizer/base_finalizer.go
karlkfi: I ok with merging this as-is and having you rewrite the stats logging for both branches in a followup PR, if you really want to get this in for 1.21.0.

I'm also ok with merging it a day late, but we'll need Janet's approval.

e2e/nomostest/new.go
sdowell: done

pkg/kinds/lookup.go
sdowell: discussed offline, moved workaround back to fake client

pkg/metadata/metadata.go
sdowell: Evidently so. Added comments

pkg/reconciler/finalizer/controller.go
sdowell: Can you add CRD validation for annotations? Maybe using [CEL](https://kubernetes.io/docs/reference/using-api/cel/)? Either way we may not be able to add validation for backwards compatibility reasons.

pkg/reconciler/finalizer/reposync_finalizer_test.go
sdowell: It was intentional to not use `SetConfigSyncMetadata` to compute the expected value. I want the assertion to be explicit/unambiguous and not rely on the business logic to test the business logic.

Relevant/recent [tott](https://goto.google.com/tott/741)

pkg/reconciler/finalizer/reposync_finalizer_test.go
sdowell: fixed here and in other places

pkg/reconciler/finalizer/controller.go
karlkfi: > Can you add CRD validation for annotations?

Not that I'm aware of. You get the annotations for free without specifying them in the CRD. I don't think you can add validation to unspecified fields.

We should validate them in the code, but it doesn't need to block in this PR.

karlkfi (approved):
Looks good. Just need @janetkuo to approve merging after feature freeze.

janetkuo (approved):
/lgtm

pkg/reconciler/finalizer/controller.go
karlkfi: b/409592878

pkg/reconciler/finalizer/rootsync_finalizer.go
```diff
@@ -72,10 +71,17 @@ func (f *RootSyncFinalizer) Finalize(ctx context.Context, syncObj client.Object)
...
 	}
 
-	if err := f.deleteManagedObjects(ctx, rs); err != nil {
-		return fmt.Errorf("deleting managed objects: %w", err)
+	switch core.GetAnnotation(rs, metadata.DeletionPropagationPolicyAnnotationKey) {
+	case metadata.DeletionPropagationPolicyForeground.String():
+		if err := f.deleteManagedObjects(ctx, rs); err != nil {
+			return fmt.Errorf("deleting managed objects: %w", err)
+		}
+		klog.Infof("Deletion of managed objects successful")
```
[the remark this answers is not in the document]
karlkfi: b/409594209

PR #1628, opened 2025-04-03
fix: omit resource statuses when status is disabled

sdowell:
Although UpdateStatus is not called and therefore the status is not updated, the Update call is larger due to the status being populated on the subresource. This causes the payload to be larger due to the status even when status is disabled.

This change fixes the issue by omitted the status subresource when status is disabled.


karlkfi (commented):
This seems like a good optimization, even if we also fix the problem in cli-utils.

Looks like we're missing some unit tests for these methods tho, otherwise I would have expected this to break them. Can you backfill them?

sdowell:
> This seems like a good optimization, even if we also fix the problem in cli-utils.
> 
> Looks like we're missing some unit tests for these methods tho, otherwise I would have expected this to break them. Can you backfill them?

Done

karlkfi (approved):
/lgtm

PR #1630, opened 2025-04-04
refactor: replace commonLabels with labels.pairs in kustomization files

mikebz:
This pull request involves several updates to `kustomization.yaml` files across different directories in the `e2e/testdata/hydration` folder. The main change is the replacement of `commonLabels` with `labels` that include a `pairs` key.

Why?

Gets rid of noise

```
# Warning: 'commonLabels' is deprecated. Please use 'labels' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

### DIFF
--- a/e2e/testdata/hydration/deprecated-GK/kustomization.yaml
+++ b/e2e/testdata/hydration/deprecated-GK/kustomization.yaml
@@ -18,5 +18,6 @@ helmCharts:
   version: 0.2.0
   releaseName: my-prometheus-operator
 
-commonLabels:
-  test-case: hydration
+labels:
+- pairs:
+    test-case: hydration
--- a/e2e/testdata/hydration/namespace-repo/base/kustomization.yaml
+++ b/e2e/testdata/hydration/namespace-repo/base/kustomization.yaml
@@ -17,5 +17,6 @@ resources:
 - role.yaml
 - networkpolicy.yaml
 
-commonLabels:
-  test-case: hydration
+labels:
+- pairs:
+    test-case: hydration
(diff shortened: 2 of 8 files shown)


mikebz:
this is ready for review now

e2e/testdata/hydration/deprecated-GK/kustomization.yaml
```diff
@@ -18,5 +18,6 @@ helmCharts:
   version: 0.2.0
   releaseName: my-prometheus-operator
 
-commonLabels:
-  test-case: hydration
+labels:
+- pairs:
```
janetkuo, line 22: Should we set `includeSelectors: true` too? That will be consistent with the original behavior of `commonLabels`.

mikebz: I compared the output of the commands to what we are expecting and this is what matches right now.  I think the selectors was a noop in this case.  In case we change the payload and test we can potentially add more instructions.

janetkuo (approved):
/lgtm

e2e/testdata/hydration/namespace-repo/base/kustomization.yaml
```diff
@@ -17,5 +17,6 @@ resources:
 - role.yaml
 - networkpolicy.yaml
 
-commonLabels:
-  test-case: hydration
\ No newline at end of file
+labels:
+- pairs:
+    test-case: hydration
```
karlkfi, line 22: unless it breaks something, all files should end with a training linebreak.

mikebz: added

karlkfi (approved):
/lgtm

PR #1631, opened 2025-04-06
feat: add completion message to main command execution

mikebz:
This pull request includes an old request to output a success message (http://b/182464654).  Added a call to `util.MustFprintf` to print "Done!" upon execution of all the commands given that there are no errors.

### DIFF
--- a/cmd/nomos/nomos.go
+++ b/cmd/nomos/nomos.go
@@ -26,6 +26,7 @@ import (
 	"kpt.dev/configsync/cmd/nomos/initialize"
 	"kpt.dev/configsync/cmd/nomos/migrate"
 	"kpt.dev/configsync/cmd/nomos/status"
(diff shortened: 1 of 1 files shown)


janetkuo (approved):
nit: We could also print more information such as repo path and that it initialized a hierarchical repo format. But given that this mode is no longer recommended, it's also fine as-is.

mikebz:
@janetkuo thank you!  Yes we could print more but now it just prints this output on all the successful commands.  In some situations that information isnt relevant.

mikebz:
I think it also needs lgtm

karlkfi:
Are there any commands where the STDOUT is piped into another process, like with yaml/json output, where printing done might break the output?

karlkfi (approved):
/lgtm

PR #1634, opened 2025-04-07
refactor: Add ManagementMode enum

karlkfi:


pkg/metadata/annotation_management_mode.go
```diff
@@ -0,0 +1,71 @@
...
+
+// IsManagementEnabled returns true if the object has the annotation
+// `configmanagement.gke.io/managed: enabled`.
+func IsManagementEnabled(obj client.Object) bool {
+	return core.GetAnnotation(obj, ManagementModeAnnotationKey) == ManagementEnabled.String()
+}
+
+// IsManagementDisabled returns true if the object has the annotation
+// `configmanagement.gke.io/managed: disabled`.
+func IsManagementDisabled(obj client.Object) bool {
```
sdowell, line 69: Should this default to true if the annotation is missing?

e2e/nomostest/clean.go
```diff
@@ -190,7 +189,7 @@ func uninstallUnmanagedPackagesOfType(nt *NT, gvk schema.GroupVersionKind) error
 	// Find the unmanaged packages
 	var rsObjs []*unstructured.Unstructured
 	for _, item := range rsObjList.Items {
-		if core.GetAnnotation(&item, metadata.ResourceManagementKey) != metadata.ResourceManagementEnabled {
+		if !metadata.IsManagementEnabled(&item) {
```
sdowell, line 192: Is the reason this uses `!metadata.IsManagementEnabled` rather than `metadata.IsManagementDisabled` to catch the default case where the annotation is unset?

pkg/applier/utils.go
```diff
@@ -41,7 +41,7 @@ func partitionObjs(objs []client.Object) ([]client.Object, []client.Object) {
 	var disabled []client.Object
 	for _, obj := range objs {
 		// i.e. managed by Config Sync
-		if obj.GetAnnotations()[metadata.ResourceManagementKey] == metadata.ResourceManagementDisabled {
+		if obj.GetAnnotations()[metadata.ManagementModeAnnotationKey] == metadata.ManagementDisabled.String() {
```
sdowell, line 44: use `metadata.IsManagementDisabled`?

pkg/syncer/differ/decorations.go
```diff
@@ -47,7 +30,7 @@ func ManagementDisabled(obj client.Object) bool {
 // managed by Config Sync, because the annotation may be copied from another resource
 // managed by Config Sync.
 func ManagedByConfigSync(obj client.Object) bool {
-	return obj != nil && ManagementEnabled(obj) && core.GetAnnotation(obj, metadata.ResourceIDKey) == core.GKNN(obj)
+	return obj != nil && metadata.IsManagementEnabled(obj) && core.GetAnnotation(obj, metadata.ResourceIDKey) == core.GKNN(obj)
 }
 
 // ManagementUnset returns true if the resource has no Nomos ResourceManagementKey.
```
sdowell, line 36: Should this function also be moved to the `metadata` package?

pkg/hydrate/clean.go
```diff
@@ -47,7 +47,7 @@ func clean(o ast.FileObject) {
 	// Remove all the annotations starting with configsync.gke.io or configmanagement.gke.io
 	// except for the configmanagement.gke.io/managed annotation.
 	for k := range annotations {
-		if metadata.HasConfigSyncPrefix(k) && k != metadata.ResourceManagementKey {
+		if metadata.HasConfigSyncPrefix(k) && k != metadata.ManagementModeAnnotationKey {
```
sdowell, line 50: It seems like this could probably be abstracted away to a helper function which checks whether the annotation is user configurable. My interpretation is that this should skip deleting any Config Sync annotations that are set by the user.

Even if today the only user-configurable CS annotation is `ManagementModeAnnotationKey`, it would be less of a future gotcha if we had this abstracted to a separate `metadata` function IMO

pkg/importer/analyzer/validation/nonhierarchical/management_annotation_validator.go
```diff
@@ -23,8 +23,8 @@ import (
 
 // ValidManagementAnnotation returns an Error if the user-specified management annotation is invalid.
 func ValidManagementAnnotation(o ast.FileObject) status.Error {
-	value, found := o.GetAnnotations()[metadata.ResourceManagementKey]
-	if found && (value != metadata.ResourceManagementDisabled) {
+	value, found := o.GetAnnotations()[metadata.ManagementModeAnnotationKey]
+	if found && (value != metadata.ManagementDisabled.String()) {
```
sdowell, line 27: Should this use `metadata.IsManagementEnabled`?

pkg/metadata/annotation_management_mode.go
```diff
@@ -0,0 +1,71 @@
...
+	// ManagementModeAnnotationKey indicating that Config Sync should manage
+	// content and lifecycle for the given resource object.
+	ManagementEnabled ManagementMode = "enabled"
+	// ManagementDisabled is the value corresponding to
+	// ManagementModeAnnotationKey indicating that Config Sync should not manage
+	// content and lifecycle for the given resource.
+	//
+	// By design, the `configmanagement.gke.io/managed: disabled` annotation
+	// should not be pushed to the cluster. Instead, we remove all the Config
+	// Sync metadata from the object on the cluster.
```
sdowell, line 47: Might be worthwhile adding a note that this value is set by the user on the object in the source repo

pkg/parse/annotations.go
```diff
@@ -50,9 +50,9 @@ func addAnnotationsAndLabels(objs []ast.FileObject, scope declared.Scope, syncNa
 		core.SetAnnotation(obj, metadata.ResourceIDKey, core.GKNN(obj))
 		core.SetAnnotation(obj, metadata.OwningInventoryKey, inventoryID)
 
-		value := core.GetAnnotation(obj, metadata.ResourceManagementKey)
-		if value != metadata.ResourceManagementDisabled {
-			core.SetAnnotation(obj, metadata.ResourceManagementKey, metadata.ResourceManagementEnabled)
+		value := core.GetAnnotation(obj, metadata.ManagementModeAnnotationKey)
+		if value != metadata.ManagementDisabled.String() {
```
sdowell, line 54: nit: use `!metadata.IsManagementDisabled`

the line below could also use `WithManagementMode`

pkg/metadata/annotation_management_mode.go
karlkfi: Depends on usage. I initially had a different GetManagementMode that made that assumption and found it confusing to use, so I removed it. It seems like most of the places that test for disabled were really checking for explicitly disabled. So I figured it was better to have these checks be explicit.

pkg/syncer/differ/namespace_test.go
```diff
@@ -39,9 +39,10 @@ func markForDeletion(nsConfig *v1.NamespaceConfig) *v1.NamespaceConfig {
 }
 
 var (
-	disableManaged    = syncertest.ManagementDisabled
-	managementInvalid = core.Annotation(metadata.ResourceManagementKey, "invalid")
-	managementEmpty   = core.Annotation(metadata.ResourceManagementKey, "")
+	enableManaged     = metadata.WithManagementMode(metadata.ManagementEnabled)
```
sdowell, line 42: nit: this might be a useful abstraction in the `metadata` package, i.e. instead of `metadata.WithManagementMode` we could have `metadata.WithManagementEnabled` and `metadata.WithManagementDisabled`

e2e/nomostest/clean.go
karlkfi: Might read better if the cases were flipped, but it's skipping managed RSyncs. So all the RSyncs on the cluster without management enabled must have been manually applied.

The annotation should never be set to any other value on the cluster. If the user sets it to disabled in the source, the annotation is removed from the previously applied object.

pkg/metadata/annotation_management_mode.go
sdowell: Ack, after seeing more of the code I think this makes sense

pkg/importer/analyzer/validation/nonhierarchical/management_annotation_validator.go
karlkfi: This one is special. The validation requires that IF the annotation is specified in the source, then the only valid value is `disabled`.

pkg/syncer/differ/decorations.go
karlkfi: This one is kinda special, and probably misnamed. I wasn't sure what to call it if I moved it.

It doesn't just check the managed annotation. It also checks the resource-id annotation, which confirms that it was created by Config Sync and not just propagated to another resource by another controller.

pkg/parse/annotations.go
karlkfi: WithManagementMode constructs a function, which makes it awkward to use in this case. So I left it as-is.

I suppose we could add a SetManagementMode...

pkg/importer/analyzer/validation/nonhierarchical/management_annotation_validator.go
karlkfi: Refactored to be more readable, but slightly less efficient.

pkg/applier/utils.go
karlkfi: Done

e2e/nomostest/clean.go
karlkfi: Swapped

pkg/hydrate/clean.go
karlkfi: Honestly, I don't know why this is here. Seems like invalid fields should cause an error, not silently strip them.

I don't really want to touch this in this PR.

pkg/metadata/annotation_management_mode.go
karlkfi: Done.

pkg/syncer/differ/namespace_test.go
karlkfi: I considered this, but there's also a `syncertest.ManagementEnabled` that does something different: adding the resource-id annotation and managed-by label. That seems overloaded for that name, but I wasn't sure what to rename it to. So I didn't want to make two methods with the same name in different packages doing subtly different things.

pkg/applier/utils.go
sdowell: Looks like this was unchanged

pkg/validate/raw/validate/management_annotation_test.go
```diff
@@ -1,64 +0,0 @@
...
-		{
-			name: "invalid management fails",
-			obj:  k8sobjects.Role(core.Annotation(metadata.ResourceManagementKey, "invalid")),
-			want: status.FakeError(nonhierarchical.IllegalManagementAnnotationErrorCode),
-		},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			err := nonhierarchical.ValidManagementAnnotation(tc.obj)
```
sdowell, line 58: It seems like this test should be verifying `ManagementAnnotation` in the adjacent file

pkg/syncer/differ/decorations.go
sdowell: I was referring to the `ManagementUnset` function, looks like you've moved it in the latest change

pkg/validate/raw/validate/management_annotation.go
```diff
@@ -15,17 +15,19 @@
...
 	"kpt.dev/configsync/pkg/status"
 )
 
 // ManagementAnnotation returns an Error if the user-specified management annotation is invalid.
+// The only valid value in the source is `disabled`.
 func ManagementAnnotation(obj ast.FileObject) status.Error {
-	value, found := obj.GetAnnotations()[metadata.ResourceManagementKey]
-	if found && (value != metadata.ResourceManagementDisabled) {
-		return nonhierarchical.IllegalManagementAnnotationError(obj, value)
+	if !metadata.IsManagementDisabled(obj) && !metadata.IsManagementUnspecified(obj) {
```
sdowell, line 28: Could this be simplified to `metadata.IsManagementEnabled`? Is the concern that we may add more enum values?

Not trying to be difficult but I find this a bit harder to read than before. Maybe a switch would make it easier to read?

karlkfi: IsManagementEnabled isn't the same thing. Disabled and Unspecified are allowed. Enabled is not allowed in the source and neither are any other values.

pkg/validate/raw/validate/management_annotation_test.go
karlkfi: reverted and refactored

pkg/validate/raw/validate/management_annotation.go
sdowell: Right, my suggestion was to replace:
`!metadata.IsManagementDisabled(obj) && !metadata.IsManagementUnspecified(obj)`
with 
`metadata.IsManagementEnabled(obj)`

But ack if we need to specifically handle the case of an invalid (non-enumerated) value

I think either using a switch or applying De Morgan's would help. It might just be my brain but I have a harder time reasoning about negated conjunctions.

e.g. maybe something like:
```golang
func ManagementAnnotation(obj ast.FileObject) status.Error {
  if metadata.IsManagementDisabled(obj) || metadata.IsManagementUnspecified(obj) {
    return nil
  }
  return nonhierarchical.IllegalManagementAnnotationError(obj,
    core.GetAnnotation(obj, metadata.ManagementModeAnnotationKey))
}
```

sdowell: I can follow up with a separate change to refactor this function.

sdowell (approved):
/lgtm

pkg/validate/raw/validate/management_annotation.go
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1636

PR #1635, opened 2025-04-08
chore: update base image versions for golang, gcloud, and docker

mikebz:
This pull request includes updates to the versions of several base images used in the `Makefile`. These changes ensure that the project uses the latest versions of these images for better security, performance, and compatibility.

Version updates:

* Updated `GOLANG_IMAGE_VERSION` to `1.24.2` for the golang container base image.
* Updated `GCLOUD_IMAGE_VERSION` to `517.0.0` for the gcloud install base image.
* Updated `DOCKER_CLI_IMAGE_VERSION` to `20.10.24` for the docker CLI install base image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,19 +38,20 @@ GO_DIR := $(OUTPUT_DIR)/go
 
 # Base image used for all golang containers
 # Uses trusted google-built golang image
-GOLANG_IMAGE_VERSION := 1.23.4
+# See go/cs-upgrade-go for updates
+GOLANG_IMAGE_VERSION := 1.24.2
 GOLANG_IMAGE := google-go.pkg.dev/golang:$(GOLANG_IMAGE_VERSION)
 # Base image used for debian containers
-# When updating you can use this command: 
-# gcloud container images list-tags gcr.io/gke-release/debian-base --filter="tags:bookworm*" 
+# When updating you can use this command:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -38,19 +38,19 @@ GO_DIR := $(OUTPUT_DIR)/go
 
 # Base image used for all golang containers
 # Uses trusted google-built golang image
-GOLANG_IMAGE_VERSION := 1.23.4
+GOLANG_IMAGE_VERSION := 1.24.2
```
sdowell, line 41: Are you planning to update the Go version to 1.24?

mikebz: let me see how much churn that introduces.

sdowell: Playbook is here: go/cs-upgrade-go (applicable if you're upgrading minor versions)

mikebz: thank you.  added the links for discoverability.  I think we can consider moving this playbook closer to the source of changes.

go.mod
```diff
@@ -1,6 +1,7 @@
 module kpt.dev/configsync
 
-go 1.23.0
+// see go/cs-upgrade-go for updates
+go 1.24.2
```
sdowell, line 4: This should not be changed in the same commit as the version bump for the image. See the playbook

Makefile
sdowell: Note it involves internal specific steps. I don't think it belongs in OSS.

PR #1636, opened 2025-04-08
refactor: ManagementAnnotation readability

sdowell:
This small change refactors the boolean logic in the ManagementAnnotation function to improve readability.


karlkfi (approved):
/lgtm

I don't really find this easier to read, and prefer to exit early for errors rather than success, but I can't find any compelling reason for why. So I'll just approve.

PR #1637, opened 2025-04-09
chore: update Debian base image version to bookworm-v1.0.4-gke.6

mikebz:
This pull request includes an update to the `Makefile` to use a newer version of the Debian base image.


sdowell (approved):
/lgtm

PR #1638, opened 2025-04-10
test: fix TestSplitRSyncsWithDeletion on CSR

sdowell:
Tests on CSR require workload identity bindings to be created for every RSync. The new TestSplitRSyncsWithDeletion test case used new RSync names that did not have workload identity bindings.

This fixes the issue by changing the RSync names to reuse already provisioned workload identity bindings.


Camila-B (approved):
/lgtm

PR #1647, opened 2025-04-22
fix: validate deletion propagation annotation

karlkfi:
b/409592878


pkg/validate/raw/validate/deletion_propagation_validator_test.go
```diff
@@ -0,0 +1,129 @@
...
+			obj: k8sobjects.RepoSyncObjectV1Beta1("test-ns", "test-name",
+				metadata.WithDeletionPropagationPolicy(metadata.DeletionPropagationPolicyOrphan)),
+			syncKind: configsync.RepoSyncKind,
+		},
+		{
+			name: "RootSync empty annotation fails",
+			obj: k8sobjects.RootSyncObjectV1Beta1("test-name",
+				metadata.WithDeletionPropagationPolicy("")),
+			syncKind: configsync.RootSyncKind,
+			wantErr: invalidSyncBuilder.
```
sdowell, line 74: This would be easier to read/review if the expected error is expressed as a literal instead of reusing the builder logic. Same applies to other assertions in this change set

See [go/tott/741](https://goto.google.com/tott/741)

karlkfi: done

sdowell: It looks like there's now another layer of indirection to using the business logic to compute the expected value. I meant to declare the expected value as a literal/constant

karlkfi: All the other RSync validation errors use the invalidSyncBuilder pattern. They do not specify constants or variables, possibly because many of them have non-constant inputs. 

The RSync validation methods also all take either the rsync object or the syncKind, which is a variable on the reconciler. Switching from variable interpolation to a switch on the kind or object type would add a default error case. I'd rather not add that into generic/reusable validation methods.

I also plan to refactor them all to use syncKind: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1649

karlkfi: If you want to do a bulk migration to use literals/constants for status errors that don't wrap resource objects, that seems reasonable, but I don't really want to do it in this PR.

sdowell: I wasn't suggesting to change the business logic, just to use a literal/constant in the test assertion. For example:

```golang
wantErr: v1beta1.ConfigSyncError{
  Code: ...,
  ErrorMessage: "some expected error message",
}
```

So that when a reader looks at the test they can understand the expected output in a straightforward way.

If the error builder doesn't make that easy then I suppose this is fine, but it makes the unit tests more ambiguous and in a sense it's using the business logic to test the business logic

karlkfi: You want the condition fields in the test NOT to use the other constants and constructors and just use string literals?

karlkfi: I guess you're not suggesting not to use constants.

I get the iidea that string literals might be more clear than the indirection of the builder pattern, but what about error constructors, like `NewError(inputs).Error()`? Is that good enough or do they need to be constants/variables?

sdowell: In this instance it seems like the constructor/builder is being used to compute the expected value of itself. Maybe as a compromise there could be another test case for the constructor to document and verify the inputs/outputs?

My understanding of the best practice is generally to use as little logic as possible to create test assertions.

I didn't expect this ask to be controversial - no worries you can just leave it as it is.

karlkfi: > In this instance it seems like the constructor/builder is being used to compute the expected value of itself.

The `TestDeletionPropagationAnnotation` test validates the validation method `DeletionPropagationAnnotation`, not the error constructor `NewDeletionPropagationAnnotationError`.

You're right that if we hard-coded the error string values in this test it would also cover the error constructor, but I don't know that that's especially valuable, given that the builders should be tested in the `status` package and there's not much value in validating a thin wrapper around `fmt.Sprintf`. We don't usually test constructors.

The benefit of not testing the error constructor is that we can change the error message without refactoring any tests.

The cost of not testing the error constructor is that it could be changed accidentally without knowing it. IMO, this kind of issue should be caught by a standard code review, where a human would need to approve the error text change anyway.

Personally, I dislike the whole `status.Error` interface and equality checks for multiple reasons and would love to simplify it, making it easier to use and test, but I don't really feel like dying on that hill today. So I opted for consistency with the existing tests instead.

If you feel passionate about "fixing" it, please do a general refactor to set the precedent.

sdowell (approved):
/lgtm

PR #1649, opened 2025-04-22
fix: remove RSync object from validation errors

karlkfi:
- RSync spec validation errors are reported in the Stalled condition on the RSync object, so they do not need to self-reference the RSync object.
- Remove RSync spec validation from source validation. This early validation happens in the reconciler today, when parsing source files, but it duplicates validation that happens in the reconciler-manager when reconciling RSyncs. Removing this validation from the reconciler means that invalid RSyncs will no longer fail to parse and apply when synced by another RSync. Instead, these RSyncs will be applied and then the reconciler-manager will fail to reconcile them, adding the normal Stalled condition with the appropriate errors. Removing this validation from the reconciler unblocks removing the RSync object from the error, otherwise we would need to have two code paths, one with the object (for the reconciler) and one without (for the reconciler manager). This way, the validation only happens in one place, rather than being duplicated by two different components.
- Move RSync validation to pkg/validate/rsync/validate
- Refactor the reconciler-manager rsync controllers to use RepoSyncSpec and RootSyncSpec to reduce code duplication.
- Add override validation to ensure non-negative resource quantifies
- Backfill some RootSyncSpec tests
- Modify the test constructors to set auth to none by default. This makes it clearer that redundant source specs don't cause errors.


pkg/validate/raw/raw.go
```diff
@@ -103,8 +101,6 @@ func Unstructured(objs *fileobjects.Raw) status.MultiError {
 		fileobjects.VisitAllRaw(validate.ManagementAnnotation),
 		fileobjects.VisitAllRaw(validate.IllegalCRD),
 		fileobjects.VisitAllRaw(validate.CRDName),
-		fileobjects.VisitAllRaw(validate.RootSync),
-		fileobjects.VisitAllRaw(validate.RepoSync),
```
janetkuo, line 107: So this validates RSyncs in an RSync? Removing the validations here would mean that the root RSync status won't show problems with invalid child RSyncs?

karlkfi: Updated the PR description to clarify.

> So this validates RSyncs in an RSync?

Yes, if the RSync is in the source.

> Removing the validations here would mean that the root RSync status won't show problems with invalid child RSyncs?

Yes, mostly. The RSync status will no longer show spec errors from child RSyncs UNLESS the RSync is synced in a non-final apply phase, then the RSync would fail to reconcile with the Stalled condition, which would surface as an error on the parent RSync `status.sync.errors` and `Syncing` condition.

If at some point we fix the problem of final apply phase reconcile errors not being exposed in the RSync status, then that would also make this behavior consistent as well.

This validation is happening in two places right now. IMO, the controller reconciling the object should handle it, as is the case with every other resource type. So this change removes the special early validation, in service of simplifying validation, removing unnecessary errors, and avoiding self-referencing RSync spec errors.

You're welcome to disagree, but if that's the case I'll have to go back and jump through some hoops to make it only include the RSync object reference on source errors from the reconciler, and not on spec validations from the reconciler-manager.

janetkuo: Thanks for the detailed explanation! It makes sense to consolidate the validation. I'm not sure why we had this check here in the first place. 

With your PR, if the child RSync config is invalid and therefore failed to be applied, the root RSync would be `Stalled`, right?

karlkfi: > With your PR, if the child RSync config is invalid and therefore failed to be applied, the root RSync would be Stalled, right?

There's two ways for an RSync to be invalid:
1. Fail to meet CRD schema requirements
2. Fail to meet programmatic validation requirements

If a child RSync fails from schema requirements, the API server will reject the apply, causing the parent RSync Syncing condition to show the error.

If a child RSync fails from validation requirements, the API server will accept the apply and the parent Syncing condition will likely succeed, but the reconciler-manager will add the Stalled condition to the child RSync, causing it not to reconcile. So it will show up as a reconcile error in the parent RSync's ResourceGroup status.

As of today, managed object reconcile errors do not show up in the RSync status, only source, render, and sync errors.

pkg/validate/raw/validate/source_spec_validator.go
```diff
@@ -46,48 +46,48 @@ func HelmValuesFileDataKeyOrDefault(key string) string {
 }
 
 // RepoSyncSpec validates the Repo Sync source specification for any obvious problems.
-func RepoSyncSpec(sourceType configsync.SourceType, git *v1beta1.Git, oci *v1beta1.Oci, helm *v1beta1.HelmRepoSync, rs client.Object) status.Error {
+func RepoSyncSpec(sourceType configsync.SourceType, git *v1beta1.Git, oci *v1beta1.Oci, helm *v1beta1.HelmRepoSync, syncKind string) status.Error {
```
janetkuo, line 49: Is this used elsewhere, besides repo_sync_validator.go?

karlkfi: Yeah, in the reconciler-manager's reposync controller. 

The reconciler-manager validates the RSyncs specs. That's why we don't need the reconciler to do it when it finds an RSync in a source repo.

janetkuo: Where is `RepoSyncSpec` called in `reposync_controller.go`? I didn't see it.

karlkfi: Good catch. Looks like `reposync_controller.go`  does call all the underlying functions, but not that parent function.

```

func (r *RepoSyncReconciler) validateSourceSpec(ctx context.Context, rs *v1beta1.RepoSync, reconcilerName string) error {
	switch rs.Spec.SourceType {
	case configsync.GitSource:
		return r.validateGitSpec(ctx, rs, reconcilerName)
	case configsync.OciSource:
		return r.validateOciSpec(ctx, rs)
	case configsync.HelmSource:
		return r.validateHelmSpec(ctx, rs)
	default:
		return validate.InvalidSourceType(r.syncGVK.Kind)
	}
}
```

Basically the same thing, but with a few extra checks.

RepoSyncSpec & RootSyncSpec have good test coverage tho. Not sure how best to break it up.

janetkuo: How about adding a comment saying that this function is no longer used for validation, and is only kept for its test coverage for the underlying validations. 

This change LGTM otherwise.

karlkfi: I did some refactoring to make the controllers use the tested validate methods instead. It also reduced some redundancy. And I added a basic non-negative check for resource quantities. It's not clear why they allow negatives upstream, but AFAICT they fail to reconcile when the container runtime tries to create the container.

janetkuo (approved):
/lgtm

PR #1650, opened 2025-04-28
test: fix TestOtelCollectorDeployment expectation

karlkfi:
Cloud Monitoring recently changed the error message to be lowercase (cl/745657726). Update our expectation to match.

### DIFF
--- a/e2e/testcases/otel_collector_test.go
+++ b/e2e/testcases/otel_collector_test.go
@@ -49,7 +49,7 @@ const (
 	DefaultMonitorKSA             = "default"
 	MonitorGSA                    = "e2e-test-metric-writer"
 	MetricExportErrorCaption      = "One or more TimeSeries could not be written"
-	UnrecognizedLabelErrorCaption = "Unrecognized metric labels"
+	UnrecognizedLabelErrorCaption = "unrecognized metric labels"
 	GCMMetricPrefix               = "custom.googleapis.com/opencensus/config_sync"
 )
 


sdowell (approved):
nit on the commit message: This failure started without a Config Sync commit change so I don't think it has to do with an otel-collector bump from us

Seems like it's probably a message from some other component outside of our control.

e2e/testcases/otel_collector_test.go
```diff
@@ -49,7 +49,7 @@ const (
 	DefaultMonitorKSA             = "default"
 	MonitorGSA                    = "e2e-test-metric-writer"
 	MetricExportErrorCaption      = "One or more TimeSeries could not be written"
```
tiffanny29631, line 51: Might need to change this line to lower case too

karlkfi: That one still uses capitals in the example log i posted.

karlkfi:
This was a recent Cloud Monitoring change. Updated the description to match.

e2e/testcases/otel_collector_test.go
tiffanny29631: Ack

tiffanny29631 (approved):
/lgtm

janetkuo (approved):
This change is fine, but `validateDeploymentLogHasFailure` should ignore cases. We could add a comment here saying that the check is fragile given that it checks the log messages.

PR #1651, opened 2025-04-29
test: add assertions to adoption test case

sdowell:
This change adds some additional assertions to verify the behavior of merging annotations/labels when adopting an already existing object with Config Sync. This is not a behavioral change but rather documents the current behavior.


janetkuo (approved):
/lgtm

PR #1652, opened 2025-05-01
feat: print applier stats in reconciler finalizer

sdowell:
This updates the reconciler finalizer to print more detailed stats on the objects it unmanages. This is accomplished by re-using the status logger of the applier for a consistent format.

### DIFF
--- a/pkg/reconciler/finalizer/base_finalizer.go
+++ b/pkg/reconciler/finalizer/base_finalizer.go
@@ -24,8 +24,12 @@ import (
 	"kpt.dev/configsync/pkg/api/configsync"
 	"kpt.dev/configsync/pkg/api/kpt.dev/v1alpha1"
 	"kpt.dev/configsync/pkg/applier"
+	"kpt.dev/configsync/pkg/applier/stats"
+	"kpt.dev/configsync/pkg/core"
 	"kpt.dev/configsync/pkg/metadata"
 	"kpt.dev/configsync/pkg/status"
+	"sigs.k8s.io/cli-utils/pkg/apis/actuation"
+	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


Camila-B (commented):
/lgtm

pkg/reconciler/finalizer/base_finalizer.go
```diff
@@ -86,43 +88,62 @@ func (bf *baseFinalizer) unmanageObjects(ctx context.Context, rsyncRef client.Ob
...
+		Group: objRef.Group,
+		Kind:  objRef.Kind,
+	})
+	key := client.ObjectKey{
+		Name:      objRef.Name,
+		Namespace: objRef.Namespace,
+	}
+	if err := bf.Client.Get(ctx, key, obj); err != nil {
+		if apierrors.IsNotFound(err) { // skip orphaning objects that don't exist
+			return actuation.ActuationSkipped, nil
```
karlkfi, line 119: I think when actuation is skipped in the applier/destroyer it returns an error that explains why.

I'm not entirely sure if that's useful or needed here, but it seems like something we might want to log, even if the reason is not useful for the stats.

pkg/reconciler/finalizer/base_finalizer.go
```diff
@@ -86,43 +88,62 @@ func (bf *baseFinalizer) unmanageObjects(ctx context.Context, rsyncRef client.Ob
...
+	klog.Info("Finalizer unmanaged objects:")
+	objStatusMap.Log(klog.V(0))
+	return errs
+}
+
+func (bf *baseFinalizer) unmanageObject(ctx context.Context, objRef v1alpha1.ObjMetadata) (actuation.ActuationStatus, error) {
+	obj := &metav1.PartialObjectMetadata{}
+	obj.SetGroupVersionKind(schema.GroupVersionKind{
+		Group: objRef.Group,
+		Kind:  objRef.Kind,
```
karlkfi, line 111: nit: add a comment that Version is not required by PartialObjectMetadata

pkg/reconciler/finalizer/base_finalizer.go
```diff
@@ -86,43 +88,62 @@ func (bf *baseFinalizer) unmanageObjects(ctx context.Context, rsyncRef client.Ob
...
-				numNotFound++
-				continue
-			}
-			errs = status.Append(errs, status.APIServerError(err, "failed to get object", obj))
-			continue
+		errs = status.Append(errs, err)
+	}
+
+	klog.Info("Finalizer unmanaged objects:")
+	objStatusMap.Log(klog.V(0))
```
karlkfi, line 103: The applier has both a map of stats as a summary and a SyncStatus object with more verbose details. Do we want both here?

```
if !syncStats.Empty() {
	klog.Infof("Applier made new progress: %s", syncStats.String())
	objStatusMap.Log(klog.V(0))
}
```

pkg/reconciler/finalizer/base_finalizer.go
sdowell: done

pkg/reconciler/finalizer/base_finalizer.go
sdowell: added logs

pkg/reconciler/finalizer/base_finalizer.go
karlkfi: I guess we don't really need to know the permutation of events that were received, since they're all "Unmanage". But it might be nice for logging consistency.

It also makes it a little easier to see when new progress is made and what it was, rather than a retry attempt that makes no progress.

pkg/reconciler/finalizer/base_finalizer.go
karlkfi: Infof?

pkg/reconciler/finalizer/base_finalizer.go
sdowell: It seems like the `syncStats` log is actually less verbose compared to objStatusMap.

Example:
```
I0501 21:15:35.502487       1 updater.go:196] Applier made new progress: ApplyEvents: 4 (Successful: 4), WaitEvents: 5 (Pending: 1, Successful: 4)
I0501 21:15:35.502526       1 status.go:68] Apply Actuations (Total: 4):\nSkipped (0),\nSucceeded (4): [/Namespace/safety-config-management-system-root-sync, /Namespace/test-ns, apps/namespaces/test-ns/Deployment/helloworld-1, rbac.authorization.k8s.io/ClusterRole/safety-config-management-system-root-sync],\nFailed (0)
I0501 21:15:35.502541       1 status.go:84] Apply Reconciles (Total: 4):\nSkipped (0),\nSucceeded (4): [/Namespace/safety-config-management-system-root-sync, /Namespace/test-ns, apps/namespaces/test-ns/Deployment/helloworld-1, rbac.authorization.k8s.io/ClusterRole/safety-config-management-system-root-sync],\nFailed (0),\nTimeout (0)
I0501 21:15:35.502553       1 status.go:98] Delete Actuations (Total: 0)
I0501 21:15:35.502564       1 status.go:114] Delete Reconciles (Total: 0)
```

I think given that the unmanager is only Applying objects, `syncStats` wouldn't add much additional/useful information

sdowell: Fair point on logging consistency, I noticed the destroyer logs syncstats as well with Foreground deletion propagation. I've added syncstats to the unmanager

pkg/reconciler/finalizer/base_finalizer.go
sdowell: fixed

karlkfi (approved):
/lgtm

PR #1655, opened 2025-05-02
chore: add Makefile rule to clean up unused images

Camila-B:
Follow Up to cl/753749202

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -125,7 +125,18 @@ make retag-images \
  REGISTRY=gcr.io/bat \
  IMAGE_TAG=bar 
 ```
+### Deleting old images
+Docker does not automatically clean up old, unused images. Our local e2e tests create several large images where a single run can take up hundreds of MB of disk space.
 
+In order to to prevent your hard drive from being filled up with these unused images, regularly run the following commands:
+```shell
+docker image prune -a   # Removes all images without at least one container associated to them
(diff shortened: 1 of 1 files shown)


Makefile.e2e
```diff
@@ -110,3 +110,9 @@ test-e2e-kind-test-group2:
 .PHONY: test-e2e-kind-test-group3
 test-e2e-kind-test-group3:
 	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=reconciliation-2,multi-repos,override-api,hydration" test-e2e-kind
+
+# This target cleans up old and unused images
+.PHONY: clean-images
+clean-images:
```
sdowell, line 116: nit: I don't really find this necessary - this is just general practice with docker development on a workstation. It's not really config sync specific

tiffanny29631: I suggested adding this yesterday to round out the local test workflow. It could also serve as a helpful reminder for developers to periodically clean up their environment. I agree that this is a standard practice and could end up removing images unrelated to Config Sync. So, rather than introducing a new make rule, perhaps we could just mention this in the comments for the test target instead?

Camila-B: I removed the make rule and I added a note about how to clean up old images to `development.md`

janetkuo (approved):
/lgtm

PR #1661, opened 2025-05-06
Remove buildx builder instance after image push

tiffanny29631:
Postsync status watch controller supports multi-arch builds by using a dedicated builder. This PR ensures the Docker builder is cleaned up afterward to prevent interference with other image builds.


sdowell (approved):
/lgtm

PR #1663, opened 2025-05-06
test: wait for GKE cluster readiness

sdowell:
The test scaffolding previously waited for cluster readiness by waiting for all operations to be complete, but that evidently is no longer sufficient to wait for cluster readiness. This change adds an additional check to make sure the cluster has the RUNNING status before handing off the cluster to the test framework.


tiffanny29631 (approved):
/lgtm

PR #1665, opened 2025-05-12
Refactor Otel Collector Googlecloud ConfigMap Generation

tiffanny29631:
To Use Go Struct and YAML Marshaling

Refactor the generation of the otel-collector Googlecloud ConfigMap to construct the configuration using a Go map and marshal it to YAML, replacing the previous approach of using a static YAML string. This change improves maintainability and reduces the risk of formatting errors in the generated config. Add GooglecloudCollectorConfigYAML() and deprecate the old CollectorConfigGooglecloud string constant. This refactor makes the config generation more robust and easier to extend or validate in the future.

### DIFF
--- a/pkg/metrics/otel.go
+++ b/pkg/metrics/otel.go
@@ -14,6 +14,10 @@
 
 package metrics
 
+import (
+	"sigs.k8s.io/yaml"
+)
+
 const (
 	// OpenTelemetry is the app label for all otel resources.
 	OpenTelemetry = "opentelemetry"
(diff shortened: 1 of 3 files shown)


pkg/reconcilermanager/controllers/otel_controller.go
```diff
@@ -140,8 +140,12 @@ func (r *OtelReconciler) configureGooglecloudConfigMap(ctx context.Context) ([]b
 			metadata.SystemLabel: "true",
 			metadata.ArchLabel:   "csmr",
 		}
+		configYAML, yamlErr := metrics.CollectorConfigGooglecloudYAML()
+		if yamlErr != nil {
+			return status.APIServerErrorf(yamlErr, "failed to marshal otel-collector-googlecloud config to YAML")
```
janetkuo, line 145: Is this an API server error?

tiffanny29631: Missed the message. Changed to regular error for marshaling.

pkg/metrics/otel.go
```diff
@@ -14,6 +14,10 @@
 
 package metrics
 
+import (
+	"gopkg.in/yaml.v3"
```
janetkuo, line 18: This package is archived, use `sigs.k8s.io/yaml` instead.

tiffanny29631: Fixed

pkg/metrics/otel.go
```diff
@@ -29,324 +33,209 @@ const (
...
-      processors: [batch]
-      exporters: [prometheus]
-    metrics/kubernetes:
-      receivers: [opencensus]
-      processors: [batch, filter/kubernetes, metricstransform/kubernetes, resourcedetection]
-      exporters: [googlecloud/kubernetes]`
 )
+
+// CollectorConfigGooglecloudYAML returns the otel-collector-googlecloud config as YAML.
+func CollectorConfigGooglecloudYAML() (string, error) {
```
sdowell, line 39: It looks like valid configs can be generated programmatically, e.g. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckextension/factory.go#L20

Would it be possible for us to use that to generate canonical/valid configs and then marshal to YAML?

sdowell, line 39: Can you add a unit test for this function which validates the output yaml string against a static/hardcoded test yaml string? That will make it easier for a human reviewer to inspect the actual output yaml

tiffanny29631: Tried it, and following this pattern will lead to adding many more dependencies, which could make the build larger and slower. It would also tightly couple us with the upstream design, complicating maintenance, especially since we have numerous exporters and processors. Given that the Config Sync ConfigMap is slightly customized and this change is a refactor, I prefer to keep it minimal while ensuring it remains relatively readable for future maintenance.

tiffanny29631: Okie just pushed two new commits to the repository so that we can revert if we don't like them.

The second commit in the chain specifically updates the otel-controller unit test. It eliminates the use of a hard-coded hash when comparing the desired ConfigMap with the one retrieved from the fake client, as well as when comparing the hashed ConfigMap annotation.
The third commit introduces a human-readable example ConfigMap that can be generated with the rule upon code changes. Including a hard-coded ConfigMap in the unit test would necessitate manual adjustments every time otel.go is modified, which could lead to human error.

sdowell: Ack, sgtm

sdowell: This seems a bit more overkill than I was expecting (with the make target and new gen-otel-config binary).

> Including a hard-coded ConfigMap in the unit test would necessitate manual adjustments every time otel.go is modified, which could lead to human error.

That's the idea of golden/approval/snapshot testing (minus the human error part). The idea is that if the Go map changes, the human-readable yaml must be updated in lockstep or the tests will fail. This guarantees a human-readable diff any time the ConfigMap changes.

I think a yaml string defined as a Go variable would be fine. If the unit test fails it can print out the "got" yaml and a developer can just copy/paste to the "expected" yaml string variable.

Alternatively the Go unit test could compare against a separate yaml file, but that requires a bit more scaffolding

tiffanny29631: Sure. I’ve added a static string for the config data. One of the main reasons for this refactor is to avoid maintaining a hardcoded config string, since even minor issues like incorrect indentation or whitespace can cause problems with marshaling. However, I agree that if we have to keep it somewhere, it’s preferable for it to be in test code rather than production.

pkg/reconcilermanager/controllers/otel_controller_test.go
```diff
@@ -167,7 +509,19 @@ func TestOtelReconcilerGooglecloud(t *testing.T) {
...
+	require.Equal(t, wantConfigMap.ObjectMeta.Labels, gotConfigMap.ObjectMeta.Labels, "ConfigMap labels")
+	require.Equal(t, wantConfigMap.ObjectMeta.Name, gotConfigMap.ObjectMeta.Name, "ConfigMap name")
+	require.Equal(t, wantConfigMap.ObjectMeta.Namespace, gotConfigMap.ObjectMeta.Namespace, "ConfigMap namespace")
+
+	// Compare the YAML content as objects, not strings
+	var wantObj, gotObj map[string]interface{}
+	require.NoError(t, yaml.Unmarshal([]byte(wantConfigMap.Data["otel-collector-config.yaml"]), &wantObj))
+	require.NoError(t, yaml.Unmarshal([]byte(gotConfigMap.Data["otel-collector-config.yaml"]), &gotObj))
+	if !reflect.DeepEqual(wantObj, gotObj) {
+		t.Errorf("ConfigMap otel-collector-config.yaml does not match:\n%s", cmp.Diff(wantObj, gotObj))
```
sdowell, line 523: What's the reason for comparing them as objects as opposed to strings? I expect the output is not as easy to read.

It would be nice if this output the actual yaml on failure to make it easier to update (copy/paste) the expected yaml string above.

pkg/reconcilermanager/controllers/otel_controller_test.go
```diff
@@ -167,7 +521,14 @@ func TestOtelReconcilerGooglecloud(t *testing.T) {
 	gotConfigMap := &corev1.ConfigMap{}
 	err := fakeClient.Get(ctx, cmKey, gotConfigMap)
 	require.NoError(t, err, "ConfigMap[%s] not found", cmKey)
-	asserter.Equal(t, wantConfigMap, gotConfigMap, "ConfigMap")
+
+	// Compare the YAML content as strings
+	gotYAML := gotConfigMapToYAML(gotConfigMap)
+	if configYAML != gotYAML {
```
sdowell, line 527: Suggest using `cmp.Diff` to compare strings and print a diff. That will be easier to view the difference compared to the full expected yaml. in other words:

- print the `cmp.Diff` output for easy human readable diff
- print the full actual yaml for easy copy/paste if the developer wants to update the expected yaml

tiffanny29631: Done

sdowell: a few minor nits:
- the typical pattern is `if diff := cmp.Diff(wantYAML, gotYAML); diff != "" { ...`
- use `fmt.Printf`, because `fmt.Println` prints an unintended space between args
- use `t.Fatalf` to fail immediately

tiffanny29631: Done

sdowell (approved):
/lgtm

PR #1671, opened 2025-05-15
chore: update Go version to 1.24.3 in Makefile and go.mod

mikebz:
updating go version while I am onduty.  CI and Louhi updates forthcoming.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ GO_DIR := $(OUTPUT_DIR)/go
 
 # Base image used for all golang containers
 # Uses trusted google-built golang image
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module kpt.dev/configsync
 
-go 1.23.0
+go 1.24.3
```
sdowell, line 3: I'd suggest updating just the image version initially, otherwise CI may break in the interim (until the CI environment is updated with the new Go version)

mikebz: reverted

sdowell (approved):
/lgtm

PR #1672, opened 2025-05-19
test: default to creating clusters for KinD only if -create-clusters=false

Camila-B:
This allows for `create-clusters` to be set to `lazy` for local development


sdowell (approved):
/lgtm

PR #1681, opened 2025-05-21
ci: skip kind updates in dependabot

sdowell:
kind update also requires updating the supported image tags


mikebz:
given that we have just a few people on the core team right now I suggest giving me the power to approve PRs (I will keep it to the small ones)

Camila-B (approved):

PR #1688, opened 2025-05-21
ci: install go-junit-report at runtime

sdowell:
Updates the go-junit-report binary to be installed at runtime. This
makes it so that it is updated in lockstep with the gomod file and so
that the gke-e2e image doesn't need to be bumped to consume a new
version.


Makefile
```diff
@@ -483,6 +485,17 @@ clean-helm:
 # install kind (user-friendly target alias)
 install-kind: "$(KIND)"
 
+.PHONY: clean-go-junit-report
+clean-go-junit-report:
+	@rm -rf $(GO_JUNIT_REPORT)
+
+.PHONY: install-go-junit-report
+# install go-junit-report (user-friendly target alias)
+install-go-junit-report: "$(COSIGN)"
```
tiffanny29631, line 494: GO_JUNIT_REPORT

sdowell: good catch, fixed

tiffanny29631 (approved):
/lgtm

PR #1693, opened 2025-05-22
test: fix hasTolerations segfault

sdowell:
If the number of actual tolerations was greater than the number of expected tolerations, this would cause a segfault.

Caused a panic/segfault on this execution: https://oss.gprow.dev/view/gs/oss-prow-build-kpt-config-sync/logs/kpt-config-sync-standard-rapid/1925287984836382720

### DIFF
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -754,6 +754,12 @@ func hasTolerations(tolerations []corev1.Toleration) testpredicates.Predicate {
 		if !ok {
 			return testpredicates.WrongTypeErr(d, &appsv1.Deployment{})
 		}
+		// two lists are equal if they have the same length and each element is equal.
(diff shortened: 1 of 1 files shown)


e2e/testcases/reconciler_manager_test.go
```diff
@@ -754,6 +754,10 @@ func hasTolerations(tolerations []corev1.Toleration) testpredicates.Predicate {
 		if !ok {
 			return testpredicates.WrongTypeErr(d, &appsv1.Deployment{})
 		}
+		if len(d.Spec.Template.Spec.Tolerations) != len(tolerations) {
```
mikebz, line 757: I would add some commentary as to why the length of tolerations is actually important.

sdowell: nit: I find it fairly self-evident, but added a comment

PR #1721, opened 2025-05-29
test: add e2e test for adopting implicit namespace

sdowell:
This change adds a testcase which documents/validates the behavior of adopting a Namespace by one RootSync after it has been created by another RootSync as an implicit Namespace.


mikebz:
/lgtm

even though just in general readability the creation and watch code requires mental load on the reader of translating those objects from Go into YAML.  The scope of this topic is well beyond this PR but something to think about.

PR #1723, opened 2025-05-29
test: use Watcher for validating Stalled errors

sdowell:
This refactors the Stalled condition validation to use the Watcher, similar to the validation for other conditions. This completes the refactoring of moving the condition validation from using the previous polling mechanism to using the watch mechanism. Streamlined error code validation for Stalled errors will be added in a subsequent change.


mikebz:
Also did a cursory look in the branch for remaining references

https://github.com/search?q=repo%3Asdowell%2Fkpt-config-sync%20%20WaitForRootSyncStalledError&type=code

It looks like they are all gone

PR #1724, opened 2025-05-30
[WIP] Design Doc: Support for kpt Apply-Time-Mutation (ATM) in the Remediator

vs-odie:
This PR introduces a design document proposing support for kpt Apply-Time-Mutation (ATM) in the Config Sync remediator. Currently, only the applier honors ATM annotations, which can lead to resources being reverted to their unmutated state by the remediator. The proposed change ensures that ATM substitutions are applied consistently during both initial apply and ongoing reconciliation, improving correctness and user experience. The design also references real-world needs for cross-resource value injection, as highlighted in k8s-config-connector [Issue #101](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/101)

### DIFF
--- a/docs/design-docs/03-apply-time-mutation-remediator.md
+++ b/docs/design-docs/03-apply-time-mutation-remediator.md
@@ -1,0 +1,74 @@
+# Support for kpt Apply-Time-Mutation (ATM) in the Remediator
+
+**Authors:**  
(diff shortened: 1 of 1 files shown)


mikebz:
thank you for contributing a well thought out proposal and following the design doc pattern! Adding to me review list.  Apply time mutations have been a subject of much debate over the years.

Issue #1716, opened 2025-05-28
New "Done!" message induces an error when sourcing completions


ciaccotaco:
### Checklist

- [x] I did not find a related open issue.
- [x] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/how-to/troubleshooting-config-sync)
- [x] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).
- [x] I have attached a [nomos bugreport](https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/how-to/nomos-command#bug) to this issue.

### Bug Description

[Feature: add completion message to main command execution](https://github.com/GoogleContainerTools/kpt-config-sync/pull/1631) introduced a `Done!` message at the end of the nomos output.

This introduced an error that was [predicted](https://github.com/GoogleContainerTools/kpt-config-sync/pull/1631#issuecomment-2784937608) by @karlkfi:

> Are there any commands where the STDOUT is piped into another process, like with yaml/json output, where printing done might break the output?

When sourcing `nomos completion`, the command throws the following error:

```bash
$ source <(nomos completion zsh)
/dev/fd/11:213: command not found: Done!
```

I'm curious as to why `Done!` was added.  I cannot view the internal google `http://b` URL.  From a customer's standpoint, this message seems a bit superfluous.

### Additional Diagnostic Information

N/A

### Kubernetes Cluster Type/Version

- k8s version: 1.32
- Type of k8s cluster (GKE, EKS, etc): GKE
- If GKE, standard or autopilot: standard

### Config Sync Version

1.21.0-rc.2

### Steps to reproduce the issue

Run `source <(nomos completion zsh)`

mikebz:
thanks for reporting this!  Do you happen to have a PR or are just looking to get a fix in the next release?

sdowell:
Fixed by https://github.com/GoogleContainerTools/kpt-config-sync/pull/1717

PR #1725, opened 2025-05-30
feat: Add --format flag to nomos status

Camila-B:
* This PR adds the `--format` flag to allow users to specify the output format of the `nomos status` command. It currently only supports the `json` format with `tabular` being the default format used.
* b/261152515

### DIFF
--- a/cmd/nomos/status/status.go
+++ b/cmd/nomos/status/status.go
@@ -16,6 +16,7 @@ package status
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -44,7 +45,8 @@ var (
 	pollingInterval time.Duration
 	namespace       string
 	resourceStatus  bool
-	name            string
+	syncName        string
+	format          string
 )
 
 func init() {
@@ -53,7 +55,8 @@ func init() {
 	Cmd.Flags().DurationVar(&pollingInterval, "poll", 0*time.Second, "Continuously polls for status updates at the specified interval. If not provided, the command runs only once. Example: --poll=30s for polling every 30 seconds")
 	Cmd.Flags().StringVar(&namespace, "namespace", "", "Filters the status output by the specified RootSync or RepoSync namespace. If not provided, displays status for all RootSync and RepoSync objects.")
 	Cmd.Flags().BoolVar(&resourceStatus, "resources", true, "Displays detailed status for individual resources managed by RootSync or RepoSync objects. Defaults to true.")
-	Cmd.Flags().StringVar(&name, "name", "", "Filters the status output by the specified RootSync or RepoSync name.")
+	Cmd.Flags().StringVar(&syncName, "name", "", "Filters the status output by the specified RootSync or RepoSync name.")
+	Cmd.Flags().StringVar(&format, "format", "", fmt.Sprintf("Output format. Accepts '%s'.", flags.OutputJSON))
 }
 
 // SaveToTempFile writes the `nomos status` output into a temporary file, and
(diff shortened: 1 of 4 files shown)


e2e/testcases/cli_test.go
```diff
@@ -1167,6 +1167,22 @@ func TestNomosStatus(t *testing.T) {
...
+
+	// get status
+	cmd := nt.Shell.Command("nomos", "status", "--format", "json")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		nt.T.Log(string(out))
+		nt.T.Fatal(err)
+	}
+
+	if !strings.Contains(string(out), `"status": "SYNCED"`) {
```
sdowell, line 1181: If the output is stable, can we compare this to a static string so that the expected output is obvious to the reader?

ditto below

sdowell:
> Question: how does the e2e tests get around the fact that its launching a built binary? I remember at times the nomos API surface changes and the tests were broken. Is it solved in this case?
---
> how does the e2e tests get around the fact that its launching a built binary?

Binaries (docker images and nomos cli) are built before the e2e tests run

> I remember at times the nomos API surface changes and the tests were broken. Is it solved in this case?

I'm not sure what you're referring to

mikebz:
> I'm not sure what you're referring to

https://github.com/GoogleContainerTools/kpt-config-sync/pull/816

this commit was because the test passed locally but failed on CI: https://github.com/GoogleContainerTools/kpt-config-sync/pull/816/commits/b06bb4e101a9c9bab72474ddaacf75c7c1c71569

sdowell:
> #816
> 
> this commit was because the test passed locally but failed on CI: [b06bb4e](https://github.com/GoogleContainerTools/kpt-config-sync/commit/b06bb4e101a9c9bab72474ddaacf75c7c1c71569)

The Go e2e tests themselves don't build the binaries (docker images or CLI), so when testing locally you need to make sure you've built the relevant artifacts first. The entrypoints used by the presubmits build all of the binaries before executing the tests. It's possible that there may have been a CI issue in the past but it's currently WAI

e2e/testcases/cli_test.go
Camila-B: Most of the output is stable. However, there are some fields that are not static ex. `commit` and `lastSyncTimestamp`. Should I still use static strings to compare the fields that don't change with each test run?

sdowell: My suggestion would be to unmarshal the output and diff the whole output against an expected `map[string]interface{}`. The fields which aren't stable can be ignored in the diff

Unmarshalling the output asserts that the output is valid JSON and can be ingested programmatically (catching issues like https://github.com/GoogleContainerTools/kpt-config-sync/issues/1716), and diffing against an expected object explicitly validates/documents the fields/structure

cmd/nomos/status/status.go
```diff
@@ -53,7 +55,8 @@ func init() {
 	Cmd.Flags().DurationVar(&pollingInterval, "poll", 0*time.Second, "Continuously polls for status updates at the specified interval. If not provided, the command runs only once. Example: --poll=30s for polling every 30 seconds")
 	Cmd.Flags().StringVar(&namespace, "namespace", "", "Filters the status output by the specified RootSync or RepoSync namespace. If not provided, displays status for all RootSync and RepoSync objects.")
 	Cmd.Flags().BoolVar(&resourceStatus, "resources", true, "Displays detailed status for individual resources managed by RootSync or RepoSync objects. Defaults to true.")
-	Cmd.Flags().StringVar(&name, "name", "", "Filters the status output by the specified RootSync or RepoSync name.")
+	Cmd.Flags().StringVar(&syncName, "name", "", "Filters the status output by the specified RootSync or RepoSync name.")
+	Cmd.Flags().StringVar(&format, "format", "", fmt.Sprintf("Output format. Accepts '%s'.", flags.OutputJSON))
```
sdowell, line 59: The default value should be a string enum rather than empty string (e.g. something like `table`)

Camila-B:
Closing this PR as the FR has been deemed obsolete

PR #1727, opened 2025-06-03
test: fix workload identity for TestAdoptImplicitNamespace

sdowell:
The previous reconciler KSA did not have workload identity provisioned. This updates the reconciler KSA name to one that already has workload identity provisioned.


e2e/testcases/lifecycle_directives_test.go
```diff
@@ -338,15 +338,15 @@ func TestPreventDeletionSpecialNamespaces(t *testing.T) {
 // one RootSync can be adopted by another RootSync.
 func TestAdoptImplicitNamespace(t *testing.T) {
 	rootSyncID := nomostest.DefaultRootSyncID
-	rootSync2ID := core.RootSyncID("root-sync-2")
```
Camila-B, line 341: Nit: you should probably add the test name to https://github.com/GoogleContainerTools/kpt-config-sync/blob/1d04b235981272d4ca2ce63358c8307cf47bd370/e2e/testinfra/terraform/modules/service_account/service_account.tf#L55

PR #1734, opened 2025-06-05
fix: handle IsNoMatchError for remediator startWatch

sdowell:
This error can occur when the RESTMapper fails to look up the object type, and has been observed to happen semi-frequently for the e2e TestCRDDeleteBeforeRemoveCustomResourceV1.

The e2e test is flaky because depending on the ordering of events, the conflict metric may not be recorded. This should fix the metric flakiness for this edge case.


tiffanny29631 (approved):
/lgtm

PR #1735, opened 2025-06-05
fix: update watches for triggerWatchUpdate

sdowell:
The triggerWatchUpdate event is invoked when one of the remediator watches failed and needs to be restarted. However, the cache.watchesUpdated flag was never reset, resulting in an infinite loop in which the reconciler is triggered to update watches but does not attempt to update the watches.

The cache.watchesUpdated flag is set to false for this trigger so that the updater attempts to update the remediator watches.


tiffanny29631 (approved):
/lgtm

PR #1736, opened 2025-06-06
[cherry-pick] fix: handle IsNoMatchError for remediator startWatch (#1734)

tiffanny29631:
This error can occur when the RESTMapper fails to look up the object type, and has been observed to happen semi-frequently for the e2e TestCRDDeleteBeforeRemoveCustomResourceV1.

The e2e test is flaky because depending on the ordering of events, the conflict metric may not be recorded. This should fix the metric flakiness for this edge case.


sdowell (approved):
/lgtm

PR #1738, opened 2025-06-10
fix: handle Helm chart names of the form <repo-name>/<chart-name>

Camila-B:
* This PR adds logic to handle the case where the provided spec.helm.chart value is of the form `<repo-name>/<chart-name>`. This is needed as the **reconciler** attempts to read the rendered chart at `/repo/source/<version>/<repo-name>/<chart-name>` while the chart is rendered at  `/repo/source/<version>/<chart-name>` by **helm-sync**
* b/423072233

### DIFF
--- a/pkg/reconcilermanager/controllers/util.go
+++ b/pkg/reconcilermanager/controllers/util.go
@@ -17,6 +17,7 @@ package controllers
 import (
 	"fmt"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 	"time"
@@ -128,7 +129,8 @@ func reconcilerEnvs(opts reconcilerOptions) []corev1.EnvVar {
 		syncDir = opts.ociConfig.Dir
 	case configsync.HelmSource:
 		syncRepo = opts.helmConfig.Repo
-		syncDir = opts.helmConfig.Chart
+		syncDir = path.Base(opts.helmConfig.Chart)
+
 		if opts.helmConfig.Version != "" {
 			syncRevision = opts.helmConfig.Version
 		} else {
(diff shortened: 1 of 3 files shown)


pkg/reconcilermanager/controllers/util.go
```diff
@@ -128,7 +128,15 @@ func reconcilerEnvs(opts reconcilerOptions) []corev1.EnvVar {
 		syncDir = opts.ociConfig.Dir
 	case configsync.HelmSource:
 		syncRepo = opts.helmConfig.Repo
-		syncDir = opts.helmConfig.Chart
+		repoAndChart := strings.Split(opts.helmConfig.Chart, "/")
+
+		if len(repoAndChart) == 2 {
```
tiffanny29631: Have you tried with longer name that includes more than one `/`? I wonder if it should take the last element as chart name, instead of just the second one

Camila-B: I'm not sure if it's possible for repo names to have `/` in them. However, I changed the code to account for this case, which also simplifies the logic

tiffanny29631: lgtm, you could also update the public doc http://cloud/kubernetes-engine/enterprise/config-sync/docs/reference/rootsync-reposync-fields to reflect the definition of the field

tiffanny29631 (approved):
/lgtm

Camila-B:
As Sam provided a more thorough investigation into the bug in b/421345481 and determined that it is actually WAI, this change is no longer necessary

PR #1739, opened 2025-06-12
fix: register auth plugins for nomos

sdowell:
This re-adds the oidc plugin support for the nomos command. The oidc plugin was removed unintentionally when trying to clean up the other deprecated plugins.

Note that this plugin is also supported by kubectl.

Relevant PR: https://github.com/GoogleContainerTools/kpt-config-sync/pull/983


tiffanny29631 (approved):
/lgtm

PR #1740, opened 2025-06-12
fix: remove deprecated gcp plugin from r-g-ctrllr

sdowell:
The gcp plugin is deprecated and simply a stub that errors.


tiffanny29631 (approved):
/lgtm

PR #1742, opened 2025-06-15
refactored the nomos commands for better separation of concerns

mikebz:


cmd/nomos/bugreport/bugreport.go
sdowell, line 1: I'd really prefer if these changes preserved the git history - could this be refactored in a way that is more surgical? That would also make review easier

cmd/nomos/bugreport/cmd.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2022 Google LLC
```
sdowell: New files need the current copyright year

cmd/nomos/bugreport/bugreport.go
mikebz: the initial refactor was done with an LLM.  I think if someone wants to restart this they will probably going to need to take this over, but I am OK either way.

cmd/nomos/bugreport/cmd.go
mikebz: good catch.  Did a quick search of the things I also changed to fix more of these.  Handy script:

```
for file in $(git ls-files ./cmd/nomos); do
  if [ "$(git log -1 --pretty=format:'%an' -- "$file")" = "Mike Borozdin" ]; then
    echo "$file"
  fi
done
```

PR #1744, opened 2025-06-16
fix: add validation to spec.helm.chart

Camila-B:
This change adds validation to spec.helm.chart to ensure that it contains no slashes. A KNV1061 error is returned to the user if this condition is not met.

E2E tests validate that this error is returned for both charts hosted in OCI-based registries and traditional Helm repos

b/423072233

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


pkg/validate/rsync/validate/source_spec_validator.go
```diff
@@ -199,6 +200,12 @@ func HelmSpec(helm *v1beta1.HelmBase, syncKind string) status.Error {
 		return MissingHelmChart(syncKind)
 	}
 
+	// Validates chart name
+	re := regexp.MustCompile("^[a-z0-9-]*$")
```
mikebz: curious if this is in the Helm spec.  If so, let's link out to the spec or something else that we can cross reference in the future.

Camila-B: I added a link to https://helm.sh/docs/chart_best_practices/conventions/

Camila-B: I also changed the validation slightly to make sure that the chart name doesn't start or end with a hyphen

Camila-B:
Q: Should the validation solely focus on making sure that "/" is not used? That is, that all other conditions mentioned in https://helm.sh/docs/chart_best_practices/conventions/ should not be validated?

sdowell:
> Q: Should the validation solely focus on making sure that "/" is not used? That is, that all other conditions mentioned in https://helm.sh/docs/chart_best_practices/conventions/ should not be validated?

Yes I think that's a good suggestion, the known corner case we're trying to improve is silent failure due to incorrect directory mounting/mapping - so we can scope the validation to that for now. We can delegate to helm for other validation

Camila-B:
> > Q: Should the validation solely focus on making sure that "/" is not used? That is, that all other conditions mentioned in https://helm.sh/docs/chart_best_practices/conventions/ should not be validated?
> 
> Yes I think that's a good suggestion, the known corner case we're trying to improve is silent failure due to incorrect directory mounting/mapping - so we can scope the validation to that for now. We can delegate to helm for other validation

Ok I simplified the validation so that it only looks for any usage of "/"

pkg/validate/rsync/validate/source_spec_validator.go
```diff
@@ -199,6 +200,13 @@ func HelmSpec(helm *v1beta1.HelmBase, syncKind string) status.Error {
 		return MissingHelmChart(syncKind)
 	}
 
+	// Validates chart name such that it contains no slashes
+	// See https://helm.sh/docs/chart_best_practices/conventions/ for more details
+	re := regexp.MustCompile("^[^/]*$")
```
sdowell: Can just use `strings.Contains(helm.Chart, "/")`.

I prefer to avoid regex wherever possible, and I think it's overkill here

Camila-B: Sure done

sdowell:
LGTM, can you please update the commit message body with the context that is in the PR description?

e2e/testcases/helm_sync_test.go
```diff
@@ -135,6 +136,47 @@ func TestPublicHelm(t *testing.T) {
...
+	repo := strings.TrimSuffix(rs.Spec.Helm.Repo, "/"+repoSuffix)
+	nt.MustMergePatch(rs, fmt.Sprintf(`{"spec": {"helm": {"chart": "%s", "repo": "%s"}}}`, chart, repo))
+	nt.Must(nt.Watcher.WatchForRootSyncStalledError(rs.Name, "Validation",
+		`KNV1061: RootSync field spec.helm.chart must not contain a slash (/)`))
+}
+
+// TestOCIHelmChartNameHasRepoName verifies the Config Sync behavior for OCI-based Helm charts when helm.spec.chart is of the form <repo-name>/<chart-name>
+// This test will work only with following pre-requisite:
+// Google service account `e2e-test-ar-reader@${GCP_PROJECT}.iam.gserviceaccount.com` is created with `roles/artifactregistry.reader` for accessing images in Artifact Registry.
+func TestOCIHelmChartNameHasRepoName(t *testing.T) {
```
sdowell: nit: `TestOCIHelmChartNameContainsSlash` might be more precise

Camila-B:
> LGTM, can you please update the commit message body with the context that is in the PR description?

Done

e2e/testcases/helm_sync_test.go
Camila-B: I agree. I changed the name

pkg/reconcilermanager/controllers/reposync_controller_test.go
```diff
@@ -3775,6 +3775,9 @@ func TestRepoSyncWithOCI(t *testing.T) {
 }
 
 func TestRepoSyncSpecValidation(t *testing.T) {
+	// Mock out parseDeployment for testing.
+	parseDeployment = parsedDeployment
```
mikebz, line 3779: I think the naming here make it easy to make a mistake

PR #1745, opened 2025-06-16
[cherry-pick] fix: register auth plugins for nomos (#1739)

sdowell:
This re-adds the oidc plugin support for the nomos command. The oidc plugin was removed unintentionally when trying to clean up the other deprecated plugins.

Note that this plugin is also supported by kubectl.


tiffanny29631 (approved):
/lgtm

PR #1748, opened 2025-06-17
refactor: delete unused script file

Camila-B:
This change deletes an obsolete script file. See go/acm-publish-crd for more details.


scripts/mirror-foo-corp-to-gh.sh
```diff
@@ -1,84 +0,0 @@
...
-  REPO=$(mktemp -d /tmp/foo-corp-example-XXXXX)
-  trap cleanup EXIT
-  git clone "$REMOTE" "$REPO"
-fi
-
-if ! git -C "$REPO" checkout -t "origin/$REPO_VERSION" &>/dev/null; then
-  git -C "$REPO" checkout -B "$REPO_VERSION"
-fi
-
-EXAMPLE="$(dirname "$0")/../examples/foo-corp-example/"
```
sdowell, line 80: I think we can probably also looking at cleaning up some of the examples and the scaffolding surrounding it

sdowell (approved):
/lgtm

PR #1749, opened 2025-06-18
chore: clean up obsolete code

Camila-B:
-  Deletes unused repo-version/main.go
- Deletes Makefile rule that references it
- Deletes unused examples in foo-corp-example and only keeps foo-corp


cmd/nomos/vet/vet_test.go
```diff
@@ -92,19 +92,6 @@ func TestVet_AcmeSymlink(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestVet_FooCorp(t *testing.T) {
-	resetFlags()
-	Cmd.SilenceUsage = true
-
-	os.Args = []string{
-		"vet", // this first argument does nothing, but is required to exist.
-		"--path", examplesDir.Join(cmpath.RelativeSlash("foo-corp-example/foo-corp")).OSPath(),
```
sdowell, line 101: I think there may be value in keeping this specific test case for `nomos vet` test coverage, but the rest looks good to clean up.

sdowell (commented):
nit: this doesn't seem like a `ci` change, maybe `chore`

cmd/nomos/vet/vet_test.go
Camila-B: `TestVet_Acme` provides the same test coverage

Camila-B:
> nit: this doesn't seem like a `ci` change, maybe `chore`

Makes sense, I changed the prefix

cmd/nomos/vet/vet_test.go
sdowell: It doesn't look like the `acme` example is identical to the `foo-corp` example

Camila-B: So should I add back `foo-corp-example/foo-corp` so we can keep this testcase?

sdowell: yes, and you can probably collapse the subdirectory

Camila-B: Done

PR #1752, opened 2025-06-18
[cherry-pick]: bump nginx image used by TestDeclaredFieldsPod

mikebz:
This was an old nginx image that does not support current media types.

When trying to upgrade kind the following error was observed: media type application/vnd.docker.distribution.manifest.v1+prettyjws is no longer supported since containerd v2.1

This moves the image definition to the constant so it gets updated consistently.


sdowell (approved):
/lgtm

PR #1755, opened 2025-06-22
improve error message for large resource group handling

mikebz:
This pull request includes a minor update to the error message in the `largeResourceGroupError` function within `pkg/applier/applier_err.go`. The change simplifies the wording of the error message and adjusts the formatting for clarity because the underlying error message was too confusing.

b/418818042

### DIFF
--- a/pkg/applier/applier_err.go
+++ b/pkg/applier/applier_err.go
@@ -74,10 +74,8 @@ func SkipErrorForResource(err error, id core.ID, strategy actuation.ActuationStr
 // largeResourceGroupError indicates that the source repo has too many objects
 // to manage with a single resource group.
 func largeResourceGroupError(err error, id core.ID) status.Error {
-	e := fmt.Errorf("source repository contains too many resource objects to inventory with a single ResourceGroup: "+
-		"inventory apply failed: %v: %s. "+
-		"To mitigate, spread the source objects across multiple repositories. "+
-		"For how, see https://cloud.google.com/anthos-config-management/docs/how-to/breaking-up-repo",
-		id, err)
+	e := fmt.Errorf("the size of the ResourceGroup object for this repository is exceeding the row size. "+
+		"To mitigate follow instructions here https://cloud.google.com/anthos-config-management/docs/how-to/breaking-up-repo. "+
+		": %v", id)
 	return applierErrorBuilder.Wrap(e).Build()
 }
--- a/pkg/applier/applier_err.go
+++ b/pkg/applier/applier_err.go
@@ -80,6 +80,6 @@ func largeResourceGroupError(_ error, id core.ID) status.Error {
 	// Possibly in the future if the error changes we will start passing it back to the user.
 	e := fmt.Errorf("the size of the ResourceGroup object for this repository is exceeding the row size. "+
 		"To mitigate follow instructions here https://cloud.google.com/anthos-config-management/docs/how-to/breaking-up-repo. "+
-		": %v", id)
+		"%v", id)
 	return applierErrorBuilder.Wrap(e).Build()
 }
--- a/pkg/applier/applier_err.go
+++ b/pkg/applier/applier_err.go
@@ -78,8 +78,8 @@ func largeResourceGroupError(_ error, id core.ID) status.Error {
 	// says: Request entity too large: limit is 3145728 (eg)
 	// the actual size is 1.5mb, so it is a bit mislreading.
 	// Possibly in the future if the error changes we will start passing it back to the user.
-	e := fmt.Errorf("the size of the ResourceGroup object for this repository is exceeding the row size. "+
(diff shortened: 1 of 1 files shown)


pkg/applier/applier_err.go
```diff
@@ -74,10 +74,8 @@ func SkipErrorForResource(err error, id core.ID, strategy actuation.ActuationStr
...
 // to manage with a single resource group.
 func largeResourceGroupError(err error, id core.ID) status.Error {
-	e := fmt.Errorf("source repository contains too many resource objects to inventory with a single ResourceGroup: "+
-		"inventory apply failed: %v: %s. "+
-		"To mitigate, spread the source objects across multiple repositories. "+
-		"For how, see https://cloud.google.com/anthos-config-management/docs/how-to/breaking-up-repo",
-		id, err)
+	e := fmt.Errorf("the size of the ResourceGroup object for this repository is exceeding the row size. "+
+		"To mitigate follow instructions here https://cloud.google.com/anthos-config-management/docs/how-to/breaking-up-repo. "+
+		": %v", id)
```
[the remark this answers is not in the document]
mikebz: this is on purpose

pkg/applier/applier_err.go
```diff
@@ -73,11 +73,13 @@ func SkipErrorForResource(err error, id core.ID, strategy actuation.ActuationStr
...
 // largeResourceGroupError indicates that the source repo has too many objects
 // to manage with a single resource group.
-func largeResourceGroupError(err error, id core.ID) status.Error {
-	e := fmt.Errorf("source repository contains too many resource objects to inventory with a single ResourceGroup: "+
-		"inventory apply failed: %v: %s. "+
-		"To mitigate, spread the source objects across multiple repositories. "+
-		"For how, see https://cloud.google.com/anthos-config-management/docs/how-to/breaking-up-repo",
-		id, err)
+func largeResourceGroupError(_ error, id core.ID) status.Error {
+	// the error is not used because it's confusing to the user.  The current core error
```
sdowell, line 77: This seems like an anti-pattern to completely swallow the real error and replace it with a different error. The `misleading error` is coming from the apiserver and is the same thing a user would see if they tried to apply the oversized CR with `kubectl`. If the error is misleading then I think the action item is to improve the error string upstream in apiserver

mikebz: I agree that working with etcd and kubernetes layers is probably the right course of actions here, however it seems like that kind of a change is many months away and wouldn't serve the customers that have current versions of Kubernetes for some time.

Also agreed that generally swallowing errors is discouraged but I think the UX here is what error our user sees so we should probably figure out what leads to the least amount of support load.  As it stands today this is one of the most common scenarios so we should evaluate the potential remediations.

pkg/applier/applier_err.go
```diff
@@ -73,11 +73,13 @@ func SkipErrorForResource(err error, id core.ID, strategy actuation.ActuationStr
 
 // largeResourceGroupError indicates that the source repo has too many objects
 // to manage with a single resource group.
-func largeResourceGroupError(err error, id core.ID) status.Error {
-	e := fmt.Errorf("source repository contains too many resource objects to inventory with a single ResourceGroup: "+
-		"inventory apply failed: %v: %s. "+
-		"To mitigate, spread the source objects across multiple repositories. "+
-		"For how, see https://cloud.google.com/anthos-config-management/docs/how-to/breaking-up-repo",
-		id, err)
+func largeResourceGroupError(_ error, id core.ID) status.Error {
```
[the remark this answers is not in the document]
mikebz: the error field is added there speicifically in hopes that in the future we will be able to pass the error to the caller.

mikebz:
let me know if we want to close this or if there are iterations in the same direction.  All good either way but hanging PRs probably increase the cognitive load for seeing what needs to be reviewed and require rebases.

pkg/applier/applier_err.go
sdowell: I'm not totally convinced this is an improvement over the current behavior. I think I'd be more open to it if the underlying error is still logged but not surfaced on the RSync status, but even then it still seems like a bit of a code smell

pkg/applier/applier_err.go
```diff
@@ -73,11 +73,13 @@ func SkipErrorForResource(err error, id core.ID, strategy actuation.ActuationStr
...
+	// the error is not used because it's confusing to the user.  The current core error
+	// says: Request entity too large: limit is 3145728 (eg)
+	// the actual size is 1.5mb, so it is a bit mislreading.
```
tiffanny29631, line 79: I vote for piping the err from API server as is, and add clarification wording if the err message can be misleading, while improving the behavior from upstream. Not sure what is 1.5mb, is it an example or it's the actually size limit?

PR #1769, opened 2025-07-01
ci: group dependabot kustomize updates

sdowell:
kustomize contains multiple modules that are versioned together and can be grouped for updates to minimize the number of dependabot PRs


Camila-B (approved):
/lgtm

PR #1777, opened 2025-07-02
POC: using MutatingAdmissionPolicy to schedule CS pods

sdowell:
This change documents a proof-of-concept of configuring the node placement of Config Sync system pods using MutatingAdmissionPolicy.

This works out of the box, but is currently an alpha api and is behind a feature gate. It is targeted for beta in k8s 1.34.


e2e/testcases/mutating_admission_policy_test.go
```diff
@@ -0,0 +1,93 @@
...
+// and requires a feature gate.
+func TestMutatingAdmissionPolicy(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.Reconciliation2,
+		ntopts.SyncWithGitSource(nomostest.DefaultRootSyncID, ntopts.Unstructured),
+		ntopts.RequireKind(t))
+	rootSyncGitRepo := nt.SyncSourceGitReadWriteRepository(nomostest.DefaultRootSyncID)
+
+	mapFile := filepath.Join(".", "..", "..", "examples", "mutating-admission-policies", "config-sync-node-placement.yaml")
+	nt.T.Cleanup(func() {
+		nt.Must(nt.Shell.Kubectl("delete", "--ignore-not-found", "-f", mapFile))
```
tiffanny29631, line 44: Should the reconciler-manager have a reset to restore the pod state in a shared env?

sdowell: It's not strictly necessary because the node affinity is a no-op, but I added a cleanup step to be thorough

tiffanny29631 (approved):
/lgtm

PR #1778, opened 2025-07-02
test: refactor Wait to return error

sdowell:
This refactors the Wait test helper functions to return an error rather than calling Fatal within the test helper. This makes it simpler for the caller to decide how to handle the error and understand the control flow from the test.

Also refactors some of the code to use newer constructs like taskgroup.

### DIFF
--- a/e2e/testcases/acme_test.go
+++ b/e2e/testcases/acme_test.go
@@ -60,7 +60,7 @@ func TestAcmeCorpRepo(t *testing.T) {
 
 	checkResourceCount(nt, kinds.Namespace(), "", len(nsToFolder), nil, configSyncManagementAnnotations)
 	for namespace, folder := range nsToFolder {
-		checkNamespaceExists(nt, namespace, configSyncManagementLabels(namespace, folder), configSyncManagementAnnotations)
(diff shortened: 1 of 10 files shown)


e2e/testcases/acme_test.go
```diff
@@ -241,8 +241,8 @@ func checkResource(nt *nomostest.NT, obj client.Object, namespace, name string,
 	return nil
 }
 
-func checkNamespaceExists(nt *nomostest.NT, name string, labels, annotations map[string]string) {
-	nomostest.Wait(nt.T, "namespace exists", nt.DefaultWaitTimeout, func() error {
+func checkNamespaceExists(nt *nomostest.NT, name string, labels, annotations map[string]string) error {
```
tiffanny29631, line 244: How about moving this function to the nomostest pkg to group the helpers

sdowell: I'd prefer to remove this function and replace it with `WatchObject` - I can do that as a separate change

tiffanny29631: SG!

tiffanny29631 (approved):
/lgtm

e2e/testcases/acme_test.go
sdowell: https://github.com/GoogleContainerTools/kpt-config-sync/pull/1781

PR #1779, opened 2025-07-02
Update logic for function checking log messages in tests

LogicalShark:


e2e/testcases/postsync_logging_test.go
```diff
@@ -320,37 +319,36 @@ func TestCloudLoggingStatusEventRepoSync(t *testing.T) {
...
 			}
 		}
 	}
-	if count != expectedCount {
-		return fmt.Errorf("Expected %d occurrences of error log entry %s, but found %d",
-			expectedCount, messages, count)
+	for _, msg := range messages {
+		if count, ok := messageCounts[msg]; !ok || count != 1 {
+			return fmt.Errorf("failed to find single occurrence of each message in logs, counted:\n%v\nin logs:\n%v", messageCounts, logs)
+		}
```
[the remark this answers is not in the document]
LogicalShark: WAI

e2e/testcases/postsync_logging_test.go
```diff
@@ -320,37 +319,36 @@ func TestCloudLoggingStatusEventRepoSync(t *testing.T) {
...
-				allFound = true
-			}
-			if allFound {
-				count++
+		if !(strings.Contains(line, commit) && strings.Contains(line, resourceName) && strings.Contains(line, resourceNamespace)) {
+			continue
+		}
+		for _, msg := range messages {
+			if strings.Contains(line, msg) {
+				messageCounts[msg]++
```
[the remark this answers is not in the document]
LogicalShark: Prefer to provide all info

e2e/testcases/postsync_logging_test.go
```diff
@@ -320,37 +319,36 @@ func TestCloudLoggingStatusEventRepoSync(t *testing.T) {
...
 			}
 		}
 	}
-	if count != expectedCount {
-		return fmt.Errorf("Expected %d occurrences of error log entry %s, but found %d",
-			expectedCount, messages, count)
+	for _, msg := range messages {
+		if count, ok := messageCounts[msg]; !ok || count != 1 {
+			return fmt.Errorf("failed to find single occurrence of each message in logs, counted:\n%v\nin logs:\n%v", messageCounts, logs)
```
sdowell: How many logs are typically in this array? I'm concerned this error may be overly verbose, granted it would probably be useful for debugging

e2e/testcases/postsync_logging_test.go
```diff
@@ -320,37 +319,36 @@ func TestCloudLoggingStatusEventRepoSync(t *testing.T) {
...
 }
 
-// assertLogEntryHasCount checks if the specified messages appear in the logs
-// exactly the expected number of times, matching the given commit, name, and namespace.
-func assertLogEntryHasCount(logs []string, expectedCount int, commit, rname, rnamespace string, messages []string) error {
-	count := 0
-
+// assertOneLogLinePerMessage checks if each specified message appears in exactly one line
+// among the log entries which match the given commit, resource name, and namespace.
+func assertOneLogLinePerMessage(logs []string, commit, resourceName, resourceNamespace string, messages []string) error {
```
sdowell: Looking at the usages of this, I think this function can be simplified to just verifying a single message instead of an array of messages (e.g. `assertLogsContainMessageOnce`). I'm thinking for each of the existing assertions that pass in more than one message, we can just validate the first message in the array (for each assertion the first message seems to be the most meaningful). This should make the function easier to understand without weakening the assertion much IMO.

WDYT?

BTW this function definitely looks like it needed a refactor - good catch 👍

LogicalShark: Sounds good, simplifies it a lot. For the one test where the other message seemed relevant I added a second assertion

e2e/testcases/postsync_logging_test.go
LogicalShark: Probably not necessary, removed

e2e/testcases/postsync_logging_test.go
```diff
@@ -275,8 +275,8 @@ func TestStatusEventLogRepoSync(t *testing.T) {
 		nt.T.Fatal(err)
 	}
 
-	logMessages := []string{msg, "\"generation\":1"}
-	nt.Must(assertLogEntryHasCount(logs, 1, commit, repoSyncID.Name, repoSyncID.Namespace, logMessages))
+	nt.Must(assertOneLogLineWithMessage(logs, commit, repoSyncID.Name, repoSyncID.Namespace, msg))
+	nt.Must(assertOneLogLineWithMessage(logs, commit, repoSyncID.Name, repoSyncID.Namespace, "\"generation\":1"))
```
sdowell, line 279: I'm not sure how important the assertion is for the `generation` in the log, but as long as it consistently passes this seems fine

PR #1783, opened 2025-07-08
fix: use intended 1hr backoff cap for oci-sync/helm-sync

sdowell:
fix: use intended 1hr backoff cap for oci-sync/helm-sync

Casting the default duration to float64 was resulting in the incorrect
units being used (nanoseconds instead of seconds). This bug resulted in
a much larger backoff cap than intended.

This change properly casts the value to seconds and results in the
intended default backoff cap of 1 hour.

---

refactor: consolidate error backoff for sync containers

This change consolidates the function used to construct the error
backoff for the helm-sync and oci-sync containers. A unit test is added
to verify that the Cap is set as intended.

### DIFF
--- a/cmd/helm-sync/main.go
+++ b/cmd/helm-sync/main.go
@@ -78,7 +78,7 @@ var (
 )
 
 func errorBackoff() wait.Backoff {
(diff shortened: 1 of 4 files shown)


tiffanny29631 (approved):
/lgtm

PR #1784, opened 2025-07-09
fix: force live get of ignore-mutation objects when apply fails

sdowell:
The current design for ignore-mutation objects passes in the latest object from the cluster into the applier. A caching mechanism is used to minimize live Get requests to the cluster, and the cached object is updated by the remediator when it receives a watch event.

There is currently no fallback reconciliation mechanism if the object is stale and the remediator failed to update the object cache (for example, if the remediator missed the watch event). This change adds a fallback mechanism by clearing the ignore-mutation object from the cache whenever the apply fails for such an object.

Note that it is still possible for the object to become stale between the time that the object is fetched and the applier tries to apply the object, but this fallback mechanism should help guarantee eventual consistency when the apply fails due to a stale cache.

### DIFF
--- a/pkg/applier/applier.go
+++ b/pkg/applier/applier.go
@@ -207,7 +207,7 @@ func (s *supervisor) UpdateStatusMode(ctx context.Context) error {
 	})
 }
 
-func (s *supervisor) processApplyEvent(ctx context.Context, e event.ApplyEvent, syncStats *stats.ApplyEventStats, objectStatusMap ObjectStatusMap, unknownTypeResources map[core.ID]struct{}, resourceMap map[core.ID]client.Object) status.Error {
+func (s *supervisor) processApplyEvent(ctx context.Context, e event.ApplyEvent, syncStats *stats.ApplyEventStats, objectStatusMap ObjectStatusMap, unknownTypeResources map[core.ID]struct{}, resourceMap map[core.ID]client.Object, declaredResources *declared.Resources) status.Error {
 	id := idFrom(e.Identifier)
 	syncStats.Add(e.Status)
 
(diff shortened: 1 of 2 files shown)


pkg/applier/applier.go
```diff
@@ -231,6 +231,15 @@ func (s *supervisor) processApplyEvent(ctx context.Context, e event.ApplyEvent,
...
 		objectStatus.Actuation = actuation.ActuationFailed
 		handleMetrics(ctx, "update", e.Error)
+		// If apply failed for an ignore-mutation object, delete it from the ignore cache.
+		// Normally the cached object should be updated by the remediator when it
+		// receives a watch event - This is a fallback to force a live lookup the
+		// next time the applier runs.
+		iObj, found := declaredResources.GetIgnored(id)
+		if found {
+			klog.Infof("Deleting object '%v' from the ignore cache (apply failed)", core.GKNN(iObj))
+			declaredResources.DeleteIgnored(id)
```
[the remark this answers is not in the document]
sdowell: Good robot, sent https://github.com/GoogleContainerTools/kpt-config-sync/pull/1785

pkg/applier/applier_test.go
```diff
@@ -915,6 +915,7 @@ func TestProcessApplyEvent(t *testing.T) {
 	deploymentObj := newDeploymentObj()
```
mikebz, line 964: as a person who has never seen this test before I find it a little hard to figure out what the intent of this is.  Maybe a little context would help.  The comments as they stand seem to just outline what is happening next

sdowell: Could you clarify which intent you'd like to see more context on?

pkg/applier/applier_test.go
```diff
@@ -965,6 +973,39 @@ func TestProcessApplyEvent(t *testing.T) {
...
+			GVK: metav1.GroupVersionKind{
+				Group:   "configsync.test",
+				Version: "v1",
+				Kind:    "Test",
+			},
+		}},
+	}
+	testutil.AssertEqual(t, expectedCSE, err.ToCSE(), "expected CSEs to match")
+	// The object should be removed from the IgnoredObjects cache when ApplyFailed
+	testutil.AssertEqual(t, 0, len(resources.IgnoredObjects()))
```
tiffanny29631, line 1057: Would be helpful to fix the object, have a successful apply and check the ignoremutation object is added back to the cache.

pkg/applier/applier_test.go
mikebz: I guess when you don't know what the intent is it's hard to clarify which intent you are missing.  The test by naming seems like a happy path test:  TestProcessApplyEvent

Maybe folks with more context in the code @tiffanny29631 can take the lead on this review actually.

pkg/applier/applier_test.go
sdowell: Given this is a race condition, there isn't really a way to deterministically test this with a e2e/integration test (at least with the existing e2e test framework). This unit test just validates that the event is processed in the correct way in the scenario that a race condition occurs.

The `TestApplyMutationIgnoredObjects` test case above verifies that a live lookup is performed to update the cache if the object is missing from the cache. I've added another test case to that test which verifies that the object is evicted from the cache for ApplyFailed

tiffanny29631: Ack, thanks.

PR #1785, opened 2025-07-09
fix: use write lock for DeleteIgnored

sdowell:
This function mutates the map in the critical section and therefore should not use a read lock.


tiffanny29631 (approved):
/lgtm

PR #1789, opened 2025-07-10
test: fix GetDeploymentPod for failed Pods

sdowell:
Sometimes a single replica Deployment can have multiple Pods when an older Pod is failed and the new Pod is running. This change updates GetDeploymentPod when in this intermediate state to return the running Pod and ignore the failed Pod(s).

This is a fix for some test flakes which have been observed

### DIFF
--- a/e2e/nomostest/testkubeclient/client.go
+++ b/e2e/nomostest/testkubeclient/client.go
@@ -206,15 +206,21 @@ func (tc *KubeClient) GetDeploymentPod(deploymentName, namespace string, retrytT
 		if err != nil {
 			return err
 		}
-		if len(pods.Items) != 1 {
-			return fmt.Errorf("deployment has %d pods, expected 1: %s\n%s",
-				len(pods.Items), deploymentNN, log.AsYAML(pods))
+		numRunning := 0
+		runningPodIdx := -1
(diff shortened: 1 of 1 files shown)


e2e/nomostest/testkubeclient/client.go
```diff
@@ -206,15 +206,21 @@ func (tc *KubeClient) GetDeploymentPod(deploymentName, namespace string, retrytT
...
-			return fmt.Errorf("deployment has %d pods, expected 1: %s\n%s",
-				len(pods.Items), deploymentNN, log.AsYAML(pods))
+		numRunning := 0
+		runningPodIdx := -1
+		// There should be exactly one Pod with status.phase: Running.
+		// It is possible there are other pods that are Terminating.
+		for idx, po := range pods.Items {
+			if po.Status.Phase == corev1.PodRunning {
+				numRunning++
+				runningPodIdx = idx
```
tiffanny29631, line 216: Are the pods sorted in a way that the fresh running pod will show up last in the slice? Was wondering if we need to look at the age of pod instead of a static assignment when a running pod is found

sdowell: I assume the Pods are unsorted, however it shouldn't matter because this only accepts a single running pod

PR #1791, opened 2025-07-11
Add startupProbe to otel-agent for reliable startup

tiffanny29631:
The otel-agent container can take extra time to bind to port 13133, especially when the cluster is under heavy load or resources are constrained. Without a startup probe, the pod may be restarted prematurely by Kubernetes if the readiness or liveness probes fail during this slow startup period, leading to CrashLoopBackOff errors.

By introducing a startupProbe, we provide a dedicated grace period for the otel-agent to initialize and bind to its health check port before readiness and liveness checks begin. This ensures the pod is not restarted unnecessarily during startup, improving reliability and reducing the risk of CrashLoopBackOff due to transient startup delays.

### DIFF
--- a/manifests/templates/reconciler-manager-configmap.yaml
+++ b/manifests/templates/reconciler-manager-configmap.yaml
@@ -195,6 +195,13 @@ data:
            volumeMounts:
            - name: otel-agent-config-reconciler-vol
              mountPath: /conf
+           startupProbe:
+             httpGet:
+               path: /
+               port: 13133
+             failureThreshold: 30   # Allow up to 5 minutes (30 * 10s)
+             periodSeconds: 10
+             initialDelaySeconds: 0
            readinessProbe:
              httpGet:
                path: /
--- a/manifests/templates/reconciler-manager.yaml
+++ b/manifests/templates/reconciler-manager.yaml
@@ -91,6 +91,13 @@ spec:
         volumeMounts:
         - name: otel-agent-config-vol
           mountPath: /conf
+        startupProbe:
+          httpGet:
+            path: /
+            port: 13133
+          failureThreshold: 30   # Allow up to 5 minutes (30 * 10s)
+          periodSeconds: 10
+          initialDelaySeconds: 0
         readinessProbe:
           httpGet:
             path: /
--- a/manifests/templates/resourcegroup-manifest.yaml
+++ b/manifests/templates/resourcegroup-manifest.yaml
@@ -275,6 +275,13 @@ spec:
         - containerPort: 55678
         - containerPort: 8888
         - containerPort: 13133
+        startupProbe:
+          httpGet:
+            path: /
+            port: 13133
+          failureThreshold: 30   # Allow up to 5 minutes (30 * 10s)
+          periodSeconds: 10
+          initialDelaySeconds: 0
         readinessProbe:
           httpGet:
             path: /


manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -195,6 +195,13 @@ data:
            volumeMounts:
            - name: otel-agent-config-reconciler-vol
              mountPath: /conf
+           startupProbe:
```
sdowell: Could you clarify what the intent is with this change? The commit message mentions liveness probe, however there is no liveness probe defined for these containers (since [2023](https://github.com/GoogleContainerTools/kpt-config-sync/pull/788)).

Also I expect the `readinessProbe` on these containers may be superfluous as none of these Deployments are exposed through a Service (except for `otel-collector` Deployment)

tiffanny29631: Updated in http://b/419026380#comment3. The Readiness probe for otel-agent side car can also be removed to align with the config of other containers. The readiness failure of telemetry containers is not necessary, and could block the reconciler and resource group controller pods.

sdowell: > The readiness failure of telemetry containers is not necessary, and could block the reconciler and resource group controller pods.

Could you explain why the readiness probe blocks the reconciler and resource-group-controller Pods?

tiffanny29631: The PR has been updated to removal of the readiness probe of otel-agent side car in reconciler / reconciler-manager / resource group controller.

manifests/templates/reconciler-manager.yaml
```diff
@@ -91,10 +91,6 @@ spec:
         volumeMounts:
         - name: otel-agent-config-vol
           mountPath: /conf
-        readinessProbe:
```
tiffanny29631, line 95: Odd. I removed the readinessProbe from both the reconciler-manager and resource-group-controller configurations, but when checking the pod description for the failed tests, the readiness probe is still showing up. Also there are two resource-group-controller pods running, with one not being ready due to readiness failure, which seems unusual.

tiffanny29631: Same issue does not apply to the test root-reconciler, suspecting caching issue.

manifests/templates/reconciler-manager-configmap.yaml
sdowell: > The PR has been updated to removal of the readiness probe of otel-agent side car in reconciler / reconciler-manager / resource group controller.

The motivation and intent for this change is still unclear. Can you explain the *why* of this change?

tiffanny29631: The otel-agent container (sidecar) in the reconciler / resource-group-controller currently has a readiness probe configured, which is causing operational issues and false alarms during cluster startup.

Issues with Current Readiness Probe:
- False Alarms During Slow Cluster Startup: When clusters have slow startup times, the otelcontribcol image's health check component may fail to bind to the readiness probe port properly. This results in false "pod not ready" alarms even when the pod is actually functional and running.
- Inconsistent Configuration: While readiness probes are generally recommended Kubernetes best practices, the Config Sync project maintains consistency by not configuring readiness probes for other critical containers like git-sync and reconciler. The otel-agent container should follow the same pattern for consistency.
- Redundant Health Checking: The otel-agent container serves as a telemetry sidecar and doesn't provide direct user-facing services. Its readiness status doesn't necessarily indicate the overall health of the resource group controller or reconciler functionality.

The test-group-1 is experiencing inconsistent test failures because a readiness probe configuration keeps reappearing for the otel-agent container in the resource-group-controller-manager deployment, despite attempts to remove it.

If change is found unnecessary we can close it.

sdowell: Ack, can you include the justification in the commit message and PR descriptions?

---

> causing operational issues

Can you clarify what are the operational issues?

> "pod not ready" alarms

Can you clarify which alarms this is referring to?

---

The change seems reasonable but I'm a bit surprised by the amount of churn. I'd perhaps consider timeboxing this effort given it doesn't solve the underlying issue of the `otel-agent` being unresponsive

tiffanny29631: Updated, also limited the change to only remove the readinessProbe for consistent test result.

sdowell: > Updated, also limited the change to only remove the readinessProbe for consistent test result.

Could you elaborate a bit more on this? What was the cause of the prior test failures and why are the tests passing now?

If the readinessProbe is removed, I would expect the health check endpoint/port to be removed as well. Why is the health check endpoint retained?

tiffanny29631: The test failure was inconsistent and flaky between runs. You can see the test history in the PR email thread.

Previously even after completely removing the readiness configuration, some tests still showed the resource-group-controller and reconciler-manager pods with readiness settings. This caused failures because the port/health_check extension was already removed.

When I ran the same tests on my local kind cluster, this issue didn't happen, so I can't identify the root cause yet. Suggest we could do this in two steps - first remove the readiness probe, then remove the port & extension infrastructure.

tiffanny29631: Wait I might have reproduced it, allow some more time to figure out the rootcause.

e2e/testcases/cli_test.go
```diff
@@ -1301,6 +1302,23 @@ func TestNomosMigrate(t *testing.T) {
 
 	nt.T.Cleanup(func() {
 		// Restore state of Config Sync installation after test
+		// Legacy ConfigManagement sets readiness on reconciler-manager and resource-group-controller,
```
sdowell: Good catch! This could probably be pulled out as a standalone change.

This also uncovers an edge case / side effect that we should be aware of. If the client uses server side apply to apply the new manifests immediately after uninstalling ACM operator, the resulting objects may be incomplete/incorrect. We need to ensure that clients which use server side apply first apply the manifests from the previous version so the client takes ownership of the fields (e.g. `readinessProbe` etc). This should be verified for each installation path

manifests/templates/reconciler-manager.yaml
tiffanny29631: Issue caused by nomos migrate test in nomos_cli test not restoring the controller managers into a desired state. Fixed in a different PR.

e2e/testcases/cli_test.go
tiffanny29631: The test refinement is moved to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1834. 

Just to confirm, are the install methods still console, gcloud, terraform, and manual? How would someone using gcloud or the console get a previous manifest version? Would a force install of that version work? I was also curious about the remove-configmanagement, whether it would work on a managed installed config sync, like through console

sdowell: I think there's a lot to unpack here, probably better to discuss offline or in a doc

manifests/templates/reconciler-manager-configmap.yaml
```diff
@@ -195,6 +195,13 @@ data:
...
+             httpGet:
+               path: /
+               port: 13133
+             failureThreshold: 30   # Allow up to 5 minutes (30 * 10s)
+             periodSeconds: 10
+             initialDelaySeconds: 0
```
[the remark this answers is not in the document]
tiffanny29631: This field is supposed to be removed.

manifests/templates/resourcegroup-manifest.yaml
```diff
@@ -275,6 +275,13 @@ spec:
         - containerPort: 55678
         - containerPort: 8888
         - containerPort: 13133
+        startupProbe:
+          httpGet:
+            path: /
+            port: 13133
+          failureThreshold: 30   # Allow up to 5 minutes (30 * 10s)
+          periodSeconds: 10
+          initialDelaySeconds: 0
```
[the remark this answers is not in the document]
tiffanny29631: This field is supposed to be removed.

manifests/templates/reconciler-manager.yaml
```diff
@@ -91,6 +91,13 @@ spec:
         volumeMounts:
         - name: otel-agent-config-vol
           mountPath: /conf
+        startupProbe:
+          httpGet:
+            path: /
+            port: 13133
+          failureThreshold: 30   # Allow up to 5 minutes (30 * 10s)
+          periodSeconds: 10
+          initialDelaySeconds: 0
```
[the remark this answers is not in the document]
tiffanny29631: This field is supposed to be removed.

e2e/testcases/cli_test.go
```diff
@@ -1297,13 +1297,14 @@ func TestApiResourceFormatting(t *testing.T) {
 }
 
 func TestNomosMigrate(t *testing.T) {
-	nt := nomostest.New(t, nomostesting.NomosCLI, ntopts.SkipConfigSyncInstall)
+	nt := nomostest.New(t, nomostesting.NomosCLI)
```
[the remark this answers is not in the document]
tiffanny29631: This is rebased from change https://github.com/GoogleContainerTools/config-sync/pull/1834

e2e/testcases/cli_test.go
```diff
@@ -1451,11 +1452,11 @@ func TestNomosMigrate(t *testing.T) {
...
 	})
 	tg.Go(func() error {
-		return nt.Watcher.WatchForNotFound(kinds.Deployment(),
+		return nt.Watcher.WatchForCurrentStatus(kinds.Deployment(),
 			core.RootReconcilerName(configsync.RootSyncName), configsync.ControllerNamespace)
 	})
 	tg.Go(func() error {
-		return nt.Watcher.WatchForNotFound(kinds.RootSyncV1Beta1(),
+		return nt.Watcher.WatchForCurrentStatus(kinds.RootSyncV1Beta1(),
 			configsync.RootSyncName, configsync.ControllerNamespace)
```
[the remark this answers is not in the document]
tiffanny29631: This is rebased from change https://github.com/GoogleContainerTools/config-sync/pull/1834

manifests/templates/reconciler-manager-configmap.yaml
tiffanny29631: Tests are now passing, change is ready for review.

sdowell: Sounds good! As discussed previously let's address [the concerns around upgrades](https://github.com/GoogleContainerTools/config-sync/pull/1791#discussion_r2283040584) before proceeding further

tiffanny29631: I've filed b/443790079 to track this. I recommend we merge PR #1834 first, as it's an independent test fix and unrelated to the readinessProbe changes.

e2e/testcases/cli_test.go
tiffanny29631: Created b/443790079.

sdowell:
Hold this until we address the SSA->CSA issue that we discussed
/hold

tiffanny29631:
/unhold

The test & deployment plan is finalized in [go/cdcs-field-ownership](http://goto.google.com/cdcs-field-ownership)

e2e/testcases/cli_test.go
```diff
@@ -1451,11 +1452,11 @@ func TestNomosMigrate(t *testing.T) {
 			configmanagement.RGControllerName, configmanagement.RGControllerNamespace)
 	})
 	tg.Go(func() error {
-		return nt.Watcher.WatchForNotFound(kinds.Deployment(),
+		return nt.Watcher.WatchForCurrentStatus(kinds.Deployment(),
```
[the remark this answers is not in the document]
tiffanny29631: Test initialization now creates test default Rootsync so the assertion needs to be updated.

e2e/testcases/cli_test.go
```diff
@@ -1451,11 +1452,11 @@ func TestNomosMigrate(t *testing.T) {
...
 			core.RootReconcilerName(configsync.RootSyncName), configsync.ControllerNamespace)
 	})
 	tg.Go(func() error {
-		return nt.Watcher.WatchForNotFound(kinds.RootSyncV1Beta1(),
+		return nt.Watcher.WatchForCurrentStatus(kinds.RootSyncV1Beta1(),
```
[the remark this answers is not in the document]
tiffanny29631: Test initialization now creates test default Rootsync so the assertion needs to be updated.

e2e/testcases/cli_test.go
```diff
@@ -1861,11 +1863,11 @@ func TestACMUninstallScript(t *testing.T) {
 			configmanagement.RGControllerName, configmanagement.RGControllerNamespace)
 	})
 	tg.Go(func() error {
-		return nt.Watcher.WatchForNotFound(kinds.Deployment(),
+		return nt.Watcher.WatchForCurrentStatus(kinds.Deployment(),
```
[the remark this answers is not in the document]
tiffanny29631: Test initialization now creates test default Rootsync so the assertion needs to be updated.

e2e/testcases/cli_test.go
```diff
@@ -1861,11 +1863,11 @@ func TestACMUninstallScript(t *testing.T) {
...
 			core.RootReconcilerName(configsync.RootSyncName), configsync.ControllerNamespace)
 	})
 	tg.Go(func() error {
-		return nt.Watcher.WatchForNotFound(kinds.RootSyncV1Beta1(),
+		return nt.Watcher.WatchForCurrentStatus(kinds.RootSyncV1Beta1(),
```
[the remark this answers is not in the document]
tiffanny29631: Test initialization now creates test default Rootsync so the assertion needs to be updated.

PR #1800, opened 2025-07-24
fix: handle ArgoCD Rollout observedGeneration type mismatch

tiffanny29631:
ArgoCD Rollouts sets status.observedGeneration as string instead of int64, causing kstatus library failures. This fix converts string values to int64 before status computation for ArgoCD Rollout objects.

Related upstream issue: https://github.com/argoproj/argo-rollouts/issues/3402

### DIFF
--- a/pkg/resourcegroup/controllers/status/status.go
+++ b/pkg/resourcegroup/controllers/status/status.go
@@ -16,6 +16,7 @@ package status
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


mikebz:
I am a bit suspicious about hard coding a specific CR name.  Do we not want to do anything in general in this case?

tiffanny29631:
> I am a bit suspicious about hard coding a specific CR name. Do we not want to do anything in general in this case?

Updated to a generalized fix of observedGeneration type.

pkg/resourcegroup/controllers/status/status.go
```diff
@@ -30,11 +31,38 @@ import (
...
 
+const (
+	// StatusField is the field name for status
+	StatusField = "status"
+	// ObservedGenerationField is the field name for observedGeneration
+	ObservedGenerationField = "observedGeneration"
+)
+
+// fixObservedGenerationType fixes the observedGeneration type issue for any resource.
+// Some controllers (like ArgoCD Rollouts) set status.observedGeneration as a string
```
sdowell, line 42: My first impression here is that it feels like this is working as intended from our end. It's standard for `status.observedGeneration` to be expressed as int64 (the same type as [metadata.generation](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L177), where the value is mapped from). To me it seems expected that the status is reported as "unknown" for a CRD that breaks the convention of a standard field.

The fundamental issue here appears to be that the ArgoCD rollouts CRD breaks the convention. The issue should really be fixed there, as the type mismatch is likely to break assumptions made by other consumers as well.

If expressing `status.observedGeneration` as a `string` is a common enough occurrence, then IMO the workaround would belong upstream in the [kstatus library](https://github.com/kubernetes-sigs/cli-utils/blob/v0.37.2/pkg/kstatus/status/generic.go#L73). This would ensure the status is consistently evaluated by other clients (including the cli-utils applier used by Config Sync). However I would still probably push back against this workaround even in `kstatus`, as it appears "incorrect" to express this field as type `string`.

tiffanny29631:
Quoting from b/ :
> The Argo Rollout switched to use string convention in early stage without much of justification https://github.com/argoproj/argo-rollouts/pull/5, from what I read the intention was to use hash to reflect more granular status, but not clear on why the value cannot be converted back to int64 to align with k8s convention, it could be that the hashed observedGeneration is no longer a strictly numeric string. We lean more on handling this exception in the upstream , until Argo responses to https://github.com/argoproj/argo-rollouts/issues/3402 or https://github.com/argoproj/argo-rollouts/discussions/976

For user currently seeing this issue, please encourage the fix in upstream Argo Rollouts until further notice.

mikebz:
Thank you Tiffany for closing the loop.  I agree that this is the right decision.

PR #1801, opened 2025-07-24
test: fix githubapp expectation in proxy e2e test

sdowell:
The fake githubapp Secret actually breaks the expectation for the RootSyncStalledError, since it makes the RootSync spec valid. This creates a race condition where the stalled error must be detected before the reconciler-manager can reconcile the contrived Secret.

The fake secret is unnecessary for the assertion and is removed.


tiffanny29631 (approved):
/lgtm

PR #1803, opened 2025-07-29
test: fix SSM credential helper entry to use explicit hostname

sdowell:
Some versions of git do not support wildcards in the credential helper host name matching. This issue was encountered in the test environment, which was not correctly using the credential helper.

This change fixes the test runner and increases the portability/explicitness of the credential helper.


Camila-B (approved):
/lgtm

PR #1804, opened 2025-07-29
test: fix tests failing with SSM gitprovider

Camila-B:
Since SSM uses 'gcpserviceaccount' for authentication, it doesn't require a Secret to be created. This change fixes some tests that were failing as they validated the existence of a Secret.

### DIFF
--- a/e2e/testcases/reconciler_manager_test.go
+++ b/e2e/testcases/reconciler_manager_test.go
@@ -444,8 +444,8 @@ func validateRepoSyncDependencies(nt *nomostest.NT, ns, rsName string) []client.
 	repoSyncDependencies = append(repoSyncDependencies, repoSyncSA)
 
 	// See nomostest.CreateNamespaceSecrets for creation of user secrets.
-	// The Secret is neither needed nor created when using CSR as the Git provider.
-	if nt.GitProvider.Type() != e2e.CSR {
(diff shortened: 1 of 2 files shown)


e2e/testcases/reconciler_manager_test.go
```diff
@@ -444,8 +444,8 @@ func validateRepoSyncDependencies(nt *nomostest.NT, ns, rsName string) []client.
 	repoSyncDependencies = append(repoSyncDependencies, repoSyncSA)
 
 	// See nomostest.CreateNamespaceSecrets for creation of user secrets.
-	// The Secret is neither needed nor created when using CSR as the Git provider.
-	if nt.GitProvider.Type() != e2e.CSR {
+	// The Secret is neither needed nor created when using CSR or SSM as the Git provider.
+	if nt.GitProvider.Type() != e2e.CSR && nt.GitProvider.Type() != e2e.SSM {
```
sdowell, line 448: Line 944 also needs to be updated.

Also nit: This check should be moved to a helper function (e.g. `IsGoogleGitProvider`) and used by the other places which have special logic based on the assumption that `gcpserviceaccount` is used for this type of gitprovider. The refactor can be done separately if preferred

Camila-B: Thanks. I'll do that in a separate PR

PR #1808, opened 2025-08-04
test: reduce flakiness of TestSurfaceFightError

sdowell:
To trigger the fight detector, the remediator needs to update the same object at least 5 times in a minute. If multiple watch events happen in too quick of succession, the updates might get batched together as a single remediation. Spacing out the watch events and creating more watch events than the minimum should make this test less flaky.


e2e/testcases/remediator_test.go
```diff
@@ -45,17 +45,19 @@ func TestSurfaceFightError(t *testing.T) {
...
-			nt.MustMergePatch(rb, `{"metadata": {"annotations": {"foo": "baz"}}}`)
-			time.Sleep(time.Second)
+			time.Sleep(500 * time.Millisecond)
 		}
 	}()
 
 	nt.T.Log("The RootSync reports a fight error")
 	nt.Must(nt.Watcher.WatchForRootSyncSyncError(configsync.RootSyncName, status.FightErrorCode,
 		"This may indicate Config Sync is fighting with another controller over the object.", nil))
+	<-doneCh // Wait for fighting goroutine to exit
```
tiffanny29631: Should the watch for error starts after the rapid updates are done?

e2e/testcases/remediator_test.go
```diff
@@ -45,17 +45,19 @@ func TestSurfaceFightError(t *testing.T) {
...
 
 	rootSyncNN := nomostest.RootSyncNN(configsync.RootSyncName)
 	rootSyncLabels, err := nomostest.MetricLabelsForRootSync(nt, rootSyncNN)
```
tiffanny29631, line 73: Need to update the metric criteria

e2e/testcases/remediator_test.go
sdowell: Not necessarily, the idea is that it watches for the error while the async updates are made

e2e/testcases/remediator_test.go
sdowell: Which metric criteria needs to change? The test is passing when I ran it locally

sdowell: The metric validation uses "hasValueAtLeast", so any value greater than 5 is acceptable. This is reasonable as the main thing we are trying to validate is that it **does** detect the fight - the exact number of update events the remediator detects is less relevant and is hard to reliably assert (e.g. some events may be batched, missed, etc)

tiffanny29631 (approved):
/lgtm

PR #1812, opened 2025-08-05
test: fix metrics flake in TestSurfaceFightError

sdowell:
The previous change fixed the flakiness with the RootSync status validation, but now the metrics validation is still flaky because the async status updates may stop before the expected number of fight metrics are emitted. This change validates the RootSync status and metric in parallel and only stops the async status updates once both conditions succeed/timeout.

Also adds an initial assertion to verify there are no fight errors at the beginning.


tiffanny29631 (approved):
/lgtm

PR #1813, opened 2025-08-06
[cherry-pick] additional cherrypicks to fix CI

Camila-B:


sdowell:
nit: we don't have any more patch releases planned for 1.21, so the cherry pick should not be necessary

Camila-B:
@sdowell so it's ok for the release testgrid to remain noisy for now?

tiffanny29631:
I'll be unduty until the 1.22 branch cut, it's fine to leave it like this.

Camila-B:
Alright sounds good!

PR #1815, opened 2025-08-06
refactor: fightThreshold global variable

Camila-B:
* Adds DefaultFightThreshold to register.go
* Adds fightThreshold member to the Detector struct

### DIFF
--- a/pkg/api/configsync/register.go
+++ b/pkg/api/configsync/register.go
@@ -90,6 +90,17 @@ const (
 
 	// DefaultHelmReleaseNamespace is the default namespace for a Helm Release which does not have a namespace specified
 	DefaultHelmReleaseNamespace = "default"
+
+	// DefaultFightThreshold is the default threshold of updates per minute to a
+	// resource before Config Sync reports a fight. A fight occurs when the
+	// Remediator and another process are both trying to manage the same
+	// resource.
+	// When a fight is detected, an error is logged, surfaced in the
+	// RSync status, and emitted as a metric.
+	// This value was chosen because a resource being updated more than 5 times
+	// per minute is a clear sign of a problem, while less frequent updates are
+	// unlikely to be fights.
+	DefaultFightThreshold = 5.0
 )
 
 // SourceFormat specifies how the Importer should parse the repository.
--- a/pkg/syncer/reconcile/fight/detector.go
+++ b/pkg/syncer/reconcile/fight/detector.go
@@ -24,20 +24,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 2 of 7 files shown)


pkg/syncer/reconcile/fight/detector.go
```diff
@@ -24,20 +24,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// fightThreshold is the threshold of updates per minute at which we log to Info
-// that the Syncer is fighting over a resource on the API Server with some
-// other process.
-//
-// This value was chosen arbitrarily as updates occurring more frequently are
-// obviously problems, and we don't really care about less frequent updates.
-var fightThreshold = 5.0
```
sdowell, line 33: The change overall seems reasonable, but just a minor note - While global variables like this are generally frowned upon in many (perhaps most) other programming languages, package level variables like this in Go are typically considered idiomatic

All that to say - I think it would also have been an acceptable solution to just unify the declaration of the default value

pkg/syncer/reconcile/fight/detector.go
```diff
@@ -59,13 +45,18 @@ type Detector struct {
 	fights map[core.ID]*fight
 
 	fLogger *logger
+
+	// fightThreshold is the maximum allowed rate of updates to a resource
+	// per minute before we begin logging errors to the user.
```
sdowell: nit: the errors aren't just logged - it's surfaced on the RSync status and also emitted as a metric

pkg/api/configsync/register.go
```diff
@@ -90,6 +90,13 @@ const (
 
 	// DefaultHelmReleaseNamespace is the default namespace for a Helm Release which does not have a namespace specified
 	DefaultHelmReleaseNamespace = "default"
+
+	// DefaultFightThreshold is the default threshold of updates per minute at which we log to Error
+	// that the Syncer is fighting over a resource on the API Server with some
```
sdowell: nit: I believe the reference to `Syncer` is a legacy artifact from the old "monorepo" architecture. Since we're changing this I'd suggest updating the description to clarify this is used by the remediator, surfaces the error on the RSync status, and emits a metric

Camila-B: Done

pkg/syncer/reconcile/fight/detector.go
Camila-B: Got it. I updated the comment

pkg/syncer/reconcile/fight/detector.go
Camila-B: Thanks! I wasn't aware of that but it's good to know for the future

sdowell (approved):
/lgtm

PR #1820, opened 2025-08-08
refactor: decouple business logic for nomos version command

Camila-B:
This PR decouples the business logic of `nomos version` from Cobra command, in a similar fashion to https://github.com/GoogleContainerTools/kpt-config-sync/pull/1742


pkg/version/version.go
```diff
@@ -14,5 +14,234 @@
...
+		var pathErr *os.PathError
+		if errors.As(err, &pathErr) {
+			err = pathErr
+		}
+
+		fmt.Printf("failed to create client configs: %v\n", err)
+	}
+
+	return allCfgs, err
+}
```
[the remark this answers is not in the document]
Camila-B: Done

mikebz:
this seems like it's just dealing with a subset of commands.  Is that because more changes are coming?  Do we have a larger plan?

cmd/nomos/version/version.go
sdowell, line 1: Thank you for the PR! Could you clarify what are the goals/motivations for this refactor?

Camila-B:
I added more context to the changes introduced by this PR in the description

cmd/nomos/version/version.go
sdowell: It's not clear how this improves modularity or maintainability - or why these functions need to be more modular. Are these being moved to a package so that the code can be used as a Go library? My understanding of this code is that it's only really intended to be used for the `nomos version` command.

If this is part of a larger scale refactor, could you draft up a more concrete proposal of the desired end state?

Camila-B: That's fair. Given those points, I'll go ahead and close this PR as it doesn't actually seem that the changes are necessary for the version command

PR #1831, opened 2025-08-19
test: Add retry / skip logic and limit check for Bitbucket repository deletion

tiffanny29631:
The Bitbucket backend occasionally has issues creating or removing repos,
leaving them stuck in pending/starting state, causing CI to hang on failed
cleanup operations. Before the root cause is figured out, we make this change
to improve reliability:

- Add retry logic with 30s timeout for repository deletion
- Skip deletion of repositories that are "currently not available" 
- Add 1000 repo limit check to indicate workspace overflow
- Add TODO linking to known issue

### DIFF
--- a/e2e/nomostest/gitproviders/bitbucket.go
+++ b/e2e/nomostest/gitproviders/bitbucket.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/uuid"
 	"go.uber.org/multierr"
 	"kpt.dev/configsync/e2e"
+	"kpt.dev/configsync/e2e/nomostest/retry"
 	"kpt.dev/configsync/e2e/nomostest/testlogger"
 )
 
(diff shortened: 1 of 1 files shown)


e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -140,31 +141,69 @@ func (b *BitbucketClient) DeleteRepositories(names ...string) error {
...
+		_, err := retry.Retry(30*time.Second, func() error {
+			return deleteSingleRepo(accessToken, name)
+		})
 		if err != nil {
-			errs = multierr.Append(errs, fmt.Errorf("%s: %w", string(out), err))
-		}
-		if len(out) != 0 {
-			errs = multierr.Append(errs, errors.New(string(out)))
+			// Skip repositories that are currently unavailable
+			if strings.Contains(err.Error(), "Repository currently not available") {
```
sdowell, line 155: It's not clear to me what skipping a failed deletion fixes. The repo name is deterministic so if there's a dangling repo, the next test execution will encounter the dirty state and fail. Could you clarify what the intent is behind this change?

tiffanny29631: This is a temporary workaround to prevent freshtestenv from failing at startup. It skips the deletion of repositories older than 24 hours if they fail with a "currently not available" error, which happens when a repo is in a bad state. This reduces the need for manual Bitbucket workspace updates until the underlying issue is resolved by Bitbucket.

sdowell: Looks like there was a misunderstanding - to clarify:

For most of the git provider implementations the repository name is static/deterministic and reused between test executions - I was operating under this assumption which would mean that if a repo is in a dirty/bad state subsequent executions would still fail.

It looks like gitlab and bitbucket are special cases in that we generate a UUID in the repository name, so the repository names are unique/nondeterministic for each test execution

PR #1834, opened 2025-08-19
Cleanup legacy controller deployments in test teardown

tiffanny29631:
Improve test reliability by deleting legacy reconciler-manager and resource-group-controller deployments in nomos migrate test cleanup. 

Legacy ConfigManagement installed during nomos migrate test sets readiness on these controllers, which isn’t reliably cleaned up after tests. Removing both deployments ensures a fresh Config Sync install instead of patching, reducing flakiness and side effects between tests.

### DIFF
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -43,6 +43,7 @@ import (
 	"kpt.dev/configsync/e2e/nomostest/ntopts"
 	"kpt.dev/configsync/e2e/nomostest/policy"
 	"kpt.dev/configsync/e2e/nomostest/syncsource"
+
 	"kpt.dev/configsync/e2e/nomostest/taskgroup"
 	nomostesting "kpt.dev/configsync/e2e/nomostest/testing"
 	"kpt.dev/configsync/e2e/nomostest/testpredicates"
@@ -1301,6 +1302,23 @@ func TestNomosMigrate(t *testing.T) {
 
 	nt.T.Cleanup(func() {
 		// Restore state of Config Sync installation after test
+		// Legacy ConfigManagement sets readiness on reconciler-manager and resource-group-controller,
+		// which isn't reliably cleaned up after tests. Delete both to ensure a fresh
+		// install instead of patching.
+		rmDeployment := k8sobjects.DeploymentObject(
+			core.Name(reconcilermanager.ManagerName),
+			core.Namespace(configsync.ControllerNamespace),
+		)
+		rgDeployment := k8sobjects.DeploymentObject(
+			core.Name(configmanagement.RGControllerName),
+			core.Namespace(configmanagement.RGControllerNamespace),
+		)
+		if err := nt.KubeClient.Delete(rmDeployment); err != nil && !apierrors.IsNotFound(err) {
+			nt.T.Error(err)
+		}
+		if err := nt.KubeClient.Delete(rgDeployment); err != nil && !apierrors.IsNotFound(err) {
+			nt.T.Error(err)
+		}
 		if err := nomostest.InstallConfigSync(nt); err != nil {
 			nt.T.Fatal(err)
 		}
@@ -1469,6 +1487,23 @@ func TestNomosMigrateMonoRepo(t *testing.T) {
 	nt.T.Cleanup(func() {
 		// Restore state of Config Sync installation after test.
 		// This also emulates upgrading to the current version after migrating
+		// Legacy ConfigManagement sets readiness on reconciler-manager and resource-group-controller,
+		// which isn’t reliably cleaned up after tests. Delete both to ensure a fresh
+		// install instead of patching.
+		rmDeployment := k8sobjects.DeploymentObject(
+			core.Name(reconcilermanager.ManagerName),
+			core.Namespace(configsync.ControllerNamespace),
+		)
+		rgDeployment := k8sobjects.DeploymentObject(
+			core.Name(configmanagement.RGControllerName),
+			core.Namespace(configmanagement.RGControllerNamespace),
+		)
+		if err := nt.KubeClient.Delete(rmDeployment); err != nil && !apierrors.IsNotFound(err) {
+			nt.T.Error(err)
+		}
+		if err := nt.KubeClient.Delete(rgDeployment); err != nil && !apierrors.IsNotFound(err) {
+			nt.T.Error(err)
+		}
 		if err := nomostest.InstallConfigSync(nt); err != nil {
 			nt.T.Fatal(err)
 		}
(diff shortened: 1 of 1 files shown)


e2e/testcases/cli_test.go
```diff
@@ -1301,6 +1302,23 @@ func TestNomosMigrate(t *testing.T) {
...
+			core.Name(configmanagement.RGControllerName),
+			core.Namespace(configmanagement.RGControllerNamespace),
+		)
+		if err := nt.KubeClient.Delete(rmDeployment); err != nil && !apierrors.IsNotFound(err) {
+			nt.T.Error(err)
+		}
+		if err := nt.KubeClient.Delete(rgDeployment); err != nil && !apierrors.IsNotFound(err) {
+			nt.T.Error(err)
+		}
 		if err := nomostest.InstallConfigSync(nt); err != nil {
```
sdowell: Deleting and recreating specific subcomponents is going to be error prone, as it doesn't guarantee the e2e client claims ownership of all fields for all objects which could have drifted. Perhaps using client side apply here when reinstalling would be a more thorough option

tiffanny29631: Mmm without the other PR it's hard to know whether further changes are effective or just false positive, some natural result from the standalone branching. I might need to add some check for the readiness field as well to verify it's actually removed.

sdowell: I'd suggest rebasing that PR on top of this one, then you can validate it works from the tests in the other PR

tiffanny29631:
Note: should only merge if https://github.com/GoogleContainerTools/kpt-config-sync/pull/1791 has stable success presubmit result.

/hold

e2e/nomostest/config_sync.go
```diff
@@ -245,6 +246,34 @@ func InstallConfigSync(nt *NT) error {
...
+
+	nt.T.Logf("Applying multi-repo configmaps from %s", multiConfigMaps)
+	out, err = nt.Shell.Kubectl("apply", "-f", multiConfigMaps)
+	if err != nil {
+		return fmt.Errorf("failed to apply multi-repo configmaps: %v\n%s", err, out)
+	}
+
+	// Apply the admission webhook manifest
+	nt.T.Logf("Applying admission webhook manifest from %s", admissionWebhookManifest)
+	out, err = nt.Shell.Kubectl("apply", "-f", admissionWebhookManifest)
```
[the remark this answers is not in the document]
tiffanny29631: done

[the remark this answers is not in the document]
tiffanny29631: done

e2e/nomostest/config_sync.go
```diff
@@ -245,6 +246,34 @@ func InstallConfigSync(nt *NT) error {
 	return nil
 }
 
+// InstallConfigSyncFromManifest installs ConfigSync on the test cluster by directly
+// applying the manifest file using kubectl client-side apply
+func InstallConfigSyncFromManifest(nt *NT) error {
```
sdowell: This will revert some of the modifications made by the usual [InstallConfigSync](https://github.com/GoogleContainerTools/config-sync/blob/main/e2e/nomostest/config_sync.go#L209) function used by the test scaffolding. Can this be consolidated to use the same function? I expect you could add an option to use `Update` instead of Patch/Apply

tiffanny29631: Could you elaborate on the 'same function'? I assume you were not suggesting merging the two install functions, the KubeClient.Apply does not achieve necessary field management or removes any legacy fields. Is it now `kubectl update` instead of `kubectl apply` that you suggest?

sdowell: I'm suggesting it would use the client-go `Update` (KubeClient.Update) method instead of the client-go `Apply` method (KubeClient.Apply)

tiffanny29631: The go-client update has a few edge cases too.

When working on the configsync manifest, CRD like clusters.clusterregistry.k8s.io disallows update when there's no resourceversion attached. Something like 
```
    cli_test.go:1477: 2025-09-10 05:42:53.626931518 +0000 UTC ERROR: customresourcedefinitions.apiextensions.k8s.io "clusters.clusterregistry.k8s.io" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

The operator also removes the admission webhook, which will need a special handle when restoring the shared test env, a create instead of update. We could also use preventDrift: true when configuring the configmanagement? 

Anyways not sure if the effort is worthy for a deprecated use case test.

tiffanny29631: OK added the resourceVersion, hopefully it solves the issue

sdowell: Yes you have to set the resourceVersion when using update. Are you missing context on working with k8s client libraries? If so I can take over the PR - although I don't think this deliverable is really needed until we solve the upgrade issue

tiffanny29631: Good chance to learn. The latest change seemed to be working.

e2e/nomostest/config_sync.go
```diff
@@ -78,6 +78,16 @@ const (
 	shortSyncPollingPeriod = 5 * time.Second
 )
 
+// InstallMethod defines how Config Sync should be installed
+type InstallMethod string
+
+const (
+	// InstallMethodApply uses server-side apply (default)
```
sdowell, line 85: It doesn't look like this is currently the default. If the intent is to retain default behavior, I'd suggest updating the function signature to accept a variadic list of options (for example `func InstallConfigSync(nt *NT, opts ...InstallConfigSyncOpts) error {`)

e2e/nomostest/config_sync.go
```diff
@@ -230,16 +240,36 @@ func parseConfigSyncManifests(nt *NT) ([]client.Object, error) {
...
 		nt.T.Logf("installConfigSync obj: %v", core.GKNN(o))
-		if err := nt.KubeClient.Apply(o); err != nil {
-			return err
+		switch method {
+		case InstallMethodApply:
+			if err := nt.KubeClient.Apply(o); err != nil {
+				return err
+			}
+		case InstallMethodUpdate:
+			currentObj := o.DeepCopyObject().(client.Object)
```
sdowell: A DeepCopy should not be necessary here for every object, can this be switched to populating an empty object (e.g. `Unstructured`)?

e2e/nomostest/config_sync.go
```diff
@@ -230,16 +240,36 @@ func parseConfigSyncManifests(nt *NT) ([]client.Object, error) {
...
+			if err := nt.KubeClient.Get(currentObj.GetName(), currentObj.GetNamespace(), currentObj); err != nil {
+				if apierrors.IsNotFound(err) {
+					if err := nt.KubeClient.Create(o); err != nil {
+						return err
+					}
+				} else {
+					return err
+				}
+			} else {
+				// Attach existing resourceVersion to the object
```
sdowell: The comment doesn't add much context, I'd suggest explaining "why" it's attaching the `resourceVersion` to the object.

In this instance, the current state of the object on the cluster does not matter and the intent is to fully update to the original configuration declared in the manifest

e2e/nomostest/config_sync.go
tiffanny29631: Done

e2e/nomostest/config_sync.go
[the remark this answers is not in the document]
tiffanny29631: Deepcopy replaced

e2e/nomostest/config_sync.go
tiffanny29631: Comment updated.

e2e/nomostest/config_sync.go
```diff
@@ -230,16 +240,36 @@ func parseConfigSyncManifests(nt *NT) ([]client.Object, error) {
...
+				o.SetResourceVersion(currentObj.GetResourceVersion())
+				if err := nt.KubeClient.Update(o); err != nil {
+					return err
+				}
```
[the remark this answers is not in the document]
tiffanny29631: Given the intent is to fully update the status to the original declaration in manifest, and it's test setup code which runs in controlled environment, the current approach has relatively good balance. The race condition window is small, and if it does occur, failing the test is probably the right behavior rather than silently handling conflicts.

e2e/nomostest/config_sync.go
tiffanny29631: Updated

sdowell:
Hold this until we address the SSA->CSA issue that we discussed (in the interim using SSA in the test will catch regressions)
/hold

PR #1822, opened 2025-08-13
fix: logic of when to print monorepo notice

Camila-B:
* Fixes the logic in `nomos bugreport` such that the monorepo notice is printed only if the ConfigManagement object is present and configured for mono repo mode.
* Adds unit tests to validate the new behavior
* b/422835662

PR #1837, opened 2025-08-20
refactor: Remove monorepo structs

Camila-B:
* This change removes monorepo API Go structs that are no longer used
* Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1822


pkg/api/configmanagement/v1/types.go
```diff
@@ -311,91 +119,6 @@ type NamespaceSelectorList struct {
...
-}
-
-// +genclient
-// +genclient:nonNamespaced
-// +kubebuilder:object:root=true
-// +kubebuilder:resource:scope=Cluster
-// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-
 // Repo holds configuration and status about the Nomos source of truth.
 type Repo struct {
```
sdowell, line 123: Is `Repo` still being used somewhere?

Camila-B: Yes it is currently used in the `nomos init' command https://github.com/GoogleContainerTools/kpt-config-sync/blob/97ad6702f9f11161ba7655f701aa8aecda7044c8/cmd/nomos/initialize/init.go#L100

sdowell: That looks like another chunk of code that can be cleaned up - the `Repo` resource is a legacy monorepo CRD

Camila-B: It seems to be part of hierarchical repos https://github.com/GoogleContainerTools/config-sync/blob/bf82814340d4248168084258399f390f0dc1119d/pkg/validate/fileobjects/tree.go#L37 However, the object doesn't seem to be used anywhere in the code. Am I good to still remove all references to Repo and to update the hierarchical repo doc accordingly? https://cloud.google.com/kubernetes-engine/enterprise/config-sync/docs/concepts/hierarchical-repo

sdowell: We can take a closer look at `Repo` as a separate follow up, I think this warrants some due diligence before removing. Please rebase this change and then we can proceed with review

Camila-B: I agree. I went ahead and rebased this branch

sdowell (approved):
Looks good, thanks for cleaning this up!
/lgtm

PR #1819, opened 2025-08-07
fix: Prevent adding ownership condition for NotFound objects

cowsking:
This commit fixes a bug in the ResourceGroup controller where the ownership condition was occasionally added to the status for objects which were not found.

This happened when the object was cached as NotFound by the FilteredWatcher. The Reconcile loop would then use the cached object and call setResStatus, which incorrectly added the ownershipCondition.

The fix ensures that if the cached status is NotFound, the reconciler skips the call to setResStatus.

A unit test, TestComputeStatus, has been added to verify this behavior.

PR #1839, opened 2025-08-21
test: fix TestStressResourceGroup expectation

sdowell:
A recent change intentionally removed the ownership condition from NotFound objects. This change updates the test expectation to be consistent with the new behavior.

Note that this was not caught in presubmits because the stress tests do not run as presubmits.

See https://github.com/GoogleContainerTools/kpt-config-sync/pull/1819


tiffanny29631 (approved):
/lgtm

PR #1844, opened 2025-08-26
feat: remove legacy monorepo-related code

Camila-B:
* This PR removes most references to the legacy monorepo-related code from the nomos CLI
* Depends on #1822
* b/424602791


cmd/nomos/status/client.go
```diff
@@ -164,67 +164,10 @@ func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace st
 		cs.Error = c.namespaceRepoClusterStatus(ctx, cs, namespace)
 	} else if isOss || (cs.isMulti != nil && *cs.isMulti) {
 		c.multiRepoClusterStatus(ctx, cs)
-	} else {
-		c.monoRepoClusterStatus(ctx, cs)
```
sdowell, line 168: What does the behavior here look like now if someone were to run this against a legacy cluster with monorepo installed?

I think it may be a better UX to surface a warning/error message in this case to help make the user aware of the legacy installation

cmd/nomos/status/status.go
```diff
@@ -148,9 +148,10 @@ func clusterStates(ctx context.Context, clientMap map[string]*ClusterClient) (ma
 			stateMap[name] = unavailableCluster(name)
```
tiffanny29631: This function name is identical to the one in client.go, which could cause confusion. Let's refactor it to be more specific about its purpose here.

cmd/nomos/status/client.go
tiffanny29631: Fly by comment: It eventually gets called in the printStatus where the mono repo notice will be attached,  should be good?

tiffanny29631 (commented):
Mostly lgtm with nit only, as long as the removal of the mono status does not break the current tests.

cmd/nomos/status/status.go
Camila-B: Which function name?

tiffanny29631: Ah I misread, it's clusterStates and clusterStatus, nevermind.

cmd/nomos/status/client.go
Camila-B: Yes that's correct. I added a unit test for `clusterStates` which also validates the behavior

cmd/nomos/status/status_test.go
```diff
@@ -0,0 +1,286 @@
...
+	"kpt.dev/configsync/pkg/api/configsync/v1beta1"
+	"kpt.dev/configsync/pkg/core"
+	"kpt.dev/configsync/pkg/core/k8sobjects"
+	csfake "kpt.dev/configsync/pkg/generated/clientset/versioned/fake"
+	"kpt.dev/configsync/pkg/kinds"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func newTestClusterClient(t *testing.T, k8sObjs []runtime.Object, csObjs []runtime.Object, cmObj *unstructured.Unstructured) *ClusterClient {
```
sdowell, line 48: Rather than 3 separate arguments, how about a single `objs ...runtime.Object` argument?

cmd/nomos/status/client.go
sdowell: Thanks for adding the new unit tests!

I don't think we have complete test coverage for this scenario currently, but it looks like the code would segfault on [this line](https://github.com/GoogleContainerTools/kpt-config-sync/blob/v1.22.0/cmd/nomos/status/status.go#L186) when trying to print the status table.

I'm thinking we could surface a table entry with an informative message/error, and retain the current assumption that a table entry is emitted for each cluster name that is passed in. WDYT?

cmd/nomos/status/status_test.go
Camila-B: The issue with combining all 3 is that `k8sfake.NewClientset` uses its own scheme and doesn't recognize any of the Config Sync types. For example, this is the error message that is surfaced when I try to add all objects to the fake clientset: `panic: no kind is registered for the type v1.Repo in scheme`

cmd/nomos/status/client.go
Camila-B: @sdowell thanks for catching that! I fixed the code so that it handles that case. I also added a new unit test `TestPrintStatus` that validates the behavior with monorepo clusters. Currently, it prints out the monorepo notice at the top of the table.

sdowell: Curious what is the motivation for omitting the monorepo clusters from the table? Ack on it printing the monorepo notice at the top of the table, but I think it would be a preferable UX and more consistent code structure to keep the invariant that exactly one table entry is printed for each cluster that is passed in. WDYT?

What I had in mind was just printing a table entry with a static Status/Error message for each mono repo cluster

Camila-B: I thought that it would be redundant to have a table entry for each monorepo cluster along with the monorepo notice. However, I'm ok with having a table entry for the sake of consistency. I went ahead and updated the code and TestPrintStatus to print an Error in the table for each monorepo cluster

cmd/nomos/status/status.go
```diff
@@ -184,11 +184,14 @@ func printStatus(ctx context.Context, writer *tabwriter.Writer, clientMap map[st
 	// Print status for each cluster.
 	for _, name := range names {
 		state := stateMap[name]
-		if name == currentContext {
-			// Prepend an asterisk for the users' current context
-			state.Ref = "*" + name
+
+		if state != nil {
```
sdowell: This change is no longer needed, right? My understanding is now we are back to the assumption that every cluster state should be valid and non-nil

cmd/nomos/status/client.go
sdowell: Ack SGTM

cmd/nomos/status/client.go
```diff
@@ -165,64 +165,10 @@ func (c *ClusterClient) clusterStatus(ctx context.Context, cluster, namespace st
...
 		cs.status = util.ErrorMsg
-		cs.Error = err.Error()
-		return
-	}
-
-	if len(repoList.Items) == 0 {
-		cs.status = util.UnknownMsg
-		cs.Error = "Repo resource is missing"
-		return
+		cs.Error = "The cluster is still running in the legacy mode"
```
tiffanny29631: "This cluster is running in legacy mono-repo mode, which is no longer supported"

cmd/nomos/status/cluster_state_test.go
```diff
@@ -3073,6 +2993,9 @@ gke_sample-project_europe-west1-b_cluster-2
 		t.Run(tc.name, func(t *testing.T) {
 			var buffer bytes.Buffer
 			name = "root-sync-2"
+			t.Cleanup(func() {
```
tiffanny29631, line 2996: Why resetting the name?

Camila-B: If it's not cleaned up, it affects the results of `TestPrintStatus`. I added a comment to clarify this

cmd/nomos/status/client.go
Camila-B: Done

cmd/nomos/status/status.go
Camila-B: Right. I removed this check

PR #1846, opened 2025-08-27
feat: remove obsolete errors

Camila-B:
Removes errors that are no longer used and marks them as deprecated in `nomoserrors/examples/examples.go`


sdowell (approved):
/lgtm

PR #1852, opened 2025-08-28
test: Use hermetic helm chart for wordpress in TestHydrateHelmComponents

tiffanny29631:
The [Bitnami catalog](https://hub.docker.com/r/bitnami/wordpress) recently announced that it is moving most of its container images to a legacy registry without updating corresponding [charts](https://artifacthub.io/packages/helm/bitnami/wordpress/15.2.35). To prevent future disruptions from these upstream changes, this modifies the `TestHydrateHelmComponents` test to use a local Helm chart.

### DIFF
--- a/e2e/testdata/hydration/compiled-json/helm-components/wordpress/deployment_my-wordpress.json
+++ b/e2e/testdata/hydration/compiled-json/helm-components/wordpress/deployment_my-wordpress.json
@@ -166,7 +166,7 @@
 									}
 								],
 								"envFrom": null,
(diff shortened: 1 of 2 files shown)


tiffanny29631:
We might want to keep a copy of the container image as well in case of similar refactors in the future.

mikebz:
I would say that it's a step towards more test coverage as is..  so my preference is to merge this and possibly do another PR if we want to have our own copy.

sdowell (approved):
/lgtm

PR #1870, opened 2025-09-09
fix: Update import path and dependencies

cowsking:
This change replaces the kpt.dev/configsync vanity URL with the direct GitHub path github.com/GoogleContainerTools/config-sync in the post-sync example. This is necessary to ensure that the example can be built independently of the main Config Sync repository.

Additionally, this change updates the Go version and dependencies in the post-sync example to align with the latest versions used in the main repository.


examples/post-sync/go.mod
```diff
@@ -1,48 +1,40 @@
...
+module github.com/GoogleContainerTools/config-sync/examples/post-sync
 
-go 1.23.0
+go 1.24.0
+
+toolchain go1.24.4
 
 require (
-	github.com/go-logr/logr v1.4.2
+	github.com/GoogleContainerTools/config-sync v1.21.0
```
tiffanny29631: Does this work? I thought the import would be `kpt.dev/configsync` as in https://pkg.go.dev/kpt.dev/configsync

tiffanny29631:
To properly test this, the e2e test image should be updated as well. Eventually, we will need to update the image reference in https://github.com/GoogleContainerTools/config-sync/blob/main/e2e/nomostest/sync-status-watch-controller.go#L44 and have the e2e tests use the updated version. The build rule is located at https://github.com/GoogleContainerTools/config-sync/blob/main/Makefile.build#L340.

Please note that to conserve resources, there is no automated build for this standalone controller; it is intended to be built manually only when updates are made.

examples/post-sync/go.mod
cowsking: It should work. We are using GitHub because we are removing the kpt.dev branding

Camila-B:
@cowsking are you able to run `make build-local`? I checked out the PR and tried to run it but got an error

cowsking:
> @cowsking are you able to run `make build-local`? I checked out the PR and tried to run it but got an error

It gave me the message: *** No rule to make target 'build-local'. Stop. Did you encounter the same error?

Camila-B:
> > @cowsking are you able to run `make build-local`? I checked out the PR and tried to run it but got an error
> 
> It gave me the message: *** No rule to make target 'build-local'. Stop. Did you encounter the same error?

This is the make rule that I was referring to https://github.com/GoogleContainerTools/config-sync/blob/04033ef810fd478a4e283cb23c14f6b279d4eee7/examples/post-sync/Makefile#L36

PR #1873, opened 2025-09-11
fix: improve `nomos status` output when no sync objects are found

Camila-B:
This change refactors output of `nomos status` to provide a more accurate output when no RootSync and/or RepoSyncs are found on the cluster.

* Removes UNKNOWN status when no RootSync and/or RepoSync is found on the cluster but Config Sync is running
* Removes "No RootSync resources found" error when a RepoSync exists on the cluster and vice versa
* Replaces error messages with concrete errors to improve the overall error handling logic
* Adds additional test cases to validate the error handling and output of `nomos status`

### DIFF
--- a/cmd/nomos/status/client.go
+++ b/cmd/nomos/status/client.go
@@ -17,6 +17,7 @@ package status
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"sort"
@@ -35,6 +36,7 @@ import (
 	typedv1 "github.com/GoogleContainerTools/config-sync/pkg/generated/clientset/versioned/typed/configmanagement/v1"
 	"github.com/GoogleContainerTools/config-sync/pkg/kinds"
 	"github.com/Masterminds/semver"
+	"go.uber.org/multierr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


cmd/nomos/status/status_test.go
```diff
@@ -150,23 +149,54 @@ func repoSyncObj(ns, name string) *v1beta1.RepoSync {
 	return repoSyncObj
 }
 
+func resourceGroupObject(name, ns string) *unstructured.Unstructured {
```
sdowell, line 152: Is there a reason this returns unstructured instead of a typed object? It would be preferable to use the typed object

cmd/nomos/status/client.go
```diff
@@ -188,116 +205,118 @@ func (c *ClusterClient) syncingConditionSupported(ctx context.Context) bool {
...
 		sort.Slice(repos, func(i, j int) bool {
 			return repos[i].scope < repos[j].scope || (repos[i].scope == repos[j].scope && repos[i].syncName < repos[j].syncName)
 		})
 		cs.repos = append(cs.repos, repos...)
+	} else if errs == nil && len(syncs) == 0 {
+		errs = ErrNoRepoSyncsFound
 	}
 
-	if len(errs) > 0 {
+	if errs != nil && !errors.Is(errs, ErrNoRepoSyncsFound) {
```
sdowell: I think the logic here can be simplified, e.g.
```golang
if errs != nil {
  cs.status = util.ErrorMsg
} else if len(syncs) == 0 {
  errs = ErrNoRepoSyncsFound
}
```

Similar logic can be applied above to the other function

cmd/nomos/status/client.go
```diff
@@ -188,116 +205,118 @@ func (c *ClusterClient) syncingConditionSupported(ctx context.Context) bool {
...
 		cs.status = util.ErrorMsg
-		repoErr = strings.Join(errs, ", ")
-	} else if len(cs.repos) == 0 {
-		cs.status = util.UnknownMsg
```
sdowell, line 296: Just curious, what is the motivation for removing the "Unknown" status?

cmd/nomos/status/status_test.go
Camila-B: Currently, nomos status uses Unstructured for ResourceGroups. https://github.com/GoogleContainerTools/config-sync/blob/8c2fa436678613949ba35d1917f4837c82317577/cmd/nomos/status/client.go#L116 I think moving from using unstructured to `v1alpha1.ResourceGroup` could be done in a followup PR

cmd/nomos/status/client.go
Camila-B: Done

cmd/nomos/status/client.go
Camila-B: With the previous implementation, the ClusterState status was being set to UNKNOWN even when no RootSyncs were found but a RepoSync was and vice versa. For the situation where no RootSyncs or RepoSyncs are found on the cluster, I also removed the UNKNOWN status for parity with the Config Sync UI b/421014597. However, I'm ok with going back to showing an UNKNOWN status in this specific situation

cmd/nomos/status/status_test.go
sdowell: Thanks for clarifying, would you mind following up with that refactor?

cmd/nomos/status/client.go
sdowell: Ack, thank you for clarifying

PR #1875, opened 2025-09-16
test: log ResourceGroup diff on unexpected updates

sdowell:
This change updates TestResourceGroupApplyStatus to always log the diff when an unexpected ResourceGroup update happens. This test has been somewhat flaky and this additional logging is intended to help understand the root cause of the test flake.


tiffanny29631 (approved):
/lgtm

PR #1877, opened 2025-09-17
test: fix flake in TestResourceGroupApplyStatus

sdowell:
After adding the diff logging, it looks like the root cause for this test flake is due to a watch event ordering issue. Namely, the test begins waiting for the resourceVersion to not change before the r-g-c has actually updated the status. The fix is to add a resourceVersion check to the prior wait condition, such that it does not pass prematurely on a stale watch event.


sdowell:
Holding while I experiment with another approach
/hold

PR #1883, opened 2025-09-23
feat: native ignore-mutation support in applier

sdowell:
This change reimplements the ignore-mutation feature using the native
support in the cli-utils applier. This removes the need for the
workaround of caching the object in Config Sync and passing in the
cached object to the applier.

This fixes several race conditions and side effects associated with the
caching solution.

Depends on new functionality in cli-utils: https://github.com/kubernetes-sigs/cli-utils/commit/5895ad6c17dd06b99c1e5c8af17e2b12a3504fad

Fixes: https://github.com/GoogleContainerTools/config-sync/issues/1668


vendor/sigs.k8s.io/cli-utils/pkg/apply/filter/prevent-update-filter.go
```diff
@@ -0,0 +1,83 @@
...
+		} else if err != nil { // unexpected error - fatal
+			return NewFatalError(fmt.Errorf("failed to get current object from cluster: %w", err))
+		}
+		// Object exists - skip apply
+		return &AnnotationPreventedUpdateError{
+			Annotation: common.LifecycleMutationAnnotation,
+			Value:      common.IgnoreMutation,
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
sdowell: It does return the error for other potential errors

pkg/applier/applier.go
```diff
@@ -317,11 +308,52 @@ func (s *supervisor) handleApplySkippedEvent(obj *unstructured.Unstructured, id
...
+	if err := s.clientSet.Client.Get(ctx, key, metaObj); err != nil {
+		return err
+	}
+
+	existing := metaObj.DeepCopy()
+	if metadata.UpdateConfigSyncMetadata(obj, metaObj) {
+		if err := s.clientSet.Client.Patch(ctx, metaObj, client.MergeFrom(existing),
+			client.FieldOwner(configsync.FieldManager)); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
sdowell: It does handle the error

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -340,16 +344,21 @@ func (r *reconciler) computeStatus(
 
 			// Update the reconcile status based on the Strategy & Actuation
 			// from the last apply attempt, and the newly computed kstatus.
-			if reconcile, err := UpdateReconcileStatusToReflectKstatus(resStatus); err != nil {
+			if reconcile, err := UpdateReconcileStatusToReflectKstatus(resStatus, ignoreMutation); err != nil {
```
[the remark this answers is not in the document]
sdowell: It does pass in the ignoreMutation flag

pkg/applier/applier.go
```diff
@@ -317,11 +308,52 @@ func (s *supervisor) handleApplySkippedEvent(obj *unstructured.Unstructured, id
 		return KptManagementConflictError(obj)
 	}
 
+	var annotationErr *filter.AnnotationPreventedUpdateError
+	if errors.As(err, &annotationErr) {
+		// For applies this is desired behavior, not unexpected. The following logic
+		// re-applies just the CS metadata to ensure metadata does not drift.
+		klog.Info("Got AnnotationPreventedUpdateError")
+		if err := s.updateObjectMetadata(context.TODO(), obj); err != nil {
```
tiffanny29631: Should a context get passed down here?

pkg/applier/applier.go
```diff
@@ -317,11 +308,52 @@ func (s *supervisor) handleApplySkippedEvent(obj *unstructured.Unstructured, id
...
+				fmt.Errorf("updating Config Sync metadata for ignore mutation object: %w", err),
+				id, actuation.ActuationStrategyApply)
+		}
+		return nil
+	}
+
 	return SkipErrorForResource(err, id, actuation.ActuationStrategyApply)
 }
 
+func (s *supervisor) updateObjectMetadata(ctx context.Context, obj *unstructured.Unstructured) error {
```
tiffanny29631, line 327: Is the current unit test sufficient or updateObjectMetadata worths its own unit test, i.e. when patching fails or no drift? I assume there are apply tests in place in unit and e2e to ensure the behavior does not change with the removal of cache, could you double check on the prune?

pkg/applier/applier.go
```diff
@@ -317,11 +308,52 @@ func (s *supervisor) handleApplySkippedEvent(obj *unstructured.Unstructured, id
...
+	// Using PartialObjectMetadata optimizes the client calls (uses metadata client under the hood)
+	metaObj := &metav1.PartialObjectMetadata{}
+	metaObj.Name = obj.GetName()
+	metaObj.Namespace = obj.GetNamespace()
+	metaObj.SetGroupVersionKind(obj.GroupVersionKind())
+	key := client.ObjectKey{
+		Name:      obj.GetName(),
+		Namespace: obj.GetNamespace(),
+	}
+	if err := s.clientSet.Client.Get(ctx, key, metaObj); err != nil {
```
tiffanny29631, line 337: Will IsNotFound error be a case or it rarely happens at this stage?

pkg/applier/applier.go
sdowell: Updated to propagate the context

pkg/applier/applier.go
sdowell: Added another unit test case to cover this scenario.

This behavior is thoroughly covered by e2e tests, so that should give confidence that this doesn't break the existing functionality. This change shouldn't impact prune behavior but that also has test coverage

pkg/applier/applier.go
sdowell: That can happen but is generally unexpected - that means the object was deleted after the applier looked up the object but before this code processes the ApplySkipped event.

In that scenario it should be preferable to treat it as an error - it will trigger a re-invocation of the applier which will then recreate the object (if it still does not exist).

tiffanny29631:
Nit: suggest linking the change from cli-utils in commit message. Otherwise lgtm.

sdowell:
> Nit: suggest linking the change from cli-utils in commit message. Otherwise lgtm.

Added a link in the PR description

tiffanny29631 (approved):
/lgtm

PR #1887, opened 2025-09-26
ci: Create Helm chart for E2E tests

Camila-B:
* Creates a minimal Helm chart to be used by E2E tests 
* Creates a Make target which packages and uploads the chart to the publicly accessible GCS bucket `kpt-config-sync-ci-test-helm-charts` (that already exists)


Makefile.oss.prow
```diff
@@ -59,6 +59,10 @@ push-test-oci-images-public: "$(CRANE)"
 	cd $(KUSTOMIZE_COMPONENTS_DIR) && crane append -f <(tar -f - -c .) -t $(KUSTOMIZE_COMPONENTS_PUBLIC_GCR_IMAGE)
 	cd $(KUSTOMIZE_COMPONENTS_DIR) && crane append -f <(tar -f - -c .) -t $(KUSTOMIZE_COMPONENTS_PUBLIC_AR_IMAGE)
 
+.PHONY: push-test-helm-chart-public
+push-test-helm-chart-public: "$(HELM)"
```
tiffanny29631, line 63: What is the intended lifecycle for the test Helm charts? Do you plan for them to be built and published automatically whenever changes are made, or will they be managed and published in a different way?

e2e/testdata/helm-charts/simple-pause/v1.1.0/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A simple test chart to be used for Config Sync E2E tests.
+name: simple-pause
+version: 1.1.0
```
tiffanny29631, line 5: 1.0.1?

scripts/package-test-helm-chart.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Ensures the temporary directory is cleaned up on exit
+trap cleanup EXIT
+
+# Copy chart source to the temporary directory.
+cp -R "${CHART_SRC_DIR}" "${TMP_DIR}/"
+cd "${TMP_DIR}"
+
+echo "Packaging Helm chart in ${TMP_DIR}"
+helm package ./${CHART_NAME}/*
```
tiffanny29631, line 38: How will versioning of the test Helm charts be handled to prevent overwriting existing versions in the chart repository? Is there automated logic in place to detect and prevent re-publishing the same chart version, or to remind contributors to bump the version in Chart.yaml on every change?

Makefile.oss.prow
Camila-B: I don't expect there to be a need to modify the Helm charts. They should be uploaded once to the bucket and remain unchanged. However with this make rule, it would be a manual process to package and re-upload them as needed.

e2e/testdata/helm-charts/simple-pause/v1.1.0/Chart.yaml
Camila-B: Oh that was before I changed the version to 1.1.0. I'll go ahead and change the directory name to match

scripts/package-test-helm-chart.sh
Camila-B: I added the `-n` flag to prevent existing versions of the chart from getting overwritten. If a new version needs to be created for whatever reason, a new directory with the chart's configs can be created

Makefile.oss.prow
tiffanny29631: Ack

scripts/package-test-helm-chart.sh
tiffanny29631: Sounds good

PR #1888, opened 2025-09-26
test: use own remote chart for Helm tests

Camila-B:
Depends on: https://github.com/GoogleContainerTools/config-sync/pull/1887
* Replaces all references to the Bitnami Wordpress Helm chart with our own self hosted chart while maintaining a similar level of test coverage

### DIFF
--- a/e2e/testcases/helm_sync_test.go
+++ b/e2e/testcases/helm_sync_test.go
@@ -61,14 +61,13 @@ const (
 )
 
 // TestPublicHelm can run on both Kind and GKE clusters.
-// It tests Config Sync can pull from public Helm repo without any authentication.
+// It tests Config Sync can pull from a public Helm repo without any authentication.
 func TestPublicHelm(t *testing.T) {
-	t.Skip("Temporarily skipping failing test")
 	nt := nomostest.New(t, nomostesting.SyncSource,
 		ntopts.SyncWithGitSource(nomostest.DefaultRootSyncID, ntopts.Unstructured))
 
-	rs := rootSyncForWordpressHelmChart(nt, nil)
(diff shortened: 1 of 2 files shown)


e2e/testcases/helm_sync_test.go
```diff
@@ -959,12 +916,12 @@ func rootSyncForWordpressHelmChart(nt *nomostest.NT, valuesMutator func(map[stri
...
 		HelmBase: v1beta1.HelmBase{
-			Repo:        "https://charts.bitnami.com/bitnami",
+			Repo:        nomostesting.ConfigSyncTestHelmChartsAddress,
 			Chart:       chartID.Name,
 			Version:     chartID.Version,
-			ReleaseName: "my-wordpress",
+			ReleaseName: "my-simple-pause",
 			Auth:        configsync.AuthNone,
 			Values: &apiextensionsv1.JSON{
 				Raw: out,
```
sdowell, line 927: nit: I expect this test should no longer face memory issues given the new minimal helm repo, so the helm-sync memory increase below can probably be adjusted/removed. That can be done as a follow up if you prefer

Camila-B: I agree. I'll do it as a followup

e2e/testcases/helm_sync_test.go
```diff
@@ -242,24 +236,18 @@ image:
 	cm2 := k8sobjects.ConfigMapObject(core.Name(cmName), core.Namespace(configsync.ControllerNamespace))
```
tiffanny29631, line 236: nit: Test loggings can be updated to reflect the change

PR #1896, opened 2025-10-06
test: add Must to Validate calls

Camila-B:
This change replaces all calls to Validate in managed_resources_test.go with nt.Must(nt.Validate())

### DIFF
--- a/e2e/testcases/managed_resources_test.go
+++ b/e2e/testcases/managed_resources_test.go
@@ -39,6 +39,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// driftDetectionTimeout is the timeout to use when waiting for the remediator to act
+	driftDetectionTimeout = 5 * time.Second
+)
+
 // This file includes tests for drift correction and drift prevention.
 //
 // Drift Correction uses the following metadata:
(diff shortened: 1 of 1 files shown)


e2e/testcases/managed_resources_test.go
```diff
@@ -118,21 +118,16 @@ func TestDriftKubectlApplyClusterScoped(t *testing.T) {
...
+		testwatcher.WatchPredicates(testpredicates.HasExactlyAnnotationKeys(
 			metadata.ManagementModeAnnotationKey,
 			metadata.ResourceIDKey,
 			metadata.ResourceManagerKey,
-			corev1.LastAppliedConfigAnnotation))
-	if err != nil {
-		nt.T.Fatal(err)
-	}
+			corev1.LastAppliedConfigAnnotation)),
+		testwatcher.WatchTimeout(10*time.Second)))
```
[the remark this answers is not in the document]
sdowell: Change looks good but I'd agree with this suggestion for defining a constant

e2e/testcases/managed_resources_test.go
```diff
@@ -118,21 +123,16 @@ func TestDriftKubectlApplyClusterScoped(t *testing.T) {
...
 
-	// Wait 5 seconds so that the remediator can process the event.
-	time.Sleep(5 * time.Second)
-
 	// Remediator SHOULD NOT delete the namespace, since its
 	// `configsync.gke.io/manager` annotation indicates it is managed by a
 	// RootSync that does not exist.
-	err = nt.Validate("test-ns2", "", &corev1.Namespace{},
-		testpredicates.HasExactlyAnnotationKeys(
+	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), "test-ns2", "",
```
sdowell: On closer inspection this actually isn't exactly verifying the same thing as before.

What the previous assertion does is verify that the remediator *does not do X* (in this case, delete the namespace). What the new assertion does is just verify the Namespace exists, which in most cases which will just return immediately (even if the remediator deletes the namespace 1 second later). To maintain the behavior from before the assertion would need to be inverted, for example:

- `WatchForNotFound` for X seconds
- If `WatchForNotFound` returns nil, fail the test because the object was deleted by the remediator
- If `WatchForNotFound` times out, succeed because the remediator did not delete the object

The only optimization here would be in the case that there's a bug in the remediator and the object gets deleted, in which case the assertion would return a little sooner. TBH it's not clear whether this is really an improvement, the tradeoff is that the failure case exits sooner at the cost of added code complexity

Camila-B: That makes sense. I reverted back to using `sleep` and `Validate` and added `Must` around each call to`Validate` instead

e2e/testcases/managed_resources_test.go
```diff
@@ -708,7 +689,7 @@ func TestDriftRemoveApplySetPartOfLabel(t *testing.T) {
 	nt.Must(nt.Watcher.WatchObject(kinds.Namespace(), namespace, "", testwatcher.WatchPredicates(
 		testpredicates.HasLabel(metadata.ApplySetPartOfLabel, rootSync1ApplySetID),
 		testpredicates.HasAnnotation("season", "summer"),
-	), testwatcher.WatchTimeout(10*time.Second)))
+	), testwatcher.WatchTimeout(driftDetectionTimeout)))
```
sdowell, line 692: nit: This is reducing the timeout and it's not clear that it's intentional. Will this make the test more flaky with a shorter timeout?

Camila-B: I don't believe so. It seems that the value of 10 seconds was set arbitrary for these 2 test cases and not because it actually requires up to 10 seconds for the remediator to act

sdowell (approved):
/lgtm

PR #1898, opened 2025-10-06
refactor: Use ResourceGroup type instead of Unstructured

Camila-B:
This change replaces all uses of Unstructured with ResourceGroup in the nomos CLI. A separate change will replace such references in the Applier


pkg/core/k8sobjects/resourcegroup.go
```diff
@@ -15,16 +15,31 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ResourceGroupObject initializes a ResourceGroup.
-func ResourceGroupObject(opts ...core.MetaMutator) *unstructured.Unstructured {
-	result := resourcegroup.Unstructured("", "", "")
-	defaultMutate(result)
+// ResourceGroupObjectV1Alpha1 initializes a kptv1alpha1.ResourceGroup.
+func ResourceGroupObjectV1Alpha1(ns, name string, opts ...core.MetaMutator) *kptv1alpha1.ResourceGroup {
```
sdowell: nit: `v1alpha1` is the only version currently so this could probably stay as `ResourceGroupObject`

Camila-B: Ack. I renamed it back to ResourceGroupObject

cmd/nomos/status/cluster_state_test.go
```diff
@@ -1438,13 +1436,13 @@ func TestRepoState_NamespaceRepoStatus(t *testing.T) {
...
 				commit:            "abc123",
-				resources:         exampleResources(""),
+				resources:         exampleResourceStatuses(""),
 			},
 		},
 		{
 			name:                      "[accurate status with syncing condition supported] repo is synced",
 			gitSpec:                   git,
-			resourceGroup:             k8sobjects.ResourceGroupObject(core.Namespace("bookstore"), core.Name("repo-sync"), withResourcesAndCommit("abc123")),
+			resourceGroup:             k8sobjects.ResourceGroupObject("bookstore", "repo-sync", k8sobjects.WithRGResourceStatuses(exampleResourceStatuses("abc123")...)), //k8sobjects.ResourceGroupObject(core.Namespace("bookstore"), core.Name("repo-sync"), withResourcesAndCommit("abc123")),
```
sdowell: nit: clean up commented out code

cmd/nomos/status/status_test.go
```diff
@@ -149,32 +150,22 @@ func repoSyncObject(ns, name string) *v1beta1.RepoSync {
 	return repoSyncObj
 }
 
-func resourceGroupObject(name, ns string) *unstructured.Unstructured {
-	rg := k8sobjects.ResourceGroupObject(core.Name(name), core.Namespace(ns))
-	resources := []interface{}{map[string]interface{}{
```
sdowell, line 154: The prior implementation also set the spec fields

cmd/nomos/status/cluster_state_test.go
Camila-B: Done

cmd/nomos/status/status_test.go
Camila-B: Ack. I added the resource back to the spec

PR #1901, opened 2025-10-07
test: update check for project marked for deletion

Camila-B:
This is a followup to https://github.com/GoogleContainerTools/config-sync/pull/1824 which adds another similar error message that can occur when a project is marked for deletion

### DIFF
--- a/e2e/nomostest/gitproviders/gitlab.go
+++ b/e2e/nomostest/gitproviders/gitlab.go
@@ -161,7 +161,7 @@ func (g *GitlabClient) DeleteRepositories(names ...string) error {
 				continue
 			}
 			// Check if project is already marked for deletion (this is also a success case)
-			if strings.Contains(response, "Project has been already marked for deletion") {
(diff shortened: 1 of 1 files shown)


e2e/nomostest/gitproviders/gitlab.go
```diff
@@ -255,3 +255,10 @@ func (g *GitlabClient) GetObsoleteRepos() ([]string, error) {
 
 	return result, nil
 }
+
+// projectMarkedForDeletion checks if the GitLab API response indicates that
+// the project has already been marked for deletion
+func projectMarkedForDeletion(response string) bool {
+	return strings.Contains(response, "Project has been already marked for deletion") ||
```
sdowell: So both of these are possible response bodies from the gitlab API?

Camila-B: Upon further consideration, I think the previous error message has been updated to use the new phrasing. I updated the code to reflect this

sdowell (approved):
/lgtm

PR #1907, opened 2025-10-14
test: Split SyncSource feature

Camila-B:
This splits the SyncSource feature into SyncSourceGit, SyncSourceOCI and SyncSourceHelm respectively

### DIFF
--- a/Makefile.e2e
+++ b/Makefile.e2e
@@ -91,7 +91,7 @@ test-e2e-kind-test-group1:
 # This target runs the second group of e2e tests.
 .PHONY: test-e2e-kind-test-group2
 test-e2e-kind-test-group2:
-	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=sync-source,reconciliation-1" test-e2e-kind
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=sync-source-git,sync-source-helm,sync-source-oci,reconciliation-1" test-e2e-kind
 
 # This target runs the third group of e2e tests.
 .PHONY: test-e2e-kind-test-group3
--- a/e2e/nomostest/testing/feature.go
+++ b/e2e/nomostest/testing/feature.go
@@ -38,8 +38,12 @@ const (
 	Reconciliation1 = "reconciliation-1"
 	// Reconciliation2 verifies the second part of the reconciliation test.
 	Reconciliation2 = "reconciliation-2"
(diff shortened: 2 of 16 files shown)


Makefile.e2e
```diff
@@ -86,14 +86,14 @@ test-e2e-kind: config-sync-manifest-local build-kind-e2e
...
 .PHONY: test-e2e-kind-test-group1
 test-e2e-kind-test-group1:
-	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=acm-controller,selector,lifecycle,nomos-cli,drift-control" test-e2e-kind
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=sync-source-oci,acm-controller,selector,lifecycle,nomos-cli,drift-control" test-e2e-kind
 
 # This target runs the second group of e2e tests.
 .PHONY: test-e2e-kind-test-group2
 test-e2e-kind-test-group2:
-	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=sync-source,reconciliation-1" test-e2e-kind
+	$(MAKE) E2E_ARGS="$(E2E_ARGS) --test-features=sync-source-git,reconciliation-1" test-e2e-kind
```
sdowell: minor nit: I'd suggest just keeping all three in `group2` so that we aren't restructuring which tests run in which presubmit job in this change.

FYI the primary factor of splitting the groups here is to balance the execution time of each test subset. Ideally we want each presubmit test group to complete in a similar amount of time.

Camila-B: Sounds good, done

e2e/nomostest/testing/feature.go
```diff
@@ -38,8 +38,12 @@ const (
 	Reconciliation1 = "reconciliation-1"
 	// Reconciliation2 verifies the second part of the reconciliation test.
 	Reconciliation2 = "reconciliation-2"
-	// SyncSource verifies the various source configs, including different source types, sync directory, branch, and etc.
-	SyncSource = "sync-source"
+	// SyncSourceGit verifies syncing from a Git repository
+	SyncSourceGit = "sync-source-git"
```
sdowell, line 42: I went through all the test cases to see if there's anything else that should fall into this bucket -> I'd also suggest classifying `TestMultipleRemoteBranchesOutOfSync` as `sync-source-git`

Camila-B: Thanks! That makes more sense than classifying it as `nomostesting.ACMController`

PR #1908, opened 2025-10-15
test: reduce flakiness of TestStressLargeRequest

sdowell:
This test can be flaky on autopilot, where it times out waiting for the large number of objects to be synced. This change does two things to reduce the flakiness:
- Increases the CPU request/limit. The reconciler currently has the default of ~700m, and from the cluster metrics is currently being CPU throttled.
- Increases the sync timeout from 30 minutes to 40 minutes. On successful runs it usually completes in around 25 minutes. This adds a bit more headroom.


Camila-B (approved):
/lgtm

PR #1912, opened 2025-10-23
test: Reuse Bitbucket repos

Camila-B:
* Creates new repos based on the test cluster name and RSync namespace and name, similar to what is done by other gitproviders
* Removes the deletion of Bitbucket repos since they can now be reused
* Replaces the use of curl with the http package

### DIFF
--- a/e2e/nomostest/gitproviders/bitbucket.go
+++ b/e2e/nomostest/gitproviders/bitbucket.go
@@ -45,20 +45,20 @@ type BitbucketClient struct {
 	refreshToken string
 	logger       *testlogger.TestLogger
 	workspace    string
-	// repoPrefix is used to avoid overlap
-	repoPrefix string
+	// repoSuffix is used to avoid overlap
+	repoSuffix string
 	httpClient *http.Client
 }
 
 // newBitbucketClient instantiates a new Bitbucket client.
-func newBitbucketClient(repoPrefix string, logger *testlogger.TestLogger) (*BitbucketClient, error) {
+func newBitbucketClient(repoSuffix string, logger *testlogger.TestLogger) (*BitbucketClient, error) {
 	if *e2e.BitbucketWorkspace == "" {
 		return nil, errors.New("bitbucket workspace cannot be empty; set with -bitbucket-workspace flag or E2E_BITBUCKET_WORKSPACE env var")
 	}
 	client := &BitbucketClient{
 		logger:     logger,
 		workspace:  *e2e.BitbucketWorkspace,
-		repoPrefix: repoPrefix,
+		repoSuffix: repoSuffix,
 		httpClient: &http.Client{
 			Timeout: 10 * time.Second,
 		},
@@ -78,7 +78,7 @@ func newBitbucketClient(repoPrefix string, logger *testlogger.TestLogger) (*Bitb
 }
 
 func (b *BitbucketClient) fullName(name string) string {
-	return util.SanitizeBitbucketRepoName(b.repoPrefix, name)
+	return util.SanitizeBitbucketRepoName(b.repoSuffix, name)
 }
 
 // Type returns the provider type.
@@ -220,7 +220,7 @@ func (b *BitbucketClient) refreshAccessToken() (string, error) {
 		return "", fmt.Errorf("reading token refresh response body: %w", err)
 	}
 	if resp.StatusCode != http.StatusOK {
-		return "", fmt.Errorf("refresh token failed with status %d: %s", resp.StatusCode, string(body))
+		return "", fmt.Errorf("refresh token failed with status %d", resp.StatusCode)
 	}
 
 	var output map[string]interface{}
@@ -230,12 +230,12 @@ func (b *BitbucketClient) refreshAccessToken() (string, error) {
 
 	accessToken, ok := output["access_token"]
 	if !ok {
-		return "", fmt.Errorf("no access_token in response: %s", string(body))
+		return "", fmt.Errorf("no access_token in response")
 	}
 
 	accessTokenStr, ok := accessToken.(string)
 	if !ok {
-		return "", fmt.Errorf("access_token is not a string: %v", accessToken)
+		return "", fmt.Errorf("access_token is not a string: %T", accessToken)
 	}
 
 	return accessTokenStr, nil
--- a/e2e/nomostest/gitproviders/git-provider.go
+++ b/e2e/nomostest/gitproviders/git-provider.go
@@ -51,7 +51,8 @@ type GitProvider interface {
 func NewGitProvider(t testing.NTB, provider, clusterName string, logger *testlogger.TestLogger, shell *testshell.TestShell) GitProvider {
 	switch provider {
 	case e2e.Bitbucket:
-		client, err := newBitbucketClient(clusterName, logger)
+		repoSuffix := *e2e.GCPProject + "/" + clusterName
+		client, err := newBitbucketClient(repoSuffix, logger)
 		if err != nil {
 			t.Fatal(err)
 		}
--- a/e2e/nomostest/gitproviders/util/reponame.go
+++ b/e2e/nomostest/gitproviders/util/reponame.go
@@ -27,16 +27,40 @@ const (
 	repoNameHashLen         = 8
 )
 
-// SanitizeRepoName replaces all slashes with hyphens, and truncates the name.
+// SanitizeGCPRepoName replaces all slashes with hyphens, and truncates the name.
 // repo name may contain between 3 and 63 lowercase letters, digits and hyphens.
-func SanitizeRepoName(repoPrefix, name string) string {
-	return sanitize(repoPrefix, name, defaultRepoNameMaxLen)
+// This is used by CSR and SSM which are per-project resources
+// The repo name will be of the form cs-e2e-<repoPrefix>-<name>-<hash>
+func SanitizeGCPRepoName(repoPrefix, name string) string {
+	if name == "" {
+		return name // Requires at least the base repo name
+	}
+	fullName := "cs-e2e"
+
+	if repoPrefix != "" {
+		fullName += "-" + repoPrefix
+	}
+	fullName += "-" + name
+	hashStr := hashName(fullName)
+
+	return sanitize(fullName, hashStr, defaultRepoNameMaxLen)
 }
 
 // SanitizeBitbucketRepoName replaces all slashes with hyphens, and truncates the name for Bitbucket.
 // repo name may contain between 3 and 62 lowercase letters, digits and hyphens.
-func SanitizeBitbucketRepoName(repoPrefix, name string) string {
-	return sanitize(repoPrefix, name, bitbucketRepoNameMaxLen)
+// The repo name will be of the form <name>-<repoSuffix>-<hash>
+func SanitizeBitbucketRepoName(repoSuffix, name string) string {
+	if name == "" {
+		return name // Requires at least the base repo name
+	}
+
+	fullName := name
+	if repoSuffix != "" {
+		fullName += "-" + repoSuffix
+	}
+	hashStr := hashName(fullName)
+
+	return sanitize(fullName, hashStr, bitbucketRepoNameMaxLen)
 }
 
 func hashName(fullName string) string {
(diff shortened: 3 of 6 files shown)


e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -93,154 +97,155 @@ func (b *BitbucketClient) SyncURL(name string) string {
...
-			return b.deleteSingleRepo(accessToken, name)
-		})
-		if err != nil {
-			// Skip repositories that are currently unavailable
-			if strings.Contains(err.Error(), "Repository currently not available") {
-				fmt.Printf("[BITBUCKET] Skipping repository %s: %v\n", name, err)
-				continue
-			}
-			errs = multierr.Append(errs, err)
+	if resp.StatusCode == http.StatusOK { //Already exists
```
sdowell: In the case that the repo already exists, is there logic in place to ensure a clean initial state of the repository? For example a hard reset / force push to a known initial state

Camila-B: It seems that the logic already exists in `setupTestCase` with the call to `initRepository` and `initialCommit` 

https://github.com/GoogleContainerTools/config-sync/blob/0694e5d29ac5e9feba9fb8c51c4d1f733d0325cc/e2e/nomostest/new.go#L450

https://github.com/GoogleContainerTools/config-sync/blob/0694e5d29ac5e9feba9fb8c51c4d1f733d0325cc/e2e/nomostest/new.go#L484

e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -93,154 +97,155 @@ func (b *BitbucketClient) SyncURL(name string) string {
...
-	out, err := exec.Command("curl", "-sX", "POST",
-		"-H", "Content-Type: application/json",
-		"-H", fmt.Sprintf("Authorization:Bearer %s", accessToken),
-		"-d", fmt.Sprintf("{\"scm\": \"git\",\"project\": {\"key\": \"%s\"},\"is_private\": \"true\"}", bitbucketProject),
-		fmt.Sprintf("https://api.bitbucket.org/2.0/repositories/%s/%s", b.workspace, repoName)).CombinedOutput()
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
 
+	req, err := http.NewRequest(http.MethodGet, repoURL, nil)
```
sdowell: minor nit: the http request logic could be broken out to a helper function to improve the flow/readability

e2e/nomostest/gitproviders/util/reponame.go
```diff
@@ -22,19 +22,35 @@ import (
...
+	return sanitize(repoPrefix, name, bitbucketRepoNameMaxLen, "")
+}
+
+func hashName(fullName string) string {
 	hashBytes := sha1.Sum([]byte(fullName))
-	hashStr := hex.EncodeToString(hashBytes[:])[:repoNameHashLen]
+	return hex.EncodeToString(hashBytes[:])[:repoNameHashLen]
+}
+
+func sanitize(repoPrefix, name string, maxLen int, hashSeparator string) string {
```
sdowell: nit: Would you mind adding some unit tests for these functions to verify and document the behavior?

e2e/nomostest/gitproviders/git-provider.go
```diff
@@ -51,7 +51,7 @@ type GitProvider interface {
 func NewGitProvider(t testing.NTB, provider, clusterName string, logger *testlogger.TestLogger, shell *testshell.TestShell) GitProvider {
 	switch provider {
 	case e2e.Bitbucket:
-		client, err := newBitbucketClient(logger)
+		client, err := newBitbucketClient(clusterName, logger)
```
tiffanny29631: Looks like the repos will be grouped by cluster name now, quick thought is that the ci-main and ci-release projects have same set of cluster names, will there be collisions, i.e. autopilot-rapid-1 from both project are preforming tests with the same repo in bitbucket?

e2e/nomostest/gitproviders/bitbucket.go
Camila-B: Thanks! Done

e2e/nomostest/gitproviders/git-provider.go
Camila-B: You're right, there is the possibility of a collision there. I think that the repo name should also contain a reference to the project

e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -93,154 +101,143 @@ func (b *BitbucketClient) SyncURL(name string) string {
 
 // CreateRepository calls the POST API to create a remote repository on Bitbucket.
 // The remote repo name is unique with a prefix of the local name.
-func (b *BitbucketClient) CreateRepository(localName string) (string, error) {
-	u, err := uuid.NewRandom()
+func (b *BitbucketClient) CreateRepository(name string) (string, error) {
```
mikebz, line 104: I'd suggest that if the method no longer creates something it's better to pick a different name than just Create.  Create should probably error out if something exists.

e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -93,154 +101,143 @@ func (b *BitbucketClient) SyncURL(name string) string {
...
-// deleteSingleRepo deletes a single repository
-func (b *BitbucketClient) deleteSingleRepo(accessToken, name string) error {
-	out, err := exec.Command("curl", "-sX", "DELETE",
-		"-H", fmt.Sprintf("Authorization:Bearer %s", accessToken),
-		fmt.Sprintf("https://api.bitbucket.org/2.0/repositories/%s/%s",
-			b.workspace, name)).CombinedOutput()
+	defer func() {
+		if closeErr := resp.Body.Close(); closeErr != nil {
+			// Log the error as just printing it might be missed.
+			fmt.Printf("failed to close response body: %v\n", err)
```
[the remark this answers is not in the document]
Camila-B: Done

e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -93,154 +101,143 @@ func (b *BitbucketClient) SyncURL(name string) string {
...
-		return errors.New(string(out))
-	}
+	return resp, nil
+}
+
+// DeleteRepositories is a no-op because Bitbucket repo names are determined by the
+// test cluster name and RSync namespace and name, so they can be reset and reused
+// across test runs
+func (b *BitbucketClient) DeleteRepositories(_ ...string) error {
+	fmt.Println("[BITBUCKET] Skip deletion of repos")
```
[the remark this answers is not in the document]
Camila-B: Done

e2e/nomostest/gitproviders/git-provider.go
Camila-B: I added the project name to the repo name

e2e/nomostest/gitproviders/util/reponame.go
Camila-B: I refactored the sanitize function and added additional tests

e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -93,154 +101,144 @@ func (b *BitbucketClient) SyncURL(name string) string {
...
 	}
 
-	out, err := exec.Command("curl", "-sX", "POST",
-		"-H", "Content-Type: application/json",
-		"-H", fmt.Sprintf("Authorization:Bearer %s", accessToken),
-		"-d", fmt.Sprintf("{\"scm\": \"git\",\"project\": {\"key\": \"%s\"},\"is_private\": \"true\"}", bitbucketProject),
-		fmt.Sprintf("https://api.bitbucket.org/2.0/repositories/%s/%s", b.workspace, repoName)).CombinedOutput()
-
+	// Check if repository already exists
+	resp, err := b.sendRequest(http.MethodGet, repoURL, accessToken, nil)
```
tiffanny29631, line 115: minor: This 'get-or-create' logic looks like it could introduce a race condition. If two parallel tests try to create the same (new) repo at the same time, couldn't they both get a 404 and then both try to POST, causing one to fail? It's possible that the failed test eventually passes in the next runs but could introduce a new source of flakiness. We might need a way to make this 'get-or-create' operation atomic, perhaps by handling the 409 Conflict on the POST as a 'success'?

e2e/nomostest/gitproviders/util/reponame.go
```diff
@@ -22,19 +22,41 @@ import (
...
 )
 
-// SanitizeRepoName replaces all slashes with hyphens, and truncate the name.
+// SanitizeGCPRepoName replaces all slashes with hyphens, and truncates the name.
 // repo name may contain between 3 and 63 lowercase letters, digits and hyphens.
-func SanitizeRepoName(repoPrefix, name string) string {
+// This is used by CSR and SSM which are per-project resources
+func SanitizeGCPRepoName(repoPrefix, name string) string {
 	fullName := "cs-e2e-" + repoPrefix + "-" + name
+	hashStr := hashName(fullName)
```
tiffanny29631, line 44: Minor nitpick: what happens if name or repoSuffix is an empty string? Same question applies to both repo

e2e/nomostest/gitproviders/bitbucket.go
Camila-B: I don't think that this can happen given that repo names will be unique and based on the cluster name + project name + R*Sync name

e2e/nomostest/gitproviders/util/reponame.go
```diff
@@ -22,19 +22,41 @@ import (
...
+
+	return sanitize(fullName, hashStr, defaultRepoNameMaxLen)
+}
+
+// SanitizeBitbucketRepoName replaces all slashes with hyphens, and truncates the name for Bitbucket.
+// repo name may contain between 3 and 62 lowercase letters, digits and hyphens.
+func SanitizeBitbucketRepoName(repoSuffix, name string) string {
```
sdowell, line 52: Just curious - what's the rationale for using suffix here rather than prefix?

Camila-B: It's more of a stylistic choice than anything else. If the repo name were to use a prefix, the end result would be of the form`kpt-config-sync-ci-main-autopilot-rapid-latest-5dd666db` vs `config-management-system-root-sync-kpt-config-sync-ci-6485bfa0`. Given that in either format, the hash generated is unique, I can change it back to using a prefix for consistency's sake

sdowell: The suffix seems reasonable, I'd suggest just documenting the rationale in a comment since it's a new pattern

e2e/nomostest/gitproviders/bitbucket.go
tiffanny29631: SG

e2e/nomostest/gitproviders/util/reponame.go
Camila-B: Both values should be provided. However, I modified the functions to require at least the base repo name and to handle the case where repoSuffix/repoPrefix is empty. I also added unit tests to cover this scenario

e2e/nomostest/gitproviders/bitbucket.go
```diff
@@ -93,154 +101,144 @@ func (b *BitbucketClient) SyncURL(name string) string {
...
 
 	accessToken, ok := output["access_token"]
 	if !ok {
-		return "", fmt.Errorf("no access_token: %s", string(out))
+		return "", fmt.Errorf("no access_token in response: %s", string(body))
+	}
+
+	accessTokenStr, ok := accessToken.(string)
+	if !ok {
+		return "", fmt.Errorf("access_token is not a string: %v", accessToken)
```
sdowell: The access token is (semi) sensitive and probably shouldn't be included in an error, since that makes it likely to be logged. You could print the type (%T) here instead of the value

I'd also probably be cautious about logging/returning the raw output from the response above

Camila-B: Good point, I modified the error messages to no longer include the access token nor the raw response

sdowell (approved):
/lgtm

PR #1922, opened 2025-10-31
test: add output to AR error message

Camila-B:
Updates the error returned when trying to create a new Artifact Registry repository so that it also includes the output from the failed gcloud command.


sdowell (approved):
/lgtm

Issue #1633, opened 2025-04-07
Nomos vet fails to validate OPA Gatekeeper custom resource when ConstraintTemplate is in the same commit


philipsabri:
When validating a set of Kubernetes configuration files using `nomos vet`, specifically involving OPA Gatekeeper resources, a validation failure occurs if a `ConstraintTemplate` (which defines a Custom Resource Definition) and a corresponding Constraint (an instance of the custom resource defined by the template) are present together in the same batch of files being validated (e.g., as part of a single commit or PR).

This behavior seems specific to Gatekeeper `ConstraintTemplates`. `nomos vet` does not exhibit the same failure when validating a standard `CustomResourceDefinition` object alongside an instance of its defined kind present in the same batch of files. It is likely that similar issues would occur with other tools that generate CRDs indirectly, such as [kro](https://github.com/kro-run/kro/).

This forces users to either split the ConstraintTemplate definition and its corresponding Constraint usage into separate commits/validation steps (which hinders atomic changes).

Steps to reproduce
1. Install OPA gatekeeper
2. run nomos vet on this example
```yaml
kind: ConstraintTemplate
apiVersion: templates.gatekeeper.sh/v1
metadata:
  name: validateuserinfo
spec:
  crd:
    spec:
      names:
        kind: ValidateUserInfo
  targets:
    - target: admission.k8s.gatekeeper.sh
      rego: |
        package k8svalidateuserinfo
        violation[{"msg": msg}] {
          username := input.review.userInfo.username
          not startswith(username, "system:")
          msg := sprintf("username is not allowed to perform this operation: %v", [username])
        }
---
kind: ValidateUserInfo
apiVersion: constraints.gatekeeper.sh/v1
metadata:
  name: always-validate
```

3. Error

```
Error: KNV1021: No CustomResourceDefinition is defined for the type "ValidateUserInfo.constraints.gatekeeper.sh" in the cluster.
Resource types that are not native Kubernetes objects must have a CustomResourceDefinition.

Config Sync will retry until a CustomResourceDefinition is defined for the type "ValidateUserInfo.constraints.gatekeeper.sh" in the cluster.

source: /workspaces/kpt-config-sync/philip/cluster/test.yaml
metadata.name: always-validate
group: constraints.gatekeeper.sh
version: v1
kind: ValidateUserInfo

For more information, see https://g.co/cloud/acm-errors#knv1021
```

[...]

PR #1926, opened 2025-11-06
feat: Enhance nomos vet with flag validation and raw resource processing

cowsking:
This commit expands the nomos vet command's capabilities by:
   - Introducing a flag validation, including specific handling for --skip-group.
   - Adding new unit tests within the vet package, alongside new test cases (TestVetSkipGroup, TestVetFlagExclusion), to ensure correct behavior and validation of these enhancements.

These changes specifically address issue #1633, where nomos vet failed to validate OPA Gatekeeper custom resources when a ConstraintTemplate and its corresponding Constraint were present in the same commit. The improved raw resource processing and  flag handling now allow nomos vet to correctly process such configurations, preventing the KNV1021 error.

### DIFF
--- a/cmd/nomos/flags/flags.go
+++ b/cmd/nomos/flags/flags.go
@@ -40,6 +40,9 @@ const (
 	// SkipAPIServerFlag is the flag name for SkipAPIServer below.
 	SkipAPIServerFlag = "no-api-server-check"
 
+	// NoAPIServerCheckForGroupFlag is the flag name for SkipAPIServerCheckForGroup below.
+	NoAPIServerCheckForGroupFlag = "no-api-server-check-for-group"
+
 	// OutputYAML specifies exporting the output in YAML format.
 	OutputYAML = "yaml"
 
@@ -63,6 +66,9 @@ var (
 
 	// SkipAPIServer directs whether to try to contact the API Server for checks.
 	SkipAPIServer bool
+
+	// SkipAPIServerCheckForGroup contains the list of API Groups to skip API server validation for.
+	SkipAPIServerCheckForGroup []string
 
 	// SourceFormat indicates the format of the Git repository.
 	SourceFormat string
@@ -99,6 +105,12 @@ func AddPath(cmd *cobra.Command) {
 func AddSkipAPIServerCheck(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&SkipAPIServer, SkipAPIServerFlag, false,
 		"If true, disables talking to the API Server for discovery.")
+}
+
+// AddNoAPIServerCheckForGroup adds the --no-api-server-check-for-group flag.
+func AddNoAPIServerCheckForGroup(cmd *cobra.Command) {
+	cmd.Flags().StringSliceVar(&SkipAPIServerCheckForGroup, NoAPIServerCheckForGroupFlag, nil,
+		"Accepts a comma-separated list of API Groups will be skipped during processing")
 }
 
 // AllClusters returns true if all clusters should be processed.
--- a/cmd/nomos/vet/vet.go
+++ b/cmd/nomos/vet/vet.go
@@ -35,6 +35,7 @@ func init() {
 	flags.AddClusters(Cmd)
 	flags.AddPath(Cmd)
 	flags.AddSkipAPIServerCheck(Cmd)
+	flags.AddNoAPIServerCheckForGroup(Cmd)
 	flags.AddSourceFormat(Cmd)
 	flags.AddOutputFormat(Cmd)
 	flags.AddAPIServerTimeout(Cmd)
@@ -75,6 +76,13 @@ returns a non-zero error code if any issues are found.
   nomos vet --path=my/directory
   nomos vet --path=/path/to/my/directory`,
 	Args: cobra.ExactArgs(0),
+	PreRunE: func(_ *cobra.Command, _ []string) error {
+		if flags.SkipAPIServer && len(flags.SkipAPIServerCheckForGroup) > 0 {
+			// If --no-api-server-check is specified, throw the error
+			return fmt.Errorf("cannot specify both --%s and --%s", flags.SkipAPIServerFlag, flags.NoAPIServerCheckForGroupFlag)
+		}
+		return nil
+	},
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		// Don't show usage on error, as argument validation passed.
 		cmd.SilenceUsage = true
--- a/cmd/nomos/vet/vet_test.go
+++ b/cmd/nomos/vet/vet_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/GoogleContainerTools/config-sync/cmd/nomos/flags"
@@ -50,6 +51,12 @@ func resetFlags() {
 	keepOutput = false
 	outPath = flags.DefaultHydrationOutput
 	flags.OutputFormat = flags.OutputYAML
+	flags.SkipAPIServerCheckForGroup = nil
+}
+
+func resetFlagExclusivityTestFlags() {
+	flags.SkipAPIServer = false
+	flags.SkipAPIServerCheckForGroup = nil
 }
 
 var examplesDir = cmpath.RelativeSlash("../../../examples")
@@ -175,6 +182,92 @@ func TestVet_MultiCluster(t *testing.T) {
 				// Execute wraps the string output as a simple error.
 				wantError := errors.New(tc.wantError.Error())
 				require.Equal(t, wantError, err)
+			}
+		})
+	}
+}
+
+func TestVet_FlagExclusivity(t *testing.T) {
+	testCases := []struct {
+		name                     string
+		flags                    []string
+		expectError              bool
+		expectedError            string
+		expectedWarning          string
+		expectGroupsToBeNil      bool
+		initialNoAPIServerGroups []string
+	}{
+		{
+			name:          "Both flags used, expect error",
+			flags:         []string{"--no-api-server-check", "--no-api-server-check-for-group=constraints.gatekeeper.sh"},
+			expectError:   true,
+			expectedError: fmt.Sprintf("cannot specify both --%s and --%s", flags.SkipAPIServerFlag, flags.NoAPIServerCheckForGroupFlag),
+		},
+		{
+			name:        "Only --no-api-server-check used",
+			flags:       []string{"--no-api-server-check"},
+			expectError: false,
+		},
+		{
+			name:                     "Only --no-api-server-check-for-group used",
+			flags:                    []string{"--no-api-server-check-for-group=constraints.gatekeeper.sh"},
+			expectError:              false,
+			initialNoAPIServerGroups: []string{"constraints.gatekeeper.sh"},
+		},
+		{
+			name:        "No flags used",
+			flags:       []string{},
+			expectError: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resetFlags()
+			// We need a valid directory for the command to run, even though we're testing flag validation.
+			tmpDir, err := os.MkdirTemp("", "nomos-vet-test")
+			require.NoError(t, err)
+			t.Cleanup(func() {
+				err := os.RemoveAll(tmpDir)
+				require.NoError(t, err)
+			})
+
+			resetFlagExclusivityTestFlags()
+
+			// Set the flags for the current test case.
+			var groups []string
+			for _, f := range tc.flags {
+				if f == "--no-api-server-check" {
+					flags.SkipAPIServer = true
+				} else if strings.HasPrefix(f, "--no-api-server-check-for-group=") {
+					parts := strings.SplitN(f, "=", 2)
+					groups = append(groups, strings.Split(parts[1], ",")...)
+				}
+			}
+			flags.SkipAPIServerCheckForGroup = groups
+			if tc.initialNoAPIServerGroups != nil {
+				flags.SkipAPIServerCheckForGroup = tc.initialNoAPIServerGroups
+			}
+
+			output := new(bytes.Buffer)
+			Cmd.SetOut(output)
+			Cmd.SetErr(output)
+
+			err = Cmd.PreRunE(Cmd, nil)
+
+			if tc.expectError {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tc.expectedError)
+			} else {
+				require.NoError(t, err)
+			}
+
+			if tc.expectedWarning != "" {
+				require.Equal(t, tc.expectedWarning, output.String())
+			}
+
+			if tc.expectGroupsToBeNil {
+				require.Nil(t, flags.SkipAPIServerCheckForGroup)
 			}
 		})
 	}
--- a/e2e/testcases/cli_test.go
+++ b/e2e/testcases/cli_test.go
@@ -1278,6 +1278,34 @@ func TestNomosStatusNameFilter(t *testing.T) {
 	}
 }
 
+func TestVetSkipGVK(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI,
+		ntopts.SyncWithGitSource(nomostest.DefaultRootSyncID, ntopts.Unstructured),
+	)
+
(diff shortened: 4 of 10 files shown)


pkg/validate/fileobjects/raw.go
```diff
@@ -78,9 +80,22 @@ func (r *Raw) Scoped() (*Scoped, status.MultiError) {
...
 		s, err := scoper.GetObjectScope(obj)
 		if err != nil {
+			// Check if the error is a KNV1021 and if the GVK should be skipped.
+			// The error message for KNV1021 typically indicates an unknown GVK.
+			gk := obj.GroupVersionKind().GroupKind()
+
+			if gvkutil.Matches(gk, skipGVKErrors) {
+				klog.V(6).Infof("ignoring KNV1021 error for %s/%s due to --no-api-server-check-for-group flag: %v", gk.Group, gk.Kind, err)
+				continue // Skip appending this error
+			}
```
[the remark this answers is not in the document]
cowsking: Should --no-api-server-check-for-group selectively ignore only the KNV1021 error for specified groups to prevent masking other validation errors? This approach differs from the broader --no-api-server-check, which disables all API server communication. If we adhere to the design of --no-api-server-check, I believe it should remain broad.

pkg/validate/fileobjects/raw.go
```diff
@@ -78,9 +80,22 @@ func (r *Raw) Scoped() (*Scoped, status.MultiError) {
 		DynamicNSSelectorEnabled: r.DynamicNSSelectorEnabled,
 		NSControllerState:        r.NSControllerState,
 	}
+	// Parse the patterns from the flag once.
+	// We ignore the error from ParsePatterns for now, assuming valid input or empty.
+	skipGVKErrors, _ := gvkutil.ParsePatterns(flags.NoAPIServerCheckForGroup)
```
sdowell: This package library shouldn't directly access flags, the flag should be passed in as an arg/struct value. It should follow a similar pattern as [SkipAPIServer](https://github.com/GoogleContainerTools/config-sync/blob/main/cmd/nomos/vet/vet_impl.go#L99). This will help with separation of concerns and testability

e2e/testdata/gatekeeper-skip-gvk/constraint-template.yaml
sdowell, line 1: I expect there to be an e2e test case that uses this file. Will you be adding test coverage?

pkg/validate/fileobjects/raw.go
```diff
@@ -78,9 +80,22 @@ func (r *Raw) Scoped() (*Scoped, status.MultiError) {
...
+
 	for _, obj := range r.Objects {
 		s, err := scoper.GetObjectScope(obj)
 		if err != nil {
+			// Check if the error is a KNV1021 and if the GVK should be skipped.
+			// The error message for KNV1021 typically indicates an unknown GVK.
+			gk := obj.GroupVersionKind().GroupKind()
+
+			if gvkutil.Matches(gk, skipGVKErrors) {
```
sdowell: Please also add unit test coverage for this new functionality. It would be ideal to add unit test coverage for this function directly.

Unit tests can additionally be added at the layer of [TestUnstructured and TestHierarchical](https://github.com/GoogleContainerTools/config-sync/blob/main/pkg/validate/validate_test.go).

e2e/testdata/gatekeeper-skip-gvk/constraint-template.yaml
cowsking: Added the test case

cmd/nomos/flags/flags.go
```diff
@@ -40,6 +40,9 @@ const (
 	// SkipAPIServerFlag is the flag name for SkipAPIServer below.
 	SkipAPIServerFlag = "no-api-server-check"
 
+	// NoAPIServerCheckForGroupFlag is the flag name for NoAPIServerCheckForGroup below.
+	NoAPIServerCheckForGroupFlag = "no-api-server-check-for-group"
```
mikebz, line 44: this seems pretty long... I would consider something shorter.

e2e/testcases/cli_test.go
```diff
@@ -1278,6 +1278,37 @@ func TestNomosStatusNameFilter(t *testing.T) {
...
 }
 
+func TestVetSkipGVK(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI,
+		ntopts.SyncWithGitSource(nomostest.DefaultRootSyncID, ntopts.Unstructured),
+	)
+
+	rootRepo := nt.SyncSourceGitReadWriteRepository(nomostest.DefaultRootSyncID)
+
+	if err := rootRepo.Copy("../testdata/gatekeeper-skip-gvk", "acme"); err != nil {
```
Camila-B: Nit: You can use `nt.Must()` here to simplify the code 
```suggestion
	nt.Must(rootRepo.Copy("../testdata/gatekeeper-skip-gvk", "acme"))
```

e2e/testcases/cli_test.go
```diff
@@ -1278,6 +1278,37 @@ func TestNomosStatusNameFilter(t *testing.T) {
...
+	if err == nil {
+		t.Fatal("expected `nomos vet` to fail but it passed")
+	}
+	// Check for KNV1021 error.
+	if !strings.Contains(string(out), "KNV1021") {
+		t.Fatalf("expected KNV1021 error, but got: %v", string(out))
+	}
+
+	// Now, run `nomos vet` with the flag and expect it to pass.
+	out, err = nt.Shell.Command("nomos", "vet", "--path", rootRepo.Root, "--source-format=unstructured", "--no-api-server-check-for-group=templates.gatekeeper.sh,constraints.gatekeeper.sh").CombinedOutput()
```
Camila-B: `nt.Must()` can also be used here

pkg/util/gvkutil/gvk_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			patterns, err := ParsePatterns(tc.rawPatterns)
+			if (err != nil) != tc.expectErr {
+				t.Errorf("ParsePatterns() error = %v, expectErr %v", err, tc.expectErr)
+				return
+			}
+			if len(patterns) != len(tc.expected) {
```
Camila-B: Nit: you can use assert.Len() here

pkg/util/gvkutil/gvk_test.go
```diff
@@ -0,0 +1,112 @@
...
+				t.Fatalf("expected %d patterns, got %d", len(tc.expected), len(patterns))
+			}
+			for i := range patterns {
```
Camila-B: You can use `sort.Sort` and `cmp.Diff` here to compare the expected and actual patterns

cmd/nomos/flags/flags.go
cowsking: It was designed to name as skip-gvk but we wanted it to align with no-api-server-check. [go/nomos-vet-skip-gvk-dd](http://goto.google.com/nomos-vet-skip-gvk-dd)

cmd/nomos/flags/flags.go
```diff
@@ -101,6 +107,12 @@ func AddSkipAPIServerCheck(cmd *cobra.Command) {
 		"If true, disables talking to the API Server for discovery.")
 }
 
+// AddNoAPIServerCheckForGroup adds the --no-api-server-check-for-group flag.
+func AddNoAPIServerCheckForGroup(cmd *cobra.Command) {
+	cmd.Flags().StringSliceVar(&NoAPIServerCheckForGroup, NoAPIServerCheckForGroupFlag, nil,
+		"Accepts a comma-separated list of API Groups for which **all** resources will be skipped during processing (e.g., 'constraints.gatekeeper.sh'). This excludes objects in these groups from all validation and processing, not just API server validation.")
```
tiffanny29631: The description is detailed in a good way but might be verbose for a --help output, suggest keeping it simple here, and add the detailed description in public doc.

cmd/nomos/vet/vet_test.go
```diff
@@ -180,6 +182,75 @@ func TestVet_MultiCluster(t *testing.T) {
...
+			// We need a valid directory for the command to run, even though we're testing flag validation.
+			tmpDir, err := os.MkdirTemp("", "nomos-vet-test")
+			require.NoError(t, err)
+			t.Cleanup(func() {
+				err := os.RemoveAll(tmpDir)
+				require.NoError(t, err)
+			})
+
+			// Reset flags to a clean state for each test case.
+			flags.SkipAPIServer = false
```
tiffanny29631: Can these be consolidated to resetFlags?

cmd/nomos/vet/vet.go
```diff
@@ -75,15 +77,27 @@ returns a non-zero error code if any issues are found.
   nomos vet --path=my/directory
   nomos vet --path=/path/to/my/directory`,
 	Args: cobra.ExactArgs(0),
+	PreRunE: func(_ *cobra.Command, _ []string) error {
```
tiffanny29631, line 79: Is it possible to just ignore the no-api-server-check-for-group if the no-api-server-check is specified, assuming it's an inclusive action overall

cmd/nomos/flags/flags.go
cowsking: Sounds good to me

cmd/nomos/vet/vet.go
cowsking: Sounds good. I will also implement a warning message to inform the user that the no-api-server-check-for-group option will be ignored.

cmd/nomos/vet/vet.go
```diff
@@ -75,15 +77,29 @@ returns a non-zero error code if any issues are found.
   nomos vet --path=my/directory
   nomos vet --path=/path/to/my/directory`,
 	Args: cobra.ExactArgs(0),
+	PreRunE: func(cmd *cobra.Command, _ []string) error {
+		if flags.SkipAPIServer && len(flags.NoAPIServerCheckForGroup) > 0 {
+			// If --no-api-server-check is specified, ignore --no-api-server-check-for-group.
```
sdowell: I recall in the design doc it was decided to return an error if both were specified

e2e/testcases/cli_test.go
```diff
@@ -1278,6 +1278,34 @@ func TestNomosStatusNameFilter(t *testing.T) {
...
 
+func TestVetSkipGVK(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.NomosCLI,
+		ntopts.SyncWithGitSource(nomostest.DefaultRootSyncID, ntopts.Unstructured),
+	)
+
+	rootRepo := nt.SyncSourceGitReadWriteRepository(nomostest.DefaultRootSyncID)
+
+	nt.Must(rootRepo.Copy("../testdata/gatekeeper-skip-gvk", "acme"))
+	if err := rootRepo.CommitAndPush("add gatekeeper constraint template and constraint"); err != nil {
```
sdowell: nit: It doesn't seem like it should be necessary to push the commit to the remote repo. This will cause the resources to get synced to the cluster, which is an unnecessary side effect

pkg/validate/fileobjects/raw.go
```diff
@@ -78,9 +81,19 @@ func (r *Raw) Scoped() (*Scoped, status.MultiError) {
...
+			// For objects with matching GVKs, all errors from scoper.GetObjectScope are skipped.
+			// This includes, but is not limited to, unknown GVK errors such as KNV1021.
+			gk := obj.GroupVersionKind().GroupKind()
+
+			if gvkutil.Matches(gk, r.SkippedGVKs) {
+				klog.V(6).Infof("ignoring KNV1021 error for %s/%s due to --no-api-server-check-for-group flag: %v", gk.Group, gk.Kind, err)
+				continue // Skip appending this error
+			}
+
 			if r.AllowUnknownKinds {
```
sdowell: nit: I'd suggest implementing the control flow as part of the `r.AllowUnknownKinds` if/else to keep this consistent. The intent is that only one of `AllowUnknownKinds` and `SkippedGVKs` is specified, and it would be more intuitive if these followed a similar control flow (rather than using a separate `if` with `continue`)

pkg/validate/fileobjects/raw.go
```diff
@@ -78,9 +80,22 @@ func (r *Raw) Scoped() (*Scoped, status.MultiError) {
...
 	for _, obj := range r.Objects {
 		s, err := scoper.GetObjectScope(obj)
 		if err != nil {
+			// Check if the error is a KNV1021 and if the GVK should be skipped.
+			// The error message for KNV1021 typically indicates an unknown GVK.
+			gk := obj.GroupVersionKind().GroupKind()
+
+			if gvkutil.Matches(gk, skipGVKErrors) {
+				klog.V(6).Infof("ignoring KNV1021 error for %s/%s due to --no-api-server-check-for-group flag: %v", gk.Group, gk.Kind, err)
+				continue // Skip appending this error
```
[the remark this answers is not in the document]
sdowell: This comment is correct - the object should be retained with `Unknown` scope rather than skipped

pkg/validate/fileobjects/raw.go
```diff
@@ -78,9 +81,19 @@ func (r *Raw) Scoped() (*Scoped, status.MultiError) {
...
 	}
+
 	for _, obj := range r.Objects {
 		s, err := scoper.GetObjectScope(obj)
 		if err != nil {
+			// For objects with matching GVKs, all errors from scoper.GetObjectScope are skipped.
+			// This includes, but is not limited to, unknown GVK errors such as KNV1021.
+			gk := obj.GroupVersionKind().GroupKind()
+
+			if gvkutil.Matches(gk, r.SkippedGVKs) {
```
sdowell: nit: the control flow here is a bit inconsistent with the `AllowUnknownKinds` flow. The intent is that only one of `AllowUnknownKinds` or `SkippedGVKs` is specified, but this control flow doesn't really demonstrate that intuitively.

One way I would consider this is combining both "Skip all" and "Skip for certain groups" into a single interface to streamline this control flow. See this commit as an example: https://github.com/sdowell/kpt-config-sync/commit/eeaf4274f41f1d785396d965e9d5052710909857 (this commit isn't necessarily complete but serves as an example)

pkg/validate/fileobjects/raw.go
```diff
@@ -58,6 +60,9 @@ type Raw struct {
 	NSControllerState *namespacecontroller.State
 	// WebhookEnabled indicates whether Webhook configuration is enabled
 	WebhookEnabled bool
+	// AllowUnknownKindMatcher indicates which object kinds do not need to ignore
```
sdowell: nit: `do not need to` -> `should`

pkg/util/gvkutil/gvk.go
sdowell, line 1: This file and the accompanying test file are now unused and can be removed

cmd/nomos/vet/vet.go
```diff
@@ -75,15 +77,31 @@ returns a non-zero error code if any issues are found.
   nomos vet --path=my/directory
   nomos vet --path=/path/to/my/directory`,
 	Args: cobra.ExactArgs(0),
+	PreRunE: func(_ *cobra.Command, _ []string) error {
+		if flags.SkipAPIServer && len(flags.SkipAPIServerCheckForGroup) > 0 {
+			// If --no-api-server-check is specified, throw the error
+			flags.SkipAPIServerCheckForGroup = nil
```
sdowell: nit: is there a reason this needs to be set to nil?

cmd/nomos/vet/vet.go
```diff
@@ -75,15 +77,31 @@ returns a non-zero error code if any issues are found.
...
+			return fmt.Errorf("cannot specify both --%s and --%s", flags.SkipAPIServerFlag, flags.NoAPIServerCheckForGroupFlag)
+		}
+		return nil
+	},
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		// Don't show usage on error, as argument validation passed.
 		cmd.SilenceUsage = true
 
+		var matcher fileobjects.ObjectMatcher
+		if flags.SkipAPIServer {
```
sdowell: This is a duplicate code path to the setter in `hydrate.ValidateOptions`. I'd suggest consolidating this to the single code path in the `ValidateOptions` helper function

sdowell (approved):
Looks good, thanks!
/lgtm

PR #1930, opened 2025-11-10
fix: reset ResourceGroup metrics on deletion

tiffanny29631:
ResourceGroup metrics were showing stale values from deleted ResourceGroups because the controller never received deletion events. Since the pod persists, metrics continue to be emitted but with incorrect values. 

Example: When a RootSync with 30 resources is deleted, ready_resource_count will still show 30 even though the ResourceGroup no longer exists. Resetting the resource metrics to zero ensures the aggregate metric values across all ResourceGroups remain correct.

Root controller now sends deletion events, and ResourceGroup controller resets resource metrics to zero for deleted ResourceGroups, ensuring aggregate metric values remain correct.

### DIFF
--- a/e2e/testcases/reconciler_finalizer_test.go
+++ b/e2e/testcases/reconciler_finalizer_test.go
@@ -37,6 +37,9 @@ import (
 	"github.com/GoogleContainerTools/config-sync/pkg/core/k8sobjects"
 	"github.com/GoogleContainerTools/config-sync/pkg/kinds"
 	"github.com/GoogleContainerTools/config-sync/pkg/metadata"
+	rgmetrics "github.com/GoogleContainerTools/config-sync/pkg/resourcegroup/controllers/metrics"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	prometheusmodel "github.com/prometheus/common/model"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/resourcegroup/controllers/root/root_controller.go
```diff
@@ -92,7 +93,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			}
+			// Send deletion event to channel so ResourceGroup controller can clean up metrics
+			r.Logger(ctx).V(3).Info("Sending deletion event to ResourceGroup controller")
+			r.channel <- event.GenericEvent{Object: &v1alpha1.ResourceGroup{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: req.NamespacedName.Namespace,
+					Name:      req.NamespacedName.Name,
+				},
+			}}
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
Camila-B: I think this is a good comment to address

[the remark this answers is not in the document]
tiffanny29631: Done

pkg/resourcegroup/controllers/resourcegroup/resourcegroup_controller.go
```diff
@@ -162,6 +169,18 @@ func updateResourceMetrics(ctx context.Context, nn types.NamespacedName, statuse
 	metrics.RecordKCCResourceCount(ctx, nn, int64(kccCount))
 }
 
+// resetResourceMetrics resets all resource metrics to zero when a ResourceGroup is deleted.
+// This ensures the metrics stop being emitted and can expire.
```
sdowell, line 173: Does a value of zero look the same as "unset" for each of these metrics from an observability standpoint?

tiffanny29631: Yes, these are the resource related metrics that are in the form of a gauge (only records last value), and are updated upon each status compute. Resetting them is counted as 'unset' the values for a deleted resourcegroup. 

Reminds me that I also need to add an e2e for verification, will update you.

tiffanny29631: E2E test added!

e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -757,6 +760,257 @@ func deleteObject(nt *nomostest.NT, obj client.Object) error {
 	return nil
 }
 
+// TestReconcilerFinalizer_ResourceGroupMetricsReset tests that when a ResourceGroup
+// is deleted, all resource-related metrics are reset to 0.
+func TestReconcilerFinalizer_ResourceGroupMetricsReset(t *testing.T) {
+	nt := nomostest.New(t, nomostesting.MultiRepos)
```
sdowell: nit: suggest using `Unstructured` repository for testing instead of hierarchical

e2e/testcases/reconciler_finalizer_test.go
```diff
@@ -757,6 +760,257 @@ func deleteObject(nt *nomostest.NT, obj client.Object) error {
...
+
+	nt.T.Log("Verifying resourcegroup metrics are non-zero before deletion")
+	nt.Must(validateResourceGroupMetricsNonZero(nt, rootSyncKey))
+
+	// Tail reconciler logs and print if there's an error.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go nomostest.TailReconcilerLogs(ctx, nt, nomostest.RootReconcilerObjectKey(rootSyncKey.Name))
+
+	// Enable deletion propagation
```
sdowell: Is the deletion propagation relevant to the test case? I believe the `ResourceGroup` should get cleaned up regardless of the deletion propagation policy

tiffanny29631: Done

e2e/testcases/reconciler_finalizer_test.go
tiffanny29631: Done

PR #1935, opened 2025-11-13
test: use retry backoff when creating a Gitlab repo

Camila-B:
This change mitigates the GitLab API rate limiting requests, resulting in 429 Too Many Requests errors during repository creation.

### DIFF
--- a/e2e/nomostest/gitproviders/gitlab.go
+++ b/e2e/nomostest/gitproviders/gitlab.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/GoogleContainerTools/config-sync/e2e"
 	"github.com/GoogleContainerTools/config-sync/e2e/nomostest/gitproviders/util"
+	"github.com/GoogleContainerTools/config-sync/e2e/nomostest/retry"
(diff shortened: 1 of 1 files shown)


e2e/nomostest/gitproviders/gitlab.go
```diff
@@ -80,61 +81,67 @@ func (g *GitlabClient) SyncURL(name string) string {
...
 
-	// Check if the repository already exists
-	getRequestCtx, getRequestCancel := context.WithTimeout(context.Background(), gitlabRequestTimeout)
-	defer getRequestCancel()
-	resp, err := g.sendRequest(getRequestCtx, http.MethodGet, repoURL, nil)
-	if err != nil {
-		return "", err
-	}
+	// Handles 429 errors with retry backoff
+	_, err = retry.Retry(2*time.Minute, func() error {
```
tiffanny29631: Is 2min enough for the timeout?

Camila-B: I think it should be as the limit is 25 requests a minute https://docs.gitlab.com/user/gitlab_com/#rate-limits-on-gitlabcom. I increased it just in case though.

PR #1936, opened 2025-11-14
feat: Add support for token auth in oci-sync

Camila-B:


cmd/oci-sync/main.go
```diff
@@ -52,6 +52,10 @@ var flOneTime = flag.Bool("one-time", util.EnvBool("OCI_SYNC_ONE_TIME", false),
 	"exit after the first sync")
 var flMaxSyncFailures = flag.Int("max-sync-failures", util.EnvInt("OCI_SYNC_MAX_SYNC_FAILURES", 0),
 	"the number of consecutive failures allowed before aborting (the first sync must succeed, -1 will retry forever after the initial sync)")
+var flUsername = flag.String("username", util.EnvString("OCI_SYNC_USERNAME", ""),
+	"the username to use for oci authentication")
+var flPassword = flag.String("password", util.EnvString("OCI_SYNC_PASSWORD", ""),
+	"the password or personal access token to use for oci authentication")
```
tiffanny29631, line 58: Does the password needs nil check is username is set?

Camila-B: Yes. I added a check for the username and password

pkg/reconcilermanager/controllers/reposync_controller.go
```diff
@@ -1018,6 +1032,11 @@ func (r *RepoSyncReconciler) validateNamespaceSecret(ctx context.Context, repoSy
 		authType = repoSync.Spec.Helm.Auth
 		namespaceSecretName = v1beta1.GetSecretName(repoSync.Spec.Helm.SecretRef)
 	}
+	if repoSync.Spec.SourceType == configsync.OciSource {
```
sdowell: nit: `else if` for consistency

pkg/validate/rsync/validate/source_spec_validator_test.go
```diff
@@ -456,6 +469,33 @@ func TestValidateRepoSyncSpec(t *testing.T) {
 			obj:     repoSyncWithOci(withGit()),
 			wantErr: nil,
 		},
+		{
```
sdowell, line 481: nit: could also use test cases for `IllegalSecretRef`

pkg/reconcilermanager/controllers/reposync_controller.go
Camila-B: I changed it to a switch statement

pkg/validate/rsync/validate/source_spec_validator_test.go
Camila-B: Done. I added a case for both RootSync and RepoSync spec validation

PR #1937, opened 2025-11-17
test: return unexpected errors when checking if Bitbucket repo exists

Camila-B:
This helps deal with errors that may occur when checking if a Bitbucket repository exists such as `429 Too Many Requests`


tiffanny29631 (approved):
/lgtm

PR #1940, opened 2025-11-18
refactor: replace unstructured with ResourceGroup type

Camila-B:


pkg/resourcegroup/resourcegroup.go
```diff
@@ -72,3 +46,12 @@ func IsStatusDisabled(resgroup *v1alpha1.ResourceGroup) bool {
 	val, found := annotations[metadata.StatusModeAnnotationKey]
 	return found && val == metadata.StatusDisabled.String()
 }
+
+// ObjectKey returns a key appropriate for fetching a ResourceGroup in the given
+// namespace.
+func ObjectKey(name, namespace string) client.ObjectKey {
```
sdowell: nit: this helper function doesn't really seem needed since it's just a wrapper for as basic struct instantiation

pkg/applier/applier.go
```diff
@@ -186,21 +187,23 @@ func NewSupervisor(cs *ClientSet, scope declared.Scope, syncName string, reconci
 // trying to add any new resources from source.
 func (s *supervisor) UpdateStatusMode(ctx context.Context) error {
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		u := newInventoryUnstructured(s.syncName, s.syncNamespace)
-		key := client.ObjectKeyFromObject(u)
-		err := s.clientSet.Client.Get(ctx, key, u)
+		var rg kptv1alpha1.ResourceGroup
+		rg.SetGroupVersionKind(kinds.ResourceGroup())
```
sdowell: Is it required to set the GVK on the object? I expected the client should be able to get that from the type

pkg/resourcegroup/resourcegroup.go
Camila-B: Good point. I removed it.

pkg/applier/applier.go
Camila-B: No. I removed setting the GVK

sdowell: It looks like this line is still setting the GVK

Camila-B: I removed that line as well

PR #2002, opened 2026-01-16
feat(e2e): skip SSM repository deletion and improve creation logging

cowsking:
- Make `DeleteRepositories` a no-op for the SSM provider, as repository names are derived from the cluster and RSync names and can be reused.
- Update `CreateRepository` to log an info message instead of returning an error if the `gcloud describe` command fails with an error other than `NOT_FOUND`.

### DIFF
--- a/e2e/nomostest/gitproviders/secure_source_manager.go
+++ b/e2e/nomostest/gitproviders/secure_source_manager.go
@@ -121,17 +121,11 @@ func (c *SSMClient) CreateRepository(name string) (string, error) {
 	return fullName, nil
 }
 
-// DeleteRepositories calls the gcloud SDK to delete the provided repositories from SSM.
-func (c *SSMClient) DeleteRepositories(names ...string) error {
-	for _, name := range names {
-		_, err := c.shell.ExecWithDebug("gcloud", "beta", "source-manager", "repos",
-			"delete", name,
-			"--region", c.region,
-			"--project", c.project)
-		if err != nil {
-			return fmt.Errorf("deleting source repository: %w", err)
-		}
-	}
+// DeleteRepositories is a no-op because SSM repo names are determined by the
+// test cluster name and RSync namespace and name, so it can be reused if it
+// failed to be deleted after the test.
+func (c *SSMClient) DeleteRepositories(_ ...string) error {
+	c.shell.Logger.Info("[SSM] Skip deletion of repos")
 	return nil
 }
 


e2e/nomostest/gitproviders/git-provider.go
```diff
@@ -48,7 +48,7 @@ type GitProvider interface {
 }
 
 // NewGitProvider creates a GitProvider for the specific provider type.
-func NewGitProvider(t testing.NTB, provider, clusterName string, logger *testlogger.TestLogger, shell *testshell.TestShell) GitProvider {
+func NewGitProvider(t testing.NTB, provider, clusterName string, logger *testlogger.TestLogger, shell *testshell.TestShell, uniqueID string) GitProvider {
```
tiffanny29631: Could you confirm how 'repo ID collision' was the root cause of the flakiness?

Since tests run sequentially on a given cluster, reusing the repository (as originally designed) should be safe.

e2e/nomostest/gitproviders/secure_source_manager.go
```diff
@@ -135,9 +141,8 @@ func (c *SSMClient) DeleteRepositories(names ...string) error {
 	return nil
 }
 
-// DeleteObsoleteRepos is a no-op because SSM repo names are determined by the
-// test cluster name and RSync namespace and name, so it can be reused if it
-// failed to be deleted after the test.
+// DeleteObsoleteRepos is a no-op because SSM repo names are unique per test
```
tiffanny29631: If we generate unique repositories for every test run, leaving this as a no-op will cause repositories to accumulate indefinitely - we should implement the actual deletion logic here to prevent resource exhaustion. In the original design the repository gets reused between tests so the delete can be a no-op.

cowsking: I see. just spoke with Camila, and I might try modifying the deletion process in SSM first to see if it works.

e2e/nomostest/gitproviders/secure_source_manager.go
```diff
@@ -107,7 +107,7 @@ func (c *SSMClient) CreateRepository(name string) (string, error) {
 		return fullName, nil // repo already exists, skip creation
 	}
 	if !strings.Contains(string(out), "NOT_FOUND") {
-		return fullName, fmt.Errorf("describing source repository: %w", err)
+		c.shell.Logger.Infof("describing source repository %q: %v", fullName, err)
```
Camila-B: Why is this logged instead of returning an error?

cowsking: It would log the error and then proceed to try gcloud create. does this sound correct to you?

Camila-B (approved):
/lgtm

PR #2014, opened 2026-01-22
Bypass inconsistent SSM repository state using a stable version suffix

cowsking:
This PR updates the repository name generation for Secure Source Manager (SSM) in the E2E test framework by appending a stable suffix (e.g., -v2). This change addresses an inconsistent state in the kpt-config-sync-ci-release project where specific repository names have become unusable—reporting as "already existing" during creation but "not found" during describe or delete operations.

### DIFF
--- a/e2e/nomostest/gitproviders/git-provider.go
+++ b/e2e/nomostest/gitproviders/git-provider.go
@@ -74,7 +74,8 @@ func NewGitProvider(t testing.NTB, provider, clusterName string, logger *testlog
 
 		projectNumber := strings.Split(string(out), "\n")[0]
 
-		return newSSMClient(clusterName, shell, projectNumber)
+		// Add suffix for repoPrefix to fix SSM testgrid failure
+		return newSSMClient(clusterName+"-v2", shell, projectNumber)
 	default:
 		return &LocalProvider{}
 	}


e2e/nomostest/gitproviders/secure_source_manager.go
```diff
@@ -60,8 +60,9 @@ func newSSMClient(repoPrefix string, shell *testshell.TestShell, projectNumber s
 	}
 }
 
+// Add suffix for fullName to fix SSM testgrid failure 
 func (c *SSMClient) fullName(name string) string {
-	return util.SanitizeGCPRepoName(c.repoPrefix, name)
+	return util.SanitizeGCPRepoName(c.repoPrefix+"-v2", name)
```
Camila-B: I think you should add the prefix here instead https://github.com/GoogleContainerTools/config-sync/blob/eeefb46d4dd134df1c809c17cb78174e13ade8cd/e2e/nomostest/gitproviders/git-provider.go#L77

e2e/nomostest/gitproviders/secure_source_manager.go
```diff
@@ -60,8 +60,9 @@ func newSSMClient(repoPrefix string, shell *testshell.TestShell, projectNumber s
 	}
 }
 
+// Add suffix for fullName to fix SSM testgrid failure
 func (c *SSMClient) fullName(name string) string {
-	return util.SanitizeGCPRepoName(c.repoPrefix, name)
+	return util.SanitizeGCPRepoName(c.repoPrefix+"-v2", name)
```
tiffanny29631: Is the error more on the provider side or Config Sync? Should we have better error handling instead of adding a constant suffix, which IIUC might need to change if the `repo-a-v2` gets stuck in an inconsistent state in the future?

Camila-B: It's an error on the SSM side. With https://github.com/GoogleContainerTools/config-sync/pull/2002, repos theoretically shouldn't go into an inconsistent state as the repo deletion is what caused the issue in the first place

cowsking: The error is strictly on the provider side (SSM) as they confirm that `an internal call to delete a "control plane repository record" failed with a generic::invalid_argument error`

While we are pushing for this change, their team said they will bring this up in our next on-call handoff to check if this behaviour can be avoided, or surface this error to the customer, as it will break other config sync related customers too.

e2e/nomostest/gitproviders/secure_source_manager.go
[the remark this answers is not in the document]
tiffanny29631: Please address this

PR #2141, opened 2026-04-22
fix: set SSL_CERT_FILE for git-sync for custom CA during GitHub App Auth

samirtahir91:
This PR updates the git-sync reconciler to have the `SSL_CERT_FILE` env variable set to custom CA certificate when the root/repo sync has as custom ca cert defined.

Without is Github App Auth fails due to SSL errors.

Issue - https://github.com/GoogleContainerTools/config-sync/issues/2140


samirtahir91:
@tiffanny29631 - Are you happy to review please?

tiffanny29631 (commented):
Nice fix! Since we have an existing test suite in gitsync_env_test.go, could we add a test case to TestGitSyncEnvVars to verify this change?

samirtahir91:
> Nice fix! Since we have an existing test suite in gitsync_env_test.go, could we add a test case to TestGitSyncEnvVars to verify this change?

Added test, can you review please

tiffanny29631 (approved):
/lgtm

PR #2233, opened 2026-07-20
feat: support disabling monitoring for RootSync and RepoSync

cowsking:
This commit introduces the ability to opt-out of monitoring sidecars (specifically otel-agent) for RootSync and RepoSync objects via the `spec.monitoring.enabled field` in the declarative API.

When monitoring is disabled, the config-sync controllers will omit the otel-agent container from the reconciler pods and set the environment variable `DISABLE_MONITORING=true` in the remaining vital containers to silence telemetry exporter logic. All Unit and E2E tests for this logic have been implemented accordingly.


e2e/testcases/disable_monitoring_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	nt.Must(nt.Watcher.WatchObject(kinds.Deployment(),
+		rootReconcilerName.Name, rootReconcilerName.Namespace,
+		testwatcher.WatchPredicates(
+			testpredicates.DeploymentHasContainer(metrics.OtelAgentName),
+			testpredicates.DeploymentMissingEnvVar(reconcilermanager.Reconciler, "DISABLE_MONITORING"),
+		),
+	))
+	nt.Must(nt.WatchForAllSyncs())
+}
```
tiffanny29631, line 96: While nt.WatchForAllSyncs() validates deployment rollouts and Git sync status, it does not query metric endpoints. Verifying Prometheus metrics via nomostest.ValidateStandardMetricsForRootSync (or RepoSync) is necessary in the E2E test to confirm that metric scraping actually stops when disabled and resumes when re-enabled. For complete confidence, full telemetry verification (e.g., querying Google Cloud Monitoring or checking Otel-Collector pipeline health in otel_collector_test.go) would be a great addition to guarantee end-to-end metric resumption.

e2e/testcases/disable_monitoring_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// disable monitoring
+	nt.MustMergePatch(rootSyncV1, `{"spec": {"monitoring": {"enabled": false}}}`)
+
+	nt.Must(nt.Watcher.WatchObject(kinds.Deployment(),
+		rootReconcilerName.Name, rootReconcilerName.Namespace,
+		testwatcher.WatchPredicates(
+			testpredicates.DeploymentMissingContainer(metrics.OtelAgentName),
+			testpredicates.DeploymentHasEnvVar(reconcilermanager.Reconciler, "DISABLE_MONITORING", "true"),
+		),
+	))
```
tiffanny29631, line 71: Verify volume omit, for both test.

pkg/kmetrics/register.go
```diff
@@ -32,6 +32,10 @@ const (
 
 // RegisterOTelExporter creates the OTLP metrics exporter.
 func RegisterOTelExporter(ctx context.Context, containerName string) (*otlpmetricgrpc.Exporter, error) {
+	if os.Getenv("DISABLE_MONITORING") == "true" {
+		err := InitializeOTelKustomizeMetrics()
+		return nil, err
+	}
```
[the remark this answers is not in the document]
tiffanny29631: Please fix, the nil pointer panic applies to the 4 entry points : 

cmd/reconciler/main.go
cmd/hydration-controller/main.go 
cmd/reconciler-manager/main.go
pkg/resourcegroup/controllers/runner/run.go

pkg/api/configsync/v1alpha1/sync_types.go
```diff
@@ -249,3 +249,11 @@ type ResourceRef struct {
 	// +optional
 	GVK metav1.GroupVersionKind `json:"gvk,omitempty"`
 }
+
+// MonitoringSpec defines the monitoring configuration.
+type MonitoringSpec struct {
```
tiffanny29631, line 254: The API descriptions for monitoring and enabled are currently a bit tautological (monitoring defines the monitoring configuration). Following Kubernetes API documentation conventions, it would be clearer to describe the exact behavior, affected sidecars, and defaults

cmd/reconciler-manager/main.go
```diff
@@ -161,26 +161,30 @@ func main() {
 	})
 	setupLog.Info("RootSync controller registration scheduled")
 
-	otelCredentialProvider := &auth.CachingCredentialProvider{
-		Scopes: traceapi.DefaultAuthScopes(),
-	}
+	if os.Getenv("DISABLE_MONITORING") != "true" {
```
tiffanny29631, line 164: Notice that the CRD API uses positive naming (spec.monitoring.enabled: false) while the internal environment variable uses negative opt-out naming (DISABLE_MONITORING=true). While spec.monitoring.enabled follows K8s API conventions and DISABLE_MONITORING=true provides backward-compatible defaults when unset, the difference might confuse contributors. Could we add a short code comment in pkg/reconcilermanager/controllers/util.go explaining this naming rationale?

e2e/testcases/disable_monitoring_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/GoogleContainerTools/config-sync/pkg/api/configsync"
+	"github.com/GoogleContainerTools/config-sync/pkg/api/configsync/v1beta1"
+	"github.com/GoogleContainerTools/config-sync/pkg/core"
+	"github.com/GoogleContainerTools/config-sync/pkg/core/k8sobjects"
+	"github.com/GoogleContainerTools/config-sync/pkg/kinds"
+	"github.com/GoogleContainerTools/config-sync/pkg/metrics"
+	"github.com/GoogleContainerTools/config-sync/pkg/reconcilermanager"
+)
+
+func deploymentMissingVolume(volumeName string) testpredicates.Predicate {
```
tiffanny29631: Suggest moving the predicates to e2e/nomostest/testpredicates/predicates.go

pkg/kmetrics/register.go
```diff
@@ -32,6 +32,10 @@ const (
 
 // RegisterOTelExporter creates the OTLP metrics exporter.
 func RegisterOTelExporter(ctx context.Context, containerName string) (*otlpmetricgrpc.Exporter, error) {
+	if os.Getenv("DISABLE_MONITORING") == "true" {
```
tiffanny29631, line 35: Meant to be "!="? I wonder why the test was not catching this

pkg/metrics/register.go
```diff
@@ -33,6 +33,11 @@ const (
 
 // RegisterOTelExporter creates the OTLP metrics exporter.
 func RegisterOTelExporter(ctx context.Context, containerName string) (*otlpmetricgrpc.Exporter, error) {
+	if os.Getenv("DISABLE_MONITORING") == "true" {
```
tiffanny29631, line 36: "!="?

pkg/metrics/register.go
```diff
@@ -33,6 +33,11 @@ const (
...
+		klog.V(5).Infof("METRIC DEBUG: Monitoring is disabled via DISABLE_MONITORING env var")
+		err := InitializeOTelMetrics()
```
tiffanny29631, line 41: Why init if monitoring is disabled?

pkg/kmetrics/register.go
cowsking: In `cmd/reconciler-manager/main.go`, we use `if os.Getenv("DISABLE_MONITORING") != "true"` because we are conditionally *executing* and wrapping a block of code (and persisting general initialization flow afterwards). 

Here in `pkg/kmetrics/register.go` (and `pkg/metrics/register.go`), we use `== "true"` as an early return to guard clause. If monitoring is explicitly disabled, we initialize the safe Noop variables and `return` early. The rest of the function operates under the assumption that monitoring is active. 

Returning nil here is by design, no exporter exists when monitoring is disabled. All four call sites in this PR already nil-check OTel Collector Exporter before Shutdown, so this can't panic.

pkg/metrics/register.go
cowsking: The recording functions in `record.go` call these global instruments unconditionally and skipping init would leave them `nil` and panic on the first `Record()` call. Since we skip `SetMeterProvider` here, the instruments fall back to the global `no-op` provider, so recordings are safe `no-ops` and nothing is exported. Added a comment in the code to clarify.

cmd/reconciler-manager/main.go
cowsking: Thank you for the suggestions! Added the comment

pkg/api/configsync/v1alpha1/sync_types.go
cowsking: Acked. Updated the godoc to explicitly describe that it removes the otel-agent sidecar and sets DISABLE_MONITORING=true, rather than just saying it configures monitoring.

e2e/testcases/disable_monitoring_test.go
cowsking: Ack, done. Moved the predicates to e2e/nomostest/testpredicates/predicates.go.

tiffanny29631 (approved):
/lgtm

PR #2242, opened 2026-07-28
feat(nomos): distinguish synced and current polling

### LINKED ISSUE #2006: Feature Request: Add –poll-until to ‘nomos status’ command
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature

Add –poll-until to ‘nomos status’ command

Reconciliation is a tricky thing in Kubernetes clusters. A way to think about that is using temporal logic: F ψ ≡ true U ψ (eventually ψ becomes true)

In this case, eventually resource x will be reconciled with the cluster. Of course, one cannot wait forever for the reconciliation to happen.

With that in mind, poll-until would be able construct something like:

$ timeout 2m nomos status –poll-until=complete

Print, and wait, till all resources are complete(i.e.: reconciled with cluster). If that doesn’t happen in 2 minutes, terminate.

### Importance

_No response_

tomatotomata:
Fixes #2006

## Summary

Adds explicit completion targets for `nomos status`:

- `--poll-until=synced` waits for the selected RootSync and RepoSync objects to report `SYNCED` without errors.
- `--poll-until=current` also waits for every reported managed resource to become `Current`. `complete` remains an alias for `current`.

The `--name` filter scopes the completion check. An empty reachable cluster does not block other contexts, while a requested name must be found. In `current` mode, a `Failed` resource returns an error instead of polling forever; `Unknown` remains incomplete. Use `synced` when resource readiness is not required.

## Validation

- `go test ./cmd/nomos/status -run 'TestPollUntilReached|TestValidatePollUntil|TestClusterStates' -count=1` passed.
- `git diff --check` passed.
- The full status package run reaches an existing Windows timezone-sensitive `TestPrintStatus` expectation: the fixture expects UTC while the output uses the local Riyadh timezone. The focused polling and cluster-state tests pass, and the PR presubmits will validate the pushed head.

### DIFF
--- a/cmd/nomos/status/cluster_state.go
+++ b/cmd/nomos/status/cluster_state.go
@@ -46,6 +46,26 @@ type ClusterState struct {
 	isMulti *bool
 }
 
+// allSynced reports whether every reachable cluster has at least one repository
+// and every repository has completed successfully. A cluster or repository
+// error keeps polling active so callers can observe a later successful state.
+func allSynced(states map[string]*ClusterState) bool {
+	if len(states) == 0 {
+		return false
+	}
+	for _, state := range states {
+		if state == nil || state.Error != "" || len(state.repos) == 0 {
+			return false
+		}
+		for _, repo := range state.repos {
+			if repo == nil || repo.status != syncedMsg || len(repo.errors) > 0 {
+				return false
+			}
+		}
+	}
+	return true
+}
+
 func (c *ClusterState) printRows(writer io.Writer) {
 	util.MustFprintf(writer, "\n")
 	util.MustFprintf(writer, "%s\n", c.Ref)
--- a/cmd/nomos/status/cluster_state.go
+++ b/cmd/nomos/status/cluster_state.go
@@ -61,6 +61,11 @@ func allSynced(states map[string]*ClusterState) bool {
 			if repo == nil || repo.status != syncedMsg || len(repo.errors) > 0 {
 				return false
 			}
+			for _, resource := range repo.resources {
+				if resource.Status != kptv1alpha1.Current {
+					return false
+				}
+			}
 		}
 	}
 	return true
--- a/cmd/nomos/status/poll_until_test.go
+++ b/cmd/nomos/status/poll_until_test.go
@@ -14,7 +14,11 @@
 
 package status
 
-import "testing"
+import (
+	"testing"
+
+	kptv1alpha1 "github.com/GoogleContainerTools/config-sync/pkg/api/kpt.dev/v1alpha1"
+)
 
 func TestAllSynced(t *testing.T) {
 	tests := []struct {
@@ -34,6 +38,16 @@ func TestAllSynced(t *testing.T) {
 			states: map[string]*ClusterState{
 				"cluster": {repos: []*RepoState{{status: pendingMsg}}},
 			},
+		},
+		{
+			name: "repository with non-current resource",
+			states: map[string]*ClusterState{
+				"cluster": {repos: []*RepoState{{
+					status: syncedMsg,
+					resources: []kptv1alpha1.ResourceStatus{{
+						Status: kptv1alpha1.Failed,
+					}},
+				}}},
 		},
 		{
 			name: "cluster error",
--- a/cmd/nomos/status/poll_until_test.go
+++ b/cmd/nomos/status/poll_until_test.go
@@ -1,0 +1,71 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package status
+
+import "testing"
+
+func TestAllSynced(t *testing.T) {
+	tests := []struct {
+		name   string
+		states map[string]*ClusterState
+		want   bool
+	}{
+		{
+			name: "all repositories synced",
+			states: map[string]*ClusterState{
+				"cluster": {repos: []*RepoState{{status: syncedMsg}}},
+			},
+			want: true,
+		},
+		{
+			name: "pending repository",
+			states: map[string]*ClusterState{
+				"cluster": {repos: []*RepoState{{status: pendingMsg}}},
+			},
+		},
+		{
+			name: "cluster error",
+			states: map[string]*ClusterState{
+				"cluster": {Error: "unavailable", repos: []*RepoState{{status: syncedMsg}}},
+			},
+		},
+		{
+			name: "empty state",
+			states: map[string]*ClusterState{
+				"cluster": {},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			if got := allSynced(test.states); got != test.want {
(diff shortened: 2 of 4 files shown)


cmd/nomos/status/cluster_state.go
```diff
@@ -46,6 +46,26 @@ type ClusterState struct {
...
+// error keeps polling active so callers can observe a later successful state.
+func allSynced(states map[string]*ClusterState) bool {
+	if len(states) == 0 {
+		return false
+	}
+	for _, state := range states {
+		if state == nil || state.Error != "" || len(state.repos) == 0 {
+			return false
+		}
+		for _, repo := range state.repos {
```
tiffanny29631, line 60: Thank you for working on this! From what I read, allSynced returns true as soon as the repository is SYNCED (applied). However, it does not verify if the actual underlying resources have finished reconciling (e.g., reaching status Current in the Resource Group). If a deployment is successfully applied but goes into CrashLoopBackOff, this logic will still consider it "Complete".

Could you clarify if the intent is to wait for Apply completion or full Resource Readiness? If it's readiness, we should also iterate over repo.resources. If it's just apply, we should explicitly document this behavior so automation doesn't falsely assume workloads are ready to serve traffic.

tomatotomata:
Followed up on the readiness question in commit `64ff1554` (already pushed): `--poll-until=complete` now waits for both repository `SYNCED` status and every reported managed resource to reach `Current`. The design doc explicitly documents that this is full resource readiness and that it is independent of the `--resources` display flag. A regression test covers a non-current resource.

The Google CLA bot still requires the account holder to complete the CLA; that external signature step cannot be performed by the agent.

cmd/nomos/status/cluster_state.go
tomatotomata: Thanks — the intended meaning is Config Sync resource reconciliation, not application readiness. A workload can still be unhealthy after its managed resource is applied, so `--poll-until=complete` must not promise readiness such as avoiding CrashLoopBackOff.

The current PR head (`64ff1554`) now requires every ResourceGroup entry to be `Current` in addition to a `SYNCED` repository, with a regression case for a failed managed resource. The design note now makes that boundary explicit.

tomatotomata:
The presubmit failure was a syntax error in the new table literal at `poll_until_test.go:64`. I fixed the nested literal structure in `8603456a` and pushed it. The focused status tests pass locally; the full package run on this Windows host also exposes an existing timezone-sensitive `TestPrintStatus` expectation, so I have not treated that local result as a full pass.

Could you please rerun the presubmit on the new head?

/retest

tomatotomata:
I traced the latest results to two separate causes. The required presubmit still wanted Go's simplified composite-literal form in `poll_until_test.go`, so I applied `gofmt -s` and pushed `20db06ab`. The e2e group 3 failure happened while the runner was building its Docker image and could not fetch the Ubuntu `manpages` package from `archive.ubuntu.com`, so that result looks environment-related rather than caused by this change. The focused status tests and `git diff --check` pass on the new head, and the push has already started a fresh full run. Let me know what you think.

cmd/nomos/status/cluster_state.go
```diff
@@ -46,6 +46,31 @@ type ClusterState struct {
 	isMulti *bool
 }
 
+// allSynced reports whether every reachable cluster has at least one repository
+// and every repository has completed successfully. A cluster or repository
+// error keeps polling active so callers can observe a later successful state.
+func allSynced(states map[string]*ClusterState) bool {
```
tiffanny29631: This ignores the `--name` flag. `allSynced` should accept the `name` filter (or filter `state.repos` before evaluating).

cmd/nomos/status/cluster_state.go
```diff
@@ -46,6 +46,31 @@ type ClusterState struct {
...
+	if len(states) == 0 {
+		return false
+	}
+	for _, state := range states {
+		if state == nil || state.Error != "" || len(state.repos) == 0 {
```
tiffanny29631: Requiring `len(state.repos) > 0` on every reachable cluster causes infinite polling if:
1. The user has multiple contexts in their kubeconfig and one cluster does not have any sync objects configured.
2. The user runs `nomos status --namespace <ns>` on a namespace that has no `RepoSync` objects.
We should distinguish between "not yet populated" vs. "no sync objects configured on this cluster/namespace".

cmd/nomos/status/cluster_state.go
```diff
@@ -46,6 +46,31 @@ type ClusterState struct {
...
+			return false
+		}
+		for _, repo := range state.repos {
+			if repo == nil || repo.status != syncedMsg || len(repo.errors) > 0 {
+				return false
+			}
+			for _, resource := range repo.resources {
+				if resource.Status != kptv1alpha1.Current {
```
tiffanny29631: Requiring all managed resources to reach `kptv1alpha1.Current` introduces two catches:
1. Resources/CRDs that don't report standard kstatus conditions often remain in `Unknown` status and will never reach `Current`, causing this loop to hang forever.
2. Workloads failing readiness (e.g. CrashLoopBackOff) report `Failed` or `InProgress` in kstatus, which keeps polling active indefinitely instead of surfacing an error.
Could we either:
- Support explicit values such as `--poll-until=synced` (RootSync/RepoSync applied) vs `--poll-until=current` (all resources ready)?
- Or fail fast when a resource enters a terminal `Failed` state?

cmd/nomos/status/poll_until_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+package status
+
+import (
+	"testing"
+
+	kptv1alpha1 "github.com/GoogleContainerTools/config-sync/pkg/api/kpt.dev/v1alpha1"
+)
+
+func TestAllSynced(t *testing.T) {
```
tiffanny29631: Please add test cases verifying:
1. Behavior when the `--name` filter is specified.
2. Behavior when a managed resource is `Unknown` vs `Failed` vs `Current`.
3. Behavior with multiple clusters where one cluster has no sync objects.

cmd/nomos/status/status.go
```diff
@@ -161,7 +174,7 @@ func clusterStates(ctx context.Context, clientMap map[string]*ClusterClient) (ma
 // and then prints a formatted status row for each one. If there are any errors reported by either
 // object, those are printed in a second table under the status table.
 // nolint:errcheck
-func printStatus(ctx context.Context, writer *tabwriter.Writer, clientMap map[string]*ClusterClient, names []string) {
+func printStatus(ctx context.Context, writer *tabwriter.Writer, clientMap map[string]*ClusterClient, names []string) bool {
```
tiffanny29631: Nit: `printStatus` should only be responsible for formatting and writing output. Having it return a `bool` for sync state couples presentation with business logic.

tiffanny29631 (commented):
Please update the PR description and title to reflect latest changes

tomatotomata:
I changed the polling target to separate successful Config Sync application from managed-resource readiness. `--poll-until=synced` stops once the selected repositories report `SYNCED`; `--poll-until=current` also waits for every reported resource to be `Current`, with `complete` kept as an alias. A `Failed` resource now returns an error instead of polling forever, while `Unknown` remains incomplete in current mode.

The completion check now respects `--name`, and a reachable cluster with no sync objects no longer holds up other contexts. I added coverage for name selection, empty clusters, `Current`/`Unknown`/`Failed`, and the separate synced target. I also split state collection from rendering so `printStatus` only formats the output.

The focused tests and `git diff --check` pass on `13cf2d1f`. The full status-package run still hits the UTC-versus-local-time expectation in `TestPrintStatus` on this Windows host; the new presubmits are running on this head. I updated the title and description to reflect the expanded behavior. Let me know if you think the `synced`/`current` split fits the issue.

PR #2244, opened 2026-07-29
docs: add disabling monitoring guide and benchmark report

cowsking:


docs/custom-metric-filter.md
```diff
@@ -307,9 +307,71 @@ curl http://localhost:8675/metrics
...
+2. Delete the monitoring namespace:
+
+```shell
+kubectl delete ns config-management-monitoring
+```
+
+When `DISABLE_MONITORING=true` is injected:
+- The `reconciler-manager` skips registering Otel controllers.
+- Reconciler Pods will no longer include `otel-agent` sidecar containers.
+- Metric recording calls fallback safely to OpenTelemetry Noop providers without panic or network connection errors.
```
tiffanny29631: I feel this section is too detail in technical side, can be removed.

docs/custom-metric-filter.md
```diff
@@ -307,9 +307,71 @@ curl http://localhost:8675/metrics
...
+
+1. Statically inject the `DISABLE_MONITORING=true` environment variable into the Deployment manifests of both `reconciler-manager` and `resource-group-controller-manager`:
+
+```yaml
+env:
+- name: DISABLE_MONITORING
+  value: "true"
+```
+
+2. Delete the monitoring namespace:
```
tiffanny29631, line 354: Recommend noting the users to confirm all the syncs are configured to disable monitoring before this step

docs/custom-metric-filter.md
```diff
@@ -307,9 +307,71 @@ curl http://localhost:8675/metrics
...
+
+### Option 2: Complete Cluster-Wide Removal (OSS Administrators)
+
+If an open-source cluster administrator wants to safely delete the `config-management-monitoring` namespace entirely and stop all OpenTelemetry controller activity:
+
+1. Statically inject the `DISABLE_MONITORING=true` environment variable into the Deployment manifests of both `reconciler-manager` and `resource-group-controller-manager`:
+
+```yaml
+env:
+- name: DISABLE_MONITORING
```
tiffanny29631, line 350: It'd be convenient if provide a kubectl command for both controllers

docs/custom-metric-filter.md
```diff
@@ -307,9 +307,71 @@ curl http://localhost:8675/metrics
...
+
+### Option 1: Per-Sync Declarative Disabling (RootSync / RepoSync)
+
+You can turn off monitoring for individual `RootSync` or `RepoSync` objects by setting `spec.monitoring.enabled: false` in their specification:
+
+```yaml
+apiVersion: configsync.gke.io/v1beta1
+kind: RootSync
+metadata:
+  name: root-sync
```
tiffanny29631: Convert the root-sync to a rootsync name placeholder. Same for the namespace - some reposync might not be in this namespace.

docs/custom-metric-filter.md
cowsking: Ack. Removed the technical implementation details section as suggested, and updated the example YAML to use custom-root-sync along with a Note disclaimer.

docs/custom-metric-filter.md
cowsking: Ack

docs/custom-metric-filter.md
cowsking: Ack. Added a Note callout before step 2 advising users to verify that DISABLE_MONITORING=true is applied to all manager controllers

docs/custom-metric-filter.md
cowsking: Ack. Added `kubectl set env` commands for both `reconciler-manager` and `resource-group-controller-manager` in Step 1 to make it easy for administrators to dynamically set. Thanks for the suggestion!

tiffanny29631:
/hold
Please unhold on release day.

tiffanny29631 (approved):
/lgtm

PR #2255, opened 2026-08-12
fix: register explicit bucket boundaries for duration histograms

cowsking:
Following the OpenCensus to OpenTelemetry metrics migration, duration histogram instruments in Config Sync lacked explicit bucket boundaries. This caused the OpenTelemetry Go SDK to fall back to default millisecond-scale boundaries {0, 5, 10, 25, ...}, which for second-unit metrics resulted in all sub-5-second values falling into the (0, 5] bucket and skewing Prometheus histogram_quantile queries (e.g. static 4.75s).

This change:
- Configures explicit sub-second bucket boundaries for core duration metrics (APICallDuration, ReconcileDuration, RemediateDuration) and long duration metrics (ParserDuration, ApplyDuration).
- Configures explicit bucket boundaries for ResourceGroup ReconcileDuration.
- Fixes `pkg/kmetrics` `kustomize_build_latency`: the histogram recorded raw **nanosecond** values while declaring a **millisecond** unit, so every recording landed in the overflow (+Inf) bucket regardless of bucket configuration. It now records the duration in **seconds** (consistent with all other Config Sync duration metrics), declares the unit as `s`, and registers explicit bucket boundaries. Note this changes the scale of the exported values (previously nanosecond-scale numbers labeled as ms).
- Adds unit tests to verify histogram bucket bounds and sub-second data point placement.

### DIFF
--- a/pkg/kmetrics/exec.go
+++ b/pkg/kmetrics/exec.go
@@ -120,7 +120,7 @@ func runKustomizeBuild(ctx context.Context, sendMetrics bool, inputDir string, c
 		resourceCount, err := kustomizeResourcesGenerated(output)
 		if err == nil && sendMetrics {
 			RecordKustomizeResourceCount(ctx, resourceCount)
-			RecordKustomizeExecutionTime(ctx, float64(executionTime))
+			RecordKustomizeExecutionTime(ctx, float64(executionTime.Milliseconds()))
 		}
 		outputs <- output
 		errors <- kustomizeErr
@@ -139,14 +139,14 @@ func runKustomizeBuild(ctx context.Context, sendMetrics bool, inputDir string, c
 	return <-outputs, <-errors
 }
 
-func runCommand(cmd *exec.Cmd) (int64, string, error) {
+func runCommand(cmd *exec.Cmd) (time.Duration, string, error) {
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	now := time.Now()
 	err := cmd.Run()
-	executionTime := time.Since(now).Nanoseconds()
+	executionTime := time.Since(now)
 	if err != nil {
 		return executionTime, stdout.String(), errors.New(stderr.String())
 	}
--- a/pkg/kmetrics/metrics.go
+++ b/pkg/kmetrics/metrics.go
@@ -67,6 +67,15 @@ var (
 	KustomizeExecutionTime metric.Float64Histogram
 )
 
+var (
+	// KustomizeBuildLatencyBounds defines the millisecond bounds for the kustomize
+	// build latency histogram. These must match the pre-OTel-migration view
+	// aggregation (view.Distribution(0, 10, 20, ..., 10240)) that the Monarch
+	// metric descriptor was registered with; changing the bounds or the unit
+	// causes the exporter to fail with a bucket options mismatch.
+	KustomizeBuildLatencyBounds = []float64{0, 10, 20, 40, 80, 160, 320, 640, 1280, 2560, 5120, 10240}
+)
+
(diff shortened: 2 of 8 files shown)


pkg/metrics/metrics.go
```diff
@@ -51,6 +51,14 @@ const (
 	InternalErrorsName = "internal_errors_total"
 )
```
tiffanny29631, line 53: Thanks for working on the fix, could you check if the fix apply to pkg/kmetrics as well?

cowsking: Good catch, thanks! Pushed an update to include `pkg/kmetrics`: added explicit second-scale bucket bounds for `kustomize_build_latency`, fixed a pre-existing bug in `exec.go` where raw nanoseconds were recorded as duration and added unit test coverage.

pkg/kmetrics/exec.go
```diff
@@ -120,7 +120,7 @@ func runKustomizeBuild(ctx context.Context, sendMetrics bool, inputDir string, c
 		resourceCount, err := kustomizeResourcesGenerated(output)
 		if err == nil && sendMetrics {
 			RecordKustomizeResourceCount(ctx, resourceCount)
-			RecordKustomizeExecutionTime(ctx, float64(executionTime))
+			RecordKustomizeExecutionTime(ctx, executionTime.Seconds())
```
tiffanny29631: Record in ms, to align with pkg/kmetrics/metrics.go

pkg/kmetrics/metrics.go
```diff
@@ -152,8 +157,9 @@ func InitializeOTelKustomizeMetrics() error {
 	// Initialize histogram instrument
 	KustomizeExecutionTime, err = meter.Float64Histogram(
 		"kustomize_build_latency",
-		metric.WithDescription("Kustomize build latency"),
-		metric.WithUnit("ms"),
+		metric.WithDescription("Kustomize build latency in seconds"),
+		metric.WithUnit("s"),
```
tiffanny29631: Keep metric.WithUnit("ms"), set KustomizeBuildLatencyBounds to match its original millisecond exponential buckets. Context go/config-sync-monarch-metrics some metrics are predefined in Monarch, changes in unit might cause error like `INVALID_ARGUMENT: The bucket options / unit do not match the metric descriptor`

pkg/metrics/metrics.go
```diff
@@ -51,6 +51,14 @@ const (
...
 
+var (
+	// DistributionBounds defines the bounds for a histogram distribution measuring short durations.
+	DistributionBounds = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
+
+	// LongDistributionBounds defines the bounds for a histogram distribution measuring long durations.
+	LongDistributionBounds = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 5, 10, 30, 60, 300, 600, 1200, 1800, 3600, 5400}
```
tiffanny29631: Could you check whether LongDistributionBounds (parser_duration_seconds, apply_duration_seconds) should start at 1 instead of .005 to match Monarch's pre-registered descriptor ([1, 5, 10, ...]).

pkg/kmetrics/exec.go
cowsking: Done. Updated to `float64(executionTime.Milliseconds())`

pkg/kmetrics/metrics.go
cowsking: Done. Kept the `"ms"` unit and restored the exponential buckets (`[0, 10, 20, ..., 10240]`) to match the Monarch descriptor.

pkg/metrics/metrics.go
cowsking: Updated `LongDistributionBounds` to start at `1` (`[1, 5, 10, ..., 5400]`) to align

tiffanny29631 (approved):
/lgtm

